PR #25, opened 2022-08-18
[3.2] Implement JSON Snapshot Reader

ClaytonCalabrese:
https://github.com/eosnetworkfoundation/mandel/pull/710 ports a way to convert a snapshot to json. This PR enhances the --snapshot option to support reading a snapshot in JSON format if the snapshot has .json extension.

Resolves: https://github.com/eosnetworkfoundation/mandel/issues/771

Previous discussion: https://github.com/eosnetworkfoundation/mandel/pull/732


spoonincode:
Is there any reasonable way to sniff the file and determine binary vs JSON, instead of relying on the file name?

heifner:
> Is there any reasonable way to sniff the file and determine binary vs JSON, instead of relying on the file name?

Maybe, is that worth doing? The `snapshot-to-json` generates with a `.json` format.

spoonincode:
> Maybe, is that worth doing?

Maybe. Being prescriptive about filenames can break more flexible invocations. Like maybe someone could set up a pipeline to modify a snapshot without ever writing it to disk via something like
```
nodeos --snapshot <(leap-util --snapshot-to-json snapshot.bin | program_mutates_json_somehow) --delete-all
```

But this could just be academic. The binary snapshots need to be seekable at the moment; I'm not sure if the JSON ones seek around or not, and it may be hard to accommodate something like the above command even for JSON, idk

libraries/chain/include/eosio/chain/snapshot.hpp
```diff
@@ -5,6 +5,7 @@
 #include <fc/variant_object.hpp>
 #include <boost/core/demangle.hpp>
 #include <ostream>
+#include <memory>
 
 namespace eosio { namespace chain {
```
heifner, line 10: Please rebase this PR to main (or start a new PR) and incorporate `libraries/chain/include/eosio/chain/snapshot.hpp` & `libraries/chain/snapshot.cpp` changes from https://github.com/AntelopeIO/leap/pull/24 into it. Then we will be able to get the PR into `main`. The `--snapshot-to-json` portion of #24 can be added to `leap-util` later (keep that PR up so it can be easily incorporated into `leap-util`.

libraries/testing/include/eosio/testing/snapshot_suites.hpp
```diff
@@ -106,5 +106,68 @@ struct buffered_snapshot_suite {
...
+      ,storage(storage)
+      {
+
+      }
+
+      std::shared_ptr<write_storage_t> storage;
+   };
+
+   static std::string temp_file() {
+      static fc::temp_directory temp_dir;
```
heifner, line 129: Don't use a `static` here. Instead have the reader create its own `temp_directory` and then its destructor can remove it. Using a `static` here means that every call to `temp_file()` returns the same file, which seems counterintuitive.

heifner (changes requested):

libraries/testing/include/eosio/testing/snapshot_suites.hpp
ClaytonCalabrese: It has to be used here and not within the reader struct because it is needed in the static function `get_reader` which constructs the reader. This code is also only relevant in the `snapshot tests`.

heifner: I think it is worth cleaning up. Probably need to refactor the `get_reader` (just remove and use types directly).

heifner: Sorry, I complete misread how this was being used.

heifner:
[@spoonincode sorry, somehow edited your comment instead of quoting it. I didn't even think that was possible.]

> But this could just be academic. The binary snapshots need to be seekable at the moment; I'm not sure if the JSON ones seek around or not, and it may be hard to accommodate something like the above command even for JSON, idk

We could just try reading as binary which should quickly fail as it will not find the magic number of the binary file, then switch over to JSON.

PR #29, opened 2022-08-19
[3.2] Refactor producer_plugin to consolidate controller push_transaction calls

heifner:
`producer_plugin` had duplicate code around calling chain `controller` `push_transaction` which resulted in slight differences in handling of error conditions; particularly around logging of failures. This PR refactors the calling of controller `push_transaction` into a common `push_transaction` call so that all instances of handling incoming transactions are handled the same way. Now all transactions are logged regardless of the stage of execution. Also ordering of log messages are now consistent with `transaction` logged before `transaction_success_tracing`/`transaction_failure_tracing` which is logged before `transaction_trace_success`/`transaction_trace_failure`.

Includes a slight change to log output format of https://github.com/eosnetworkfoundation/mandel/pull/733, new example:
```
debug 2022-08-19T13:21:48.872 nodeos    producer_plugin.cpp:1838      log_trx_results      ] [TRX_TRACE] Block 12 for producer eosio is REJECTING tx: 5766e8a589b7725f6d0dd9ea8044f62bedd18754b6902a310402e8fd49ecbcea, auth: eosio, action: eosio:unlinkauth, the owner of the linked permission needs to be the actor of the declared authorization
```

Includes a fix for `channels::transaction_ack` not always being signaled for expired or transactions with invalid signatures.
Includes a change to set `disable-api-persisted-trx=true` when running in `HEAD` mode as transaction side-effects are undone anyway when in `HEAD` mode so no need to ever speculatively execute them.

Resolves https://github.com/eosnetworkfoundation/mandel/issues/297


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1914,12 +1781,215 @@ bool producer_plugin_impl::remove_expired_blacklisted_trxs( const fc::time_point
...
+         uint64_t prev_billed_plus100_us = prev_billed_cpu_time_us + EOS_PERCENT( prev_billed_cpu_time_us, 100 * config::percent_1 );
+         if( prev_billed_plus100_us < max_trx_time.count() ) max_trx_time = fc::microseconds( prev_billed_plus100_us );
+      }
+   }
+
+   auto trace = chain.push_transaction( trx, block_deadline, max_trx_time, prev_billed_cpu_time_us, false, sub_bill );
+   push_result pr;
+   if( trace->except ) {
+      if( exception_is_exhausted( *trace->except ) ) {
+         //_unapplied_transactions.add_incoming( trx, persist_until_expired, return_failure_traces, next );
```
heifner, line 1929: remove commented code

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1914,12 +1781,215 @@ bool producer_plugin_impl::remove_expired_blacklisted_trxs( const fc::time_point
...
+   } else {
+      fc_dlog( _trx_successful_trace_log, "Subjective bill for success ${a}: ${b} elapsed ${t}us, time ${r}us",
+               ("a",first_auth)("b",sub_bill)("t",trace->elapsed)("r", fc::time_point::now() - start));
+      log_trx_results( trx, trace, nullptr, sub_bill, start );
+      if( persist_until_expired && !_disable_persist_until_expired ) {
+         // if this trx didn't fail/soft-fail and the persist flag is set
+         // ensure it is applied to all future speculative blocks as well.
+         // No need to subjective bill since it will be re-applied.
+         pr.persist = true;
+         //_unapplied_transactions.add_persisted( trx );
```
heifner, line 1974: Remove commented out code.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1914,12 +1748,223 @@ bool producer_plugin_impl::remove_expired_blacklisted_trxs( const fc::time_point
...
+   if (!details.empty()) {
+      details = fc::format_string("${d}", fc::mutable_variant_object() ("d", details), true);  // true for limiting the formatted string size
+   }
+
+   return "action: " + contract_name + ":" + act_name + ", " + details;
+}
+
+void producer_plugin_impl::log_trx_results( const transaction_metadata_ptr& trx,
+                                            const transaction_trace_ptr& trace, const fc::exception_ptr& except_ptr,
+                                            const fc::time_point& start )
```
heifner, line 1781: Create a version that only takes a `trace` and a version that only takes a `except_ptr` since they are never both provided.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1914,12 +1748,223 @@ bool producer_plugin_impl::remove_expired_blacklisted_trxs( const fc::time_point
...
+{
+   auto get_billed_cpu = [&](const transaction_metadata_ptr& trx, const transaction_trace_ptr& trace) -> uint32_t {
+      if( trace && trace->receipt ) {
+         return trace->receipt->cpu_usage_us;
+      } else if (trx) {
+         return trx->billed_cpu_time_us;
+      }
+      return 0;
+   };
```
heifner, line 1790: This does not have to be a lambda, make it a simple `if`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -487,20 +501,29 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-                                                          next{std::move(next)}, trx]() mutable {
+                                                          next{std::move(next)}, trx=trx]() mutable {
             if( future.valid() ) {
                future.wait();
                app().post( priority::low, [self, future{std::move(future)}, persist_until_expired, next{std::move( next )}, trx{std::move(trx)}, return_failure_traces]() mutable {
                   auto exception_handler = [self, &next, trx{std::move(trx)}](fc::exception_ptr ex) {
-                     fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Speculative execution is REJECTING tx: ${txid}, auth: ${a} : ${why} ",
-                             ("txid", trx->id())("a",trx->get_transaction().first_authorizer())("why",ex->what()));
+                     self->log_trx_results( trx, nullptr, ex, 0, fc::time_point::now() );
                      next(ex);
```
heifner, line 526: `ex` can be moved

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -487,20 +501,29 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                                                                  read_only ? transaction_metadata::trx_type::read_only : transaction_metadata::trx_type::input,
                                                                  chain.configured_subjective_signature_length_limit() );
 
+         if( !read_only ) {
+            next = [this, trx, next{std::move(next)}]( const std::variant<fc::exception_ptr, transaction_trace_ptr>& response ) {
```
heifner, line 505: Clear `unapplied_transaction_queue` on shutdown so that `this` does not outlive this lambda.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1914,12 +1748,223 @@ bool producer_plugin_impl::remove_expired_blacklisted_trxs( const fc::time_point
...
+      contract_name = actions[0].account.to_string();
+      act_name = actions[0].name.to_string();
+   }
+
+   details = except_ptr ? except_ptr->top_message() : (trace && trace->except) ? trace->except->top_message() : std::string();
+   if (!details.empty()) {
+      details = fc::format_string("${d}", fc::mutable_variant_object() ("d", details), true);  // true for limiting the formatted string size
+   }
+
+   return "action: " + contract_name + ":" + act_name + ", " + details;
```
heifner, line 1776: Add comment that this format is parsed by external user tools.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1914,12 +1748,223 @@ bool producer_plugin_impl::remove_expired_blacklisted_trxs( const fc::time_point
...
+                                     || ( persist_until_expired && _disable_subjective_api_billing )
+                                     || ( !persist_until_expired && _disable_subjective_p2p_billing )
+                                     || trx->read_only;
+
+   uint32_t sub_bill = 0;
+   if( !disable_subjective_billing )
+      sub_bill = _subjective_billing.get_subjective_bill( first_auth, fc::time_point::now() );
+
+   auto prev_billed_cpu_time_us = trx->billed_cpu_time_us;
+   if( prev_billed_cpu_time_us > 0 && _pending_block_mode == pending_block_mode::producing ) {
```
heifner, line 1894: move `_pending_block_mode == pending_block_mode::producing` as first check of `if`

plugins/producer_plugin/producer_plugin.cpp
heifner: done

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -517,210 +540,49 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+                     FC_LOG_MESSAGE( error, "duplicate transaction ${id}", ("id", id))));
+               log_trx_results( trx, except_ptr );
+               next( std::move(except_ptr) );
                return true;
             }
 
             if( !chain.is_building_block()) {
-               _unapplied_transactions.add_incoming( trx, persist_until_expired, return_failure_traces, next );
+               _unapplied_transactions.add_incoming( trx, persist_until_expired, return_failure_trace, next );
                return true;
```
heifner, line 575: Consider moving these into a one function.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1119,6 +981,9 @@ void producer_plugin::plugin_startup()
    EOS_ASSERT( my->_producers.empty() || my->chain_plug->accept_transactions(), plugin_config_exception,
               "node cannot have any producer-name configured because no block production is possible with no [api|p2p]-accepted-transactions" );
 
+   // persisting transactions only makes sense for SPECULATIVE mode.
+   if( !my->_disable_persist_until_expired ) my->_disable_persist_until_expired = chain.get_read_mode() != db_read_mode::SPECULATIVE;
+
```
heifner, line 986: Make a note of this change in the PR description.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2147,14 +2149,13 @@ void producer_plugin_impl::process_scheduled_and_incoming_trxs( const fc::time_p
    }
 }
 
-bool producer_plugin_impl::process_incoming_trxs( const fc::time_point& deadline, size_t& pending_incoming_process_limit )
+bool producer_plugin_impl::process_incoming_trxs( const fc::time_point& deadline, unapplied_transaction_queue::iterator& itr, size_t& pending_incoming_process_limit )
 {
    bool exhausted = false;
    if( pending_incoming_process_limit ) {
```
heifner, line 2155: Don't think `pending_incoming_process_limit` is needed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2112,7 +2114,7 @@ void producer_plugin_impl::process_scheduled_and_incoming_trxs( const fc::time_p
                        "[TRX_TRACE] Block ${block_num} for producer ${prod} is REJECTING scheduled tx: ${txid}, time: ${r}, auth: ${a} : ${details}",
```
heifner, line 2114: See if this can use the trx_log_result instead of duplicating very similar log output.

plugins/producer_plugin/producer_plugin.cpp
heifner: There would actually not be much left in the function if that was factored out.

plugins/producer_plugin/producer_plugin.cpp
heifner: Removed

plugins/producer_plugin/producer_plugin.cpp
heifner: Since transaction is not available, it feels to forced to try and make this work.

oschwaldp-oci (approved):

Issue #21, opened 2022-08-18
Compile warnings in trace_api_plugin


linh2931:
Minor warnings:

```
plugins/trace_api_plugin/compressed_file.cpp:138:66: warning:   comparison of integer  expressions of different signedness: ‘std::__tuple_element_t<0,   std::tuple<long unsigned  int, long unsigned int> >’ {aka ‘long unsigned int’} and ‘long int’ [-Wsign-compare]

plugins/trace_api_plugin/compressed_file.cpp:289:30: warning:   comparison of integer  expressions of different signedness: ‘int’ and ‘const long  unsigned int’ [-Wsign-compare]

trace_api_plugin/trace_api_plugin.cpp:363:12: warning:  variable ‘cfg_options’ set but not used [-Wunused-but-set-variable]
```

PR #35, opened 2022-08-22
[3.2] fix compile warnings in trace_api_plugin

linh2931:
Resolve https://github.com/AntelopeIO/leap/issues/21

Fix the following warnings:

```
plugins/trace_api_plugin/compressed_file.cpp:138:66: warning:   comparison of integer  expressions of different signedness: ‘std::__tuple_element_t<0,   std::tuple<long unsigned  int, long unsigned int> >’ {aka ‘long unsigned int’} and ‘long int’ [-Wsign-compare]

plugins/trace_api_plugin/compressed_file.cpp:289:30: warning:   comparison of integer  expressions of different signedness: ‘int’ and ‘const long  unsigned int’ [-Wsign-compare]

trace_api_plugin/trace_api_plugin.cpp:363:12: warning:  variable ‘cfg_options’ set but not used [-Wunused-but-set-variable]
```


plugins/trace_api_plugin/trace_api_plugin.cpp
```diff
@@ -360,7 +360,7 @@ struct trace_api_plugin_impl {
    :common(common) {}
 
    static void set_program_options(appbase::options_description& cli, appbase::options_description& cfg) {
-      auto cfg_options = cfg.add_options();
+      cfg.add_options();
```
spoonincode, line 363: Can this line just be entirely removed?

linh2931: Thanks! The whole function `trace_api_plugin_impl::trace_api_plugin_impl` can be removed.

plugins/trace_api_plugin/compressed_file.cpp
```diff
@@ -109,7 +109,7 @@ struct compressed_file_impl {
       }
    }
 
-   void seek( long loc, fc::cfile& file ) {
+   void seek( uint64_t loc, fc::cfile& file ) {
```
spoonincode, line 112: Does it make sense to change the `long remaining = loc` to `uint64_t remaining = loc`? It's not a warning but it may be worthwhile keeping the types the same as they were before

linh2931: Yes, that is even better. Will change it to `auto remaining = loc;`

spoonincode (approved):

Issue #20, opened 2022-08-18
Compile warnings in deep_mind.cpp, abi_serializer.cpp, producer_plugin.cpp, auth_tests.cpp


linh2931:
Minor warnings:

```
libraries/chain/deep_mind.cpp:46:4: warning: control reaches end of non-void function [-Wreturn-type]

libraries/chain/abi_serializer.cpp:80:42: warning: ‘void eosio::chain::abi_serializer::set_abi(const eosio::chain::abi_def&, const fc::microseconds&)’ is deprecated: use the overload with yield_function_t[=create_yield_function(max_serialization_time)] [-Wdeprecated-declarations]

plugins/producer_plugin/producer_plugin.cpp:1959:43: warning: comparison of integer expressions of different signedness: ‘uint64_t’ {aka ‘long unsigned int’} and ‘int64_t’ {aka  ‘long int’} [-Wsign-compare]
 1959 |                if( prev_billed_plus100_us < max_trx_time.count() ) max_trx_time = fc::microseconds( prev_billed_plus100_us );

unittests/auth_tests.cpp:235:18: warning: variable ‘new_owner_priv_key’ set but not used [-Wunused-but-set-variable]
```

PR #39, opened 2022-08-22
Compile warnings fixes in deep_mind.cpp, abi_serializer.cpp, producer_plugin.cpp, auth_tests.cpp

linh2931:
Resolve https://github.com/AntelopeIO/leap/issues/20

Fix following compile warnings:

```
libraries/chain/deep_mind.cpp:46:4: warning: control reaches end of non-void function [-Wreturn-type]

libraries/chain/abi_serializer.cpp:80:42: warning: ‘void eosio::chain::abi_serializer::set_abi(const eosio::chain::abi_def&, const fc::microseconds&)’ is deprecated: use the overload with yield_function_t[=create_yield_function(max_serialization_time)] [-Wdeprecated-declarations]

plugins/producer_plugin/producer_plugin.cpp:1959:43: warning: comparison of integer expressions of different signedness: ‘uint64_t’ {aka ‘long unsigned int’} and ‘int64_t’ {aka  ‘long int’} [-Wsign-compare]
 1959 |                if( prev_billed_plus100_us < max_trx_time.count() ) max_trx_time = fc::microseconds( prev_billed_plus100_us );

unittests/auth_tests.cpp:235:18: warning: variable ‘new_owner_priv_key’ set but not used [-Wunused-but-set-variable]
```


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1955,7 +1955,7 @@ bool producer_plugin_impl::process_unapplied_trxs( const fc::time_point& deadlin
 
             auto prev_billed_cpu_time_us = trx->billed_cpu_time_us;
             if( prev_billed_cpu_time_us > 0 && !_subjective_billing.is_account_disabled( first_auth ) && !rl.is_unlimited_cpu( first_auth )) {
-               uint64_t prev_billed_plus100_us = prev_billed_cpu_time_us + EOS_PERCENT( prev_billed_cpu_time_us, 100 * config::percent_1 );
+               int64_t prev_billed_plus100_us = prev_billed_cpu_time_us + EOS_PERCENT( prev_billed_cpu_time_us, 100 * config::percent_1 );
```
heifner, line 1958: This is narrowing from `uint64_t` to `int64_t` which should be fine here as `prev_billed_cpu_time_us` is `uint32_t`. BTW, this is a very odd way to multiple a value by 2. It is written this way because originally we had a different value than `100`.

heifner (approved):

plugins/producer_plugin/producer_plugin.cpp
linh2931: Thanks for the review. Yes, I was worried about that narrowing too. Since `fc::microseconds`'s `_count` is defined as `int64_t`, it should fine.

PR #40, opened 2022-08-23
Capture data and create usable data structures storing data in performance_test_basic.py

ClaytonCalabrese:
Expanded the number of variables capture in performance_test_basic.py
Stored all data in types for future use cases.
Changed to a more class oriented design.


tests/performance_tests/performance_test_basic.py
```diff
@@ -21,26 +21,144 @@
...
+        for str in strResult:
+            intResult = re.findall(r'\d+', str)
+            total.blockLog[i].time = int(intResult[0])
+    i = -1
+    for value in trxResult:
+        i+=1
+        strResult = re.findall(r'latency:\s+.*\d+', value)
+        for str in strResult:
+            intResult = re.findall(r'-*\d+', str)
+            total.blockLog[i].latency = int(intResult[0])
```
oschwaldp-oci, line 160: This seems like it could all be done in one iteration through all `values` in `trxResult` to avoid the index gymnastics.  Can you get set all `blockData` fields in one go?

tests/performance_tests/performance_test_basic.py
```diff
@@ -97,9 +215,11 @@ def checkTotalTrx():
     testGenerationDurationSec = 60
     targetTps = 1
     transactionsSent = testGenerationDurationSec * targetTps
+    cont = chainsContainer()
 
-    waitForEmptyBlocks(node1)
-    setupTotal = checkTotalTrx()
+    # Get stats prior to transaction generation
+    cont.startBlock = waitForEmptyBlocks(node1)
+    fetchStats(cont.preData)
```
oschwaldp-oci, line 222: It would be great if we could push all log processing to the end of the test as a post processing step.  This would mean we would need to process the whole log in one fell swoop, then be able to pull out a range of blocks that we want to gather stats on.

tests/performance_tests/performance_test_basic.py
```diff
@@ -21,26 +21,144 @@
...
 
-def waitForEmptyBlocks(node1):
+class blockData():
+    def __init__(self):
+        self.transactions = 0
+        self.cpu = 0
+        self.net = 0
+        self.elapsed = 0
+        self.time = 0
+        self.latency = 0
```
oschwaldp-oci, line 31: Not sure exactly where it would fit best in the data structure, potentially in `blockData`, but it would be good to capture the block number and block id so that we can interact with known blocks and known sequences of blocks.

tests/performance_tests/performance_test_basic.py
```diff
@@ -97,9 +215,11 @@ def checkTotalTrx():
     testGenerationDurationSec = 60
     targetTps = 1
     transactionsSent = testGenerationDurationSec * targetTps
+    cont = chainsContainer()
 
-    waitForEmptyBlocks(node1)
-    setupTotal = checkTotalTrx()
+    # Get stats prior to transaction generation
+    cont.startBlock = waitForEmptyBlocks(node1)
```
oschwaldp-oci, line 221: I realize the variable naming was already in place, but would it be good to rename/identify a node as either a producer or validation node?  I think this would make understanding the context of the test, which node type is used for which purpose, more easily discernable.

tests/performance_tests/performance_test_basic.py
```diff
@@ -21,26 +21,144 @@
...
+        self.latency = 0
+
+class chainData():
+    def __init__(self):
+        self.blockLog = []
+    def sum_transactions(self):
+        total = 0
+        for block in self.blockLog:
+            total += block.transactions
+        return total
```
oschwaldp-oci, line 40: Instead of having this always dynamically calculated, should there be internal variables that track and store these data points which are updated each time a `blockData` is added to the `blockLog`?  This might be premature. Just worth thinking about.

oschwaldp-oci (changes requested):

tests/performance_tests/performance_test_basic.py
```diff
@@ -21,26 +21,144 @@
 relaunchTimeout = 30
 emptyBlockGoal = 5
 
-def waitForEmptyBlocks(node1):
+class blockData():
```
jgiszczak, line 24: This should be a Python 3 [data class](https://docs.python.org/3/library/dataclasses.html).

tests/performance_tests/performance_test_basic.py
```diff
@@ -21,26 +21,144 @@
...
+        total = 0
+        for block in self.blockLog:
+            total += block.time
+        return total
+    def sum_latency(self):
+        total = 0
+        for block in self.blockLog:
+            total += block.latency
+        return total
+    def print_stats(self):
```
jgiszczak, line 66: Add a `__str__(self)` method which calls this method.  Then `chainsContainer` can simply `print(self.preData)` and `print(self.totalData)` as needed.

tests/performance_tests/performance_test_basic.py
```diff
@@ -21,26 +21,144 @@
...
+        return self.totalData.sum_cpu() - self.preData.sum_cpu()
+    def total_net(self):
+        return self.totalData.sum_net() - self.preData.sum_net()
+    def total_elapsed(self):
+        return self.totalData.sum_elapsed() - self.preData.sum_elapsed()
+    def total_time(self):
+        return self.totalData.sum_time() - self.preData.sum_time()
+    def total_latency(self):
+        return self.totalData.sum_latency() - self.preData.sum_latency()
+    def print_stats(self):
```
jgiszczak, line 92: Add a `__str__(self)` method here, as above.  Consider replacing `print_stats()` entirely with the special method name.  Then down at the bottom, you can simply call `print(cont)`.

tests/performance_tests/performance_test_basic.py
```diff
@@ -21,26 +21,144 @@
 relaunchTimeout = 30
 emptyBlockGoal = 5
 
-def waitForEmptyBlocks(node1):
+class blockData():
+    def __init__(self):
```
jgiszczak, line 25: Once converted to a dataclass, this manually written `__init__` can be removed.  Use the generated method, which accepts arguments, to set all of the members when the class is constructed and inserted into the `chainData` class, per @oschwaldp-oci 's review comment.

tests/performance_tests/performance_test_basic.py
jgiszczak: Definitely do this and eliminate all the boilerplate methods.

tests/performance_tests/performance_test_basic.py
```diff
@@ -21,26 +21,144 @@
...
+        print("Chain time: ", self.sum_time())
+        print("Chain latency: ", self.sum_latency())
+
+class chainsContainer():
+    def __init__(self):
+        self.preData = chainData()
+        self.totalData = chainData()
+        self.startBlock = 0
+        self.ceaseBlock = 0
+    def total_transactions(self):
```
jgiszczak, line 80: Reduce all the boilerplate methods into a single `total(self, attrname)` method which calls the built-in `getattr()` function and invoke it with each of the new cache attributes you'll be adding to `chainData`.

jgiszczak (changes requested):

tests/performance_tests/performance_test_basic.py
```diff
@@ -21,26 +21,144 @@
...
-def checkTotalTrx():
-    total = 0
-    f = open("var/lib/node_00/stderr.txt")
-    stringResult = re.findall(r'trxs:\s+\d+', f.read())
-    for value in stringResult:
-        intResult = re.findall(r'\d+', value)
-        total += int(intResult[0])
+def fetchStats(total):
+    i = -1
+    f = open("var/lib/node_01/stderr.txt")
```
jgiszczak, line 117: Use `with open("...") as f` idiom and eliminate the subsequently redundant explicit call to `close()`.

tests/performance_tests/performance_test_basic.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+from itertools import chain
```
jgiszczak, line 3: Remove unused import.

tests/performance_tests/performance_test_basic.py
```diff
@@ -14,33 +15,68 @@
 from Node import Node
 from Node import ReturnType
 from TestHelper import TestHelper
+from dataclasses import dataclass
+from typing import List
```
jgiszczak, line 19: Remove unused import.

tests/performance_tests/performance_test_basic.py
```diff
@@ -14,33 +15,68 @@
...
+        self.ceaseBlock = 0
+        self.totalTransactions = 0
+        self.totalNet = 0
+        self.totalCpu = 0
+        self.totalElapsed = 0
+        self.totalTime = 0
+        self.totalLatency = 0
+    def updateTotal(self, block):
+        print("Block is ", block)
+        self.totalTransactions += int(block[2])
```
jgiszczak, line 51: Remove redundant conversions to int.  Since all of these members have been initialized to integer values, anyone inadvertently passing in unconverted strings will promptly get a TypeError exception, not an unexpectedly long string.

jgiszczak (changes requested):

oschwaldp-oci (commented):
I'm happy when @jgiszczak is.

tests/performance_tests/performance_test_basic.py
```diff
@@ -14,33 +14,66 @@
...
-    f = open("var/lib/node_00/stderr.txt")
-    stringResult = re.findall(r'trxs:\s+\d+', f.read())
-    for value in stringResult:
-        intResult = re.findall(r'\d+', value)
-        total += int(intResult[0])
-    f.close()
-    return total
+def fetchStats(total):
+    with open("var/lib/node_01/stderr.txt") as f:
+        blockResult = re.findall(r'Received block ([0-9a-fA-F]*).* #(\d+) .*trxs: (\d+),.*, net: (\d+), cpu: (\d+), elapsed: (\d+), time: (\d+), latency: (-?\d+) ms', f.read())
```
heifner, line 72: This will not match 2.x.x or even 3.1.0 output; they do not contain `net`, or `cpu` or `elapsed` or `time`

jgiszczak (approved):
Approved pending Kevin's comment being addressed.

PR #45, opened 2022-08-24
Added unit tests for the tps_tester

ndcgundlach:
Added unit tests to the tps_tester to verify that it is maintaining or trying to maintain the desired rate.


tests/trx_generator/trx_provider.hpp
```diff
@@ -42,63 +42,64 @@ namespace eosio::testing {
...
 
             if (keep_running) {
-               fc::microseconds time_to_sleep{next_run - fc::time_point::now()};
-               if (time_to_sleep.count() > min_sleep_us) {
+               fc::microseconds time_to_sleep{stats.next_run - fc::time_point::now()};
+               if (time_to_sleep.count() >= min_sleep_us) {
+                  stats.last_sleep = time_to_sleep.count();
                   std::this_thread::sleep_for(std::chrono::microseconds(time_to_sleep.count()));
+               } else {
+                  stats.last_sleep = time_to_sleep.count();
```
heifner, line 102: This could be moved out of `if` and remove `else`.

tests/trx_generator/trx_provider.hpp
```diff
@@ -42,63 +42,64 @@ namespace eosio::testing {
...
 
-   struct null_trx_generator {
-      void generate_and_send() {}
-   };
 
-   constexpr int64_t min_sleep_us = 100;
+   constexpr int64_t min_sleep_us = 1;
 
    template<typename G, typename M>
    struct trx_tps_tester {
```
oschwaldp-oci, line 60: Are you anticipating using `trx_tps_tester` outside of the unit tests?  Should this be how we run and observe generation rate during actual perf runs?  I didn't see it currently incorporated anywhere in the main code path.

tests/trx_generator/trx_provider.hpp
```diff
@@ -42,63 +42,64 @@ namespace eosio::testing {
...
-      G _generator;
-      M _monitor;
+      std::shared_ptr<G> _generator;
+      std::shared_ptr<M>_monitor;
```
heifner, line 62: Why mock out a monitor? Seems like that is what you want to test. Also a space should be added before `_monitor`

heifner (changes requested):

tests/trx_generator/trx_generator_tests.cpp
```diff
@@ -0,0 +1,192 @@
...
+           ("rt", runtime_us.count())("mx", maximum_runtime_us ) );
+      BOOST_REQUIRE_LT(monitor->_calls.back().last_sleep, 0);
+   }
+}
+
+BOOST_AUTO_TEST_CASE(tps_med_run_med_tps_med_delay)
+{
+   constexpr uint32_t test_duration_s = 30;
+   constexpr uint32_t test_tps = 10000;
+   constexpr uint32_t trx_delay_us = 10;
```
oschwaldp-oci, line 132: Would it be valuable to add a test where the `trx_delay_us` is higher?

tests/trx_generator/trx_provider.hpp
ndcgundlach: Actually that is a copy pasta error. It should be: 
stats.last_sleep = 0

tests/trx_generator/trx_provider.hpp
ndcgundlach: This will be the piece that calls the refactored main in the trx_generator.

tests/trx_generator/trx_generator_tests.cpp
ndcgundlach: Yep, do you have a suggested delay?

oschwaldp-oci: I don't honestly.  Just wasn't sure if we should test multiple values just to have some variability in testing.

tests/trx_generator/trx_provider.hpp
ndcgundlach: Actually, I take that back.  I am using that value in the test to determine that the tps tester knows that it is behind schedule.  I updated the member name to properly reflect the meaning of the value.

tests/trx_generator/trx_generator_tests.cpp
```diff
@@ -0,0 +1,223 @@
...
+      ilog("couldn't sustain transaction rate.  ran ${rt}us vs expected max ${mx}us",
+           ("rt", runtime_us.count())("mx", maximum_runtime_us ) );
+      BOOST_REQUIRE_LT(monitor->_calls.back().time_to_next_trx_us, 0);
+   }
+}
+BOOST_AUTO_TEST_CASE(tps_med_run_med_tps_30us_delay)
+{
+   constexpr uint32_t test_duration_s = 15;
+   constexpr uint32_t test_tps = 3000;
+   constexpr uint32_t trx_delay_us = 10;
```
oschwaldp-oci, line 195: Should this be 30?

heifner (approved):

tests/trx_generator/trx_provider.hpp
```diff
@@ -42,64 +42,67 @@ namespace eosio::testing {
...
 
-            keep_running = (_monitor.monitor_test(stats) && stats.trxs_left);
+            keep_running = (_monitor->monitor_test(stats) && stats.trxs_left);
 
             if (keep_running) {
-               fc::microseconds time_to_sleep{next_run - fc::time_point::now()};
-               if (time_to_sleep.count() > min_sleep_us) {
+               fc::microseconds time_to_sleep{stats.next_run - fc::time_point::now()};
+               if (time_to_sleep.count() >= min_sleep_us) {
                   std::this_thread::sleep_for(std::chrono::microseconds(time_to_sleep.count()));
```
oschwaldp-oci, line 103: may need to include thread header

oschwaldp-oci (approved):

PR #50, opened 2022-08-25
Performance harness cluster configurable logging

ClaytonCalabrese:
It was desirable to be able to reduce the logging level in performance harness so that there were fewer impacts on throughput. As such, a new configuration option was added to cluster that sets logging level. Additionally, performance_test_basic's logging level has been set to "info".


oschwaldp-oci (approved):

programs/eosio-launcher/main.cpp
```diff
@@ -1146,6 +1152,18 @@ launcher_def::write_logging_config_file(tn_node_def &node) {
   if (!bfs::exists(dd)) {
     bfs::create_directories(dd);
   }
+  fc::log_level ll = fc::log_level::debug;
+  if (logging_level == "all") {
```
jgiszczak, line 1156: Use `from_variant()` free function available in `log_message.hpp` to convert from string.  (The necessary `#include` should already be present.)  `log_level` is serialized and deserialized as the human-readable string rather than the integer from the enum.

jgiszczak (changes requested):

Issue #48, opened 2022-08-25
http-max-bytes-in-flight-mb = 4096 fails, should generate config error on nodeos startup


matthewdarwin:
Followup from: https://github.com/AntelopeIO/leap/issues/46

`http-max-bytes-in-flight-mb = 4096` is too big, but `http-max-bytes-in-flight-mb = 4095` works fine.

If there is a maximum limit, there should be a hard fail.  Realistically setting it this big is a bit crazy.

```
$ cleos get info
error 2022-08-25T13:45:41.077 cleos     main.cpp:4121                 operator()           ] Failed with error: Too many bytes in flight: 897 (429)
```

stephenpdeos:
The team believes there are valid use cases for high values like this, so we'll seek to continue support and get to the root cause of this failure.

heifner:
Note any changes also need to be folded into https://github.com/AntelopeIO/leap/pull/22 if needed.

PR #59, opened 2022-08-26
[3.2] http-max-bytes-in-flight-mb fixes

linh2931:
Resolve https://github.com/AntelopeIO/leap/issues/48

- allow http-max-bytes-in-flight-mb bigger than 4GB
- make -1 indicate unlimited
- add range check

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -217,7 +217,7 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
          std::optional<eosio::chain::named_thread_pool>   thread_pool;
          std::atomic<size_t>                         bytes_in_flight{0};
          std::atomic<int32_t>                        requests_in_flight{0};
-         size_t                                      max_bytes_in_flight = 0;
+         uint64_t                                    max_bytes_in_flight = 0;
          int32_t                                     max_requests_in_flight = -1;
          fc::microseconds                            max_response_time{30*1000};
 
@@ -750,8 +750,8 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
              "Specify if Access-Control-Allow-Credentials: true should be returned on each request.")
             ("max-body-size", bpo::value<uint32_t>()->default_value(my->max_body_size),
              "The maximum body size in bytes allowed for incoming RPC requests")
-            ("http-max-bytes-in-flight-mb", bpo::value<uint32_t>()->default_value(500),
-             "Maximum size in megabytes http_plugin should use for processing http requests. 429 error response when exceeded." )
+            ("http-max-bytes-in-flight-mb", bpo::value<int64_t>()->default_value(500),
+             "Maximum size in megabytes http_plugin should use for processing http requests. -1 for unlimited. 429 error response when exceeded." )
             ("http-max-in-flight-requests", bpo::value<int32_t>()->default_value(-1),
              "Maximum number of requests http_plugin should use for processing http requests. 429 error response when exceeded." )
             ("http-max-response-time-ms", bpo::value<uint32_t>()->default_value(30),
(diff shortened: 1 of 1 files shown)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -217,7 +217,7 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
          std::optional<eosio::chain::named_thread_pool>   thread_pool;
          std::atomic<size_t>                         bytes_in_flight{0};
          std::atomic<int32_t>                        requests_in_flight{0};
-         size_t                                      max_bytes_in_flight = 0;
+         uint64_t                                    max_bytes_in_flight = 0;
```
heifner, line 220: This is later compared with a `size_t`, I think it is clearer as a `size_t`.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -776,7 +776,16 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
          EOS_ASSERT( my->thread_pool_size > 0, chain::plugin_config_exception,
                      "http-threads ${num} must be greater than 0", ("num", my->thread_pool_size));
 
-         my->max_bytes_in_flight = options.at( "http-max-bytes-in-flight-mb" ).as<uint32_t>() * 1024 * 1024;
+         // http-max-bytes-in-flight-mb is allowed to be -1,
+         // that's why int64_t is used
```
heifner, line 780: Would not mention this except I have the other comment. This could be on the same line or even just remove the second line of the comment.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -776,7 +776,16 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
...
+         auto max_bytes_mb = options.at( "http-max-bytes-in-flight-mb" ).as<int64_t>();
+         EOS_ASSERT( max_bytes_mb == -1 || (max_bytes_mb >= 0 && max_bytes_mb < std::numeric_limits<int64_t>::max() / (1024 * 1024)), chain::plugin_config_exception,
+                     "http-max-bytes-in-flight-mb ${max_bytes_mb} must be equal to or greater than 0 and less than ${max}, or -1", ("max_bytes_mb", max_bytes_mb) ("max", std::numeric_limits<int64_t>::max() / (1024 * 1024)) );
+         if ( max_bytes_mb == -1 ) {
+            my->max_bytes_in_flight = std::numeric_limits<uint64_t>::max();
```
heifner, line 785: This should change to `size_t` as well.

heifner (approved):

spoonincode:
This arguably could be a 3.1 candidate.

plugins/http_plugin/http_plugin.cpp
linh2931: Thanks. I will change back to size_t (was thinking make it explicitly 64 bits)

plugins/http_plugin/http_plugin.cpp
linh2931: Will change.

linh2931:
> 

If we do this for 3.1, should I close this PR, do 3.1 first, and then merge 3.1 to main?

spoonincode:
I'm on the fence about it. #48 does make it sound like it could be treated as a defect since setting it to a high value silently ends up breaking HTTP by not behaving as expected. I'm merely suggesting it could be considered for 3.1 I'm not sure I have a strong feeling either way.

linh2931:
Since it is not a big effort and prevents a potentially broken HTTP, I will do a 3.1 and then main.

linh2931:
Close this. Use https://github.com/AntelopeIO/leap/pull/64 instead, which is merged from the fix in release 3.1 https://github.com/AntelopeIO/leap/pull/63.

PR #64, opened 2022-08-27
[3.1 -> main] fix a number of issues with http-max-bytes-in-flight-mb configuration

linh2931:
Resolve https://github.com/AntelopeIO/leap/issues/48

- allow `http-max-bytes-in-flight-mb` bigger than 4GB
- make `-1` indicate unlimited
- validate value of `http-max-bytes-in-flight-mb`


spoonincode (approved):

linh2931:
This is merged from https://github.com/AntelopeIO/leap/pull/63

PR #71, opened 2022-08-30
[3.2] replace git commands from nodeos' & launcher's cmake with libversion usage

spoonincode:
libversion is the canonical method by which components should be retrieving the build version. But both nodeos' & eosio-launcher's CMakeLists were running git to retrieve a version. Remove that and replace with libversion usage instead. libversion needed a method added to retrieve the git hash as nodeos' get_info server_version returns the first 4 bytes of the git hash.

### DIFF
--- a/libraries/version/src/version.cpp
+++ b/libraries/version/src/version.cpp
@@ -12,4 +12,9 @@ namespace eosio { namespace version {
       return version;
    }
 
+   const std::string& version_hash() {
+      static const std::string vhash{_version_hash()};
+      return vhash;
+   }
+
 } }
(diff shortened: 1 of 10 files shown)


libraries/version/src/version.cpp
```diff
@@ -12,4 +12,9 @@ namespace eosio { namespace version {
       return version;
    }
 
+   const std::string& version_hash() {
+      static const std::string vhash{_version_hash()};
```
heifner, line 16: Seems like an odd thing to keep around for the life of the application. Looks like a copy paste from above, but the same thing goes for those as well.

spoonincode: Yeah I was just following how it was done before, it does seem unnecessary

spoonincode: @heifner is refactoring libversion something you want to see done?

heifner: Not a huge priority, so this is fine for now.

heifner (approved):

PR #79, opened 2022-08-31
[3.2] Fix documentation for get_scheduled_transactions in plugins/chain_api_plugin/chain.swagger.yaml

pnx:
Rename `get_scheduled_transaction` to `get_scheduled_transactions` (notice the trailing "s") as `get_scheduled_transaction` does not exist.

Also fix the description.


heifner (approved):
Thanks.

PR #100, opened 2022-09-01
Refactor and Separate Python Script for Performance Harness

ClaytonCalabrese:
This PR splits the performance_test_basic script into 3 parts.
The first is the test itself, performance_test_basic.py
The second is generic helper functions, log_reader.py
The third is a script for processing a passed nodeos log into the data structures from log_reader.py.

There is also a bugfix to how range is calculated to capture that the top of range is not inclusive.


tests/performance_tests/read_log_data.py
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env python3
+ 
+import argparse
+import log_reader
+
+parser = argparse.ArgumentParser(add_help=False)
+parser.add_argument("log_path", type=str, help="Path to nodeos log to scrape")
+parser.add_argument("--start_block", type=int, help="First significant block number in the log", default=2)
```
oschwaldp-oci, line 8: Since our common use pattern in `performance_test_basic.py` has us specifying `startBlock` and `ceaseBlock` it might be good to go ahead an mirror that into this script by providing the `--cease_block` as well.

oschwaldp-oci (changes requested):

oschwaldp-oci (approved):

oschwaldp-oci (approved):

oschwaldp-oci (approved):

PR #106, opened 2022-09-06
[3.1] Fix test failure due to trx hitting near end of block production time

### LINKED ISSUE #105: Investigate trx_finality_status_test failure
https://github.com/AntelopeIO/leap/runs/8138707376?check_suite_focus=true

heifner:
It is possible for a trx to arrive at the BP node a few milliseconds before block signing which results in an almost immediate inclusion in the block. The test was not expecting the trx to make it to a block so quickly.

Resolves #105 
Resolves https://github.com/eosnetworkfoundation/mandel/issues/488

### DIFF
--- a/tests/trx_finality_status_test.py
+++ b/tests/trx_finality_status_test.py
@@ -136,8 +136,8 @@ try:
     postInfo = None
     transferAmount=10
 
-    # ensuring that prod0's producer is active, which will give sufficient time to identify the transaction as "LOCALLY_APPLIED" before it travels
-    # through the chain of nodes to node0 to be added to a block
+    # Ensuring that prod0's producer is active, which will give more time to identify the transaction as "LOCALLY_APPLIED" before it travels
+    # through the chain of nodes to node0 to be added to a block. It is still possible to hit the end of a block and state be IN_BLOCK here.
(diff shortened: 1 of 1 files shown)


linh2931 (approved):

oschwaldp-oci (commented):
It looks like the intent here was to be able to validate a `LOCALLY_APPLIED` transaction finality lifecycle event occurs on a transaction.  With the change herein, it looks like that state check will not be enforced.  Should the test be altered to more deterministically force the `LOCALLY_APPLIED` state transition?

heifner:
> It looks like the intent here was to be able to validate a `LOCALLY_APPLIED` transaction finality lifecycle event occurs on a transaction. With the change herein, it looks like that state check will not be enforced. Should the test be altered to more deterministically force the `LOCALLY_APPLIED` state transition?

It is unclear how that could be done 100% with this test. The `trx_finality_status_forked_test.py` I think does this better in a more deterministic way. Also, `IN_BLOCK` is perfectly valid state for this trx, so I think it makes sense to allow for either `LOCALLY_APPLIED` or `IN_BLOCK` at this check.

oschwaldp-oci (approved):

PR #111, opened 2022-09-07
[3.2] desubmodule wasm-spec-tests

spoonincode:
Desubmodules AntelopeIO/wasm-spec-tests@112c682. There are no other users of this repository across AntelopeIO and it's not a separate released product, so we've decided to desubmodule its contents in to leap's repository directly.


heifner:
Consider renaming so switching between branches is less of a headache.

heifner (approved):

PR #120, opened 2022-09-08
CPACK: Improve deb package filenames

pnx:
This PR fixes the format for deb package filenames to follow the standard debian convention (described [here](https://www.debian.org/doc/manuals/debian-faq/pkg-basics.en.html#pkgname) in detail): `<PackageName>_<VersionNumber>_<DebianArchitecture>.deb`

What is being  changed in the original package name is:
* First hyphen is changed to underscore. (right after the package name)
* Last hyphen before the architecture is changed to underscore.
* Use `dpkg --print-architecture` to get the correct value (e.g. `amd64` instead of `x86_64` provided by cmake's `${CMAKE_SYSTEM_PROCESSOR}`)


package.cmake
```diff
@@ -21,6 +21,20 @@ if(CMAKE_VERSION VERSION_GREATER_EQUAL "3.9" AND EXISTS /etc/os-release)
       string(APPEND CPACK_PACKAGE_FILE_NAME "-el${CMAKE_MATCH_1}")
    endif()
 endif()
+
+# Fix debian package filename as it should have the format:
+# <PackageName>_<VersionNumber>_<DebianArchitecture>.deb
+
+# Find architecture using dpkg
```
spoonincode, line 28: I believe this is already available via CMake's `CPACK_DEBIAN_PACKAGE_ARCHITECTURE` variable. Can you try using that instead?

spoonincode (changes requested):

package.cmake
pnx: The problem with `CPACK_DEBIAN_PACKAGE_ARCHITECTURE` is that it should be set in the script. cpack will use `dpkg --print-architecture` only if its empty, after it has parsed `package.cmake`

We also set `CPACK_DEBIAN_<COMPONENT>_FILE_NAME` filename ourself. so cpack has no way to inject it's default architecture value into the filename.

i could however set `CPACK_DEBIAN_PACKAGE_ARCHITECTURE` instead of `DEB_ARCH`.

spoonincode (approved):

PR #128, opened 2022-09-09
[3.1] Test: verify on the correct fork before verifying trx status

### LINKED ISSUE #127: Investigate trx_finality_status_forked_test failure
https://github.com/AntelopeIO/leap/runs/8272427791?check_suite_focus=true

heifner:
The `trx_finality_status_forked_test.py` test purposely introduces a fork to verify trx is forked out and reports the correct state. The failure was for a case when another fork happened after the first fork delaying the time for the trx to reach irreversible.

Add verification to the test that the trx is on the expected fork before verifying it is irreversible.

Resolves #127

### DIFF
--- a/tests/trx_finality_status_forked_test.py
+++ b/tests/trx_finality_status_forked_test.py
@@ -240,6 +240,7 @@ try:
         f"\n\nprod 0 info: {json.dumps(prodNodes[0].getInfo(), indent=1)}\n\nprod 1 info: {json.dumps(prodNodes[1].getInfo(), indent=1)}"
 
     afterForkInBlockState = retStatus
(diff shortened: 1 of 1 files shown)


ClaytonCalabrese (approved):

oschwaldp-oci (approved):

spoonincode:
Likely fixes eosnetworkfoundation/mandel#616 you think?

heifner:
Hmm, no that looks like something different. Do we have logs for that? Or can we get logs for a new run?

spoonincode:
I can try some time to reproduce. No rush I will keep that issue open.

PR #129, opened 2022-09-09
Nodeos log scraping test

ClaytonCalabrese:
Adds a test to ensure that Nodeos_log_scraping is done properly in log_reader.  This is done by unzipping the two attached logs and checking them verses expected values. If this test starts to fail, it is an indicator that log_reader has been changed in a way that will not work with an old version of logging.


tests/performance_tests/nodeos_log_scraping_test.py
```diff
@@ -0,0 +1,44 @@
...
+# Also ensures that all versions of nodeos logs can be handled
+import log_reader
+
+testSuccessful = False
+
+# Test log scraping for current log format
+dataCurrent = log_reader.chainData()
+dataCurrent.startBlock = None
+dataCurrent.ceaseBlock = None
+log_reader.scrapeLog(dataCurrent, "tests/performance_tests/sample_nodeos_log.txt.gz")
```
oschwaldp-oci, line 12: I would recommend naming the nodeos log file w/version information.

tests/performance_tests/nodeos_log_scraping_test.py
```diff
@@ -0,0 +1,44 @@
...
+expectedCurrent.totalLatency = -9398
+
+dataCurrent.assertEquality(expectedCurrent)
+
+
+# Test log scraping from a 2.0.14 log format
+dataOld = log_reader.chainData()
+dataOld.startBlock = None
+dataOld.ceaseBlock = None
+log_reader.scrapeLog(dataOld, "tests/performance_tests/sample_nodeos_old_log.txt.gz")
```
oschwaldp-oci, line 31: I would recommend naming the nodeos log file w/version information.

tests/performance_tests/nodeos_log_scraping_test.py
```diff
@@ -0,0 +1,44 @@
...
+expectedOld = log_reader.chainData()
+expectedOld.startBlock = 2
+expectedOld.ceaseBlock = 93
+expectedOld.totalTransactions = 129
+# Net, Cpu, Elapsed, and Time are not logged in the old logging and will thus be 0
```
oschwaldp-oci, line 36: In addition to the comment, would it be better to explicitly set Net, Cpu, Elapsed, and Time on the `expectedOld` object to codify the expected results so that, if someone were to inadvertently replace the old version log file with a newer version and missed the comment they wouldn't simply thing these values were missed and add them?

oschwaldp-oci (approved):

PR #145, opened 2022-09-12
[3.2] Log block trx summary for speculative and BP blocks

### LINKED ISSUE #119: Add more logging of subjective information on non-producing nodes
On a block producing node we have many details logged when producing a block (with debug logging enabled).

To aid in debugging of network issues in non-producing nodes, it would be great if there as some sort of summary per speculative block of how many trxs failed and for what reason and for which accounts.

To ease log scraping, the log format between producing and non-producing nodes should be the same, in an ideal world.

heifner:
Add additional logging to report transaction execution time per block including speculative blocks. Also report block idle time waiting on incoming transactions.

The output is enabled by `producer_plugin` logger at `debug` level.

Example output:
```
debug 2022-09-12T18:47:37.241 test_trx_ producer_plugin.cpp:187       report               ] Block trx idle: 231428us out of 241425us, success: 1128, 6622us, fail: 174, 2321us, other: 1054us
```

* The `Block trx idle:` is time the node is waiting for trxs to come into a node and be scheduled for execution since the end of the last trx processed.
* The `out of` is the time since the last block completion. Should be close to the configured CPU Effort for a given block. For example an empty block with 80% CPU effort `Block trx idle: 400351us out of 400541us, success: 0, 0us, fail: 0, 0us, other: 190us`. The complete "idle" time for this block would be 500ms. The idle time of the CPU effort is not reported. The value should be thought of as the time the node was able according to configuration to process trxs.
* `success: {n}, {t}us` is the number of successful transactions processed during this block and how long they took to execute.
* `fail: {n}, {t}us` is the number of transactions that failed processing during this block and how long they took to execute before they failed.
* `other:` is the time of (`out of` - `Block trx idle` - `success time` - `fail time`). The time nodeos is logging or doing other book keeping. It is the block time not accounted for in the other reported times.

Resolves #119

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -151,10 +151,12 @@ public:
 
    void add_fail_time( const fc::microseconds& fail_time ) {
       trx_fail_time += fail_time;
+      ++trx_fail_num;
    }
 
    void add_success_time( const fc::microseconds& time ) {
       trx_success_time += time;
+      ++trx_success_num;
    }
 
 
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -226,6 +249,12 @@ class account_failures {
 
    std::map<account_name, account_failure> failed_accounts;
    uint32_t max_failures_per_account = 3;
+   mutable fc::microseconds block_idle_time;
```
spoonincode, line 252: Quite unusual to see this. How bad is it to unwind the const train to not make this mutable? A cursory look, seems like `report() const` is called from both `abort_block()` & `produce_block()` (neither of which are const already)

heifner: Thanks. Removed mutable/const which is much cleaner.

spoonincode (approved):

PR #147, opened 2022-09-12
Add TPS scoring

oschwaldp-oci:
TPS scoring herein includes min, max, average, and standard deviation using measurements of pairs of consecutive blocks over a defined window of blocks during a testing run.

### DIFF
--- a/tests/performance_tests/log_reader.py
+++ b/tests/performance_tests/log_reader.py
@@ -3,6 +3,7 @@
 import os
 import sys
 import re
+import numpy
 
 harnessPath = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sys.path.append(harnessPath)
@@ -13,6 +14,13 @@ from dataclasses import dataclass
 Print = Utils.Print
 errorExit = Utils.errorExit
 cmdError = Utils.cmdError
+
+@dataclass
+class stats():
+    min: int = 0
+    max: int = 0
+    avg: int = 0
+    sigma: int = 0
 
 @dataclass
 class blockData():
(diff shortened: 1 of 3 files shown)


tests/performance_tests/log_reader.py
```diff
@@ -70,3 +78,51 @@ def scrapeLog(total, path):
                         total.updateTotal(int(value[2]), 0, 0, 0, 0, int(v2Logging[0]))
                 else:
                     print("Error: Unknown log format")
+
+def findPrunedRangeOfInterest(data: chainData, numAddlBlocksToDrop=0):
```
jgiszczak, line 82: This method can be eliminated in favor of constructing the list to iterate in place in the calling function in one line:
```
filtered = filter(lambda block: block.transactions != 0, data.blockLog)
```
and then iterating it as:
```
for first, second in zip(*[iter(filtered[numAddlBlocksToDrop:-numAddlBlocksToDrop])]*2):
    consecutiveBlockTps.append(first.transactions + second.transactions)
```
which has the benefit of not double-counting transactions as the current code does.

The `zip(*[iter(sequenceToIterate)]*n)` idiom to step through an iterable in groups of `n` items is documented in the Python documentation for the `zip()` function, so it's fairly common to see in code which doesn't use `more_itertools`.  The additional wrinkle here is just slice notation to chop off additional blocks to drop at the beginning and end of the list.

Generally speaking, you don't usually need to use container indexes in Python code.

tests/performance_tests/log_reader.py
```diff
@@ -70,3 +78,51 @@ def scrapeLog(total, path):
...
+
+    if startBlockIndex >= endBlockIndex:
+        print(f"Error: Invalid block index range start: {startBlockIndex} end: {endBlockIndex}")
+        return stats()
+
+    for i in range(startBlockIndex, endBlockIndex+1):
+        if i + 1 < endBlockIndex:
+            consecutiveBlockTps.append(data.blockLog[i].transactions + data.blockLog[i+1].transactions)
+
+    return stats(numpy.min(consecutiveBlockTps) , numpy.max(consecutiveBlockTps), numpy.average(consecutiveBlockTps), numpy.std(consecutiveBlockTps))
```
jgiszczak, line 128: I suggest including in the statistics the number of blocks considered in the statistics and the number of blocks with zero transactions as a check.  My suggested implementation will omit blocks with zero transactions which occur in the middle of the test run.  As implemented today, none should exist, but that could change.  Having the total block count and the empty block count should be generally useful for seeing if a run behaved as expected.

jgiszczak (changes requested):

tests/performance_tests/log_reader.py
```diff
@@ -70,3 +78,51 @@ def scrapeLog(total, path):
...
+    startBlockIndex,endBlockIndex = findPrunedRangeOfInterest(data, numAddlBlocksToDrop)
+
+    if startBlockIndex >= endBlockIndex:
+        print(f"Error: Invalid block index range start: {startBlockIndex} end: {endBlockIndex}")
+        return stats()
+
+    for i in range(startBlockIndex, endBlockIndex+1):
+        if i + 1 < endBlockIndex:
+            consecutiveBlockTps.append(data.blockLog[i].transactions + data.blockLog[i+1].transactions)
+
```
jgiszczak, line 127: I think you'd prefer to do a single conversion from Python list to numpy array, so rather than passing `consecutiveBlockTps` to each numpy function: `npTps = numpy.array(consecutiveBlockTps, dtype=numpy.uint32)` and pass `npTps` to each function.  I don't think the interpreter can be depended upon to make that optimization for you.

tests/performance_tests/log_reader.py
oschwaldp-oci: The additional data points in the `stats` is a great idea.  I'll add that.

`As implemented today, none should exist` -- I disagree here, and that is why it is implemented the way it is. Blocks with zero transactions in the middle of a test run should not be filtered out as that is a very interesting data point to consider.

tests/performance_tests/log_reader.py
oschwaldp-oci: Good point. I'll update.

jgiszczak (approved):

PR #150, opened 2022-09-13
Build framework for leap-util

766C6164:
This is initial proposed implementation of leap-util utility comprising following features:
- CLI11 cli parsing library, patched to support auto-completion
- partition subcommand handlers by separate files
- added blocklog implementation eosio-blocklog as subcommand "blockllog"
- included bash autocompletion configuration script for leap-util (its tested on Ubuntu and FreeBSD and works fine)
- moved cli11 to interface only lib, bot cleos and leap-util use it  now

This PR still work in progress and things will change, notes about changes will be added in comments.


spoonincode:
With both cleos & leap-util using CLI11, does it make sense to have just one instance of it? Does it make sense to have that one instance be a submodule from upstream?

766C6164:
> With both cleos & leap-util using CLI11, does it make sense to have just one instance of it? Does it make sense to have that one instance be a submodule from upstream?

Yes! This is the plan. And cleos will get autocompletion and --help-all for free.

programs/leap-cli11/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+add_library(leap-cli11 INTERFACE)
+include_directories(${CMAKE_CURRENT_SOURCE_DIR}/include)
+target_include_directories(leap-cli11
```
spoonincode, line 3: You probably just need `target_include_directories(leap-cli11 INTERFACE include)` here (current source dir is implied). not seeing off hand why you need `include_directories()` either

766C6164: Thank you, fixed, as you said, it works perfectly with just `target_include_directories(leap-cli11 INTERFACE include)`

programs/leap-util/actions_blocklog.cpp
```diff
@@ -0,0 +1,339 @@
...
+      const auto duration = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::high_resolution_clock::now() - _start).count() / 1000;
+      ilog("eosio-blocklog - ${desc} took ${t} msec", ("desc", _desc)("t", duration));
+   }
+
+   const std::chrono::high_resolution_clock::time_point _start;
+   const std::string _desc;
+};
+
+
+void BlocklogActions::setup(CLI::App& app) {
```
heifner, line 49: Here, and throughout, use snake_case instead of CamelCase.

766C6164: Done, everything snake_cased, thank you!

766C6164:
* added readme to leap-CLI11
* genesys subcommand moved to blocklog
* blocklog -> block_log
* chain- > chain_state
* disabled snapshot subcommand
* refactored blocklog opts to be subcommands

programs/CMakeLists.txt
```diff
@@ -1,5 +1,7 @@
+add_subdirectory( leap-cli11 )
```
linh2931, line 1: Should `leap-cli11` be in `libraries` directory? It is not a program.

programs/leap-util/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+target_link_libraries( leap-util
+        PRIVATE appbase version
+        PRIVATE eosio_chain chain_plugin fc leap-cli11 producer_plugin ${CMAKE_DL_LIBS} ${PLATFORM_SPECIFIC_LIBS} )
+
+copy_bin( leap-util )
+install( TARGETS
+   leap-util
+
+   COMPONENT base
```
spoonincode, line 17: I'm not sure this is in the right location for `COMPONENT`, does it work like this? Something similar to nodeos' cmake `install()` is probably all that is required here since it's only `RUNTIME` I think

programs/leap-cli11/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_library(leap-cli11 INTERFACE)
+target_include_directories(leap-cli11 INTERFACE include)
```
spoonincode, line 2: since `leap-cli11` is just a library, would it be a better fit in `libraries/` instead of `programs/`?

programs/leap-cli11/README.md
```diff
@@ -0,0 +1,26 @@
+
+## instructions for building custom leap-cli11 library
+
+leap-cli11 interface only library created in order to simplify integration of modified version of command line parsing library cli11
+
+to update included in this library include/cli11/CLI11.hpp file it needs to be (re)geenerated from repository containing forked/modified version of it, in particular:
+
+
+```bash
+git clone git@github.com:AntelopeIO/CLI11.git
```
spoonincode, line 10: since it's a public repo I'd recommend changing to https://

programs/leap-cli11/include/cli11/CLI11.hpp
```diff
@@ -1,20 +1,16 @@
-#pragma once
-
-// CLI11: Version 1.9.0
+// CLI11: Version 2.2.0
 // Originally designed by Henry Schreiner
 // https://github.com/CLIUtils/CLI11
 //
 // This is a standalone header file generated by MakeSingleHeader.py in CLI11/scripts
-// from: v1.9.0
+// from: 4c832a2
```
spoonincode, line 6: this is not the same version as in AntelopeIO/CLI11's main

spoonincode:
If I just run `leap-util chain-state` nothing is printed, and exit status is 0. This seems to apply to some other invocations too, like `leap-util chain-state build-info` which does nothing and has 0 exit status.

programs/leap-util/actions/chain.cpp
```diff
@@ -0,0 +1,88 @@
...
+   if(!opt->build_output_file.empty()) {
+      bfs::path p = opt->build_output_file;
+      if(p.is_relative()) {
+         p = bfs::current_path() / p;
+      }
+      fc::json::save_to_file(chainbase::environment(), p, true);
+      ilog("Saved build info JSON to '${path}'", ("path", p.generic_string()));
+   }
+   if(opt->build_just_print) {
+      ilog("\nBuild environment JSON:\n${e}", ("e", fc::json::to_pretty_string(chainbase::environment())));
```
spoonincode, line 84: imo leap-util should be more "CLI toolish" and avoid the "log spam" that comes along with using fc log. For example, as a standalone tool I'd expect to be able to do something like `leap-util chain-state build-info -p | jq .boost_version` but that's not possible with fc logging getting in the way.

spoonincode:
I know we were just aiming for 1:1 of current functionality, but there are some missed opportunities to improve the commands to make operator lifes' easier. Just based on a cursory look two things stood out:

#### build info
 `build-info` command is missing a very useful ability to probe what the environment is for an existing state file (as is, it only prints what `leap-util` was compiled with).

For example, I could image an operator trying to do something like
```
if [[ $(leap-util chain-state build-info -p) != $(leap-util chain-state --state-file /var/nodeos/shared_state.bin build-info -p) ]]; then
  echo "uh oh, looks like next time I launch nodeos I'll need a snapshot" && exit 1
fi
```

(the above assumes some canonical string representation of the environment, which might be kinda bad, and doesn't capture that in addtion to environment there could be other reasons the state isn't compatible). Maybe instead, just as an off the cuff thought,  there would be a
```
if ! leap-util chain-state is-compatible /var/nodeos/shared_state.bin; then
  echo "uh oh, looks like next time I launch nodeos I'll need a snapshot" && exit 1
fi
```

#### block log query
The `block-log` command could use a way to query properties of an existing block log. Like `--block-log-version` or `--block-log-first-block` or `--block-log-chain-id` or such.

spoonincode:
when I just run `leap-util block-log genesis` some unexpected behavior occurs
```
./leap-util block-log genesis
warn  2022-09-29T22:08:26.313 leap-util blocklog.cpp:164              do_genesis           ] No blocks.log found at '/home/spoon/f/leap/build.empty/programs/leap-util/blocks/blocks.log'. Using default genesis state.
info  2022-09-29T22:08:26.313 leap-util blocklog.cpp:171              do_genesis           ] Genesis JSON:
{
  "initial_timestamp": "2018-06-01T12:00:00.000",
  "initial_key": "EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV",
  "initial_configuration": {
    "max_block_net_usage": 1048576,
    "target_block_net_usage_pct": 1000,
    "max_transaction_net_usage": 524288,
    "base_per_transaction_net_usage": 12,
    "net_usage_leeway": 500,
    "context_free_discount_net_usage_num": 20,
    "context_free_discount_net_usage_den": 100,
    "max_block_cpu_usage": 200000,
    "target_block_cpu_usage_pct": 1000,
    "max_transaction_cpu_usage": 150000,
    "min_transaction_cpu_usage": 100,
    "max_transaction_lifetime": 3600,
    "deferred_trx_expiration_window": 600,
    "max_transaction_delay": 3888000,
    "max_inline_action_size": 524288,
    "max_inline_action_depth": 4,
    "max_authority_depth": 6
  }
}
error 2022-09-29T22:08:26.314 leap-util blocklog.cpp:138              run_subcommand       ] 3190000 block_log_exception: Block log exception
No blocks found in block log
    {}
    leap-util  blocklog.cpp:293 read_log

```
That alone is not what I'd expect. And then if I run it again...
```
./leap-util block-log genesis
error 2022-09-29T22:09:02.493 leap-util blocklog.cpp:138              run_subcommand       ] 3190001 block_log_unsupported_version: unsupported version of block log
Unsupported version of block log. Block log version is 0 while code supports version(s) [1,3]
    {"version":0,"min":1,"max":3}
    leap-util  block_log.cpp:1026 extract_chain_context
```

The tool made a new `blocks` directory. Something like the `block-log genesis` command should be read-only.

programs/leap-cli11/bash-completion/completions/cleos
```diff
@@ -0,0 +1,26 @@
+#/usr/bin/env bash
+
+_cleos_complete()
+{
+    # Get cmdline while ignoring the last word
+    cmdline=${COMP_LINE}
+    if [[ "${cmdline:${cmdlinex}-1}" != " " ]]; then
```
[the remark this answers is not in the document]
766C6164: Thats sort of bash hack I brought in, replaced with ${cmdline:(-1)} which means the same and do not rely on whitespace. Thank you!

programs/leap-cli11/bash-completion/completions/leap-util
```diff
@@ -0,0 +1,26 @@
+#/usr/bin/env bash
+
+_leap_complete()
+{
+    # Get cmdline while ignoring the last word
+    cmdline=${COMP_LINE}
+    if [[ "${cmdline:${cmdlinex}-1}" != " " ]]; then
```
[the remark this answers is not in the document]
766C6164: Same as above, ty!

programs/leap-util/actions/blocklog.cpp
```diff
@@ -0,0 +1,383 @@
...
+   sub->add_option("--first,-f", opt->first_block, "The first block number to log or the first to keep if trim-blocklog.");
+   sub->add_option("--last,-l", opt->last_block, "The last block number to log or the last to keep if trim-blocklog.");
+   sub->add_option("--output-dir", opt->output_dir, "The output directory for the block log extracted from blocks-dir.");
+
+   // flags
+   sub->add_flag("--no-pretty-print", opt->no_pretty_print, "Do not pretty print the output.  Useful if piping to jq to improve performance.");
+   sub->add_flag("--as-json-array", opt->as_json_array, "Print out json blocks wrapped in json array (otherwise the output is free-standing json objects).");
+
+   // subcommands
+   sub->add_subcommand("make-index", "    Create blocks.index from blocks.log. Must give 'blocks-dir'. Give 'output-file' relative to current directory or absolute path (default is <blocks-dir>/blocks.index).")->callback([this, cb]() {opt->make_index=true; cb(); });
```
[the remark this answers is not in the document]
766C6164: Good catch, ty!

programs/leap-util/actions/blocklog.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+   // options
+   sub->add_option("--blocks-dir", opt->blocks_dir, "The location of the blocks directory (absolute path or relative to the current directory).");
+   sub->add_option("--output-file,-o", opt->output_file, "The file to write the output to (absolute or relative path).  If not specified then output is to stdout.");
+   sub->add_option("--first,-f", opt->first_block, "The first block number to log or the first to keep if trim-blocklog.");
+   sub->add_option("--last,-l", opt->last_block, "The last block number to log or the last to keep if trim-blocklog.");
+   sub->add_option("--output-dir", opt->output_dir, "The output directory for the block log extracted from blocks-dir.");
+
+   // flags
+   sub->add_flag("--no-pretty-print", opt->no_pretty_print, "Do not pretty print the output.  Useful if piping to jq to improve performance.");
```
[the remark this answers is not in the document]
766C6164: Boolean flags are always false if not set by default (https://cliutils.github.io/CLI11/book/chapters/flags.html)

programs/leap-util/actions/blocklog.hpp
```diff
@@ -0,0 +1,46 @@
...
+struct blocklog_options {
+   std::string blocks_dir = "blocks";
+   std::string output_file = "";
+   int first_block = 0;
+   int last_block = std::numeric_limits<uint32_t>::max();
+   std::string output_dir;
+
+   bool no_pretty_print;
+   bool as_json_array;
+   bool make_index;
```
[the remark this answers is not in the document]
766C6164: done, ty

programs/leap-util/actions/blocklog.hpp
```diff
@@ -0,0 +1,46 @@
...
+using namespace eosio::chain;
+
+struct blocklog_options {
+   std::string blocks_dir = "blocks";
+   std::string output_file = "";
+   int first_block = 0;
+   int last_block = std::numeric_limits<uint32_t>::max();
+   std::string output_dir;
+
+   bool no_pretty_print;
```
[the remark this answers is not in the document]
766C6164: Put false for read-ability, but I believe it's reduntant to cli11 handling of flags, which will be false by default

programs/leap-util/main.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "actions/snapshot.hpp"
+
+#include <memory>
+
+#include "leap_formatter.hpp"
+
+int main(int argc, char** argv) {
+   fc::logger::get(DEFAULT_LOGGER).set_log_level(fc::log_level::debug);
+
+   CLI::App app{"Command Line Leap Utility"};
```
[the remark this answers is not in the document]
766C6164: Changed :) Thank you!

programs/leap-util/actions/chain.cpp
```diff
@@ -0,0 +1,88 @@
...
+#include <eosio/chain/global_property_object.hpp>
+#include <eosio/chain/permission_link_object.hpp>
+#include <eosio/chain/resource_limits.hpp>
+#include <eosio/chain/snapshot.hpp>
+#include <eosio/chain/wasm_interface.hpp>
+#include <eosio/chain_plugin/chain_plugin.hpp>
+#include <eosio/chain_plugin/trx_finality_status_processing.hpp>
+#include <eosio/chain_plugin/trx_retry_db.hpp>
+#include <eosio/producer_plugin/producer_plugin.hpp>
+#include <fc/io/json.hpp>
```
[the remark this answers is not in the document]
766C6164: Thank you, cleaned up a bunch of headers...

programs/leap-util/CMakeLists.txt
766C6164: Thank you! Changed.

programs/leap-cli11/CMakeLists.txt
766C6164: Totally makes sense, moving... Thank you!

programs/leap-cli11/README.md
766C6164: Changed, ty!

programs/leap-cli11/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+Resulting single-header will be located in:
+
+```cpp
+build/include/CLI11.hpp
+```
+
+And is ready to be copied to include/cli11/CLI11.hpp of leap-cli11 library
+
+Automated CLI11 subproject build / import of CLI11.hpp header will be added in feature versions.
```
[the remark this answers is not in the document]
766C6164: Yes, fixed the typo. Ty!

programs/leap-util/actions/chain.cpp
766C6164: Both done - added constraints to chain-state, ditched ilog, now jq works :) Thank you!

programs/leap-cli11/include/cli11/CLI11.hpp
766C6164: Updated, thank you! It's 1d3b0a7 now.

programs/leap-util/actions/blocklog.hpp
```diff
@@ -0,0 +1,49 @@
...
+   bool no_pretty_print = false;
+   bool as_json_array = false;
+
+   // subcommands
+   bool make_index;
+   bool trim_blocklog;
+   bool extract_blocks;
+   bool smoke_test;
+   bool vacuum;
+   bool genesis;
```
heifner, line 27: I think you should initialize these all to `false`. Some are set in callbacks which if not called leaves them un-initialized causing undefined behavior.

766C6164: Thank you, good suggestion, done!

programs/leap-util/actions/snapshot.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <boost/exception/diagnostic_information.hpp>
+#include <boost/filesystem.hpp>
+#include <boost/filesystem/path.hpp>
+
+void snapshot_actions::setup(CLI::App& app) {
+   auto* sub = app.add_subcommand("snapshot", "snapshot utility");
+   sub->add_subcommand("to-json", "convert snapshot file to convert to json format");
+
+   // options
+   sub->add_option("--input-file,-i", opt->input_file, "snapshot file to convert to json format, writes to <file>.json if output file not specified (tmp state dir used), and exit.")->required();
```
spoonincode, line 18: I don't quite follow what this description of the option is trying to tell me. It seems to be saying if I give an input file without an output file, the output file will be basenameofinput.json; but that's in conflict of the `output-file` description which says no output file means output goes to stdout. And I don't know what "tmp state dir" means in this context.

More generally, since the options are attached to `snapshot`, not `to-json`,  the descriptions are probably too narrowly defined solely for the `to-json` use case. There are a number of useful operations on a snapshot file we can add in the future that don't have anything to do with json.

linh2931 (approved):
Blocklog needs more refactoring like specific options for a subcommand, ... @766C6164 and talked about them.

spoonincode:
I wonder if `bash-completion` package should be added to `CPACK_DEBIAN_BASE_PACKAGE_SUGGESTS`. Not sure if that's a typical thing for debian/ubuntu packages to do. Not critical for now.

larryk85 (approved):

PR #160, opened 2022-09-14
Update Python Perf Harness to run CI/CD configuration.

oschwaldp-oci:
Make duration and tps configurable to python test script.
Update defaults to be in line with expected typical CICD run

### Base Test Case (CI/CD)
This configuration gets run as test in CI/CD.

- number of seconds to run → 30
- target tps → 1000
- 1 producer node, 1 observer node

### DIFF
--- a/tests/performance_tests/performance_test_basic.py
+++ b/tests/performance_tests/performance_test_basic.py
@@ -7,6 +7,7 @@ harnessPath = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 sys.path.append(harnessPath)
 
 from TestHarness import Cluster, TestHelper, Utils, WalletMgr
+from TestHarness.TestHelper import AppArgs
 import log_reader
 
 Print = Utils.Print
@@ -27,9 +28,12 @@ def waitForEmptyBlocks(node):
             emptyBlocks = 0
     return node.getHeadBlockNum()
 
+appArgs=AppArgs()
+extraArgs = appArgs.add(flag="--target-tps", type=int, help="The target transfers per second to send during test. Default: 1000", default=1000)
+extraArgs = appArgs.add(flag="--test-duration-sec", type=int, help="The duration of transfer trx generation for the test in seconds. Default: 30", default=30)
 args=TestHelper.parse_args({"-p","-n","-d","-s","--nodes-file"
                             ,"--dump-error-details","-v","--leave-running"
-                            ,"--clean-run","--keep-logs"})
+                            ,"--clean-run","--keep-logs"}, applicationSpecificArgs=appArgs)
 
 pnodes=args.p
 topo=args.s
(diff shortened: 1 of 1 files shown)


tests/TestHarness/Cluster.py
```diff
@@ -276,10 +277,14 @@ def launch(self, pnodes=1, unstartedNodes=0, totalNodes=1, prodCount=1, topo="me
...
 
-        cmdArr.append("--max-block-cpu-usage")
-        cmdArr.append(str(160000000))
-        cmdArr.append("--max-transaction-cpu-usage")
-        cmdArr.append(str(150000000))
+        if genesisPath is None:
+            cmdArr.append("--max-block-cpu-usage")
+            cmdArr.append(str(500000))
+            cmdArr.append("--max-transaction-cpu-usage")
+            cmdArr.append(str(375000))
```
heifner, line 284: I would go with 475000 instead.

oschwaldp-oci: Addressed: d159aaf6c9a7fc51ddba90f91b065016006125e0

tests/performance_tests/genesis.json
```diff
@@ -0,0 +1,23 @@
+{
+  "initial_timestamp": "2018-03-02T12:00:00.000",
+  "initial_key": "EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV",
+  "initial_configuration": {
+    "max_block_net_usage": 4194304,
```
heifner, line 5: This still limits to around ~32K I believe, I would go with 41943040 which is likely what you meant.

heifner (approved):

tests/performance_tests/genesis.json
oschwaldp-oci: Addressed: [d4c60a0](https://github.com/AntelopeIO/leap/pull/160/commits/d4c60a04349294f3df7e87e43e9fd38993b26866)

tests/performance_tests/performance_test_basic.py
```diff
@@ -27,9 +28,13 @@ def waitForEmptyBlocks(node):
             emptyBlocks = 0
     return node.getHeadBlockNum()
 
+appArgs=AppArgs()
+extraArgs = appArgs.add(flag="--target-tps", type=int, help="The target transfers per second to send during test. Default: 1000", default=1000)
```
jgiszczak, line 32: I think if you look at the `--help` output that the Python library is already displaying the default value independently of the help string so including it in the help string is redundant, here and in the next option.

jgiszczak (changes requested):

tests/performance_tests/performance_test_basic.py
oschwaldp-oci: Addressed: [b8cddd0](https://github.com/AntelopeIO/leap/pull/160/commits/b8cddd079ef86d2069cbfb713a5ebac073a534dc)

jgiszczak (approved):

PR #164, opened 2022-09-14
[PH] Export Performance Harness Results as JSON

ClaytonCalabrese:
Takes the dataclass from the python script along with some other variables and puts them into a pretty json format for later consumption.

Necessitated conversions from numpy types to ints in order to properly use with json.

Sample output:

```
{
  "TPS": {
    "avg": 1,
    "emptyBlocks": 145,
    "max": 21,
    "min": 0,
    "numBlocks": 215,
    "sigma": 3
  },
  "args": "Namespace(cease_block=None, log_path='var/lib/node_01/stderr.txt', num_blocks_to_prune=2, save_json=True, start_block=2)",
  "env": "Linux posix 5.10.16.3-microsoft-standard-WSL2",
  "nodeosVersion": "3_2_0_dev"
}
````

### CODE REVIEW

File: tests/performance_tests/log_reader.py [hunk 1]
```diff
@@ -149,4 +151,17 @@ def scoreTransfersPerSecond(data: chainData, numAddlBlocksToDrop=0) -> stats:
...
+    with open("../CMakeLists.txt", 'rt') as r:
+        regex_match = re.findall(r'VERSION_MAJOR ([0-9]+).*\n.*VERSION_MINOR ([0-9]+).*\n.*VERSION_PATCH ([0-9]+).*\n.*VERSION_SUFFIX (.*)\)', r.read())
+        json_string += f"\"nodeosVersion\":\"{regex_match[0][0]}_{regex_match[0][1]}_{regex_match[0][2]}_{regex_match[0][3]}\","
+    json_string += f"\"env\":\"{system()} {os.name} {release()}\","
+    json_string += f"\"args\":\"{args}\","
+    json_string += "\"TPS\":"
+    json_string += json.dumps(asdict(data))
+    json_string += "}"
+    with open('data.json', 'wt') as f:
+        f.write(json.dumps(json.loads(json_string), sort_keys=True, indent=2))
```
jgiszczak, line 167: Rather than doing a `dumps()` of a `loads()` of a string, it's much more readable to simply construction a Python dictionary and fill it with data, then pass that directly to `dumps()`. 

```
js = {}

js['args'] = args
# other members

with open(  # etc.
   f.write(json.dumps(js, sort_keys=True, indent=2))
```

tests/performance_tests/log_reader.py
```diff
@@ -149,4 +151,17 @@ def scoreTransfersPerSecond(data: chainData, numAddlBlocksToDrop=0) -> stats:
     npCBTAEC = np.array(consecBlkTrxsAndEmptyCnt, dtype=np.uint)
 
     # Note: numpy array slicing in use -> [:,0] -> from all elements return index 0
-    return stats(np.min(npCBTAEC[:,0]), np.max(npCBTAEC[:,0]), np.average(npCBTAEC[:,0]), np.std(npCBTAEC[:,0]), np.sum(npCBTAEC[:,1]), len(prunedBlockDataLog))
+    return stats(int(np.min(npCBTAEC[:,0])), int(np.max(npCBTAEC[:,0])), int(np.average(npCBTAEC[:,0])), int(np.std(npCBTAEC[:,0])), int(np.sum(npCBTAEC[:,1])), int(len(prunedBlockDataLog)))
+
+def exportAsJSON(data, args):
+    json_string = "{"
+    with open("../CMakeLists.txt", 'rt') as r:
```
spoonincode, line 158: Assumes buildddir is subdir of sourcedir. Can you `nodeos --version`?

spoonincode (changes requested):

jgiszczak (changes requested):

ClaytonCalabrese:
Output after changes

```
{
  "TPS": {
    "avg": 1,
    "emptyBlocks": 145,
    "max": 21,
    "min": 0,
    "numBlocks": 215,
    "sigma": 3
  },
  "args": "Namespace(cease_block=None, log_path='var/lib/node_01/stderr.txt', num_blocks_to_prune=2, save_json=True, start_block=2)",
  "env": "Linux posix 5.10.16.3-microsoft-standard-WSL2",
  "nodeosVersion": "v3.2.0-dev"
}
```

jgiszczak:
I think it's not only safe but necessary to assume the location of `nodeos`.  Running it without an explicit path leaves it up to the shell to find, which could easily be an older installed version rather than the built version being tested.  The regression test harness is already explicit about which `nodeos` and `cleos` it is using.

Having said that, you might see if you can extract that out of the test harness and use that.  If that path isn't readily accessible it should be.  Feel free to add a convenience accessor to the test harness code if there isn't one.

tests/performance_tests/log_reader.py
```diff
@@ -149,4 +151,13 @@ def scoreTransfersPerSecond(data: chainData, numAddlBlocksToDrop=0) -> stats:
     npCBTAEC = np.array(consecBlkTrxsAndEmptyCnt, dtype=np.uint)
 
     # Note: numpy array slicing in use -> [:,0] -> from all elements return index 0
-    return stats(np.min(npCBTAEC[:,0]), np.max(npCBTAEC[:,0]), np.average(npCBTAEC[:,0]), np.std(npCBTAEC[:,0]), np.sum(npCBTAEC[:,1]), len(prunedBlockDataLog))
+    return stats(int(np.min(npCBTAEC[:,0])), int(np.max(npCBTAEC[:,0])), int(np.average(npCBTAEC[:,0])), int(np.std(npCBTAEC[:,0])), int(np.sum(npCBTAEC[:,1])), int(len(prunedBlockDataLog)))
+
+def exportAsJSON(data, args):
+    js = {}
+    js['nodeosVersion'] = os.popen("nodeos --version").read().replace("\n", "")
```
spoonincode, line 158: Wouldn't this run the nodeos that's in the path? Not the one in the builddir? It seems like `Utils.EosServerName` is what you want assuming the script's working directory is the builddir root (which is how all the tests work, and what I assumed this script was expecting due to the previous `../CMakeLists.txt`)

spoonincode:
Depending on the purpose of this json file, you could consider `--full-version` instead to capture the commit hash.

tests/performance_tests/read_log_data.py
```diff
@@ -19,3 +20,5 @@
 
 stats = log_reader.scoreTransfersPerSecond(data, args.num_blocks_to_prune)
 print(f"TPS: {stats}")
+if args.save_json:
+    log_reader.exportAsJSON(stats, args)
```
oschwaldp-oci, line 24: Shouldn't this option also be included in `performance_test_basic.py`?

tests/performance_tests/read_log_data.py
```diff
@@ -8,6 +8,7 @@
 parser.add_argument("--start_block", type=int, help="First significant block number in the log", default=2)
 parser.add_argument("--cease_block", type=int, help="Last significant block number in the log")
 parser.add_argument("--num-blocks-to-prune", type=int, default=2, help="The number of potentially non-empty blocks, in addition to leading and trailing size 0 blocks, to prune from the beginning and end of the range of blocks of interest for evaluation.")
+parser.add_argument("--save-json", type=bool, default=False, help="Whether to save stats as json")
```
oschwaldp-oci, line 11: Should a path argument be provided to dictate where the json file should be written/saved?  If so, would probably be good here and in `performance_test_basic.py`. Could provide a sane default for if only `--save-json` is specified.

tests/performance_tests/log_reader.py
```diff
@@ -149,4 +151,13 @@ def scoreTransfersPerSecond(data: chainData, numAddlBlocksToDrop=0) -> stats:
     npCBTAEC = np.array(consecBlkTrxsAndEmptyCnt, dtype=np.uint)
 
     # Note: numpy array slicing in use -> [:,0] -> from all elements return index 0
-    return stats(np.min(npCBTAEC[:,0]), np.max(npCBTAEC[:,0]), np.average(npCBTAEC[:,0]), np.std(npCBTAEC[:,0]), np.sum(npCBTAEC[:,1]), len(prunedBlockDataLog))
+    return stats(int(np.min(npCBTAEC[:,0])), int(np.max(npCBTAEC[:,0])), int(np.average(npCBTAEC[:,0])), int(np.std(npCBTAEC[:,0])), int(np.sum(npCBTAEC[:,1])), int(len(prunedBlockDataLog)))
```
oschwaldp-oci, line 154: Is the `int` cast around `len(...)` needed? It is not a `np` type.

oschwaldp-oci (changes requested):

tests/performance_tests/performance_test_basic.py
```diff
@@ -29,7 +29,7 @@ def waitForEmptyBlocks(node):
     return node.getHeadBlockNum()
 
 appArgs = AppArgs()
-appArgs.add(flag="--save-json", type=bool, default=False, help="Whether to save stats as json")
+appArgs.add(flag = "--save-json", type=str, default="data.json", help="Path to save json output")
```
oschwaldp-oci, line 32: Looks like most places aren't using spaces around the `=` in these cases, thus `flag="--save-json" might be more normalized with the rest of the codebase.  Also, most places put the default case as the last argument.

oschwaldp-oci (changes requested):

oschwaldp-oci (approved):
Looks like there may be a slight merge conflict to fix before merging, but lgtm otherwise.

jgiszczak (approved):

.cicd/platforms/ubuntu18.Dockerfile
```diff
@@ -15,10 +15,14 @@ RUN apt-get update && apt-get upgrade -y && \
                                                       ninja-build                 \
                                                       pkg-config                  \
                                                       python3                     \
+                                                      python3-pip          \
```
spoonincode, line 18: please clean up the whitespace here

oschwaldp-oci: Done.

.cicd/platforms/ubuntu22.Dockerfile
```diff
@@ -14,4 +14,7 @@ RUN apt-get update && apt-get upgrade -y && \
                                                       llvm-11-dev          \
                                                       ninja-build          \
                                                       pkg-config           \
+                                                      python3-pip          \
                                                       zstd
+
+RUN python3 -m pip install numpy
```
spoonincode, line 20: (no need to take action on this comment at the moment -- not until ready to start preparing this for a main merge)

This would seem to add a new requirement to build and test Leap: it needs to be documented in the Readme and such. It's also a bit unfortunate as we've gone 5 years without needing non-standard python packages. I'm not sure if maybe we can detect the presence of this package and simply not clutch in the single test that requires it; doing this opens up other issues though.

oschwaldp-oci: Updating documentation is in the roadmap downstream a bit.  I'll make sure to note these specific updates to take care of at that point. Thanks.

Regarding `numpy` - While it may seem unfortunate from a certain perspective, it really is the `fundamental package for scientific computing with Python` and very widely adopted and used when doing the types of computing being asked for in a performance monitoring tool.  It is a standard Ubuntu package and extremely well supported, even to the point that there are boost bindings available for it also as a package.

spoonincode: That's true, maybe consider going ahead and migrating to the Ubuntu package instead of the pip package. At least for 20 & 22 where it would eliminate the pip stuff altogether.

oschwaldp-oci: Okay.  I'll add that to our backlog and make sure that's done before this hits `main`.  Thanks.

PR #165, opened 2022-09-14
[3.1] use a temp dir for release-build test as it can leak creation of config.ini in to default location

### LINKED ISSUE #157: release-build-test touches nodeos default config directory
```
root@dfd50a850728:/work/leap/build# ls ~/.local
ls: cannot access '/root/.local': No such file or directory
root@dfd50a850728:/work/leap/build# ctest -R release-build-test
Test project /work/leap/build
    Start 162: release-build-test
1/1 Test #162: release-build-test ...............   Passed    0.02 sec

100% tests passed, 0 tests failed out of 1

Total Test time (real) =   0.07 sec
root@dfd50a850728:/work/leap/build# ls ~/.local
share
```

We want to avoid this: it can create weird problems.

spoonincode:
Running the release-build test causes creation of config.ini in the default location. I'm treating any tests that touch the default config and data directories as a defect since we've seen problems before when they do that. This particular case is probably ultimately benign but eh it's also straightforward to fix.

Resolves #157

### DIFF
--- a/tests/release-build.sh
+++ b/tests/release-build.sh
@@ -11,7 +11,11 @@ echo 'nodeos from source should perform a "release build" which excludes asserts
 echo 'debugging symbols, and performs compiler optimizations.'
 echo ''
 
-NODEOS_DEBUG=$(programs/nodeos/nodeos --extract-build-info >(python3 -c 'import json,sys; print(str(json.load(sys.stdin)["debug"]).lower());') &> /dev/null)
+TDIR=$(mktemp -d || exit 1)
+NODEOS_DEBUG=$(programs/nodeos/nodeos --config-dir "${TDIR}" --data-dir "${TDIR}" --extract-build-info >(python3 -c 'import json,sys; print(str(json.load(sys.stdin)["debug"]).lower());') &> /dev/null)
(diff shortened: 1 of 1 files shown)


tests/release-build.sh
```diff
@@ -11,7 +11,11 @@ echo 'nodeos from source should perform a "release build" which excludes asserts
 echo 'debugging symbols, and performs compiler optimizations.'
 echo ''
 
-NODEOS_DEBUG=$(programs/nodeos/nodeos --extract-build-info >(python3 -c 'import json,sys; print(str(json.load(sys.stdin)["debug"]).lower());') &> /dev/null)
+TDIR=$(mktemp -d || exit 1)
+NODEOS_DEBUG=$(programs/nodeos/nodeos --config-dir "${TDIR}" --data-dir "${TDIR}" --extract-build-info >(python3 -c 'import json,sys; print(str(json.load(sys.stdin)["debug"]).lower());') &> /dev/null)
+#avoiding an rm -rf out of paranoia, but with the tradeoff this could change somehow in the future
```
heifner, line 16: Probably not a bad thing if it does break in the future, at least it will be obvious why and might actually prevent some accidental add of something.

heifner (approved):

PR #190, opened 2022-09-19
Add Prometheus Plugin

ndcgundlach:
Added the cmake changes necessary to build the prometheus plugin along with the skeleton code for the plugin itself.


plugins/prometheus_plugin/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+file(GLOB HEADERS "include/eosio/prometheus_plugin/*.hpp")
+add_library( prometheus_plugin
+        prometheus_plugin.cpp
+        ${HEADERS} )
+
+target_link_libraries( prometheus_plugin appbase fc http_plugin chain_plugin eosio_testing_contracts )
+target_include_directories( prometheus_plugin PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include
```
spoonincode, line 7: The includes and libraries need to be fixed up here. It looks like there were copied from somewhere else

spoonincode:
What's the reasoning for it being an optional plugin?

ndcgundlach:
> What's the reasoning for it being an optional plugin?

One less thing to build if you don't need it.

spoonincode:
Is it expensive to build or something? Optional items like this start raising some questions like:
* Should we be building it in CI?
  * Should we be building it both _on and off_ in CI?
* Should it be enabled in binary builds we ship

plugins/prometheus_plugin/CMakeLists.txt
ndcgundlach: > Is it expensive to build or something? Optional items like this start raising some questions like:
> 
> * Should we be building it in CI?
>   
>   * Should we be building it both _on and off_ in CI?
> * Should it be enabled in binary builds we ship

I don't forsee it being expensive to build.  Do you want me to always include it?

ndcgundlach:
> Is it expensive to build or something? Optional items like this start raising some questions like:
> 
> * Should we be building it in CI?
>   
>   * Should we be building it both _on and off_ in CI?
> * Should it be enabled in binary builds we ship

I don't forsee it being expensive to build.  Do you want me to have it always be included?

spoonincode:
It doesn't sound like there is a reason to have it optional for now.

spoonincode (approved):

PR #199, opened 2022-09-21
[3.1] Update cleos to return an error code on failed trx processing

### LINKED ISSUE #139: cleos return codes
As of Leap 3, `cleos` defaults to `--return-failure-trace true` which means that automation which depends on `cleos` to return a non-zero value on transaction failure will instead need to parse the traces or use option `--return-failure-trace false`.

Update `cleos` to parse transaction trace and return a non-zero error code when the trace includes an exception indicating the transaction failed to speculatively execute.

heifner:
Leap 3.1 returns transaction failure traces by default. `EOSIO/eos 2.0` would instead throw an exception. 
The [3.1 release notes](https://github.com/AntelopeIO/leap/releases/tag/v3.1.0) included the following:

> New command line options for cleos:
> - `--use-old-send-rpc`
>   - "Use old RPC `/v1/chain/send_transaction`, rather than new RPC `/v1/chain/send_transaction2`"
> - `-t,--return-failure-trace` defaults to `true` use `-t false` for previous behavior
>   - "Return partial traces on failed transactions"
>   - Note: Automation which depends on `cleos` to return a non-zero value on transaction failure will instead need to parse the traces or use option `-t false`.

This PR addresses the cleos non-zero return values for failed transaction execution. With this PR, cleos will parse the returned transaction traces and return a non-zero return value for transactions that failed to speculatively execute.

Before:
```
$ ./cleos  push action  eosio.token transfer '{"from":"eosio","to":"testera11111","quantity":"0.0001 SYS","memo":"hi"}'
failed transaction: d19173082af2e0dee7d11a638adabc646f55b81364b923b71c6d467fbff30b76  <unknown> bytes  <unknown> us
error 2022-09-21T15:20:49.283 cleos     main.cpp:601                  print_result         ] soft_except->to_detail_string(): 3040003 tx_no_auths: Transaction should have at least one required authority
transaction must have at least one authorization
    {}
    nodeos  transaction_context.cpp:753 validate_referenced_accounts

$ echo $?
0
```

After:
```
$ ./cleos  push action  eosio.token transfer '{"from":"eosio","to":"testera11111","quantity":"0.0001 SYS","memo":"hi"}'
failed transaction: 1488c901f8d1952eae4fddd509758a639cabf5f1ea86967bf6215580c264f519  <unknown> bytes  <unknown> us
error 2022-09-21T15:20:42.847 cleos.3.1 main.cpp:575                  print_result         ] soft_except->to_detail_string(): 3040003 tx_no_auths: Transaction should have at least one required authority
transaction must have at least one authorization
    {}
    nodeos  transaction_context.cpp:753 validate_referenced_accounts

$ echo $?
3
```

Resolves #139

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -548,16 +548,14 @@ int get_return_code( const fc::variant& result ) {
          const auto& except = processed["except"];
          if( except.is_object() ) {
             try {
-               auto soft_except = except.as<std::optional<fc::exception>>();
-               if( soft_except ) {
-                  auto code = soft_except->code();
-                  if( code > std::numeric_limits<int>::max() ) {
-                     r = 1;
-                  } else {
-                     r = static_cast<int>( code );
-                  }
-                  if( r == 0 ) r = 1;
+               auto soft_except = except.as<fc::exception>();
+               auto code = soft_except.code();
+               if( code > std::numeric_limits<int>::max() ) {
+                  r = 1;
+               } else {
+                  r = static_cast<int>( code );
                }
+               if( r == 0 ) r = 1;
             } catch( ... ) {
                r = 1;
             }
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -545,19 +545,22 @@ int get_return_code( const fc::variant& result ) {
    if (result.is_object() && result.get_object().contains("processed")) {
       const auto& processed = result["processed"];
       if( processed.get_object().contains( "except" ) ) {
-         try {
-            auto soft_except = processed["except"].as<std::optional<fc::exception>>();
-            if( soft_except ) {
-               auto code = soft_except->code();
-               if( code > std::numeric_limits<int>::max() ) {
-                  r = 1;
-               } else {
-                  r = static_cast<int>( code );
+         const auto& except = processed["except"];
+         if( except.is_object() ) {
+            try {
+               auto soft_except = except.as<std::optional<fc::exception>>();
+               if( soft_except ) {
+                  auto code = soft_except->code();
+                  if( code > std::numeric_limits<int>::max() ) {
+                     r = 1;
+                  } else {
+                     r = static_cast<int>( code );
+                  }
+                  if( r == 0 ) r = 1;
                }
+            } catch( ... ) {
+               r = 1;
             }
-            if( r == 0 ) r = 1;
-         } catch( ... ) {
-            r = 1;
          }
       }
    }
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -544,7 +544,7 @@ int get_return_code( const fc::variant& result ) {
    int r = 0;
    if (result.is_object() && result.get_object().contains("processed")) {
       const auto& processed = result["processed"];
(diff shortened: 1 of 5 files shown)


programs/cleos/main.cpp
```diff
@@ -538,6 +539,34 @@ void print_action_tree( const fc::variant& action ) {
    }
 }
 
+int get_return_code( const fc::variant& result ) {
+   // if a trx with a processed, then check to see if it failed execution for return value
+   int r = 0;
+   if (result.is_object() && result.get_object().contains("processed")) {
+      const auto& processed = result["processed"];
+      if( processed.get_object().contains( "except" ) ) {
```
spoonincode, line 547: Do you need to check if `processed` is an object before `get_object()`ing it? You do the `is_object()` check for `result` & `except`. And a failure here with `processed` not being an object would be outside of the try block internal to this function.

tests/testUtils.py
```diff
@@ -205,7 +205,8 @@ def checkDelayedOutput(popen, cmd, ignoreError=False):
         Utils.checkOutputFileWrite(start, cmd, output, error)
         if popen.returncode != 0 and not ignoreError:
             Utils.Print("ERROR: %s" % error)
-            raise subprocess.CalledProcessError(returncode=popen.returncode, cmd=cmd, output=error)
+            # for now just include both stderr and stdout in output, python 3.5+ supports both a stdout and stderr attributes
```
spoonincode, line 208: what is the Python 3.5+ comment here getting at -- even Ubuntu 18 is using 3.6 so if there is a 3.5+ feature we should be able to use it

programs/cleos/main.cpp
heifner: Yes, that would be better. I copy/pasted from `print_result` which does not have that. I'll add the check.

tests/testUtils.py
heifner: `stderr` was being placed in `output`. Python 3.5+ has both an `output` and a `stderr`. The better fix is to put `stdout` into `output` and `stderr` into `stderr`. However, `output` is used in 20+ places in many tests. I didn't want to make that big of a change in `release/3.1`. In `main` I plan to sperate out the `stdout` and `stderr` and update the 20+ usages of the `CalledProcessError`.

programs/cleos/main.cpp
```diff
@@ -538,6 +539,34 @@ void print_action_tree( const fc::variant& action ) {
...
+   // if a trx with a processed, then check to see if it failed execution for return value
+   int r = 0;
+   if (result.is_object() && result.get_object().contains("processed")) {
+      const auto& processed = result["processed"];
+      if( processed.is_object() && processed.get_object().contains( "except" ) ) {
+         const auto& except = processed["except"];
+         if( except.is_object() ) {
+            try {
+               auto soft_except = except.as<std::optional<fc::exception>>();
+               if( soft_except ) {
```
spoonincode, line 552: At this point we've established that `except` is an object. Do we really need to treat `except` as an `optional` type still?

heifner: No. Removed `optional`.

spoonincode (approved):

PR #201, opened 2022-09-22
[3.2] add missing incoming_persisted in unapplied transaction queue's get_trx_type

spoonincode:
This was generating a compiler warning that a value was not handled.


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1470,6 +1470,8 @@ producer_plugin::get_unapplied_transactions( const get_unapplied_transactions_pa
             return "forked";
          case trx_enum_type::aborted:
             return "aborted";
+         case trx_enum_type::incoming_persisted:
+            return "incoming_persisted";
```
heifner, line 1474: Thanks. That is what I get for letting CLion generate the switch statement for me. I wonder why it got confused.

heifner (approved):

PR #213, opened 2022-09-22
[PH] Performance harness multiple generators

ClaytonCalabrese:
Allow multiple transaction generators to be run in both performance_test_basic.py as well as in a standalone script. Various changes were needed to accommodate this.


tests/performance_tests/launch_transaction_generators.py
```diff
@@ -0,0 +1,61 @@
...
+parser.add_argument("tps_limit_per_generator", type=int, help="Maximum amount of transactions per second a single generator can have.")
+args = parser.parse_args()
+
+targetTps = args.target_tps
+numGenerators = math.ceil(targetTps/args.tps_limit_per_generator)
+tpsPerGenerator = math.floor(targetTps/numGenerators)
+modTps = targetTps%numGenerators
+cleanlyDivisible = modTps == 0
+incrementPoint = numGenerators + 1 - modTps
+num_list = []
```
oschwaldp-oci, line 37: Could we call this something more like `subprocesses` relating to the fact it is storing the return from `subprocess.Popen` and used to `.wait()`

tests/performance_tests/launch_transaction_generators.py
```diff
@@ -0,0 +1,61 @@
...
+                            [f'./tests/trx_generator/trx_generator',
+                            '--chain-id', f'{args.chain_id}',
+                            '--last-irreversible-block-id', f'{args.last_irreversible_block_id}',
+                            '--handler-account', f'{args.handler_account}',
+                            '--accounts', f'{args.account_1_name},{args.account_2_name}',
+                            '--priv-keys', f'{args.account_1_priv_key},{args.account_2_priv_key}',
+                            '--trx-gen-duration', f'{args.trx_gen_duration}',
+                            '--target-tps', f'{tpsPerGenerator}']
+                        ))
+exit_codes = [num.wait() for num in num_list]
```
oschwaldp-oci, line 61: Could we call `num` something more like `subprocess` just for clarity on what it is?

tests/TestHarness/Cluster.py
```diff
@@ -167,7 +167,7 @@ def setAlternateVersionLabels(self, file):
     # pylint: disable=too-many-statements
     def launch(self, pnodes=1, unstartedNodes=0, totalNodes=1, prodCount=1, topo="mesh", delay=1, onlyBios=False, dontBootstrap=False,
                totalProducers=None, sharedProducers=0, extraNodeosArgs="", useBiosBootFile=True, specificExtraNodeosArgs=None, onlySetProds=False,
-               pfSetupPolicy=PFSetupPolicy.FULL, alternateVersionLabelsFile=None, associatedNodeLabels=None, loadSystemContract=True, genesisPath=None, maximumP2pPerHost=0):
+               pfSetupPolicy=PFSetupPolicy.FULL, alternateVersionLabelsFile=None, associatedNodeLabels=None, loadSystemContract=True, genesisPath=None, maximumP2pPerHost=0, maximumClients=0):
```
oschwaldp-oci, line 170: Document descriptions below in doc string for `maximumP2pPerHost` and `maximumClients` please.

tests/TestHarness/Cluster.py
```diff
@@ -226,9 +226,12 @@ def launch(self, pnodes=1, unstartedNodes=0, totalNodes=1, prodCount=1, topo="me
         if sharedProducers > 0:
             producerFlag += (" --shared-producers %d" % (sharedProducers))
 
+        if maximumClients <= 0:
+            maximumClients = 25
```
oschwaldp-oci, line 230: Should the default for `maximumClients` just be 25 in the parameter list if you're going to overwrite `0` with `25` anyway?

tests/TestHarness/Cluster.py
```diff
@@ -226,9 +226,12 @@ def launch(self, pnodes=1, unstartedNodes=0, totalNodes=1, prodCount=1, topo="me
...
         if maximumP2pPerHost <= 0:
             maximumP2pPerHost = totalNodes
```
oschwaldp-oci, line 232: Should the default for `maximumP2pPerHost` be `totalNodes` in the parameter list since `0` will be overwritten anyway?

oschwaldp-oci (changes requested):

tests/TestHarness/Cluster.py
ClaytonCalabrese: I don't think I am able to set one argument equal to another as a default in the declaration

oschwaldp-oci: Great point, I didn't think that through completely.  Maybe just say in the doc string that it defaults to the `totalNodes` value if left at `0`

jgiszczak (approved):

PR #237, opened 2022-09-26
[PH] Performance harness graceful shutdown

ClaytonCalabrese:
- Causes trx_generators to be killed upon sigint.
- Outputs to console upon sigint.
- Reports in JSON whether a run was canceled early.
- Moves analysis to finally, so it is still run when sigint is invoked.

### DIFF
--- a/tests/performance_tests/performance_test_basic.py
+++ b/tests/performance_tests/performance_test_basic.py
@@ -60,6 +60,8 @@ cluster=Cluster(walletd=True, loggingLevel="info")
 cluster.setWalletMgr(walletMgr)
 
 testSuccessful = False
+completedRun = False
+
 try:
     # Kill any existing instances and launch cluster
     TestHelper.printSystemInfo("BEGIN")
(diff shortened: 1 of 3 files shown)


tests/performance_tests/performance_test_basic.py
```diff
@@ -132,27 +149,18 @@ def waitForEmptyBlocks(node):
...
+    report = log_reader.createJSONReport(guide, tpsStats, args, completedRun)
     print(report)
     if args.save_json:
         log_reader.exportAsJSON(report, args)
 
-    assert transactionsSent == data.totalTransactions , f"Error: Transactions received: {data.totalTransactions} did not match expected total: {transactionsSent}"
+    if completedRun:
+        assert transactionsSent == data.totalTransactions , f"Error: Transactions received: {data.totalTransactions} did not match expected total: {transactionsSent}"
+    else:
+        os.system("pkill trx_generator")
```
jgiszczak, line 160: This is a candidate for replacing with explicit subprocess control, but we'll wait on that until the regression test harness has support for it.

jgiszczak (approved):

oschwaldp-oci (approved):

PR #242, opened 2022-09-27
[3.2] avoid redefining _mm_crc32_u64 which can lead to compile errors

spoonincode:
Defining our own `_mm_crc32_u64` implementation can clash with some compilers. Not too surprising since that's a reserved name in global scope. In particular, AppleClang will give an error:
```
error: always_inline function '_mm_crc32_u64' requires target feature 'sse4.2', but would be inlined into function 'CityHashCrc256Long' that is compiled without support for 'sse4.2'
```

Rename our impl to just `mm_crc32_u64` and redirect to either `mm_crc32_u64` or `_mm_crc32_u64` via a new macro.

### DIFF
--- a/libraries/libfc/src/crypto/city.cpp
+++ b/libraries/libfc/src/crypto/city.cpp
@@ -38,8 +38,10 @@
 
 #if defined(__SSE4_2__) && defined(__x86_64__)
 #include <nmmintrin.h>
+#define MM_CRC32_U64I _mm_crc32_u64
 #else
-uint64_t _mm_crc32_u64(uint64_t a, uint64_t b );
+uint64_t mm_crc32_u64(uint64_t a, uint64_t b );
+#define MM_CRC32_U64I mm_crc32_u64
 #endif
 
 namespace fc {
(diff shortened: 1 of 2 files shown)


libraries/libfc/src/crypto/city.cpp
```diff
@@ -38,8 +38,10 @@
 
 #if defined(__SSE4_2__) && defined(__x86_64__)
 #include <nmmintrin.h>
+#define MM_CRC32_U64I _mm_crc32_u64
 #else
-uint64_t _mm_crc32_u64(uint64_t a, uint64_t b );
+uint64_t mm_crc32_u64(uint64_t a, uint64_t b );
+#define MM_CRC32_U64I mm_crc32_u64
```
heifner, line 44: Should we use `FC_MM_CRC32_U64I` instead?

spoonincode: I was trying to keep the same length to avoid touching a bunch of whitespace over in `CHUNK()` :sweat_smile: 

Since this define is local to a .cpp it doesn't seem like there is much risk of a clash? Possibly `mm_crc32_u64()` should be moved in to fc's namespace though?

spoonincode: actually if `_mm_crc32_u64` is just placed in to a namespace I think that's safe. maybe that's the better choice :thinking:

heifner: I didn't notice this was only in this one source file. I think it is fine as is.

heifner (approved):

PR #243, opened 2022-09-27
[3.2] remove old, unused, and currently broken alternate k1 ECC impls

spoonincode:
At some point long long ago it might have been possible to use alternate implementations of k1 ECC math by switching the `ECC_IMPL` config to `openssl` or `mixed`. My hunch is no one has ever made use of these in the EOSIO/Leap timeframe, and neither of those compile now. Maybe we could consider fixing them but `ec_recover.cpp` already uses `secp256k1` directly making usage of that library a requirement regardless.

I think it's best to go ahead and just remove what might as well be old cruft.


heifner (approved):
I stumbled upon this a few months back, glad to see it be cleaned up.

PR #245, opened 2022-09-27
[3.1] Remove #pragma(s) to ignore warnings

### LINKED ISSUE #229: reevaluate disablement of strict aliasing warning in chain_plugin
https://github.com/AntelopeIO/leap/blob/e401c23f0ee3a52f0f5362ced0bd0a394b10186f/plugins/chain_plugin/chain_plugin.cpp#L1709-L1710

A warning about strict aliasing is probably something that should be tended to, not silenced away.

heifner:
- Most of the changes are from the removal of ignoring unused-variables from `chain/config.hpp`
- Removed pragma to ignore `strict-aliasing` from `chain_plugin` which is the main reason for the changes.
  - Didn't need any code changes as this did not generate any warnings (at least on clang).
    - [ Will see what GCC does with it in cicd. ]

Resolves #229

### DIFF
--- a/unittests/abi_tests.cpp
+++ b/unittests/abi_tests.cpp
@@ -2054,8 +2054,7 @@ BOOST_AUTO_TEST_CASE(abi_account_name_in_eosio_abi)
    }
    )=====";
 
-   auto abi = eosio_contract_abi(fc::json::from_string(repeat_abi).as<abi_def>());
-   auto is_type_exception = [](fc::assert_exception const & e) -> bool { return e.to_detail_string().find("abi.types.size") != std::string::npos; };
+   eosio_contract_abi(fc::json::from_string(repeat_abi).as<abi_def>());
 
 } FC_LOG_AND_RETHROW() }
 
--- a/unittests/auth_tests.cpp
+++ b/unittests/auth_tests.cpp
@@ -341,7 +341,6 @@ BOOST_AUTO_TEST_CASE( any_auth ) { try {
 
    const auto spending_priv_key = chain.get_private_key(name("alice"), "spending");
    const auto spending_pub_key = spending_priv_key.get_public_key();
-   //const auto bob_spending_priv_key = chain.get_private_key(name("bob"), "spending");
    const auto bob_spending_pub_key = spending_priv_key.get_public_key();
 
    chain.set_authority(name("alice"), name("spending"), spending_pub_key, name("active"));
--- a/unittests/auth_tests.cpp
+++ b/unittests/auth_tests.cpp
@@ -341,7 +341,7 @@ BOOST_AUTO_TEST_CASE( any_auth ) { try {
 
    const auto spending_priv_key = chain.get_private_key(name("alice"), "spending");
    const auto spending_pub_key = spending_priv_key.get_public_key();
-   const auto bob_spending_priv_key = chain.get_private_key(name("bob"), "spending");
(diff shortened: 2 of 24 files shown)


unittests/abi_tests.cpp
```diff
@@ -2054,8 +2054,7 @@ BOOST_AUTO_TEST_CASE(abi_account_name_in_eosio_abi)
    }
    )=====";
 
-   auto abi = eosio_contract_abi(fc::json::from_string(repeat_abi).as<abi_def>());
-   auto is_type_exception = [](fc::assert_exception const & e) -> bool { return e.to_detail_string().find("abi.types.size") != std::string::npos; };
```
linh2931, line 2058: Maybe here we should report error if `is_type_exception` returns `true`, instead removing it?

unittests/auth_tests.cpp
```diff
@@ -341,7 +341,7 @@ BOOST_AUTO_TEST_CASE( any_auth ) { try {
 
    const auto spending_priv_key = chain.get_private_key(name("alice"), "spending");
    const auto spending_pub_key = spending_priv_key.get_public_key();
-   const auto bob_spending_priv_key = chain.get_private_key(name("bob"), "spending");
+   //const auto bob_spending_priv_key = chain.get_private_key(name("bob"), "spending");
```
linh2931, line 344: Why just comment out, not remove?

unittests/snapshot_tests.cpp
```diff
@@ -23,7 +23,7 @@ chainbase::bfs::path get_parent_path(chainbase::bfs::path blocks_dir, int ordina
       blocks_dir = blocks_dir.parent_path();
       leaf_dir = blocks_dir.filename();
       try {
-         auto ordinal_for_config = boost::lexical_cast<int>(leaf_dir.generic_string());
+         boost::lexical_cast<int>(leaf_dir.generic_string());
```
linh2931, line 26: why not remove?

unittests/wasm_config_tests.cpp
```diff
@@ -158,7 +158,7 @@ static const char many_exports_wast[] = R"=====(
   (func (export "apply") (param i64 i64 i64))
 )
 )=====";
-static const char one_export[] =  "(export \"fn${N}\" (func 0))";
+//static const char one_export[] =  "(export \"fn${N}\" (func 0))";
```
linh2931, line 161: why not remove?

unittests/abi_tests.cpp
heifner: That is a lambda that is not used. It is copy/pasted from other tests where it is actually used.

unittests/auth_tests.cpp
heifner: I'm NOT a fan of commented out code. But the symmetry with the others made me comment it out instead of remove. I'll go ahead and just remove it.

unittests/snapshot_tests.cpp
heifner: The `lexical_cast` can throw which is why the try-catch is there.

unittests/wasm_config_tests.cpp
heifner: Because I thought it might be useful in the future. Since this is non-trivial, I thought I would keep it commented out.

linh2931 (approved):

PR #258, opened 2022-09-29
Add Prometheus Library

ndcgundlach:
Added a new library for porting over prometheus-cpp. Ported ClientMetric (now client_metric). Added dependency in prometheus_plugin to verify all the CMake settings are correct.


libraries/prometheus/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
...
+if(PROMETHEUS_INSTALL_COMPONENT)
+   set(INSTALL_COMPONENT_ARGS COMPONENT ${PROMETHEUS_INSTALL_COMPONENT} EXCLUDE_FROM_ALL)
+endif()
+
+install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/prometheus DESTINATION ${CMAKE_INSTALL_FULL_INCLUDEDIR} ${INSTALL_COMPONENT_ARGS})
+
+install(TARGETS prometheus
+   LIBRARY DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR} ${INSTALL_COMPONENT_ARGS}
+   ARCHIVE DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR} ${INSTALL_COMPONENT_ARGS})
+
```
spoonincode, line 46: I'm not sure what the end game is with this, but as-is this entire file can be massively cleaned up. It seems like you only need lines 32-34. It's not a separate submodule so doesn't need all the project & options configuration, and it's not needed for libtester so stuff like headers don't need to be installed.

ndcgundlach: I'll clean it up.

libraries/CMakeLists.txt
```diff
@@ -3,6 +3,7 @@ set(FC_INSTALL_COMPONENT "dev")
 set(APPBASE_INSTALL_COMPONENT "dev")
 set(SOFTFLOAT_INSTALL_COMPONENT "dev")
 set(EOSVM_INSTALL_COMPONENT "dev")
+set(PROMETHEUS_INSTALL_COMPONENT "dev")
```
spoonincode, line 6: this isn't needed any longer either

ndcgundlach: removed

spoonincode (approved):

PR #261, opened 2022-09-29
[PH] Cluster: Node Specific Logging Levels

ClaytonCalabrese:
This PR enables custom log levels in Cluster on a per-node basis.

This is achieved by creating a dict in python with a key equal to the `dex` of the node, being either the node number, or bios if specifying the bios node. The value of the dict is the string equivalent of the `fc::log_level`.  Unspecified nodes will receive the default log level.


tests/TestHarness/Cluster.py
```diff
@@ -241,10 +242,10 @@ def launch(self, pnodes=1, unstartedNodes=0, totalNodes=1, prodCount=1, topo="me
                 return False
             tries = tries - 1
             time.sleep(2)
-
-        cmd="%s -p %s -n %s -d %s -i %s -f %s --unstarted-nodes %s --logging-level %s" % (
+        loggingLevelDictString = json.dumps(self.loggingLevelDict).replace(" ", "")
```
jgiszczak, line 245: You can include the `separators=(',', ':') ` argument to `json.dumps()` and eliminate the need to call `replace()`.

jgiszczak (changes requested):

jgiszczak (approved):

Issue #254, opened 2022-09-28
Clean up remaining 22 warnings in Leap build


linh2931:
After https://github.com/AntelopeIO/leap/pull/253, `Leap` build has only 22 warnings remaining. We can aim to clean up most of them before `release/3.2 ` and have a nice clean build ever since.

Issue #260, opened 2022-09-29
Use github issue instead of #warning directives in api_tests.cpp


linh2931:
`api_tests.cpp` has 3 `#warning` directives https://github.com/AntelopeIO/leap/blob/24b23ab3b98b0a616bf1f6baba2cf2fb89de4158/unittests/api_tests.cpp#L1261
`#warning TODO validate that the contract was successfully cached`

Those directives have been there since 06/07/2018. 

Remove the directives and create this issue so it can be triaged

heifner:
Could add a getter for cache size to `wasm_interface`, `controller` already has a `get_wasm_interface()`. The test could then check size of the cache before and after.

Issue #259, opened 2022-09-29
Use github issue instead of #warning directive in wasm_tests.cpp


linh2931:
`wasm_tests.cpp` has a `#warning` directive https://github.com/AntelopeIO/leap/blob/24b23ab3b98b0a616bf1f6baba2cf2fb89de4158/unittests/wasm_tests.cpp#L74
```
#warning Change this back to using TESTER
struct old_wasm_tester : tester {
   old_wasm_tester() : tester{setup_policy::old_wasm_parser} {}
};
```
Remove the directive and create this issue so it can be triaged and worked on if needed.

PR #262, opened 2022-09-29
[3.2] more compiler warnings cleanup

linh2931:
Partially resolve https://github.com/AntelopeIO/leap/issues/254

After this PR, only 4 warnings remain in submodules `abieos` and `libfc/libraries/ff/libff`.

Most of the changes are straightforward. The following two need to be noted.

1. For #warning directives in `wasm_tests.cpp` and `api_tests.cpp`, I created issues for them https://github.com/AntelopeIO/leap/issues/259 and https://github.com/AntelopeIO/leap/issues/260. I think this is a better way to handle TODOs. Leaving them in #warning will not get them to be done. I kept them as comments for easier identification if we decide to work on them.
2. To resolve the warning in `chain_plugin.cpp`, I need to introduce a new macro `FC_LOG_AND_DROP_NO_INTERPROCESS_BAD_ALLOC`. The warning was

```
In file included from /home/lh/work/leap-main/libraries/chain/include/eosio/chain/          exceptions.hpp:3,
                 from /home/lh/work/leap-main/libraries/chain/include/eosio/chain/asset.hpp:2,
                 from /home/lh/work/leap-main/plugins/chain_plugin/include/eosio/           chain_plugin/chain_plugin.hpp:3,
                 from /home/lh/work/leap-main/plugins/chain_plugin/chain_plugin.cpp:1:
/home/lh/work/leap-main/plugins/chain_plugin/chain_plugin.cpp: In member function ‘void     eosio::chain_apis::read_write::push_block(eosio::chain_apis::read_write::                   push_block_params&&, eosio::chain_apis::next_function<eosio::chain_apis::empty>)’:
/home/lh/work/leap-main/libraries/libfc/include/fc/exception/exception.hpp:481:4: warning:  exception of type ‘boost::interprocess::bad_alloc’ will be caught
  481 |    catch( const boost::interprocess::bad_alloc& ) {\
      |    ^~~~~
/home/lh/work/leap-main/libraries/libfc/include/fc/exception/exception.hpp:481:4: note: in  definition of macro ‘FC_LOG_AND_DROP’
  481 |    catch( const boost::interprocess::bad_alloc& ) {\
      |    ^~~~~
/home/lh/work/leap-main/plugins/chain_plugin/chain_plugin.cpp:2169:6: warning:    by        earlier handler for ‘boost::interprocess::bad_alloc’
 2169 |    } catch ( boost::interprocess::bad_alloc& ) {
      |      ^~~~~
```

### DIFF
--- a/libraries/libfc/include/fc/exception/exception.hpp
+++ b/libraries/libfc/include/fc/exception/exception.hpp
@@ -496,6 +496,22 @@ namespace fc
       wlog( "${details}", ("details",e.to_detail_string()) ); \
    }
 
+#define FC_LOG_AND_DROP_NO_INTERPROCESS_BAD_ALLOC( ... )  \
+   catch( fc::exception& er ) { \
+      wlog( "${details}", ("details",er.to_detail_string()) ); \
+   } catch( const std::exception& e ) {  \
+      fc::std_exception_wrapper sew( \
+                FC_LOG_MESSAGE( warn, "rethrow ${what}: ",FC_FORMAT_ARG_PARAMS( __VA_ARGS__  )("what",e.what()) ), \
+                std::current_exception(), \
+                BOOST_CORE_TYPEID(e).name(), \
+                e.what() ) ; \
+      wlog( "${details}", ("details",sew.to_detail_string()) ); \
+   } catch( ... ) {  \
+      fc::unhandled_exception e( \
+                FC_LOG_MESSAGE( warn, "rethrow", FC_FORMAT_ARG_PARAMS( __VA_ARGS__) ), \
+                std::current_exception() ); \
+      wlog( "${details}", ("details",e.to_detail_string()) ); \
+   }
 
 /**
  *  @def FC_RETHROW_EXCEPTIONS(LOG_LEVEL,FORMAT,...)
--- a/libraries/libfc/include/fc/exception/exception.hpp
+++ b/libraries/libfc/include/fc/exception/exception.hpp
@@ -478,25 +478,6 @@ namespace fc
    }
 
 #define FC_LOG_AND_DROP( ... )  \
-   catch( const boost::interprocess::bad_alloc& ) {\
-      throw;\
-   } catch( fc::exception& er ) { \
-      wlog( "${details}", ("details",er.to_detail_string()) ); \
-   } catch( const std::exception& e ) {  \
-      fc::std_exception_wrapper sew( \
-                FC_LOG_MESSAGE( warn, "rethrow ${what}: ",FC_FORMAT_ARG_PARAMS( __VA_ARGS__  )("what",e.what()) ), \
-                std::current_exception(), \
-                BOOST_CORE_TYPEID(e).name(), \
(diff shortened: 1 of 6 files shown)


libraries/libfc/include/fc/exception/exception.hpp
```diff
@@ -496,6 +496,22 @@ namespace fc
...
+                std::current_exception(), \
+                BOOST_CORE_TYPEID(e).name(), \
+                e.what() ) ; \
+      wlog( "${details}", ("details",sew.to_detail_string()) ); \
+   } catch( ... ) {  \
+      fc::unhandled_exception e( \
+                FC_LOG_MESSAGE( warn, "rethrow", FC_FORMAT_ARG_PARAMS( __VA_ARGS__) ), \
+                std::current_exception() ); \
+      wlog( "${details}", ("details",e.to_detail_string()) ); \
+   }
```
heifner, line 514: I think this could just be what `FC_LOG_AND_DROP` becomes. Just remove the `catch` for `bad_alloc` from `FC_LOG_AND_DROP`. All it does is `throw`, so they are equivalent.

libraries/libfc/src/crypto/elliptic_webauthn.cpp
```diff
@@ -88,6 +88,8 @@ struct webauthn_json_handler : public rapidjson::BaseReaderHandler<rapidjson::UT
             return true;
          case IN_NESTED_CONTAINER:
             return true;
+         default:
+            FC_THROW_EXCEPTION( exception, "String: current_state (${current_statea}) out of bound", ("current_state", current_state) );
```
heifner, line 92: I don't think we should make any of these changes to `elliptic_webauthn.cpp`. I'm concerned there is a consensus difference here.

heifner (changes requested):

libraries/libfc/src/crypto/elliptic_webauthn.cpp
spoonincode: We can also silence the warning with `__builtin_unreachable` at the end of the function, I believe. As-is, if it makes it to the end of the function it's pretty much UB anyways :shrug:

libraries/libfc/include/fc/exception/exception.hpp
linh2931: I was worried removing `bad_alloc` from `FC_LOG_AND_DROP` could affect other users of `FC_LOG_AND_DROP`. You are correct. After `bad_alloc` catch is removed, an exception of `bad_alloc` will be caught by the final `catch ( ... )`, logged, and re-thrown. Old and new versions are equivalent. Thanks.

libraries/libfc/src/crypto/elliptic_webauthn.cpp
linh2931: Thanks. I will do an `elog` and then call `__builtin_unreachable` at the end of the function.

heifner (approved):

Issue #256, opened 2022-09-29
reconsider/investigate broad suppression of warnings on softfloat & wasm-jit


spoonincode:
https://github.com/AntelopeIO/leap/blob/24b23ab3b98b0a616bf1f6baba2cf2fb89de4158/libraries/CMakeLists.txt#L10-L14

There seems to be an interest in fixing warnings, so bringing this to attention. The comment says these are third party libraries, and that is true, but they're also third party libraries we completely "own" in that they're part of the repo (wasm-jit) or forks that we maintain (softfloat). It's quite possible the warnings are benign, but these are important components of the software where maybe we'd like to know if something more critical ever showed up one day in a future compiler version or such.

PR #263, opened 2022-09-30
[3.2] reenable warnings for softfloat and wasm-jit, and fix a warning in wasm-jit

linh2931:
Resolve https://github.com/AntelopeIO/leap/issues/256

Warnings were disabled for `softfloat` and `wasm-jit`. From https://github.com/AntelopeIO/leap/issues/256:

> The comment says these are third party libraries, and that is true, but they're also third party libraries we completely "own" in that they're part of the repo (`wasm-jit`) or forks that we maintain (`softfloat`). It's quite possible the warnings are benign, but these are important components of the software where maybe we'd like to know if something more critical ever showed up one day in a future compiler version or such.

After warnings were re-enabled, only following warnings were reported. They were not serious.  Warning in `wasm-jit` is fixed in this PR. Warnings in softfloat will be fixed in a separate PR (as it is a submodule).

```
/home/lh/work/leap-main/libraries/wasm-jit/Source/WAST/NFA.cpp:517:54: warning: ‘void* memset(void*, int, size_t)’ clearing an object of non-trivial type ‘NFA::CharSet’ {aka ‘struct DenseStaticIntSet<unsigned char, 256>’}; use    assignment or value-initialization instead [-Wclass-memaccess]
  517 |   memset(classCharSets,0,sizeof(CharSet) * numClasses);
      |                                                      ^
/home/lh/work/leap-main/libraries/softfloat/source/s_roundPackToF16.c:106:2: warning: label ‘packReturn’ defined   but not used [-Wunused-label]
  106 |  packReturn:
/home/lh/work/leap-main/libraries/softfloat/source/s_roundPackToF32.c:106:2: warning: label ‘packReturn’ defined   but not used [-Wunused-label]
  106 |  packReturn:
/home/lh/work/leap-main/libraries/softfloat/source/s_roundPackToF64.c:110:2: warning: label ‘packReturn’ defined   but not used [-Wunused-label]
  110 |  packReturn:

```


libraries/wasm-jit/Source/WAST/NFA.cpp
```diff
@@ -514,7 +514,7 @@ namespace NFA
 		std::set<StateIndex> terminalStates;
 
 		CharSet* classCharSets = (CharSet*)alloca(sizeof(CharSet) * numClasses);
-		memset(classCharSets,0,sizeof(CharSet) * numClasses);
+    memset((char*)classCharSets,0,sizeof(CharSet) * numClasses);
```
spoonincode, line 517: Does this just mask the warning? It's still memseting a non trivial type.

A `CharSet` is a `DenseStaticIntSet<U8,256>` and `DenseStaticIntSet` actually does do a memset on its contents. Since this isn't performance sensitive I wonder if just replacing this with a std::vector or such makes sense

heifner: `dumpDFAGraphViz()` is never called unless `-D_DEBUG` is passed, which it is not.

spoonincode: good call. just zap it all then

linh2931: Did you mean `_DEBUG` flag would never be used and I can go ahead removing all `#ifdef _DEBUG` and the functions which were only used inside `#ifdef _DEBUG` block like `dumpDFAGraphViz`?

spoonincode: Right, `_DEBUG` will never be set, so might as well just remove this code since it's not too invasive of change.

libraries/wasm-jit/Source/Logging/Logging.cpp
```diff
@@ -10,11 +10,7 @@ namespace Log
 	static bool categoryEnabled[(Uptr)Category::num] =
 	{
 		true, // error
-		#ifdef _DEBUG // debug
-			true,
-		#else
-			false,
-		#endif
+	  false, // debug
```
spoonincode, line 13: wasm-jit has a different indentation style

linh2931: Thanks. Fixed. It was caused by a tab (wasm-jit uses tabs for indentation).

spoonincode (approved):

PR #268, opened 2022-10-01
[4.0] Regression test for: fix replay to correctly add trxs into the de-dup list

### LINKED ISSUE #521: leap-util snapshot to-json does not work for existing snapshots
```
➜  22-11-29 13:40:07~/ext/leap/cmake-build-relwithdebinfo/bin ✗ ./leap-util snapshot to-json --input-file snapshot-2022-09-12-19-eos-v6-0267673530.bin
info  2022-11-29T19:41:01.556 leap-util platform_timer_accurac:62     compute_and_print_ti ] Checktime timer accuracy: min:0us max:43us mean:6us stddev:8us
info  2022-11-29T19:41:01.556 leap-util controller.cpp:564            startup              ] Starting initialization from snapshot, this may take a significant amount of time
error 2022-11-29T19:43:07.493 leap-util subcommand.cpp:10             print_exception      ] 3250000 protocol_feature_exception: Protocol feature exception
unrecognized protocol feature digest: 0ec7e080177b2c02b278d5088611686b49d739925a92d9bfcacd7fc6b74053bd
    {"digest":"0ec7e080177b2c02b278d5088611686b49d739925a92d9bfcacd7fc6b74053bd"}
    leap-util  protocol_feature_manager.cpp:718 activate_feature
```

heifner:
See https://github.com/eosnetworkfoundation/mandel/issues/301

When replaying a block log, the transactions with expirations equal to the last irreversible block time were not being added to the trx de-dup list. This created a difference in the de-dup list vs starting from state or a snapshot. Note transactions used to be cleared from the de-dup list at the start of a block according to if they are less than (not less than or equal) the pending block time. This optimization has now been removed.

This PR adds a new regression test for replay to verify it correctly creates the same snapshot as starting from a snapshot.
It includes a fix for not being able to load snapshots with protocol features activated.

See #356 
Resolves #521


linh2931:
I will manually test your fix tomorrow or Sunday.

linh2931:
Tested the fix manually for 3 times. The problem was fixed. Before the fix, I could always reproduce the problem.

linh2931 (approved):

766C6164:
@heifner Thanks for updating leap-util to fix handling protocol features!

766C6164 (approved):

tests/CMakeLists.txt
```diff
@@ -140,6 +141,8 @@ add_test(NAME launcher_test COMMAND tests/launcher_test.py -v --clean-run WORKIN
 set_property(TEST launcher_test PROPERTY LABELS nonparallelizable_tests)
 add_test(NAME keosd_auto_launch_test COMMAND tests/keosd_auto_launch_test.py WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 set_property(TEST keosd_auto_launch_test PROPERTY LABELS nonparallelizable_tests)
+add_test(NAME nodeos_snapshot_diff_test COMMAND tests/nodeos_snapshot_diff_test.py -v --clean-run --dump-error-detail WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
```
linh2931, line 144: `--dump-error-detail` should be removed. It is not needed any more.

heifner: Thanks for catching that.

linh2931 (approved):

PR #280, opened 2022-10-04
Feature bn256 benchmark

plabanca:
Added benchmark tests for bn256

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -30,4 +30,4 @@
 	url = https://github.com/AntelopeIO/libff
 [submodule "libraries/libfc/libraries/bn256"]
 	path = libraries/libfc/libraries/bn256
-	url = https://github.com/AntelopeIO/bn256.git
+	url = https://github.com/AntelopeIO/bn256
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,6 @@
 [submodule "libraries/libfc/libraries/ff"]
 	path = libraries/libfc/libraries/ff
 	url = https://github.com/AntelopeIO/libff
+[submodule "libraries/libfc/libraries/bn256"]
+	path = libraries/libfc/libraries/bn256
+	url = https://github.com/AntelopeIO/bn256.git
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -1,5 +1,5 @@
 file(GLOB BENCHMARK "*.cpp")
-add_executable( benchmark ${BENCHMARK} bn_256.cpp)
+add_executable( benchmark ${BENCHMARK} )
 
 target_link_libraries( benchmark fc bn256 Boost::program_options )
 target_include_directories( benchmark PUBLIC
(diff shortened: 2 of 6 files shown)


spoonincode:
the bn256 CMake needs some attention:
https://github.com/AntelopeIO/bn256/blob/c3ae49fb432055650bcbaaa4d446e29e3562d7a5/src/CMakeLists.txt#L1
unnecessary since this is already specified at root CMakeLists.txt

https://github.com/AntelopeIO/bn256/blob/c3ae49fb432055650bcbaaa4d446e29e3562d7a5/src/CMakeLists.txt#L3
that should be removed

https://github.com/AntelopeIO/bn256/blob/c3ae49fb432055650bcbaaa4d446e29e3562d7a5/src/CMakeLists.txt#L4
Instead of using `${Boost_INCLUDE_DIRS}` should use `target_link_libraries(bn256 Boost::boost)`. However if that repo is intended for other projects to use, or to be built standalone, you really also need a `find_package(boost)` somewhere in the project (regardless of Boost_INCLUDE_DIRS or Boost::boost)

https://github.com/AntelopeIO/bn256/blob/c3ae49fb432055650bcbaaa4d446e29e3562d7a5/test/CMakeLists.txt#L2
should be removed

https://github.com/AntelopeIO/bn256/blob/c3ae49fb432055650bcbaaa4d446e29e3562d7a5/test/CMakeLists.txt#L14
LINK_PUBLIC is depreciated, there should be no need for it

https://github.com/AntelopeIO/bn256/blob/c3ae49fb432055650bcbaaa4d446e29e3562d7a5/test/CMakeLists.txt#L6
this line shouldn't be needed at all once you convert to Boost::boost usage and get rid of LINK_PUBLIC

https://github.com/AntelopeIO/bn256/blob/c3ae49fb432055650bcbaaa4d446e29e3562d7a5/test/CMakeLists.txt#L24
the TARGET_FILE stuff is not required

benchmark/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 file(GLOB BENCHMARK "*.cpp")
-add_executable( benchmark ${BENCHMARK} )
+add_executable( benchmark ${BENCHMARK} bn_256.cpp)
```
spoonincode, line 2: Unneeded -- `BENCHMARK` is a glob of all cpp files

.gitmodules
```diff
@@ -28,3 +28,6 @@
 [submodule "libraries/libfc/libraries/ff"]
 	path = libraries/libfc/libraries/ff
 	url = https://github.com/AntelopeIO/libff
+[submodule "libraries/libfc/libraries/bn256"]
+	path = libraries/libfc/libraries/bn256
+	url = https://github.com/AntelopeIO/bn256.git
```
spoonincode, line 33: All of the other submodules are without `.git` at the end. It doesn't affect anything, but for consistency consider making it like the others.

plabanca: resolve with https://github.com/AntelopeIO/leap/pull/280/commits/1404bb83e7c81bc50a2d4b0c7ba70b787870badc

benchmark/CMakeLists.txt
plabanca: resolve with https://github.com/AntelopeIO/leap/pull/280/commits/f1f96883e1bae16a13161a456f201695e91f6498

PR #283, opened 2022-10-04
don't skip the "build platforms" job even when there are no platforms to build

spoonincode:
When there are no platforms to build for the platform cache, it conceptually makes sense to skip the platform building job matrix. But skipping a job in the middle of the graph like this causes some grief downstream: we needed to start doing things like `if: always() && needs.Build.result == 'success'` for dependent jobs because the default `success()` rule didn't like to see an ancestor skipped.

But this `if: always()` hack comes with a nasty side effect: it's impossible to cancel the workflow. Apparently a job with `if: always()` is not cancellable. This is particularly annoying when something like NP Tests fail and you have to wait until LR Tests finish before you can retry the workflow.

So instead, the Build Platforms job will now inject a dummy entry in to its matrix when there are no platforms to build. This allows the job to still pass when it has no work to do and all dependent jobs don't need the `if: always()` hack. Thus making the workflow cancellable.

At some point, the discover & build jobs should be bundled in to a reusable workflow to encapsulate away some of the gory details.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,23 +34,24 @@ jobs:
   build-platforms:
     name: Build Platforms
     needs: d
-    if: needs.d.outputs.missing-platforms != '[]'
     strategy:
       fail-fast: false
       matrix:
-        platform: ${{fromJSON(needs.d.outputs.missing-platforms)}}
-    runs-on: ["self-hosted", "enf-x86-beefy"]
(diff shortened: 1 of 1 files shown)


kj4ezj (approved):
:pizza:

spoonincode:
I'm holding up this change until runners can be more robust in handling cancelled jobs.

oschwaldp-oci:
>I'm holding up this change until runners can be more robust in handling cancelled jobs.
> changes effectively incorporated in #1703

Out of curiosity, does this mean that the runners now more robustly handle cancelled jobs?

PR #304, opened 2022-10-07
[PH] Add some additional parameters to performance harness output

ClaytonCalabrese:
Adds a couple of parameters and does a little math to present information to performance harness output.

### DIFF
--- a/tests/performance_tests/log_reader.py
+++ b/tests/performance_tests/log_reader.py
@@ -7,6 +7,9 @@ import numpy as np
 import json
 from datetime import datetime
 import glob
+import multiprocessing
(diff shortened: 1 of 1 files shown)


oschwaldp-oci (approved):
Looks good.

At some point I'd like to get the `numGenerators` and `tpsPerGenerator` values passed in so that they are only calculated in one place.  So either move those calculations out of `launch_transaction_generators.py` into `performance_test_basic.py` and pass them into `launch_transaction_generators.py` and `createJSONReport` or get the values back from `launch_transaction_generators.py`.  We can tackle this at a later date though.

PR #310, opened 2022-10-10
Add prometheus-cpp as a submodule

ndcgundlach:
Added prometheus-cpp as a submodule, created a new prometheus-core library in leap that only pulls the parts of prometheus that are needed.


spoonincode:
Says adding it as submodule, but it seems like the repo contents were added too?

ndcgundlach:
> Says adding it as submodule, but it seems like the repo contents were added too?

Yeah, I b0rked it up.  I am killing this PR and creating a new one.

PR #314, opened 2022-10-12
[PH] Add method to wait and check blocks for transactions

ClaytonCalabrese:
Adds a function that will wait for a list of transactions to appear in a block. Checks incoming blocks for transactions and removes them from the list. Returns upon timeout or when list has all been seen.

### DIFF
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -498,11 +498,11 @@ class Node(object):
                     transIds.pop(trx['id'])
         return transIds
 
-    def waitForTransactionsInBlockRange(self, transIds, startBlock=2, maxFutureBlocks=60):
+    def waitForTransactionsInBlockRange(self, transIds, startBlock=2, maxFutureBlocks=None):
         lastBlockProcessed = startBlock
-        overallFinalBlock = sys.maxsize
+        overallFinalBlock = self.getHeadBlockNum()
         if maxFutureBlocks is not None:
-            overallFinalBlock = self.getHeadBlockNum() + maxFutureBlocks
+            overallFinalBlock = overallFinalBlock + maxFutureBlocks
         while len(transIds) > 0:
             currentLoopEndBlock = self.getHeadBlockNum()
             if currentLoopEndBlock >  overallFinalBlock:
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -6,6 +6,7 @@ import os
 import re
 import json
 import signal
+import sys
 
 from datetime import datetime
 from datetime import timedelta
@@ -488,6 +489,32 @@ class Node(object):
         lam = lambda: self.isTransInAnyBlock(transId)
         ret=Utils.waitForBool(lam, timeout)
         return ret
+
+    def checkBlocksForTransactions(self, transIds, lastBlockProcessed, endBlock):
+        for blockNum in range(endBlock, lastBlockProcessed + 1, -1):
+            block = self.processCurlCmd("trace_api", "get_block", f'{{"block_num":{blockNum}}}', silentErrors=False, exitOnError=True)
+            if block['transactions']:
+                for trx in block['transactions']:
+                    if trx['id'] in transIds:
+                        transIds.pop(trx['id'])
+        return transIds
+
+    def waitOnBlockTransactions(self, transIds, startBlock=1, timeout=30):
+        lastBlockProcessed = startBlock
+        finalBlock = sys.maxsize
+        if timeout is not None:
+            finalBlock = self.getHeadBlockNum() + timeout * 2
+        while len(transIds) > 0:
+            self.waitForHeadToAdvance()
+            endBlock = self.getHeadBlockNum()
+            if endBlock > finalBlock:
+                endBlock = finalBlock
+            print(lastBlockProcessed, endBlock)
+            transIds = self.checkBlocksForTransactions(transIds, lastBlockProcessed, endBlock)
+            lastBlockProcessed = endBlock
+            if endBlock == finalBlock:
+                break
+        return transIds
 
     def waitForTransFinalization(self, transId, timeout=None):
         """Wait for trans id to be finalized."""
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -490,30 +490,32 @@ class Node(object):
         ret=Utils.waitForBool(lam, timeout)
         return ret
 
-    def checkBlocksForTransactions(self, transIds, lastBlockProcessed, endBlock):
-        for blockNum in range(endBlock, lastBlockProcessed + 1, -1):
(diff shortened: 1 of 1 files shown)


tests/TestHarness/Node.py
```diff
@@ -489,6 +490,32 @@ def waitForTransInBlock(self, transId, timeout=None):
...
+                        transIds.pop(trx['id'])
+        return transIds
+
+    def waitOnBlockTransactions(self, transIds, startBlock=1, timeout=30):
+        lastBlockProcessed = startBlock
+        finalBlock = sys.maxsize
+        if timeout is not None:
+            finalBlock = self.getHeadBlockNum() + timeout * 2
+        while len(transIds) > 0:
+            self.waitForHeadToAdvance()
```
oschwaldp-oci, line 508: Probably don't want to add this additional wait into the algorithm for the first time through the while loop.

tests/TestHarness/Node.py
```diff
@@ -489,6 +490,32 @@ def waitForTransInBlock(self, transId, timeout=None):
...
+            endBlock = self.getHeadBlockNum()
+            if endBlock > finalBlock:
+                endBlock = finalBlock
+            print(lastBlockProcessed, endBlock)
+            transIds = self.checkBlocksForTransactions(transIds, lastBlockProcessed, endBlock)
```
oschwaldp-oci, line 513: For this algorithm to work with basing the timeout on block production, it would appear that this function should only process one block at a time and the loop/range be managed from withing `waitOnBlockTransactions`.  This will have some benefits:

1. Can detect and manage timeout more accurately
2. Can stop querying for additional blocks in initial range if, say, all transactions were found in the first block returned from `get_block`
3. Could allow for future optimizations if empirical data suggests wanting to do a search pattern that starts checking blocks in chronological reverse order until head advances, then check the newest block before continuing with the existing blocks in reverse chronological order.

oschwaldp-oci: Note: If updating to only query/process one block at a time, it will be important to add handling around the `waitForHeadToAdvance` so that it isn't always waiting when it still has blocks to process.

tests/TestHarness/Node.py
```diff
@@ -489,6 +490,34 @@ def waitForTransInBlock(self, transId, timeout=None):
...
+        while len(transIds) > 0:
+            endBlock = self.getHeadBlockNum()
+            if endBlock >  finalBlock:
+                endBlock = finalBlock
+            for blockNum in range(endBlock, lastBlockProcessed - 1, -1):
+                transIds = self.checkBlockForTransactions(transIds, blockNum)
+                if len(transIds) == 0:
+                    return transIds
+            lastBlockProcessed = endBlock
+            if endBlock == finalBlock:
```
oschwaldp-oci, line 515: Probably more accurate to use `self.getHeadBlockNum() >= finalBlock` instead of `endBlock` here in case enough time has passed that head has advanced.

tests/TestHarness/Node.py
```diff
@@ -489,6 +490,34 @@ def waitForTransInBlock(self, transId, timeout=None):
...
+        return transIds
+
+    def waitOnBlockTransactions(self, transIds, startBlock=2, timeout=30):
+        lastBlockProcessed = startBlock
+        finalBlock = sys.maxsize
+        if timeout is not None:
+            finalBlock = self.getHeadBlockNum() + timeout * 2
+        while len(transIds) > 0:
+            endBlock = self.getHeadBlockNum()
+            if endBlock >  finalBlock:
```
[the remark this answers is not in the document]
ClaytonCalabrese: No. There is no guarantee we are going 1 block at a time. We reduce to out timeout limit if we are above it, and if we don't find all the transactions we time out.

This function is checking X number of blocks for a list of transactions, not waiting X number of blocks to be produced for a list of transactions.

tests/TestHarness/Node.py
```diff
@@ -489,6 +490,34 @@ def waitForTransInBlock(self, transId, timeout=None):
...
+                endBlock = finalBlock
+            for blockNum in range(endBlock, lastBlockProcessed - 1, -1):
```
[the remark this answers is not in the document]
ClaytonCalabrese: Same as above comment

tests/TestHarness/Node.py
ClaytonCalabrese: This falls into the same group as the previous comments. Additionally, the greater than is handled by lines 508 and 509 meaning it can never be greater than by this point in the code.

tests/TestHarness/Node.py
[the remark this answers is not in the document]
oschwaldp-oci: I misunderstood `timeout` to be trying to constrain the amount of time spent in the `waitOnBlockTransactions` function call.  I now understand it to be a book end on how many additional blocks this function will wait to be produced and audited for transactions before exiting and reporting not all trxs were found.

tests/TestHarness/Node.py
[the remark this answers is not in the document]
oschwaldp-oci: I misunderstood timeout to be trying to constrain the amount of time spent in the waitOnBlockTransactions function call. I now understand it to be a book end on how many additional blocks this function will wait to be produced and audited for transactions before exiting and reporting not all trxs were found.

tests/TestHarness/Node.py
oschwaldp-oci: I misunderstood timeout to be trying to constrain the amount of time spent in the waitOnBlockTransactions function call. I now understand it to be a book end on how many additional blocks this function will wait to be produced and audited for transactions before exiting and reporting not all trxs were found.

tests/TestHarness/Node.py
```diff
@@ -489,6 +490,34 @@ def waitForTransInBlock(self, transId, timeout=None):
...
 
+    def checkBlockForTransactions(self, transIds, blockNum):
+        block = self.processCurlCmd("trace_api", "get_block", f'{{"block_num":{blockNum}}}', silentErrors=False, exitOnError=True)
+        if block['transactions']:
+            for trx in block['transactions']:
+                if trx['id'] in transIds:
+                    transIds.pop(trx['id'])
+        return transIds
+
+    def waitOnBlockTransactions(self, transIds, startBlock=2, timeout=30):
```
oschwaldp-oci, line 501: Could you rename `timeout` to be something that suggests it is a ceiling on the number of additional blocks the function is willing to wait to be produced to check for transactions before returning (`"ERROR: Transactions were missing upon expiration of waitOnblockTransactions"`)?

oschwaldp-oci (changes requested):

tests/TestHarness/Node.py
oschwaldp-oci: I misunderstood `timeout` to be trying to constrain the amount of time spent in the waitOnBlockTransactions function call. I now understand it to be a book end on how many additional blocks this function will wait to be produced and audited for transactions before exiting and reporting not all trxs were found.

oschwaldp-oci (approved):

PR #333, opened 2022-10-17
Add pull functionality

ndcgundlach:
Added the ability to do handle Prometheus pull requests.  Added metrics for the prometheus_plugin.

### DIFF
--- a/plugins/prometheus_plugin/prometheus_plugin.cpp
+++ b/plugins/prometheus_plugin/prometheus_plugin.cpp
@@ -26,13 +26,7 @@ namespace eosio {
    using namespace prometheus;
    static appbase::abstract_plugin &_prometheus_plugin = app().register_plugin<prometheus_plugin>();
 
-   struct prometheus_plugin_impl {
-      Registry _registry;
-
-      std::mutex _collectables_mutex;
-      std::vector<Collectable> _collectables;
-
-      // metrics for prometheus_plugin itself
+   struct prometheus_plugin_metrics {
       Family<Counter>& _bytes_transferred_family;
       Counter& _bytes_transferred;
       Family<Counter>& _num_scrapes_family;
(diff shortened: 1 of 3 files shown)


plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -1,19 +1,168 @@
...
 #include <fc/log/logger.hpp>
+#include <prometheus/metric_family.h>
+#include <prometheus/collectable.h>
+#include <prometheus/counter.h>
+#include <prometheus/summary.h>
+#include <prometheus/text_serializer.h>
+#include <prometheus/registry.h>
+
+namespace prometheus {
+   std::vector<MetricFamily> CollectMetrics(
```
heifner, line 11: Use snake_case for all types.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -1,19 +1,168 @@
...
+                                  std::make_move_iterator(metrics.end()));
+      }
+
+      return collected_metrics;
+   }
+
+   class MetricsHandler {
+   public:
+      std::mutex collectables_mutex_;
+      std::vector<std::weak_ptr<Collectable>> collectables_;
```
heifner, line 33: I don't see any reason to use `weak_ptr` or `shared_ptr` as these will be registered at startup and never removed. We do not unload plugins and I don't see us ever unloading a collection.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -1,19 +1,168 @@
...
+           num_scrapes_(num_scrapes_family_.Add({})),
+           request_latencies_family_(
+                 BuildSummary()
+                       .Name("exposer_request_latencies")
+                       .Help("Latencies of serving scrape requests, in microseconds")
+                       .Register(registry)),
+           request_latencies_(request_latencies_family_.Add(
+                 {}, Summary::Quantiles{{0.5, 0.05}, {0.9, 0.01}, {0.99, 0.001}})) {}
+
+      void cleanup_stale_pointers(
```
heifner, line 61: I don't think we will need this.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -1,19 +1,168 @@
...
+      }
+
+      void register_collectable(
+            const std::weak_ptr<Collectable>& collectable) {
+         std::lock_guard<std::mutex> lock{collectables_mutex_};
+         cleanup_stale_pointers(collectables_);
+         collectables_.push_back(collectable);
+      }
+
+      void remove_collectable(
```
heifner, line 78: I don't think we will need this.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -1,19 +1,168 @@
...
+         _metrics_handler->register_collectable(collectable);
+      }
+
+      std::string scrape() {return _metrics_handler->scrape();}
    };
 
    prometheus_plugin::prometheus_plugin() {
      my.reset(new prometheus_plugin_impl{});
+
+      app().get_plugin<http_plugin>().add_api({
```
heifner, line 158: This can use `add_async_api`. `add_async_api` is a terrible name for that, but it registers it to run on the http_thread instead of the main thread, which is what we will want for this.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -1,19 +1,168 @@
...
+         auto&& metrics = collectable->Collect();
+         collected_metrics.insert(collected_metrics.end(),
+                                  std::make_move_iterator(metrics.begin()),
+                                  std::make_move_iterator(metrics.end()));
+      }
+
+      return collected_metrics;
+   }
+
+   class MetricsHandler {
```
heifner, line 30: snake_case

heifner (changes requested):

PR #341, opened 2022-10-17
Complete Pinned Build Instructions + Other README Updates

kj4ezj:
From [issue 132](https://github.com/AntelopeIO/leap/issues/132), this pull request updates the build instructions in the `README.md` to be more approachable to new users or novices who wish to build from source without much Linux or BASH experience. [Here](https://github.com/AntelopeIO/leap/blob/2aac9530395773ab958a96815da0ed7efd87cbae/README.md) is what it looks like now. It may be easier to look over the new document than to try to look through the pull request diff.

I felt motivated so I also improved grammar and readability of the `README.md` overall, simplified the section headers, expanded the explanation of pinned vs. unpinned builds, expanded the section explaining the test suites, and various other things. I was careful to preserve the specific recommendations and commands added by contributors before me.

## See Also
- [Pull Request 340](https://github.com/AntelopeIO/leap/pull/340) - main
- [Pull Request 341](https://github.com/AntelopeIO/leap/pull/341) - release/3.2
- [Pull Request 342](https://github.com/AntelopeIO/leap/pull/342) - release/3.1

Issue #132, opened 2022-09-09
Pinned build: instructions are not complete 


matthewdarwin:
Reading the README for the pinned build, it seems that the details to run `git submodule update --init --recursive` is not there.

matthewdarwin:
A more detailed example would be helpful. Something like maybe:

```
git clone -b release/3.1 https://github.com/AntelopeIO/leap.git
cd leap
git checkout tags/v3.1.0
git submodule update --init --recursive
sudo ./scripts/install_deps.sh
mkdir /home/user/mybuild
mkdir /home/user/mylib
./scripts/pinned_build.sh /home/user/mylib /home/user/mylib 8
```

PR #342, opened 2022-10-17
Complete Pinned Build Instructions + Other README Updates [docs]

kj4ezj:
From [issue 132](https://github.com/AntelopeIO/leap/issues/132), this pull request updates the build instructions in the `README.md` to be more approachable to new users or novices who wish to build from source without much Linux or BASH experience. [Here](https://github.com/AntelopeIO/leap/blob/c3d3a02be2ac4cb57b30f7210739b0cd3f9536fe/README.md) is what it looks like now. It may be easier to look over the new document than to try to look through the pull request diff.

I felt motivated so I also improved grammar and readability of the `README.md` overall, simplified the section headers, expanded the explanation of pinned vs. unpinned builds, expanded the section explaining the test suites, and various other things. I was careful to preserve the specific recommendations and commands added by contributors before me.

## See Also
- ~~[Pull Request 340](https://github.com/AntelopeIO/leap/pull/340) - main~~
- ~~[Pull Request 341](https://github.com/AntelopeIO/leap/pull/341) - release/3.2~~
- [Pull Request 342](https://github.com/AntelopeIO/leap/pull/342) - release/3.1


README.md
```diff
@@ -1,126 +1,241 @@
...
+```bash
 ctest -L "long_running_tests"
 ```
+
+### Install
+Once you have [built](#build) Leap and [tested](#test) your build, you can install Leap to your system.
+
+Navigate to your Leap build directory and run this command.
+```bash
+sudo make install
```
spoonincode, line 239: The instructions above do a `make package` which will pop out a deb and/or rpm file. I think somehow we should encourage users to use those before resorting to a `make install`

README.md
```diff
@@ -1,126 +1,241 @@
...
+```
 
-Generally we recommend performing what we refer to as a "pinned build" which ensures the compiler and boost version remain the same between builds of different Leap versions (Leap requires these versions remain the same otherwise its state needs to be repopulated from a portable snapshot).
+## Source Installation
+You can also build and install Leap from source.
+
+### Prerequisites
+Recent Ubuntu LTS releases are the only operating systems that we support. Other Unix derivatives such as macOS are tended to on a best-effort basis and may not be full featured. If you aren't using Ubuntu, please visit [this page](./docs/00_install/01_build-from-source/00_build-unsupported-os.md) to explore your options.
+
+Notable requirements to build are:
```
spoonincode, line 48: Since this list if being changed to be exhaustive, can remove the "Notable" word. You'll listed everything.

README.md
```diff
@@ -1,126 +1,241 @@
...
+Choose which [release tag](https://github.com/AntelopeIO/leap/releases) or [branch](#branches) you would like to build and check it out. If you don't know, use the [latest release](https://github.com/AntelopeIO/leap/releases/latest). For example, if you want to build release 3.1.2, you would check it out using its tag.
+```bash
+git checkout v3.1.2
+```
 
-#### Building Pinned Build Binary Packages
-In the directory `<leap src>/scripts` you will find the two scripts `install_deps.sh` and `pinned_build.sh`. If you haven't installed build dependencies then run `install_deps.sh`. Then run `pinned_build.sh <dependencies directory> <leap build directory> <number of jobs>`.
+Once you are on the branch or tag (release) you want to build, make sure everything is up-to-date.
+```bash
+git pull
```
spoonincode, line 86: Depending how exhaustive we're trying to be with the instructions (and it seems now, rather exhaustive), then maybe somehow the `git pull` needs to be prior to the `git checkout` so that the latest tags are pulled before doing the checkout should someone being using an existing checkout.

README.md
```diff
@@ -1,126 +1,241 @@
...
+```bash
+git clone --recursive git@github.com:AntelopeIO/leap.git
+```
+An HTTPS and SSH clone both result in the same thing, a folder named `leap` containg our source code, it doesn't matter which one you use. Navigate into that folder.
+```bash
+cd leap
+```
+Choose which [release tag](https://github.com/AntelopeIO/leap/releases) or [branch](#branches) you would like to build and check it out. If you don't know, use the [latest release](https://github.com/AntelopeIO/leap/releases/latest). For example, if you want to build release 3.1.2, you would check it out using its tag.
+```bash
+git checkout v3.1.2
```
spoonincode, line 81: Is it worth using a non-existent version as the example version? So if someone blindly copy-pastes it over it errors out.

README.md
```diff
@@ -1,126 +1,241 @@
...
 
-The binary package will be produced in the Leap build directory that was supplied.
+### Build
+We have two types of builds for Leap, "pinned" and "unpinned." The only difference is that pinned builds use specific versions for some dependencies hand-picked by Leap engineers, whereas unpinned builds use the default dependency versions available on the build system at the time. We recommend performing a pinned build to ensure the compiler and boost version remain the same between builds of different Leap versions. Leap requires these versions to remain the same, otherwise its state needs to be recovered from a portable snapshot or the chain needs to be replayed.
 
-#### Manual (non "pinned") Build Instructions
+For the curious, the "pinned" terminology comes from our days developing on macOS where you would run `brew pin` to lock one of the Leap dependencies to a specific version despite available upgrades.
```
spoonincode, line 97: eh I'd recommend removing. Personally I don't recall homebrew pinning being the etymology of the "pinned build" nomenclature.

README.md
```diff
@@ -1,126 +1,241 @@
...
-git submodule update --init --recursive
-mkdir build
+- If you are in an Ubuntu docker container, omit `sudo` because you run as `root` by default.
+From a terminal in the root of the `leap` repo, build.
+```bash
+mkdir -p build
 cd build
 cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_PREFIX_PATH=/usr/lib/llvm-11 ..
-make -j $(nproc) package
+make -j "$(( $(nproc) - 2 ))" package
```
spoonincode, line 141: Is this a bashism? I don't recall how widely implemented `$(())` is. What happens if user only has two cores? Change seems unnecessary imo

README.md
kj4ezj: Good catch! What about `git fetch`? That should pull down new tags and branches without spending data/time on the current branch if it is not wanted.

README.md
kj4ezj: Yes, I like that idea.

README.md
kj4ezj: Will remove. Maybe I made this up? Haha.

README.md
kj4ezj: Good question. I don't know, so I tried it in DASH (used as `/bin/sh` in Ubuntu and other Debian-family Linuxes).

2022-10-17 19-08-19 - arithmetic in DASH

It worked! I don't use ZSH and I don't know if we're expected to support it, but I found online [z-shell also supports arithmetic operations](https://www.csse.uwa.edu.au/programming/linux/zsh-doc/zsh_10.html).

I also found online they still sell dual-core processors. I did not know that. I thought they stopped selling dual cores ten years ago. That's my bad. Even if that were true, you could have a dual-core virtual machine. I tried passing `0` and `-1` to `make -j` in a virtual machine and it rejected it.

It is also just obtuse syntax if we are writing this with an eye towards less experienced customers. I will remove it.

README.md
kj4ezj: I added instructions to install the binary, including a note that we recommend this route, before the `make install` instructions.

README.md
kj4ezj: Done. Idk if it is exhaustive or not, but I changed it to "Requirements to build."

README.md
kj4ezj: Added `git fetch --all --tags`. Let me know if there is some reason that won't work.

README.md
kj4ezj: Done, used `v0.0.0`.

README.md
kj4ezj: Removed.

README.md
kj4ezj: Removed.

README.md
```diff
@@ -1,126 +1,248 @@
...
+```
+Your download path may vary.
+- If you are in an Ubuntu docker container, omit `sudo` because you run as `root` by default.
 
-Generally we recommend performing what we refer to as a "pinned build" which ensures the compiler and boost version remain the same between builds of different Leap versions (Leap requires these versions remain the same otherwise its state needs to be repopulated from a portable snapshot).
+Finally, verify Leap was installed correctly.
+```bash
+nodeos --full-version
+```
+You should see a [semantic version](https://semver.org) string print out followed by a `git` commit hash with no errors. For example:
```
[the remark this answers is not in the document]
kj4ezj: From [Merriam-Webster](https://www.merriam-webster.com/dictionary/printout), "print out" is a verb whereas "printout" is a noun. I was using it as a verb here, though either usage would make sense.

I just went ahead and removed "print" altogether because the sentence still makes sense without it and I don't want non-technical users to wonder if we might actually print something to their printer. As silly as it may sound, I've wondered that in the past reading documentation so it might be best to avoid the word "print" altogether.

README.md
```diff
@@ -1,126 +1,248 @@
...
-Leap is blockchain node software and supporting tools that implements the [Antelope](https://github.com/AntelopeIO) protocol.
+## Branches
+The `main` branch is the development branch, do not use it for production. Refer to the [release page](https://github.com/AntelopeIO/leap/releases) for current information on releases, pre-releases, and obsolete releases, as well as the corresponding tags for those releases.
 
-## Repo Organization
+## Binary Installation
+This is the fastest way to get started. Download a binary for the [latest](https://github.com/AntelopeIO/leap/releases/latest) release, or visit the [releases](https://github.com/AntelopeIO/leap/releases) page to download a binary for a specific version of Leap.
 
-`main` branch is the development branch: do not use this for production. Refer to the [release page](https://github.com/AntelopeIO/leap/releases) for current information on releases, pre-releases, and obsolete releases as well as the corresponding tags for those releases.
+We currently only support Ubuntu. If you aren't using Ubuntu, please visit [this page](./docs/00_install/01_build-from-source/00_build-unsupported-os.md) to explore your options.
```
lparisc, line 10: I would use a more structured way to specify supported OSes, even if it's only one. For instance, look at https://github.com/AntelopeIO/leap/blob/main/docs/00_install/index.md#supported-operating-systems.

README.md
```diff
@@ -1,126 +1,248 @@
 # Leap
+Leap is blockchain node software and supporting tools that implement the [Antelope](https://github.com/AntelopeIO) protocol.
 
-Leap is blockchain node software and supporting tools that implements the [Antelope](https://github.com/AntelopeIO) protocol.
+## Branches
+The `main` branch is the development branch, do not use it for production. Refer to the [release page](https://github.com/AntelopeIO/leap/releases) for current information on releases, pre-releases, and obsolete releases, as well as the corresponding tags for those releases.
 
-## Repo Organization
+## Binary Installation
```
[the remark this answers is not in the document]
kj4ezj: > The change from "Software Installation" to "Binary Installation" breaks an anchor URL used in the docs. However, I can see that "binary" is more explicit.

Thank you. Yes, there are three ways to install the software so I wanted to make it explicit what we were doing. I thought about moving all the installation methods to the bottom in a single section, where "Software Installation" would be a good title, but I think we want the binary install up top so newcomers don't do an necessary build only to discover they can use binaries.

What would it take to change those broken links? Is that something I can do?

> Why not sticking to what we used before successfully: "Install from Binaries" for package installation[...]

I don't know, to be honest I don't remember what we had before. I just chose this because it is one word shorter. I am not married to it.

> [...]and "Build from Source" below for building from source, instead of "Source Installation" which is more rare in the literature since we don't really "install" from source.

I figure the end goal for a customer is to install what they built, unless building from source is an academic exercise. [The build](https://github.com/AntelopeIO/leap/blob/8c922dc6909a598020ea8106984c026c42898467/README.md#build) itself has its own section in the document. Moreover, "Source Installation" matches the pattern of "Binary Installation," so it makes it clear there are two options.

Since this section includes more than just a build, I would say "Build and Install from Source" or "Source Build and Install" would be the closest names to what you suggest that make sense for this heading without restructuring the document, which I am not opposed to.

README.md
```diff
@@ -1,126 +1,248 @@
...
+nodeos --full-version
+```
+You should see a [semantic version](https://semver.org) string print out followed by a `git` commit hash with no errors. For example:
+```
+v3.1.2-0b64f879e3ebe2e4df09d2e62f1fc164cc1125d1
+```
 
-#### Building Pinned Build Binary Packages
-In the directory `<leap src>/scripts` you will find the two scripts `install_deps.sh` and `pinned_build.sh`. If you haven't installed build dependencies then run `install_deps.sh`. Then run `pinned_build.sh <dependencies directory> <leap build directory> <number of jobs>`.
+## Source Installation
```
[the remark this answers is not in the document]
kj4ezj: I agree, we should have a single source of truth. I don't like that there wouldn't be build instructions in the `/docs` folder, but I do think it needs to be in the `README.md`. That's where people are going to look. I wish there were a way to embed markdown documents in other markdown documents. We could link to the build instructions from the `README.md`, but I think it would be better for them to live in the `README.md`.

README.md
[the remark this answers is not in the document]
lparisc: Ok I thought it was used a noun. No worries.

README.md
```diff
@@ -1,126 +1,248 @@
...
 ctest -L "long_running_tests"
 ```
+
+### Install
+Once you have [built](#build) Leap and [tested](#test) your build, you can install Leap to your system.
+
+We recommend installing the binary package you just built. Navigate to your Leap build directory in a terminal and run this command.
+```bash
+sudo apt-get update
+sudo apt-get install -y ./leap_*.deb
```
spoonincode, line 240: Just a heads up that the filename of the binaries are different in 3.1 vs 3.2+ due to #120, what is used here is the 3.2 naming, but this PR is targeting 3.1.

README.md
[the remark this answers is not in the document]
spoonincode: Another option, to avoid wholesale duplication, is for the README to be more on the concise side -- a "quick start" -- and then linking to docs for more in depth details. (The README should probably not solely be installation and building instructions as it is today, but also at least links to more resources like the doc portal, discord, etc)

README.md
spoonincode: It should be. The original text was trying to be both exhaustive and concise: requiring something like zlib or openssl is a nothingburger so it didn't deserve being in a list, but needing specific versions of LLVM is notable so it is brought out in to a list.

README.md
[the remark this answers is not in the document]
lparisc: Yes indeed, we could add those instructions back to `/docs` later once the "install from binaries" or "build from source" sections mature within the various README's (including `cdt`, `DUNE`, and `eos-system-contracts` at the ENF org). For now, we can keep those sections in the `/docs` but have small paragraphs that simply link to the corresponding sections in the README.

README.md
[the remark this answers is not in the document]
lparisc: Ok we could use "Binary Installation" here and "Source Build and Install" below (although the latter sounds weird to me). We could also use "Install from Binaries" here and "Build and Install from Source" below. Just throw a dice and pick one 😄.

README.md
kj4ezj: I am not opposed to putting it back the way it was because I agree with you. However, it did seem silly to me to put some in a bulleted list and then others in prose. If they aren't important, we should just exclude them entirely IMHO.

README.md
kj4ezj: Should that change be back-ported to this branch? It kind of seems like a big deal that we aren't following Debian-family package naming conventions, if we intend to cut further releases from this branch.

If we don't intend to cut further releases from this branch, I probably shouldn't bother updating the `README.md`.

I am happy to change the commands here, just curious about your thoughts on that.

README.md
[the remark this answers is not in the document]
kj4ezj: > Another option, to avoid wholesale duplication, is for the README to be more on the concise side -- a "quick start" -- and then linking to docs for more in depth details.

I agree with this. Perhaps we just have binary install and pinned build script instructions here with a link to docs for unpinned builds.

> The README should probably not solely be installation and building instructions as it is today, but also at least links to more resources like the doc portal, discord, etc

Honestly, I found this weird when I started this ticket. Why is this just build and install instructions? I do think those belong in a `README.md` personally, but we only have one sentence explaining what this repo even is.

> Yes indeed, we could add those instructions back to /docs later once the "install from binaries" or "build from source" sections mature within the various README's (including cdt, DUNE, and eos-system-contracts at the ENF org). For now, we can keep those sections in the /docs but have small paragraphs that simply link to the corresponding sections in the README.

This seems to contradict itself. I like the first half, personally. I can copy/pasta the updated instructions into `/docs` and we'll have two copies, or delete the build doc altogether and we'll have one copy.

I am not opposed to moving the build instructions into the `/docs` folder. However, when this story started, there were _already_ two copies of the build instructions. It isn't clear which one the customer was using, but it is beginning to feel like unifying them is a bigger discussion with a lot of stakeholders that requires some strategy and should probably be a separate ticket from the one this pull request intends to address.

For example, the build instructions make sense in the document you linked. However, the stuff I wrote up about test suites really doesn't. How to run the tests, maybe, but descriptions of the test suites is a whole other thing that deserves a document in its own right. Test suites aren't about building. I also think some summary of the test suites (like the table) would belong in a succinct `README.md`.

I suggest we update the build instructions in both places for now, then open a new ticket to track structural changes to the documentation.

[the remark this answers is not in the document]
lparisc: Ah I meant keeping the section titles in the `/docs` (not the content) with small paragraphs linking to the corresponding README sections. I'd say don't worry about the build or install instructions in the `/docs` and just focus on updating the README. I also brought up those instructions in the `/docs` to make you aware that there's applicable content there that might need to be reconciled with the instructions you're writing/fixing in the README.

README.md
spoonincode: I'm not actually so sure the 3.2 naming scheme is pedantically 'correct' either, due to the "ubuntu20.04" names stuck in it. That's not really what the revision field is for. Due to the uncertainty of this really being pedantically correct, I hadn't personally advocated for porting it to the 3.1.x branch. Especially because in 4.0 we expect to get rid of the "ubuntu20.04" in filenames thus truly making it correct at that point in time.

We might still make 3.1.3+ releases. It's still not entirely clear what our support lifecycle is: it's possible 3.1.x becomes an LTS of sorts should we go on to make a 3.3.x. Or not. 3.3x isn't a given anyways. But I think it's okay to continue to update the 3.1 readme.

README.md
```diff
@@ -1,126 +1,248 @@
...
+</details>
 
-**A Warning On Parallel Compilation Jobs (`-j` flag)**: When building C/C++ software often the build is performed in parallel via a command such as `make -j $(nproc)` which uses the number of CPU cores as the number of compilation jobs to perform simultaneously. However, be aware that some compilation units (.cpp files) in Leap are extremely complex and will consume nearly 4GB of memory to compile. You may need to reduce the level of parallelization depending on the amount of memory on your build host. e.g. instead of `make -j $(nproc)` run `make -j2`. Failures due to memory exhaustion will typically but not always manifest as compiler crashes.
+Once you have a `*.deb` file downloaded for your version of Ubuntu, you can install it as follows.
+```bash
+sudo apt-get update
+sudo apt-get install -y ~/Downloads/leap*.deb
+```
+Your download path may vary.
+- If you are in an Ubuntu docker container, omit `sudo` because you run as `root` by default.
```
spoonincode, line 32: Maybe consider placing one comment a single time in the instructions along the lines of "these instructions use sudo for commands that require root". It's very distracting to see this caveat next to so many code blocks.

spoonincode (approved):
The technical changes are accurate. I'll leave it to @lparisc to weigh in about the prose and structure changes.

README.md
kj4ezj: I agree, I didn't like that either...even though I did it. The bullet looks out-of-place in many instances.

README.md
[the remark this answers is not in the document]
kj4ezj: Done.

README.md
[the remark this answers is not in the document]
kj4ezj: > Ah I meant keeping the section titles in the /docs (not the content) with small paragraphs linking to the corresponding README sections.

Done.

README.md
kj4ezj: Done.

README.md
kj4ezj: Fixed.

README.md
kj4ezj: I consolidated the statements about `sudo` in docker to just three (iirc) and removed all the bullets.

docs/00_install/01_build-from-source/index.md
```diff
@@ -2,123 +2,19 @@
 content_title: Build Antelope from Source
 ---
 
-The shell scripts previously recommended for building the software have been removed in favor of a build process entirely driven by CMake. Those wishing to build from source are now responsible for installing the necessary dependencies. The list of dependencies and the recommended build procedure are in the README.md file. Instructions are also included for efficiently running the tests.
+The shell scripts previously recommended for building the software have been removed in favor of a build process entirely driven by CMake. Those wishing to build from source are now responsible for installing the necessary dependencies. The list of dependencies and the recommended build procedure are in the [`README.md`](../../../README.md) file. Instructions are also included for efficiently running the tests.
```
lparisc, line 5: We must use absolute URLs to prevent the link from breaking on the docs portal:
https://github.com/AntelopeIO/leap/blob/release/3.1/README.md

docs/00_install/01_build-from-source/index.md
```diff
@@ -2,123 +2,19 @@
...
-
-**A Warning On Parallel Compilation Jobs (`-j` flag)**: When building C/C++ software often the build is performed in parallel via a command such as `make -j $(nproc)` which uses the number of CPU cores as the number of compilation jobs to perform simultaneously. However, be aware that some compilation units (.cpp files) in mandel are extremely complex and will consume nearly 4GB of memory to compile. You may need to reduce the level of parallelization depending on the amount of memory on your build host. e.g. instead of `make -j $(nproc)` run `make -j2`. Failures due to memory exhaustion will typically but not always manifest as compiler crashes.
-
-Generally we recommend performing what we refer to as a "pinned build" which ensures the compiler and boost version remain the same between builds of different mandel versions (mandel requires these versions remain the same otherwise its state needs to be repopulated from a portable snapshot).
+You can also build and install Leap from source. Instructions for that currently live [here](../../../README.md#build-and-install-from-source).
```
lparisc, line 11: Absolute link:
https://github.com/AntelopeIO/leap/blob/release/3.1/README.md#build-and-install-from-source

docs/00_install/01_build-from-source/index.md
```diff
@@ -2,123 +2,19 @@
...
 
 #### Building Pinned Build Binary Packages
```
lparisc, line 13: Just prepend the absolute URL on all links (anchors are still correct):
https://github.com/AntelopeIO/leap/blob/release/3.1/README.md#...

docs/00_install/01_build-from-source/index.md
```diff
@@ -2,123 +2,19 @@
...
-In the directory `<mandel src>/scripts` you will find the two scripts `install_deps.sh` and `pinned_build.sh`. If you haven't installed build dependencies then run `install_deps.sh`. Then run `pinned_build.sh <dependencies directory> <mandel build directory> <number of jobs>`.
-
-The dependencies directory is where the script will pull the C++ dependencies that need to be built with the pinned compiler for building the pinned binaries for binary packaging.
-
-The binary package will be produced in the mandel build directory that was supplied.
+The pinned build instructions have moved [here](../../../README.md#pinned-build). You may want to look at the [prerequisites](../../../README.md#prerequisites) and our warning on parallelization using the `-j` jobs flag [here](../../../README.md#jobs-flag) before you build.
```
lparisc, line 14: Just prepend the absolute URL on all links (anchors are still correct):
https://github.com/AntelopeIO/leap/blob/release/3.1/README.md#...

docs/00_install/01_build-from-source/index.md
```diff
@@ -2,123 +2,19 @@
...
-git submodule update --init --recursive
-mkdir build
-cd build
-cmake -DCMAKE_C_COMPILER=gcc-8 -DCMAKE_CXX_COMPILER=g++-8 \
-      -DCMAKE_PREFIX_PATH="$HOME/boost1.79;/usr/lib/llvm-7/"  -DCMAKE_BUILD_TYPE=Release .. \
-make -j $(nproc) package
-```
-After building you may remove the `$HOME/boost1.79` directory, or you may keep it around until next time building the software.
-</details>
+The unpinned build instructions have moved [here](../../../README.md#unpinned-build). You may want to look at the [prerequisites](../../../README.md#prerequisites) and our warning on parallelization using the `-j` jobs flag [here](../../../README.md#jobs-flag) before you build.
```
lparisc, line 17: Just prepend the absolute URL on all links (anchors are still correct):
https://github.com/AntelopeIO/leap/blob/release/3.1/README.md#...

docs/00_install/01_build-from-source/index.md
```diff
@@ -2,123 +2,19 @@
...
-```
-cd build
-
-# These tests can't run in parallel but are recommended.
-ctest -L "nonparallelizable_tests"
-
-# These tests can't run in parallel. They also take a long time to run.
-ctest -L "long_running_tests"
-```
+Documentation on available test suites and how to run them has moved [here](../../../README.md#test).
```
lparisc, line 20: Absolute link (w/ corrected anchor):
https://github.com/AntelopeIO/leap/blob/release/3.1/README.md#step-4-test

README.md
```diff
@@ -1,126 +1,258 @@
...
-        pkg-config
+
+Once you are on the branch or release tag you want to build, make sure everything is up-to-date:
+```bash
+git pull
+git submodule update --init --recursive
+```
+
+#### Jobs Flag
+⚠️ **A Warning On Parallel Compilation Jobs (`-j` flag)** ⚠️
```
lparisc, line 99: Might want to make this a quote too by prepending this line with ">". That way we don't distract the user from performing the main task which is building and installing Leap.

README.md
```diff
@@ -1,126 +1,258 @@
...
+```
+
+#### Jobs Flag
+⚠️ **A Warning On Parallel Compilation Jobs (`-j` flag)** ⚠️  
+When building C/C++ software, often the build is performed in parallel via a command such as `make -j "$(nproc)"` which uses all available CPU threads. However, be aware that some compilation units (`*.cpp` files) in Leap will consume nearly 4GB of memory to compile. Failures due to memory exhaustion will typically, but not always, manifest as compiler crashes. Using all available CPU threads may also prevent you from doing other things on your computer during compilation. For these reasons, consider adjusting parallelization.
+
+### Step 3. Build the source code
+Select build instructions below for a [Pinned Build](#pinned-build) (preferred) or an [Unpinned Build](#unpinned-build).
+
+ℹ️ **Pinned Build vs. Unpinned Build** ℹ️
```
lparisc, line 105: Same here, might want to make this a quote too by prepending this line with ">".

README.md
kj4ezj: Done.

README.md
kj4ezj: Done.

docs/00_install/01_build-from-source/index.md
kj4ezj: Done.

docs/00_install/01_build-from-source/index.md
kj4ezj: Done.

docs/00_install/01_build-from-source/index.md
kj4ezj: Done.

docs/00_install/01_build-from-source/index.md
kj4ezj: Done.

lparisc (approved):
Looks great. Just fixed a small typo and made one cosmetic change.

README.md
```diff
@@ -1,126 +1,257 @@
...
+git clone --recursive https://github.com/AntelopeIO/leap.git
+```
+...or SSH:
+```bash
+git clone --recursive git@github.com:AntelopeIO/leap.git
+```
 
-The dependencies directory is where the script will pull the C++ dependencies that need to be built with the pinned compiler for building the pinned binaries for binary packaging.
+> ℹ️ **HTTPS vs. SSH Clone** ℹ️  
+Both an HTTPS or SSH git clone will yield the same result- a folder named `leap` containg our source code. It doesn't matter which type you use.
```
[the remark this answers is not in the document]
lparisc: Just fyi - fixed this already.

docs/00_install/01_build-from-source/index.md
```diff
@@ -2,123 +2,19 @@
 content_title: Build Antelope from Source
 ---
 
-The shell scripts previously recommended for building the software have been removed in favor of a build process entirely driven by CMake. Those wishing to build from source are now responsible for installing the necessary dependencies. The list of dependencies and the recommended build procedure are in the README.md file. Instructions are also included for efficiently running the tests.
+The shell scripts previously recommended for building the software have been removed in favor of a build process entirely driven by CMake. Those wishing to build from source are now responsible for installing the necessary dependencies. The list of dependencies and the recommended build procedure are in the [`README.md`](https://github.com/AntelopeIO/leap/blob/release%2f3.1/README.md) file. Instructions are also included for efficiently running the tests.
```
lparisc, line 5: Btw, any reason for the "%2F" hex character in the branch name, instead of the explicit slash "/"?

README.md
```diff
@@ -1,126 +1,257 @@
...
+make -j "$(nproc)" package
 ```
-After building you may remove the `$HOME/boost1.79` directory, or you may keep it around until next time building the software.
+After building, you may remove the `~/boost1.79` directory or you may keep it around for your next build.
 </details>
 
-### Running Tests
+Now you can optionally [test](#step-4---test) your build, or [install](#step-5---install) the `*.deb` binary packages, which will be in the root of your build directory.
 
-When building from source it's recommended to run at least what we refer to as the "parallelizable tests". Not included by default in the "parallelizable tests" are the WASM spec tests which can add additional coverage and can also be run in parallel.
```
iamveritas, line 105: I have a hard time chewing this `Parallelizable` word, https://en.wiktionary.org/wiki/parallelizable .
can we use `Parallel` instead?
`Parallel tests`
There's a few of occurrences, if you decide to change pls update all of them.

docs/00_install/01_build-from-source/index.md
```diff
@@ -2,123 +2,19 @@
...
-```
-cd build
-
-# These tests can't run in parallel but are recommended.
-ctest -L "nonparallelizable_tests"
-
-# These tests can't run in parallel. They also take a long time to run.
-ctest -L "long_running_tests"
-```
+Documentation on available test suites and how to run them has moved [here](https://github.com/AntelopeIO/leap/blob/release%2f3.1/README.md#test).
```
iamveritas, line 20: please use `/` instead of `%2f`

docs/00_install/01_build-from-source/index.md
iamveritas: can you please use / instead of %2f? it makes it easier to read for the brave soul adventuring to read the md file content

iamveritas (approved):
I am approving this but also leaving a few small suggestions for you to consider.

README.md
```diff
@@ -1,126 +1,257 @@
...
+
+> ⚠️ **A Warning On Parallel Compilation Jobs (`-j` flag)** ⚠️  
+When building C/C++ software, often the build is performed in parallel via a command such as `make -j "$(nproc)"` which uses all available CPU threads. However, be aware that some compilation units (`*.cpp` files) in Leap will consume nearly 4GB of memory to compile. Failures due to memory exhaustion will typically, but not always, manifest as compiler crashes. Using all available CPU threads may also prevent you from doing other things on your computer during compilation. For these reasons, consider adjusting parallelization.
```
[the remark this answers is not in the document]
spoonincode: > If a value it is not specified it uses all available threads

`make -j` doesn't use all available threads: it uses an _unlimited_ number of threads.

> so if we recommend to use half of them it is a safe way for all users

It really depends on the ratio of CPU to memory the user has. A safe recommendation might be something like "gigabytes of memory divided by 3" but this guideline is somewhat fluid and not something I'd recommend trying to nail down in text.

Really what this notice is trying to drive home is that if `make -j <highnumber>` fails, try it with a lower number. It was a common problem users encountered building 3.1 since it consumes more compilation resources than 2.0 did.

docs/00_install/01_build-from-source/index.md
kj4ezj: This is called [URL percent encoding](https://en.wikipedia.org/wiki/Percent-encoding). Here, `%2f` translates to a forward slash (`/`). The purpose of this encoding is to pass characters to a server in a URI that are otherwise not valid.

For example, in this case, we have a protocol (`https://`) and fully qualified domain name (`github.com.`) followed by a filesystem path on the server. The path is:
```js
/${ORGANIZATION}/${REPO}/blob/${BRANCH}/${FILENAME}
```
The branch is `release/3.1` which also contains a forward slash. If we include that forward slash in the URL unmodified, we are telling the server to go into the following folders.
```
AntelopeIO > leap > blob > release > 3.1 > README.md
```
With percent encoding, we are telling the server to go to this path.
```
AntelopeIO > leap > blob > release/3.1 > README.md
```
This conceptually makes more sense but, after doing some reading, I see that `git` does actually create a tree structure for references so it is probably stored this way on the server too.

2022-10-21 11-43-05 - git refs heads folder structure

I can make them literal forward slashes (`/`).

kj4ezj: Done.

README.md
```diff
@@ -1,126 +1,257 @@
...
 
-The binary package will be produced in the Leap build directory that was supplied.
+Navigate into that folder:
+```bash
+cd leap
+```
 
-#### Manual (non "pinned") Build Instructions
+### Step 2 - Checkout Release Tag or Branch
+Choose which [release tag](https://github.com/AntelopeIO/leap/releases) or [branch](#branches) you would like to build, then check it out. If you are not sure, use the [latest release](https://github.com/AntelopeIO/leap/releases/latest). For example, if you want to build release 3.1.2 then you would check it out using its tag, `v3.1.2`. In the example below, replace `v0.0.0` with your selected release tag accordingly:
```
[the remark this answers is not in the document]
kj4ezj: Agreed, I didn't like that either. Fixed.

README.md
kj4ezj: Rephrased.

docs/00_install/01_build-from-source/index.md
kj4ezj: See above comments.

Issue #351, opened 2022-10-19
Promote Dropped Transactions Due to Full Queue to a Warning


ClaytonCalabrese:
In the net plugin, `snprintf(reason, 72, "Dropping trx, too many trx in progress %lu bytes", trx_in_progress_sz);` is a message being reported which drops transactions. A transaction dropped by this will not be evaluated and as such should be reported as a warning.  Currently this message is appearing only starting at the debug level "debug", which can hide these transactions which will not make it into blocks.

heifner:
I would recommend we put this fix in 3.1+. Also recommend we go with `warn` level logging. This can potentially generate a lot of `warn` log entries, but a node in this state likely needs attention as it is not able to keep up with transaction flow.

PR #354, opened 2022-10-19
increase the logging level of dropping a trx due to a full queue to w…

ClaytonCalabrese:
Increased the logging level of dropped transactions due to a full queue to a warning so that they are not hidden in most runs. This is fitting because a dropped transaction will never been seen again and will never make it into the chain, so the user should see this. Also fixed an output bug where ${trx} wasn't printing right.

Resolves https://github.com/AntelopeIO/leap/issues/351


heifner:
Going to take a different approach to this.

PR #355, opened 2022-10-19
[3.2] Dropped trxs when overloaded logged as warning in nodeos

ClaytonCalabrese:
Currently when a transaction is rejected due to too many transactions in progress, that message only gets logged in a "debug" logging level. Rather than spamming that message at a higher level, instead add a new message which will periodically be printed as long as transactions are being rejected.

In this implementation that is the first time a transaction is rejected due to too many transactions, and every time it happens at an interval of 1 second or more.

Resolves: https://github.com/AntelopeIO/leap/issues/351

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -634,6 +634,7 @@ namespace eosio {
       string                  local_endpoint_port;
 
       std::atomic<uint32_t>   trx_in_progress_size{0};
+      fc::time_point          last_dropped_trx_msg_time;
       const uint32_t          connection_id;
       int16_t                 sent_handshake_count = 0;
       std::atomic<bool>       connecting{true};
@@ -656,7 +657,6 @@ namespace eosio {
       block_id_type                    fork_head;
       uint32_t                         fork_head_num{0};
       fc::time_point                   last_close;
-      fc::time_point                   last_dropped_trx_msg_time;
       string                           remote_endpoint_ip;
 
       connection_status get_status()const;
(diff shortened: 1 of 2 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -656,6 +656,7 @@ namespace eosio {
       block_id_type                    fork_head;
       uint32_t                         fork_head_num{0};
       fc::time_point                   last_close;
+      fc::time_point                   last_dropped_trx_msg_time;
```
heifner, line 659: This section of variables are protected by `conn_mtx`, see comment on line 652. I would move it up under `trx_in_progress_size`. It is only accessed via the connection strand, so no mutex is needed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2683,6 +2684,10 @@ namespace eosio {
          char reason[72];
          snprintf(reason, 72, "Dropping trx, too many trx in progress %lu bytes", trx_in_progress_sz);
          my_impl->producer_plug->log_failed_transaction(ptr->id(), ptr, reason);
+         if (fc::time_point::now() - fc::seconds(1) >= last_dropped_trx_msg_time) {
+            last_dropped_trx_msg_time = fc::time_point::now();
+            wlog("Speculative execution is REJECTING transactions, too many trx in progress");
```
heifner, line 2689: Remove "Speculative execution is" as this is not necessarily speculative execution. Also, go ahead and log the `trx_in_progress_sz`. For that matter, might as well just `wlog(reason);`

heifner (changes requested):

heifner:
Should target this to 3.1.

matthewdarwin:
I hope all these failures are well explained with `transaction_failure_tracing` set to debug?

heifner:
> I hope all these failures are well explained with `transaction_failure_tracing` set to debug?

Yes they are included in that log, in the form:
`[TRX_TRACE] Speculative execution is REJECTING tx: ${txid} : Dropping trx, too many trx in progress %lu bytes`

Issue #297, opened 2022-10-06
Measure replay performance difference between current state and replay optimization disabled


stephenpdeos:

heifner:
Setup block log with 2500 TPS for 2500 seconds with 30 sec trx expiration

Runs with skip_recording head block time optimization

* replayed 5594 blocks in 35 seconds, 6.31745682102628248 ms/block
* replayed 5594 blocks in 34 seconds, 6.24119774718398013 ms/block
* replayed 5594 blocks in 35 seconds, 6.27537511174682727 ms/block
* replayed 5594 blocks in 35 seconds, 6.27880922581798639 ms/block

Runs with `skip_recording = false`

* replayed 5594 blocks in 44 seconds, 7.97567173252279638 ms/block
* replayed 5594 blocks in 42 seconds, 7.60711210441623464 ms/block
* replayed 5594 blocks in 42 seconds, 7.56795512247452073 ms/block
* replayed 5594 blocks in 42 seconds, 7.55550885034865072 ms/block

(7.56+7.56+7.61)/3 - (6.24+6.28+6.28)/3 / (7.56+7.56+7.61)/3 = 
(7.58 - 6.27) / 7.58 = .17 = `17%` increase in time

heifner:
Test with EOS mainnet section to see how it compares to this.

heifner:
Runs with skip_recording head block time optimization (db mapped mode) - eos mainnet 161598568 to 161609500.
* replayed 10934 blocks in 420 seconds, 38.42357504801975665 ms/block
* replayed 10934 blocks in 411 seconds, 37.64449108204518524 ms/block
* replayed 10934 blocks in 417 seconds, 38.20364666605689052 ms/block

Runs with skip_recording = false (db mapped mode) - eos mainnet 161598568 to 161609500.
* replayed 10934 blocks in 418 seconds, 38.26522573858959220 ms/block
* replayed 10934 blocks in 417 seconds, 38.19811222903137349 ms/block
* replayed 10934 blocks in 420 seconds, 38.45242568370986902 ms/block

PR #356, opened 2022-10-19
[3.1] Remove replay optimization that skips recording transactions in dedup list

### LINKED ISSUE #291: Replay optimization to avoid adding to de-duplication index could leave index in inconsistent state
I am referring to the replay optimization that is trigger by the status of the following boolean: https://github.com/AntelopeIO/leap/blob/a8d48d406082cc6ec5ef18dddd30b5589b3e870a/libraries/chain/controller.cpp#L1531

It assumes that replay of irreversible blocks will complete up to and including a block with a timestamp that is greater than or equal to `replay_head_time`. This assumption holds if one starts nodeos with a blocks.log file and allows it to finish the replay.

However, we allow the user to interrupt the replay without leaving the chainbase database in a dirty state. The intention is that they can then continue the replay later. But nothing prevents them from truncating the blocks.log and finishing the replay to a block that has a timestamp strictly less than the original `replay_head_time` timestamp.

This can leave nodeos available for syncing and responding to API requests in a state in which its transaction de-duplication index is not in the same state (even after ignoring the entries that should have expired as of that head block time; see https://github.com/AntelopeIO/leap/issues/290).

This shouldn't be a practical safety issue because, assuming no finality violations, the only way the node will reach a block containing potentially malicious duplicate transactions that should be rejected is if they first replay through the branch of the blockchain (containing the irreversible and already known to be good blocks) that it would have originally replayed had it not been interrupted by the operator. And by the time it gets through those blocks, the state of the de-duplication index should correct itself in time to reject the block with the duplicate transaction.

Nevertheless, to provide protection to the nodes even under this usual sequence of events and against an attacker who could violate the finality of blocks, it is best to prevent a correct node from getting to a state where the chainbase database is not dirty and yet has consensus state that differs from other correct nodes in the network that are at the same head block. This also solves a more practical matter that even without finality violation a node that gets into this state could then be used to generate portable snapshots which differ from the ones generated by other nodes for the same block.

So resolving this issue along with https://github.com/AntelopeIO/leap/issues/290 also fixes the additional ways that are currently known to create an inconsistent portable snapshot beyond the ones that were tracked by https://github.com/eosnetworkfoundation/mandel/issues/301 and fixed by https://github.com/AntelopeIO/leap/pull/273.

One very simple way to fix the undesired behavior of the replay optimization is to simply get rid of that replay optimization by hard-coding `skip_record` to `false`.
https://github.com/AntelopeIO/leap/blob/a8d48d406082cc6ec5ef18dddd30b5589b3e870a/libraries/chain/controller.cpp#L1531

But we may not wish to completely remove the replay optimization. We could instead resolve the issue with the replay optimization while some its performance benefits by keeping an ephemeral transaction de-duplication store outside of chainbase during the replay of irreversible blocks. This ephemeral store would be modified as blocks were replayed to record entries as transactions were processed and also to remove any expired entries from the ephemeral store as blocks were processed. And since there is no issue with duplicate transactions when replaying already validated blocks, the store would not even need an index to quickly look up entries by transaction ID; it would just need an index on the expiration time to make it quick to remove expired entries during the replay to prevent the size of the store from growing too large. Then when exiting the replay of irreversible blocks (whether because it is done or because it exited early due to a signal provided to a process), nodeos would ensure that it mutates the chainbase database state to capture within its transaction de-duplication index all the remaining entries within the ephemeral store. After that point, the ephemeral store can be discarded and nodeos can continue forward with the next step in the process, whether that is replaying reversible blocks or just safely exiting.

- [x] #297
- [ ] Modify replay optimization to leave index in a consistent state
- [x] Performance benchmark between original optimization and modified optimization

heifner:
Remove `skip_recording` optimization of transactions in dedup list when replaying. The optimization provides little real-world measurable difference, see #297, but can cause differences if replay is interrupted, see #291 for details.

Resolves #291

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -238,7 +238,6 @@ struct controller_impl {
    protocol_feature_manager        protocol_features;
    controller::config              conf;
    const chain_id_type             chain_id; // read by thread_pool threads, value will not be changed
-   std::optional<fc::time_point>   replay_head_time;
    db_read_mode                    read_mode = db_read_mode::SPECULATIVE;
    bool                            in_trx_requiring_checks = false; ///< if true, checks that are normally skipped on replay (e.g. auth checks) cannot be skipped
    std::optional<fc::microseconds> subjective_cpu_leeway;
@@ -478,7 +477,6 @@ struct controller_impl {
 
       auto blog_head = blog.head();
       auto blog_head_time = blog_head ? blog_head->timestamp.to_time_point() : fork_db.root()->header.timestamp.to_time_point();
-      replay_head_time = blog_head_time;
       auto start_block_num = head->block_num + 1;
       auto start = fc::time_point::now();
 
@@ -546,7 +544,6 @@ struct controller_impl {
       ilog( "replayed ${n} blocks in ${duration} seconds, ${mspb} ms/block",
             ("n", head->block_num + 1 - start_block_num)("duration", (end-start).count()/1000000)
             ("mspb", ((end-start).count()/1000.0)/(head->block_num-start_block_num)) );
-      replay_head_time.reset();
 
       if( except_ptr ) {
          std::rethrow_exception( except_ptr );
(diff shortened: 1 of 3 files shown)


oschwaldp-oci (changes requested):
As this modifies behavior outside the confines of directly fixing a bug, should we have a separate implementation to go into 3.1 & 3.2 that preserves behavior and reserve this full change for main?

oschwaldp-oci (approved):

PR #360, opened 2022-10-20
[PH] Report start end times for tests

oschwaldp-oci:
Add start and end times to test reports.

Updated report examples to include start and end times. Cleaned up some of the report formatting.

Updated timestamped log directories to be in UTC so that all timestamps are easier to trace back to log files and timestamps therein.

### DIFF
--- a/tests/performance_tests/performance_test.py
+++ b/tests/performance_tests/performance_test.py
@@ -26,8 +26,8 @@ class PerfTestBasicResult:
     trxExpectMet: bool = False
     basicTestSuccess: bool = False
     logsDir: str = ""
-    testStart: datetime = ""
-    testEnd: datetime = ""
+    testStart: datetime = None
+    testEnd: datetime = None
 
 @dataclass
 class PerfTestSearchIndivResult:
@@ -66,7 +66,7 @@ def performPtbBinarySearch(tpsTestFloor: int, tpsTestCeiling: int, minStep: int,
         testSuccessful = myTest.runTest()
         if evaluateSuccess(myTest, testSuccessful, ptbResult):
             maxTpsAchieved = binSearchTarget
-            maxTpsReport = json.loads(myTest.report)
+            maxTpsReport = myTest.report
             floor = binSearchTarget + minStep
             scenarioResult.success = True
         else:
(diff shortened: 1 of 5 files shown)


tests/performance_tests/performance_test.py
```diff
@@ -26,6 +26,8 @@ class PerfTestBasicResult:
     trxExpectMet: bool = False
     basicTestSuccess: bool = False
     logsDir: str = ""
+    testStart: datetime = ""
```
jgiszczak, line 29: These should default to `None` since `datetime` is an object not a string.

jgiszczak (changes requested):

tests/performance_tests/performance_test.py
oschwaldp-oci: Addressed: [8ade49c](https://github.com/AntelopeIO/leap/pull/360/commits/8ade49c7f8211bc42226b2980bbf77a80103db5c)

tests/performance_tests/performance_test_basic.py
```diff
@@ -68,7 +68,7 @@ def __init__(self, testHelperConfig: TestHelperConfig=TestHelperConfig(), cluste
 
         self.rootLogDir = rootLogDir
         self.ptbLogDir = f"{self.rootLogDir}/{os.path.splitext(os.path.basename(__file__))[0]}"
-        self.testTimeStampDirPath = f"{self.ptbLogDir}/{datetime.now().strftime('%Y-%m-%d_%H-%M-%S')}"
+        self.testTimeStampDirPath = f"{self.ptbLogDir}/{datetime.utcnow().strftime('%Y-%m-%d_%H-%M-%S')}"
```
jgiszczak, line 71: For traceability, shouldn't this be using `self.testStart` rather than whatever time it happens to be when this executes?  That would require initializing `self.testStart` here in `__init__` rather than in `runTest()` but the difference should be negligible as the very next thing `main()` does after construction is execute `runTest()`.  Having the directory timestamp precisely match the logged timestamp will help developers tracking things down in the future and avoid difficulties in further machine processing down the road (from historical archives or some such).

jgiszczak (changes requested):

tests/performance_tests/performance_test_basic.py
oschwaldp-oci: Good call.  Addressed: [5ffdb64](https://github.com/AntelopeIO/leap/pull/360/commits/5ffdb64ab116ce895fc89395271e96aae7eda7d7)

oschwaldp-oci: Did the same for `performance_test.py`: [34e5857](https://github.com/AntelopeIO/leap/pull/360/commits/34e5857866fae97e8800720defc53b0f6b652639)

jgiszczak (approved):

PR #382, opened 2022-10-25
[3.1] Dropped trxs when overloaded logged as warning in nodeos

ClaytonCalabrese:
Currently when a transaction is rejected due to too many transactions in progress, that message only gets logged in a "debug" logging level. Rather than spamming that message at a higher level, instead add a new message which will periodically be printed as long as transactions are being rejected.

In this implementation that is the first time a transaction is rejected due to too many transactions, and every time it happens at an interval of 1 second or more.

Resolves: https://github.com/AntelopeIO/leap/issues/351

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -634,6 +634,7 @@ namespace eosio {
       string                  local_endpoint_port;
 
       std::atomic<uint32_t>   trx_in_progress_size{0};
+      fc::time_point          last_dropped_trx_msg_time;
       const uint32_t          connection_id;
       int16_t                 sent_handshake_count = 0;
       std::atomic<bool>       connecting{true};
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2683,6 +2684,10 @@ namespace eosio {
          char reason[72];
          snprintf(reason, 72, "Dropping trx, too many trx in progress %lu bytes", trx_in_progress_sz);
          my_impl->producer_plug->log_failed_transaction(ptr->id(), ptr, reason);
+         if (fc::time_point::now() - fc::seconds(1) >= last_dropped_trx_msg_time) {
+            last_dropped_trx_msg_time = fc::time_point::now();
+            wlog("Speculative execution is REJECTING transactions, too many trx in progress");
```
heifner, line 2689: Just `wlog(reason);` as we want the `trx_in_progress_sz` and this is not necessarily a speculative execution.

heifner (changes requested):

heifner (approved):

PR #385, opened 2022-10-25
[PH] Add ability to configure whether producer nodes enable trace api plugin

oschwaldp-oci:
Add feature to enable/disable trace_api_plugin on producer nodes.  To get the best performance out of a producer node, it should have the ability to be configured without trace_api_plugin and offload that responsibility to another node.

Adds command line option to `performance_test.py` and `performance_test_basic.py`:
    `--prods-enable-trace-api` - Determines whether producer nodes should have eosio::trace_api_plugin enabled" (Default: False)

When disabling trace_api_plugin on producer nodes, can make use of specificExtraNodeosArgs to enable trace_api_plugin on specific nodes.

Default enable trace_api_plugin on bios node in eosio-launcher.

Allow Cluster createAccounts to specify which node is used to validate the transactions, as a node with trace_api_plugin enabled is required.


tests/performance_tests/performance_test_basic.py
```diff
@@ -161,13 +165,15 @@ def launchCluster(self):
...
+            extraNodeosArgs=self.clusterConfig.extraNodeosArgs,
+            prodsEnableTraceApi=self.clusterConfig.prodsEnableTraceApi,
+            specificExtraNodeosArgs=self.clusterConfig.specificExtraNodeosArgs
             )
 
     def setupWalletAndAccounts(self):
         self.wallet = self.walletMgr.create('default')
         self.cluster.populateWallet(2, self.wallet)
-        self.cluster.createAccounts(self.cluster.eosioAccount, stakedDeposit=0)
+        self.cluster.createAccounts(self.cluster.eosioAccount, stakedDeposit=0, validationNodeIndex=1)
```
ClaytonCalabrese, line 176: Hardcoding this to 1 going to cause problems if we ran with multiple producers

ClaytonCalabrese (changes requested):

tests/performance_tests/performance_test_basic.py
```diff
@@ -43,11 +43,15 @@ class ClusterConfig():
...
         maximumClients: int = 0
-        loggingDict = { "bios": "off" }
+        loggingDict: dict = field(default_factory=lambda: { "bios": "off" })
+        prodsEnableTraceApi: bool = False
+        specificExtraNodeosArgs: dict = field(default_factory=dict)
         _totalNodes: int = 2
 
         def __post_init__(self):
             self._totalNodes = max(2, self.pnodes if self.totalNodes < self.pnodes else self.totalNodes)
+            if not self.prodsEnableTraceApi:
```
jgiszczak, line 53: Wouldn't all non-producer nodes want `trace_api_plugin` enabled, regardless of whether or not the producer nodes have it enabled?

jgiszczak (changes requested):

jgiszczak (approved):

tests/performance_tests/performance_test_basic.py
jgiszczak: It won't cause problems but it is incomplete if the intent is to enable it on all producer nodes.  The first node after the bios node will always have producers on it, even if it doesn't have the full complement of them.

oschwaldp-oci: Clayton is correct, simply forgot to go back after testing and remove the hardcoded for the dynamic variable for our validation node.

oschwaldp-oci: Addressed: [38f4076](https://github.com/AntelopeIO/leap/pull/385/commits/38f4076c8f698a5b6fb45f30643e061996b921f4)

This will become more dynamic as the test topology builds out to support more producers and validation nodes, etc.  For now just use the known producer and validation node IDs since the test is currently only set up to use one of each.

oschwaldp-oci: And a fix: [bd95fbc](https://github.com/AntelopeIO/leap/pull/385/commits/bd95fbc845f4a209f6c06c03e9aca826c8ca22f6)

heifner: I'm not sure I understand. I was able to run multi-producer tests with `performance_test.py` using `-p`. Does this break that? Seems like you could use `totalNodes - 1` for validation node index and just make sure `totalNodes` is always one more than configured producer nodes.

oschwaldp-oci: While the underlying Regression Test Framework supports multi-producer configurations, and the Performance Harness is build on top of that infrastructure, the Performance Harness MVP does not fully support multi-producer tests yet as it was out of scope for the MVP.  It was built in such a way that the lift to get it to multi-producer is not significant, but the scope for the MVP was limited.  Thus, while it appears and accepts Regression Test Framework underlying arguments for multi-producer, etc., the Performance Harness still expects a single producer and single validation (non-producer) node.

oschwaldp-oci: Realized minimal support for `-p` and `-n` would not be difficult, so implemented here: [68ee2ea](https://github.com/AntelopeIO/leap/pull/385/commits/68ee2eac5c3b4632acb0e82a30730dcc851ea39f)

heifner (approved):

ClaytonCalabrese (approved):

PR #397, opened 2022-10-27
[PH] Perf harness wait for transactions in block

ClaytonCalabrese:
Rather than waiting on empty blocks, instead wait for all transactions to appear in blocks. The number to determine is based on total transactions, with a fudge factor of 0.05 * TargetTPS additional blocks.


ClaytonCalabrese:
Closing to see if CI is not running because I changed target from main to fph

PR #398, opened 2022-10-27
[PH] Perf harness wait for transactions in block

ClaytonCalabrese:
Rather than waiting on empty blocks, instead wait for all transactions to appear in blocks. The number to determine is based on total transactions, with a fudge factor of 0.05 * TargetTPS additional blocks.

### DIFF
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -500,7 +500,7 @@ class Node(object):
 
     def waitForTransactionsInBlockRange(self, transIds, startBlock=2, maxFutureBlocks=None):
         lastBlockProcessed = startBlock
-        overallFinalBlock = self.getHeadBlockNum()
+        overallFinalBlock = sys.maxsize
         if maxFutureBlocks is not None:
             overallFinalBlock = overallFinalBlock + maxFutureBlocks
         while len(transIds) > 0:
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -498,11 +498,9 @@ class Node(object):
                     transIds.pop(trx['id'])
         return transIds
 
-    def waitForTransactionsInBlockRange(self, transIds, startBlock=2, maxFutureBlocks=None):
+    def waitForTransactionsInBlockRange(self, transIds, startBlock=2, maxFutureBlocks=0):
         lastBlockProcessed = startBlock
-        overallFinalBlock = sys.maxsize
-        if maxFutureBlocks is not None:
-            overallFinalBlock = overallFinalBlock + maxFutureBlocks
+        overallFinalBlock = startBlock + maxFutureBlocks
         while len(transIds) > 0:
             currentLoopEndBlock = self.getHeadBlockNum()
             if currentLoopEndBlock >  overallFinalBlock:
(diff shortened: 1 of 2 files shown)


tests/TestHarness/Node.py
```diff
@@ -500,7 +500,7 @@ def checkBlockForTransactions(self, transIds, blockNum):
 
     def waitForTransactionsInBlockRange(self, transIds, startBlock=2, maxFutureBlocks=None):
         lastBlockProcessed = startBlock
-        overallFinalBlock = self.getHeadBlockNum()
+        overallFinalBlock = sys.maxsize
```
oschwaldp-oci, line 503: I still don't really like setting this to `sys.maxsize` as it causes a situation where once could easily create an infinite wait.  With the original `self.getHeadBlockNum()`, the damage of failing to set `maxFutureBlocks` means you basically don't wait and just evaluate all currently available blocks.

tests/performance_tests/performance_test_basic.py
```diff
@@ -209,7 +210,11 @@ def runTpsTest(self) -> bool:
             ])
 
         # Get stats after transaction generation stops
-        self.data.ceaseBlock = self.waitForEmptyBlocks(self.validationNode, self.emptyBlockGoal) - self.emptyBlockGoal + 1
+        trxSent = {}
+        log_reader.scrapeTrxGenTrxSentDataLogs(trxSent, self.trxGenLogDirPath, self.quiet)
+        blocksToWait = ceil(self.expectedTransactionsSent / min(4000, 0.45 * self.targetTps))
```
oschwaldp-oci, line 215: Why not use a simple calculation based on `self.testTrxGenDurationSec` instead?  Could maybe wait `2 * self.testTrxGenDurationSec` worth of blocks?  Blocks are produced every half second, so it would be more like: `2 * 2 * self.testTrxGenDurationSec` to allow it to wait double the test generation duration.

tests/performance_tests/performance_test_basic.py
```diff
@@ -209,7 +210,11 @@ def runTpsTest(self) -> bool:
             ])
 
         # Get stats after transaction generation stops
-        self.data.ceaseBlock = self.waitForEmptyBlocks(self.validationNode, self.emptyBlockGoal) - self.emptyBlockGoal + 1
+        trxSent = {}
+        log_reader.scrapeTrxGenTrxSentDataLogs(trxSent, self.trxGenLogDirPath, self.quiet)
```
oschwaldp-oci, line 214: Should we try to find a way to only do `scrapeTrxGenTrxSentDataLogs` one time during the test and reuse the `trxSent` list?  It is also performed in `log_reader.calcAndReport`.  I guess we would still need to make a copy to pass into `waitForTransactionsInBlockRange`, which may mean the overhead of copying and plumbing it through to `calcAndReport` may not be worth it and a second scrape isn't too big of a performance hit.

oschwaldp-oci (changes requested):

tests/performance_tests/performance_test_basic.py
oschwaldp-oci: Might make it easier to understand as it would just be a multiplier against a duration in seconds instead of transaction and tps based calculations.

tests/TestHarness/Node.py
ClaytonCalabrese: Right but that is logically equal to 0. I'll simplify the logic and change it how you requested.

oschwaldp-oci (approved):

PR #410, opened 2022-10-29
[3.2] Add log/rethrow of possible exception on ship plugin shutdown

766C6164:
With addition of threading to ship plugin plugin_shutdown can potentially throw unexpected by caller exceptions. Wrapping code and rethrowing as it was done in develop-boxed would be more robust way to handle shutdown.


766C6164:
Fixed in https://github.com/AntelopeIO/leap/pull/413

PR #416, opened 2022-10-31
only use ENF runners for builds in AntelopeIO Leap repository

spoonincode:
One of the great strengths with Github Actions is that a user can fork a public repository and inherit a fully operational CI pipeline they can use for free on their public fork.

But that doesn't work for Leap currently because the workflow requires use of ENF runners which of course no other fork can utilize. Since we only use ENF runners for speed (Github's free runners do build the software as expected, afaik), let's only use ENF runners for builds in the AntelopeIO org. Forks will use Github's free runners.

Is there a better way to do this? Maybe only check the string once in the `d` job and reference a boolean elsewhere?

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -39,7 +39,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: ${{fromJSON(needs.d.outputs.missing-platforms)}}
-    runs-on: ["self-hosted", "enf-x86-beefy"]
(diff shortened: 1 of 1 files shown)


ericpassmore:
Should we close this PR as superseded by PR #1710

spoonincode:
It wasn't superseded by 1710, but this PR is grossly out of date and would require significant modifications to get it ready again. I think maybe we just close it out for now.

PR #432, opened 2022-11-01
Write a genesis file for each node.

jgiszczak:
Like it says on the tin.


tests/TestHarness/launcher.py
```diff
@@ -324,7 +326,30 @@ def write_logging_config_file(self, node):
         pass
 
     def write_genesis_file(self, node):
```
oschwaldp-oci, line 328: Seems like this loses the ability to load a complete genesis from file.  We should still support the ability to pass in a genesis file like `init_genesis` did previously and just overwrite the couple dynamic values `bioskey`, `max_block_cpu_usage`, `max_transaction_cpu_usage` if configured.

oschwaldp-oci (changes requested):

oschwaldp-oci (approved):

tests/TestHarness/launcher.py
oschwaldp-oci: Addressed in [8e6242d](https://github.com/AntelopeIO/leap/pull/432/commits/8e6242da0f33c0c03e1eaee7f95854c0033b9d2f)

PR #435, opened 2022-11-02
Add  CRYPTO_PRIMITIVES feature to bios boot python script

iamveritas:
Add  CRYPTO_PRIMITIVES feature to bios boot python script


iamveritas:
it is not a must so closing this PR

PR #441, opened 2022-11-03
[3.1] Fix test failure due to trx hitting near end of block production time

heifner:
The fix #106 did not update the `present` flag to `validateTrxState` function for the new state when hitting the condition of #106.

Resolves #419

### DIFF
--- a/tests/trx_finality_status_test.py
+++ b/tests/trx_finality_status_test.py
@@ -150,16 +150,21 @@ try:
     status.append(copy.copy(retStatus))
     startingBlockNum=testNode.getInfo()["head_block_num"]
 
-    def validateTrxState(status, present):
+    def validateTrxState(status):
+        Print(f"status: {status}")
+        present = True
+        state = getState(status)
+        if state == localState or state == unknownState:
+            present = False
         bnPresent = "block_number" in status
         biPresent = "block_id" in status
         btPresent = "block_timestamp" in status
-        desc = "" if present else "not "
+        desc = "" if present else " not"
         group = bnPresent and biPresent and btPresent if present else not bnPresent and not biPresent and not btPresent
         assert group, \
             f"ERROR: getTransactionStatus should{desc} contain \"block_number\", \"block_id\", or \"block_timestamp\" since state was \"{getState(status)}\".\nstatus: {json.dumps(status, indent=1)}"
 
-    validateTrxState(status[0], present=False)
+    validateTrxState(status[0])
 
     def validate(status, knownTrx=True):
         assert "head_number" in status and "head_id" in status and "head_timestamp" in status, \
@@ -185,7 +190,7 @@ try:
     state = getState(status[1])
     assert state == inBlockState, f"ERROR: getTransactionStatus never returned a \"{inBlockState}\" state"
 
-    validateTrxState(status[1], present=True)
+    validateTrxState(status[1])
 
     validate(status[1])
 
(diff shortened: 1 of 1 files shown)


tests/trx_finality_status_test.py
```diff
@@ -150,16 +150,21 @@ def isState(state, expectedState, allowedState=None, notAllowedState=None):
     status.append(copy.copy(retStatus))
     startingBlockNum=testNode.getInfo()["head_block_num"]
 
-    def validateTrxState(status, present):
+    def validateTrxState(status):
+        Print(f"status: {status}")
+        present = True
+        state = getState(status)
+        if state == localState or state == unknownState:
+            present = False
```
oschwaldp-oci, line 158: Combining knowledge from the fix in https://github.com/AntelopeIO/leap/pull/106, I'm wondering if, instead of removing the `present` argument, a determination needs to be done to calculate a correct `present` value for the first call to `validateTrxState` -- as it seems to be the only call that can have 2 expectations for the state. This way the test can continue to define its expectation of presence at each check.

tests/trx_finality_status_test.py
```diff
@@ -150,16 +150,21 @@ def isState(state, expectedState, allowedState=None, notAllowedState=None):
...
         biPresent = "block_id" in status
         btPresent = "block_timestamp" in status
-        desc = "" if present else "not "
+        desc = "" if present else " not"
         group = bnPresent and biPresent and btPresent if present else not bnPresent and not biPresent and not btPresent
         assert group, \
             f"ERROR: getTransactionStatus should{desc} contain \"block_number\", \"block_id\", or \"block_timestamp\" since state was \"{getState(status)}\".\nstatus: {json.dumps(status, indent=1)}"
 
-    validateTrxState(status[0], present=False)
+    validateTrxState(status[0])
```
oschwaldp-oci, line 167: Instead, suggestion would be to calculate `present` before the call to `validateTrxState` in this specific instance where it could have 2 values.  Something like:

`present = True if state == inBlockState else False`
`validateTrxState(status[0], present=present)`

tests/trx_finality_status_test.py
oschwaldp-oci: Would also recommend updating the Error message on line 149 to include both state options.

tests/trx_finality_status_test.py
```diff
@@ -185,7 +190,7 @@ def validate(status, knownTrx=True):
     state = getState(status[1])
     assert state == inBlockState, f"ERROR: getTransactionStatus never returned a \"{inBlockState}\" state"
 
-    validateTrxState(status[1], present=True)
+    validateTrxState(status[1])
```
oschwaldp-oci, line 193: revert

tests/trx_finality_status_test.py
```diff
@@ -212,7 +217,7 @@ def validate(status, knownTrx=True):
     state = getState(retStatus)
     assert state == unknownState, \
         f"ERROR: Calling getTransactionStatus after the success_duration should have resulted in an \"{irreversibleState}\" state.\nstatus: {json.dumps(retStatus, indent=1)}"
-    validateTrxState(retStatus, present=False)
+    validateTrxState(retStatus)
```
oschwaldp-oci, line 220: revert

oschwaldp-oci (changes requested):

oschwaldp-oci (approved):

PR #444, opened 2022-11-03
[4.0] Use libcurl  for cleos

huangminghuang:
This PR
- uses libcurl instead of asio+openssl for http connection
- adds new options `--http-verbose` and `--http-trace` for debugging

### DIFF
--- a/programs/cleos/CMakeLists.txt
+++ b/programs/cleos/CMakeLists.txt
@@ -1,4 +1,4 @@
-find_package(CURL REQUIRED)
+find_package(CURL 7.40.0 REQUIRED)
 
 configure_file(help_text.cpp.in help_text.cpp @ONLY)
 configure_file(config.hpp.in config.hpp ESCAPE_QUOTES)
--- a/programs/cleos/CMakeLists.txt
+++ b/programs/cleos/CMakeLists.txt
@@ -1,3 +1,5 @@
+find_package(CURL REQUIRED)
+
 configure_file(help_text.cpp.in help_text.cpp @ONLY)
 configure_file(config.hpp.in config.hpp ESCAPE_QUOTES)
 add_executable( ${CLI_CLIENT_EXECUTABLE_NAME} main.cpp httpc.cpp "${CMAKE_CURRENT_BINARY_DIR}/help_text.cpp" localize.hpp "${CMAKE_CURRENT_BINARY_DIR}/config.hpp")
@@ -13,6 +15,11 @@ target_include_directories(${CLI_CLIENT_EXECUTABLE_NAME} PUBLIC ${CMAKE_CURRENT_
 target_link_libraries( ${CLI_CLIENT_EXECUTABLE_NAME}
                        PRIVATE appbase version leap-cli11 chain_api_plugin producer_plugin chain_plugin http_plugin eosio_chain fc ${CMAKE_DL_LIBS} ${PLATFORM_SPECIFIC_LIBS} )
 
+if (CURL_FOUND)
+  target_sources(${CLI_CLIENT_EXECUTABLE_NAME} PRIVATE do_http_post_libcurl.cpp)
+  target_include_directories(${CLI_CLIENT_EXECUTABLE_NAME} PRIVATE ${CURL_INCLUDE_DIRS})
+  target_link_libraries(${CLI_CLIENT_EXECUTABLE_NAME} PRIVATE ${CURL_LIBRARIES})
+endif()
 
 copy_bin( ${CLI_CLIENT_EXECUTABLE_NAME} )
 install( TARGETS
--- a/programs/cleos/do_http_post_libcurl.cpp
+++ b/programs/cleos/do_http_post_libcurl.cpp
@@ -80,7 +80,7 @@ namespace eosio { namespace client { namespace http {
 
       auto curl = curl_easy_init();
       EOS_ASSERT(curl != 0, chain::http_exception, "curl_easy_init failed");
-      fc::make_scoped_exit([curl]() {
+      auto on_exit = fc::make_scoped_exit([curl]() {
          curl_easy_cleanup(curl);
       });
 
--- a/programs/cleos/do_http_post_libcurl.cpp
+++ b/programs/cleos/do_http_post_libcurl.cpp
@@ -3,7 +3,7 @@
 #include "do_http_post.hpp"
 #include <curl/curl.h>
 #include <eosio/chain/exceptions.hpp>
-#include <fc/scoped_exit.hpp>
+#include <map>
 
 namespace eosio { namespace client { namespace http {
 
@@ -65,6 +65,11 @@ namespace eosio { namespace client { namespace http {
       return 0;
    }
 
+   struct curl_handle : std::unique_ptr<CURL, decltype(&curl_easy_cleanup)> {
+      curl_handle(int) : std::unique_ptr<CURL, decltype(&curl_easy_cleanup)>( curl_easy_init(), &curl_easy_cleanup) {}
+   };
+
+
    std::tuple<unsigned int, std::string> do_http_post(const std::string& base_uri, const std::string& path,
                                                       const std::vector<std::string>& headers,
                                                       const std::string& postjson, bool verify_cert, bool verbose,
@@ -78,11 +83,10 @@ namespace eosio { namespace client { namespace http {
          initialized = true;
       }
 
-      auto curl = curl_easy_init();
+      static std::map<std::string, curl_handle> handles;
+      auto [itr, _] = handles.emplace(base_uri, 0);
+      auto curl = itr->second.get();
       EOS_ASSERT(curl != 0, chain::http_exception, "curl_easy_init failed");
-      auto on_exit = fc::make_scoped_exit([curl]() {
-         curl_easy_cleanup(curl);
-      });
 
       std::string uri;
 
--- a/programs/cleos/do_http_post_libcurl.cpp
+++ b/programs/cleos/do_http_post_libcurl.cpp
@@ -3,6 +3,7 @@
 #include "do_http_post.hpp"
 #include <curl/curl.h>
 #include <eosio/chain/exceptions.hpp>
+#include <fc/scoped_exit.hpp>
 
 namespace eosio { namespace client { namespace http {
 
@@ -13,6 +14,8 @@ namespace eosio { namespace client { namespace http {
       return realsize;
    }
 
+   // the following two functions are directly copy from the libcurl API overview from
+   // https://curl.se/libcurl/c/CURLOPT_DEBUGFUNCTION.html
    void dump(const char* text, FILE* stream, unsigned char* ptr, size_t size) {
       size_t       i;
       size_t       c;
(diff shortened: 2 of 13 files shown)


programs/cleos/main.cpp
```diff
@@ -2775,18 +2769,23 @@ int main( int argc, char** argv ) {
    app.add_option( "--wallet-host", obsoleted_option_host_port, localized("The host where ${k} is running", ("k", key_store_executable_name)) )->group("");
    app.add_option( "--wallet-port", obsoleted_option_host_port, localized("The port where ${k} is running", ("k", key_store_executable_name)) )->group("");
 
-   app.add_option( "-u,--url", ::default_url, localized( "The http/https URL where ${n} is running", ("n", node_executable_name)))->capture_default_str();
+   app.add_option_function<std::string>( "-u,--url", [](const std::string& url) {
+      ::default_url = url;
+      if (url.back()=='/') ::default_url.resize(url.size()-1); // remove trailing slash
```
heifner, line 2774: Need to check for empty string. Someone could pass an empty string to `--url`

programs/cleos/do_http_post_libcurl.cpp
```diff
@@ -0,0 +1,135 @@
...
+         for (c = 0; (c < width) && (i + c < size); c++) {
+            char x = (ptr[i + c] >= 0x20 && ptr[i + c] < 0x80) ? ptr[i + c] : '.';
+            fputc(x, stream);
+         }
+
+         fputc('\n', stream); /* newline */
+      }
+   }
+
+   int my_trace(CURL* handle, curl_infotype type, char* data, size_t size, void* userp) {
```
heifner, line 44: Seems like some of this code was likely copied from somewhere. Do we need to attribute the source?

programs/cleos/do_http_post_libcurl.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+      static bool initialized = false;
+
+      if (!initialized) {
+         auto res = curl_global_init(CURL_GLOBAL_DEFAULT);
+         EOS_ASSERT(res == CURLE_OK, chain::http_exception, curl_easy_strerror(res));
+         initialized = true;
+      }
+
+      auto curl = curl_easy_init();
```
heifner, line 78: I think it would be good to use a `fc::make_scoped_exit` to `curl_easy_cleanup`. There is nothing currently I see that would throw outside of `bad_alloc`, but it would possibly protect against code changes that might introduce exceptions.

heifner (changes requested):

programs/cleos/do_http_post_libcurl.cpp
spoonincode: Does calling `curl_easy_cleanup()` in here prevent keep alive though?

huangminghuang: I believe so. However, in the origin asio implementation, it creates new connection per http request. There's no keep alive mechanism either. To make that happen, some bookkeeping mechanism to associates URL and the curl handle is needed. I am not sure it's worth the effort given the fact that this is just a CLI utility.

spoonincode: I think it's as straightforward as just reusing the same curl handle and libcurl sorts it out itself -- even for different servers and whatnot. It's not unusual for cleos to make more than one request, and the roundtrip time of starting a new TLS connection is non-trivial.

But yeah, the previous code doesn't do it so improvements can be addressed later. I thought we added proper support for this in cleos at the same time as http_plugin (included in 3.2) but apparently not.

spoonincode:
(at least) README.md needs to be updated with new dependency information.

heifner:
Can you show the different output in the PR description for when nodeos is not running and cleos attempts to connect to it. Seems like we have gone from "Connection Refused" to "http exception".

huangminghuang:
> 

~/leap/build$ ./bin/cleos get info
Error 3200000: http exception
Error Details:
Couldn't connect to server

huangminghuang:
> 



> Can you show the different output in the PR description for when nodeos is not running and cleos attempts to connect to it. Seems like we have gone from "Connection Refused" to "http exception".

~/leap/build$ ./bin/cleos get info
Error 3200000: http exception
Error Details:
Couldn't connect to server

tests/CMakeLists.txt
```diff
@@ -142,6 +142,7 @@ add_test(NAME keosd_auto_launch_test COMMAND tests/keosd_auto_launch_test.py WOR
 set_property(TEST keosd_auto_launch_test PROPERTY LABELS nonparallelizable_tests)
 
 add_test(NAME trx_finality_status_test COMMAND tests/trx_finality_status_test.py -v --clean-run --dump-error-detail WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+set_tests_properties(trx_finality_status_test PROPERTIES TIMEOUT 200)
```
spoonincode, line 145: What are these test timeout changes for, are they related to the cleos changes?

spoonincode: yeah that sounds like something an issue should be created for; as that fix could be a candidate for inclusion in 3.1 and/or 3.2

programs/cleos/CMakeLists.txt
```diff
@@ -13,6 +15,11 @@ target_include_directories(${CLI_CLIENT_EXECUTABLE_NAME} PUBLIC ${CMAKE_CURRENT_
 target_link_libraries( ${CLI_CLIENT_EXECUTABLE_NAME}
                        PRIVATE appbase version leap-cli11 chain_api_plugin producer_plugin chain_plugin http_plugin eosio_chain fc ${CMAKE_DL_LIBS} ${PLATFORM_SPECIFIC_LIBS} )
 
+if (CURL_FOUND)
```
spoonincode, line 18: don't need the `if (CURL_FOUND)` now that curl is `REQUIRED` above

huangminghuang: I intentionally keep it that way to make it more clear what needs to changed whenever other library other than libcurl is used. You described that intention in the issue. I can certainly remove it if you insisted.

spoonincode:
> Can you show the different output in the PR description for when nodeos is not running and cleos attempts to connect to it. Seems like we have gone from "Connection Refused" to "http exception".

Yes it does seems the output has changed, previously
```
Failed http request to nodeos at http://127.0.0.1:8888/; is nodeos running?
  Error: connect: Connection refused [system:111 at /usr/include/boost/asio/detail/reactive_socket_service.hpp:587 in function 'connect']
```
and now
```
Error 3200000: http exception
Error Details:
Couldn't connect to server
```

I think there is likely a benefit to the prior output that is clear what server endpoint cleos failed to communicate with.

programs/cleos/CMakeLists.txt
spoonincode: It's not critical, you can leave it as is if you'd like, because yes it will be changed when I get around to proper macOS support here.

huangminghuang:
The error message has been changed to
```
$ ./bin/cleos get info
Failed http request to nodeos at http://127.0.0.1:8888; is nodeos running?
  Error: 1100000 connection_exception: Connection Exception
Couldn't connect to server
    {}
    cleos  do_http_post_libcurl.cpp:135 do_http_post
```

programs/cleos/do_http_post_libcurl.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+      static bool initialized = false;
+
+      if (!initialized) {
+         auto res = curl_global_init(CURL_GLOBAL_DEFAULT);
+         EOS_ASSERT(res == CURLE_OK, chain::http_exception, curl_easy_strerror(res));
+         initialized = true;
+      }
+
+      static std::map<std::string, curl_handle> handles;
```
spoonincode, line 86: Is this map necessary? The documentation seems to suggest doing this is not required.

huangminghuang: According to https://everything.curl.dev/libcurl/connectionreuse:

> When you are using the easy API, or, more specifically, curl_easy_perform(), libcurl will keep the pool associated with the specific easy handle. Then reusing the SAME easy handle will ensure it can reuse its connection.

spoonincode: > When libcurl is about to make a new connection for the purposes of doing a transfer, it will first check to see if there is an existing connection in the pool that it can reuse instead. The connection re-use check is done before any DNS or other name resolving mechanism is used, so it is purely host name based. If there is an existing live connection to the right host name, a lot of other properties (port number, protocol, etc) are also checked to see that it can be used.

huangminghuang: What's wrong to go with the public API contract I cited? Even if what you said is true that the cleanup is not necessary, I still prefer to have them. In this case, it becomes coding style preference because the performance is not an issue here. Similarly, when I am writing EOS contract code which links to simple allocator, I will still make sure all the memory allocated would be freed in my code even though I know perfectly that the free is actually no-op under the circumstance. It's a coding guideline I would always adhere to.

programs/cleos/do_http_post_libcurl.cpp
```diff
@@ -0,0 +1,142 @@
...
+      list = curl_slist_append(list, "Content-Type: application/json");
+
+      curl_easy_setopt(curl, CURLOPT_HTTPHEADER, list);
+      curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, write_callback);
+
+      std::string response;
+
+      curl_easy_setopt(curl, CURLOPT_WRITEDATA, (void*)&response);
+      if (verify_cert)
+         curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, 1L);
```
spoonincode, line 131: The documentation states the default value for `CURLOPT_SSL_VERIFYPEER` is 1, so this check's logic needs to be inverted.

programs/cleos/do_http_post_libcurl.cpp
```diff
@@ -0,0 +1,142 @@
...
+
+      curl_easy_setopt(curl, CURLOPT_URL, uri.c_str());
+      curl_easy_setopt(curl, CURLOPT_POST, 1L);
+      curl_easy_setopt(curl, CURLOPT_HTTP_VERSION, (long)CURL_HTTP_VERSION_1_1);
+      curl_easy_setopt(curl, CURLOPT_POSTFIELDSIZE, postjson.size());
+      curl_easy_setopt(curl, CURLOPT_POSTFIELDS, postjson.c_str());
+
+      if (trace) {
+         curl_easy_setopt(curl, CURLOPT_DEBUGFUNCTION, my_trace);
+         curl_easy_setopt(curl, CURLOPT_FOLLOWLOCATION, 1L);
```
spoonincode, line 111: Is there a reason `CURLOPT_FOLLOWLOCATION` is (only) enabled in trace mode? Such an option changes the semantics of the request -- it's not strictly a debug option.

programs/cleos/do_http_post_libcurl.cpp
spoonincode: There is nothing on that page that indicates it's the user's responsibility to strictly use the same hosts with a single handle, or that mixing hosts to a single handle would invalidate connection reuse. i.e. the documentation suggests it is sufficient to keep around a single curl handle and reuse it even when mixing different hosts (or protocols, etc).

spoonincode (approved):

heifner (approved):

PR #446, opened 2022-11-04
[PH] Update producer node configurations for performance

oschwaldp-oci:
For single producer nodes, need to override defaults either dynamically or hard-coded for the following features:

- last-block-cpu-effort-percent → 100
- cpu-effort-percent → 100
- last-block-time-offset-us → 0 
- produce-time-offset-us → 0
- signature-cpu-billable-pct → 0

Also, update docs for launch_transaction_generators.py changes that were missed in earlier revisions.

### DIFF
--- a/tests/performance_tests/README.md
+++ b/tests/performance_tests/README.md
@@ -217,8 +217,8 @@ The Performance Harness main script `performance_test.py` can be configured usin
                           (default: 2)
 * `--signature-cpu-billable-pct SIGNATURE_CPU_BILLABLE_PCT`
                           Percentage of actual signature recovery cpu to bill. Whole number percentages, e.g. 50 for 50% (default: 0)
-* `--disable-subjective-api-billing DISABLE_SUBJECTIVE_API_BILLING`
-                          Disable subjective CPU billing for API transactions (default: True)
+* `--disable-subjective-billing DISABLE_SUBJECTIVE_BILLING`
+                          Disable subjective CPU billing for API/P2P transactions (default: True)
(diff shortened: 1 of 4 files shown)


tests/performance_tests/README.md
```diff
@@ -215,6 +215,20 @@ The Performance Harness main script `performance_test.py` can be configured usin
                           The number of potentially non-empty blocks, in addition to leading and trailing size 0 blocks,
                           to prune from the beginning and end of the range of blocks of interest for evaluation.
                           (default: 2)
+* `--signature-cpu-billable-pct SIGNATURE_CPU_BILLABLE_PCT`
+                          Percentage of actual signature recovery cpu to bill. Whole number percentages, e.g. 50 for 50% (default: 0)
+* `--disable-subjective-api-billing DISABLE_SUBJECTIVE_API_BILLING`
+                          Disable subjective CPU billing for API transactions (default: True)
```
heifner, line 221: Not sure why this was added. It is fine to provide this option, but should also provide `disable-subjective-p2p-billing` which is the only one that would have any affect at the moment. So either add `disable-subjective-p2p-billing` or just add the `disable-subjective-billing` option which covers both.

heifner (changes requested):

tests/performance_tests/README.md
oschwaldp-oci: Addressed: [2982ba9](https://github.com/AntelopeIO/leap/pull/446/commits/2982ba9756542d645e14cf53b1dab581f806bcc4)

heifner (approved):

Issue #208, opened 2022-09-22
Move ship protocol out of abieos


stephenpdeos:

bhazzard:
**Why is this important?** This is important because ship_protocol.hpp has already been removed from abieos's main branch.

**In Scope:**

- We discussed on 11/16/2022 that the "ship_protocol.hpp" header file should be delivered via "EosioTester.cmake" (or whatever it is renamed to, to indicate its more broader purpose)

**Out of Scope:**

- We also discussed on 11/16/2022 that long-term we need a more mature API service so that headers don't need to be installed
- There was discussion of a refactor on 11/16/2022. The refactor is not strictly necessary. If refactoring would add non-trivial time to this task, it should not be performed.

PR #458, opened 2022-11-09
generate ship abi

huangminghuang:
This PR
- move the ship_protocol.hpp from abieos into state_history library
- generates ship abi directly from ship_protocol.hpp

Notice that the table name field in the generated ABI will conform the encoding rule of eosio name; i.e the manual crafted ABI before this PR,
```
    "tables": [
        { "name": "account", "type": "account", "key_names": ["name"] },
        { "name": "account_metadata", "type": "account_metadata", "key_names": ["name"] },
        { "name": "code", "type": "code", "key_names": ["vm_type", "vm_version", "code_hash"] },
        { "name": "contract_table", "type": "contract_table", "key_names": ["code", "scope", "table"] },
        { "name": "contract_row", "type": "contract_row", "key_names": ["code", "scope", "table", "primary_key"] },
        { "name": "contract_index64", "type": "contract_index64", "key_names": ["code", "scope", "table", "primary_key"] },
        { "name": "contract_index128", "type": "contract_index128", "key_names": ["code", "scope", "table", "primary_key"] },
        { "name": "contract_index256", "type": "contract_index256", "key_names": ["code", "scope", "table", "primary_key"] },
        { "name": "contract_index_double", "type": "contract_index_double", "key_names": ["code", "scope", "table", "primary_key"] },
        { "name": "contract_index_long_double", "type": "contract_index_long_double", "key_names": ["code", "scope", "table", "primary_key"] },
        { "name": "global_property", "type": "global_property", "key_names": [] },
        { "name": "generated_transaction", "type": "generated_transaction", "key_names": ["sender", "sender_id"] },
        { "name": "protocol_state", "type": "protocol_state", "key_names": [] },
        { "name": "permission", "type": "permission", "key_names": ["owner", "name"] },
        { "name": "permission_link", "type": "permission_link", "key_names": ["account", "code", "message_type"] },
        { "name": "resource_limits", "type": "resource_limits", "key_names": ["owner"] },
        { "name": "resource_usage", "type": "resource_usage", "key_names": ["owner"] },
        { "name": "resource_limits_state", "type": "resource_limits_state", "key_names": [] },
        { "name": "resource_limits_config", "type": "resource_limits_config", "key_names": [] }
    ]
```

After the PR, it becomes
```
  "tables": [
    { "name": "account", "key_names": [ "name" ], "type": "account" },
    { "name": "actmetadata", "key_names": [ "name" ], "type": "account_metadata" },
    { "name": "code", "key_names": [ "vm_type", "vm_version", "code_hash"],"type": "code"},
    { "name": "contracttbl","key_names": [ "code", "scope", "table"], "type": "contract_table"},
    { "name": "contractrow","key_names": [ "code","scope", "table", "primary_key"], "type": "contract_row"},
    { "name": "cntrctidx1", "key_names": [ "code","scope", "table", "primary_key"], "type": "contract_index64"},
    { "name": "cntrctidx2", "key_names": [ "code", "scope", "table", "primary_key" ], "type": "contract_index128" },
    { "name": "cntrctidx3", "key_names": [ "code", "scope", "table", "primary_key" ], "type": "contract_index256" },
    { "name": "cntrctidx4", "key_names": [ "code", "scope", "table", "primary_key" ], "type": "contract_index_double" },
    { "name": "cntrctidx5", "key_names": [ "code", "scope", "table", "primary_key" ], "type": "contract_index_long_double" },
    { "name": "global.pty", "type": "global_property" },
    { "name": "generatedtrx", "key_names": [ "sender", "sender_id" ], "type": "generated_transaction" },
    { "name": "protocolst", "type": "protocol_state" },
    { "name": "permission", "key_names": [ "owner", "name" ], "type": "permission" },
    { "name": "permlink", "key_names": [ "account", "code", "message_type" ], "type": "permission_link" },
    { "name": "rsclimits", "key_names": [ "owner" ], "type": "resource_limits" },
    { "name": "rscusage", "key_names": [ "owner" ], "type": "resource_usage" },
    { "name": "rsclimitsst", "type": "resource_limits_state" },
    { "name": "rsclimitscfg", "type": "resource_limits_config" }
  ]
```

This Change is necessary because the original ABI violates the schema used by contracts. If we keep using the original ship ABI, the tooling for parsing the contract ABI and the one for ship ABI needs to be different which can cause confusion in the future. For any 3rd party tool consume ship ABI affected by this change, they need to use the `type` field instead of the `name` field of the table.


libraries/state_history/CMakeLists.txt
```diff
@@ -1,14 +1,28 @@
...
+add_custom_command(OUTPUT ship_abi.cpp
+                   COMMAND ship_abi_gen > ship_abi.cpp
+                   DEPENDS ship_abi_gen)
+
+add_custom_target(ship_abi_app ALL
+   COMMAND echo "generate ship_abi.cpp"
+   DEPENDS ship_abi.cpp
+   VERBATIM
+)
+
```
spoonincode, line 13: This breaks ("easy") cross compilation, which will be our strategy for ARM & non-linux builds going forward.

huangminghuang: Fixed with including a generated ABI in the source tree.

spoonincode:
Removal of ship_protocol.hpp in abieos breaks users until we have a delivery mechanism for providing the ship_protocol.hpp created in this repository. What is the plan for delivery of this file?

heifner:
> Removal of ship_protocol.hpp in abieos breaks users until we have a delivery mechanism for providing the ship_protocol.hpp created in this repository. What is the plan for delivery of this file?

I guess I just assumed they would use it out of a leap build instead of out of abieos. Do you think we need to copy it into the dev package?

spoonincode:
Adding it to leap-dev initially seems like the right approach to me. In particular there is already the concept of doing a `find_package(leap)` to bring in targets and macros like `add_eosio_test()`, and we could add another `INTERFACE` library target named `ship_protocol` or such.

However there are some problems...
First, the cmake package file is going to need to be refactored: you probably only want to find boost, LLVM, etc when using the `add_eosio_test()` or `add_eosio_test_executable()` macros -- not all the time -- so that a user of `ship_protocol` target doesn't have to have boost etc installed.
Second, we still don't consider leap-dev fully official: it's not a release asset, for example. I'm not sure how we get it over the hump to be more officially supported. One criteria I would have is that its installation shouldn't interfere with doing a leap build. And today it might (it has in the past!).

huangminghuang:
IMO, the way EosioTester.cmake setup the target is messy. We should have used the standard cmake export() and install(EXPORT target) to do the job. The `find_package()` in `EosioTester.cmake` should actually be `find_dependency()`.  Abieos has already had this from my previous effort. If we can do the same thing for all the tester dependent targets, it would be a lot cleaner.  However, I think it would  be beyond the scope of this [issue](https://github.com/AntelopeIO/leap/issues/208).

spoonincode:
EosioTester.cmake is messy because the entire eosio_tester library situation is messy. It really ought to be a single shared library with a minimal interface ergo a minimal number of headers.

There are certainly many options other then a `ship_protocol` target in EosioTester. One way or another this needs to be resolved or agreed upon so that we don't break users _without a clear solution for them_ (since abieos is a continuous release branch, breaks are inevitable, but we need to have a clear way documented for users to migrate). I don't think "fish out ship_protocol.hpp from leap's source tree and manually copy it to your source files" is the kind of integration quality we are aiming for, but I suppose the rest of the team will need to chime in.

PR #442, opened 2022-11-03
[4.0] replace usage of curl in python with urllib

ClaytonCalabrese:
Replaces all instances of curl in the tests with urllib. This is done via a function that replaces `processCurlCmd` and calls into urllib.

This should result in increased efficiency because we aren't relying on an outside command line to execute things.

PR #459, opened 2022-11-10
Migrate chain queries out of Node.py into their own class

jgiszczak:
Node inherits from new Queries class to avoid intrusive changes to tests.
`processCurlCommand()` is left in Node to avoid merge conflicts with #442


oschwaldp-oci (commented):
I like this separation of concerns.  I might, however, make an argument for the `wait*` functions to remain in `Node.py` as they aren't queries in the basic definition of such but rather a higher level construct around queries.  I didn't look to see how that would affect the stated goal of "avoid intrusive changes to tests` though.

oschwaldp-oci (approved):

PR #461, opened 2022-11-11
Replace implementation of alt_bn128 host functions with faster bn256 library

huangminghuang:
This PR removes the old ff submodule and replace it with the new bn256 library.


libraries/libfc/CMakeLists.txt
```diff
@@ -1,11 +1,12 @@
 add_subdirectory( secp256k1 )
 
 set( WITH_PROCPS OFF CACHE BOOL "" FORCE)
-set( CURVE "ALT_BN128" CACHE STRING "" FORCE)
+# set( CURVE "ALT_BN128" CACHE STRING "" FORCE)
 set( USE_ASM OFF CACHE BOOL "" FORCE)
 set( IS_LIBFF_PARENT OFF CACHE BOOL "" FORCE)
 set( FF_INSTALL_COMPONENT "${FC_INSTALL_COMPONENT}")
-add_subdirectory( libraries/ff )
+# add_subdirectory( libraries/ff )
```
spoonincode, line 8: There are a number of commented out lines in this PR. They need to be removed or otherwise taken care of.

This entire cmake block can also be removed, from line 3-8

libraries/libfc/CMakeLists.txt
```diff
@@ -138,9 +155,9 @@ if(APPLE)
   find_library(security_framework Security)
   find_library(corefoundation_framework CoreFoundation)
 endif()
-target_link_libraries( fc PUBLIC ff
+target_link_libraries( fc PUBLIC bn256
                                  Boost::date_time Boost::filesystem Boost::chrono Boost::iostreams Threads::Threads
-                                 OpenSSL::Crypto OpenSSL::SSL ZLIB::ZLIB ${PLATFORM_SPECIFIC_LIBS} ${CMAKE_DL_LIBS} secp256k1 ${security_framework} ${corefoundation_framework} )
+                                 OpenSSL::Crypto OpenSSL::SSL ZLIB::ZLIB ${PLATFORM_SPECIFIC_LIBS} ${CMAKE_DL_LIBS} secp256k1 ${security_framework} ${corefoundation_framework} ${GMP_LIBRARIES})
```
spoonincode, line 160: This change looks redundant as there is a `target_link_libraries(fc PUBLIC GMP::gmp)` above?

spoonincode:
I'm not sure the EosioTester.cmake changes are sufficient -- I would have expected bn256 library to need to be part of it, and I don't even see any `install()` declarations in bn256's CMakeLists.txt. Unfortunately I don't have a moment to check manually, and it appears my previous recommendation to use eos-system-contract's CI doesn't work because the version check balks at Leap's main branch being 4.0.0 (and there is no way to disable that check currently? :confused: ).

When adding support for installing bn256 keep in mind it needs to be effectively `COMPONENT dev EXCLUDE_FROM_ALL` from leap's perspective, so you will likely need to follow the pattern that was used in fc, ff, etc

libraries/libfc/test/crypto/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ target_link_libraries( test_k1_recover fc )
 add_test(NAME test_cypher_suites COMMAND test_cypher_suites WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 add_test(NAME test_webauthn COMMAND test_webauthn WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 add_test(NAME test_hash_functions COMMAND test_hash_functions WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
-add_test(NAME test_alt_bn128 COMMAND test_alt_bn128 WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+# add_test(NAME test_alt_bn128 COMMAND test_alt_bn128 WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
```
spoonincode, line 25: a few things commented out here, not sure if they need to be restored, removed, or fixed

spoonincode:
The tester cmake changes appear to need work still, see
https://github.com/eosnetworkfoundation/eos-system-contracts/actions/runs/3635381933
for a run with latest 9af2daf

huangminghuang:
> The tester cmake changes appear to need work still, see
> https://github.com/eosnetworkfoundation/eos-system-contracts/actions/runs/3635381933
> for a run with latest [9af2daf](https://github.com/AntelopeIO/leap/commit/9af2dafa7270d295b6cdf531e76a5c562d3a582d)

fixed by commit [098e535](https://github.com/AntelopeIO/leap/pull/461/commits/098e53528c293185082bac855ec7fdcfef1458c5)

libraries/libfc/test/crypto/CMakeLists.txt
huangminghuang: The test is using libff APIs, so it needs to be removed. The bn256 submodule have the corresponding tests already.

libraries/chain/webassembly/crypto.cpp
```diff
@@ -117,51 +117,26 @@ namespace eosio { namespace chain { namespace webassembly {
    }
 
    int32_t interface::alt_bn128_add(span<const char> op1, span<const char> op2, span<char> result ) const {
-      bytes bop1(op1.data(), op1.data() + op1.size());
-      bytes bop2(op2.data(), op2.data() + op2.size());
-
-      auto maybe_err = fc::alt_bn128_add(bop1, bop2);
-      if(std::holds_alternative<fc::alt_bn128_error>(maybe_err)) {
+      if (op1.size() < 64 ||  op2.size() < 64 ||  result.size() < 64 ||
```
spoonincode, line 120: I think it's only `result` that should be  changed to `< 64`. In the previous impl `op1` and `op2` are copied in to `bytes` (vectors) that are the same size as the `op1` and `op2` inputs before being fed to `alt_bn128_add()`:
https://github.com/AntelopeIO/leap/blob/22daed5030ba5bc9bac81360c8c6ddd58b0d434f/libraries/chain/webassembly/crypto.cpp#L120-L123
That is then,
https://github.com/AntelopeIO/leap/blob/22daed5030ba5bc9bac81360c8c6ddd58b0d434f/libraries/libfc/src/crypto/alt_bn128.cpp#L142-L153
and `decode_g1_element()` does,
https://github.com/AntelopeIO/leap/blob/22daed5030ba5bc9bac81360c8c6ddd58b0d434f/libraries/libfc/src/crypto/alt_bn128.cpp#L42-L45

spoonincode (changes requested):

spoonincode:
Everything looks good from what I can tell, but since that input validation took a couple goes to get the consensus behavior aligned, it's probably prudent to add a unit test for that behavior. That is, a unit test that ensures that for `alt_bn128_add`, `op1` and `op2` can only be 64 bytes, where as `result` can be 64 or more; similar for `alt_bn128_mul`.

linh2931 (approved):

spoonincode (approved):
Looks good. But, I don't think the compile time decision in the bn256 library on whether to use BMI2 instructions is the appropriate choice. That can be resolved at a later time. I've added AntelopeIO/bn256#11 to track and discuss that.

PR #467, opened 2022-11-15
[TH] Replace hardcoded use of var/lib in tests

ClaytonCalabrese:
In an effort to allow all tests to be run in parallel, the tests that read nodeos logs need unique directories  as to not cause problems with each other. This PR uses a directory name based on the test name.


spoonincode:
(besides whatever the issue is at the moment), you probably need to make some modifications to ensure that data files are properly uploaded on failed tests. Right now certain paths are uploaded
https://github.com/AntelopeIO/leap/blob/e16bd9404f9aea9736f5482156a50c2b05684c76/.github/workflows/build.yaml#L189
But now things will be different.

spoonincode:
I'd also recommend getting this branch resynced to master so that LR tests run in a PR (easier to see if those are working properly too, vs needing to make a manual CI run).

tests/validate-dirty-db.py
```diff
@@ -39,7 +40,7 @@
 def runNodeosAndGetOutput(myTimeout=3):
     """Startup nodeos, wait for timeout (before forced shutdown) and collect output. Stdout, stderr and return code are returned in a dictionary."""
     Print("Launching nodeos process.")
-    cmd="programs/nodeos/nodeos --config-dir etc/eosio/node_bios --data-dir var/lib/node_bios --verbose-http-errors --http-validate-host=false --resource-monitor-not-shutdown-on-threshold-exceeded"
+    cmd=f"programs/nodeos/nodeos --config-dir etc/eosio/node_bios --data-dir TestLogs/validate-dirty-db.py{getpid()}/lib/node_bios --verbose-http-errors --http-validate-host=false --resource-monitor-not-shutdown-on-threshold-exceeded"
```
spoonincode, line 43: Using the pid like this might make sense for something ephemeral like the CI environment, but for local development does it cause a bunch of "stale" directories to pile up over time?

ClaytonCalabrese: I changed how the cleanup works. All test logs are placed under a directory "TestLogs", and are housed in there. For example
`TestLogs/nodeos_run_test.py12345/...`
If the test is successful, cleanup now deletes the folder at the nodeos_run_test.py12345 level. After a run of all tests, the only remaining folders should have been failures. I still need to address your other comment about compatibility with CI.

.github/workflows/build.yaml
```diff
@@ -186,7 +186,7 @@ jobs:
         uses: ./.github/actions/parallel-ctest-containers
         with:
           container: ${{fromJSON(needs.d.outputs.p)[matrix.platform].image}}
-          error-log-paths: '["build/etc", "build/var", "build/leap-ignition-wd"]'
+          error-log-paths: '["build/etc", "build/var", "build/leap-ignition-wd", "build/TestLogs/*"]'
```
spoonincode, line 189: I think it should just be `"build/TestLogs"`. The code just looks at `startsWith()` on these

https://github.com/AntelopeIO/leap/blob/e1b51aa0f34f86e190c6e50d5d0bbfb6b6ba044b/.github/actions/parallel-ctest-containers/main.mjs#L54-L58

tests/TestHarness/Cluster.py
```diff
@@ -166,7 +166,7 @@ def setAlternateVersionLabels(self, file):
     # pylint: disable=too-many-statements
     def launch(self, pnodes=1, unstartedNodes=0, totalNodes=1, prodCount=1, topo="mesh", delay=1, onlyBios=False, dontBootstrap=False,
                totalProducers=None, sharedProducers=0, extraNodeosArgs="", useBiosBootFile=True, specificExtraNodeosArgs=None, onlySetProds=False,
-               pfSetupPolicy=PFSetupPolicy.FULL, alternateVersionLabelsFile=None, associatedNodeLabels=None, loadSystemContract=True):
+               pfSetupPolicy=PFSetupPolicy.FULL, alternateVersionLabelsFile=None, associatedNodeLabels=None, loadSystemContract=True, nodeosLogPath=f"TestLogs/{os.path.basename(sys.argv[0])}{os.getpid()}/lib/"):
```
jgiszczak, line 169: I hate to bring it up, but what do you think of eliminating the vestigial `/lib/` component of the path that appears throughout?

jgiszczak (changes requested):

tests/launcher_test.py
```diff
@@ -226,8 +227,8 @@
     if assertionsFound:
         # Too many assertion logs, hard to validate how many are genuine. Make this a warning
         #  for now, hopefully the logs will get cleaned up in future.
-        Print("WARNING: Asserts in var/lib/node_00/stderr.txt")
-        #errorExit("FAILURE - Assert in var/lib/node_00/stderr.txt")
+        Print(f"WARNING: Asserts in TestLogs/{Path(__file__).stem}{os.getpid()}/node_00/stderr.txt")
+        #errorExit("FAILURE - Assert in launcher_test.py/node_00/stderr.txt")
```
jgiszczak, line 231: Another stray line of commented code.

jgiszczak (approved):

PR #468, opened 2022-11-15
Add text/plain content type to http_plugin

ndcgundlach:
Added the ability to set a content type for an api in the http plugin.  The default is application/json, but now text/plain is also available.  

Changed prometheus_plugin api to be text/plain.

### DIFF
--- a/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
@@ -75,6 +75,7 @@ bool allow_host(const http::request<http::string_body>& req, T& session,
    return true;
 }
 
+
 // Handle HTTP conneciton using boost::beast for TCP communication
 // Subclasses of this class (plain_session, ssl_session, etc.)
 // use the Curiously Recurring Template Pattern so that
@@ -98,6 +99,19 @@ protected:
 
    // whether response should be sent back to client when an exception occurs
    bool is_send_exception_response_ = true;
+   http_content_type content_type_ = http_content_type::json; // default to json
+
+   void set_content_type_header() {
+      switch (content_type_) {
+         case http_content_type::plaintext:
+            res_->set(http::field::content_type, "text/plain");
+            break;
+
+         case http_content_type::json:
+         default:
+            res_->set(http::field::content_type, "application/json");
+      }
+   }
 
    template<
          class Body, class Allocator>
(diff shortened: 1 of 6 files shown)


plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -150,10 +163,12 @@ class beast_http_session : public detail::abstract_conn {
             if(plugin_state_->logger.is_enabled(fc::log_level::all))
                plugin_state_->logger.log(FC_LOG_MESSAGE(all, "resource: ${ep}", ("ep", resource)));
             std::string body = req.body();
-            handler_itr->second(derived().shared_from_this(),
+            auto content_type = handler_itr->second.content_type;
```
huangminghuang, line 166: I would expect you to parse the content type from the request header and send response based on the parsed content type. However, you didn't seem to do that.

ndcgundlach: We can add that.  But, currently the code always returns json.  Probably a lot of api clients specify */*.  So we have to fall back to something.  The content type I added for api is the something.  I really don't see the use case for returning a json encoded prometheus string, though.

huangminghuang: You need to keep two different maps for handlers, like `url_json_handlers` and `url_text_handlers`, and when you need to call the handler, you provide different http response handler to the user provided handler. For example, if the request content type is json, you find the handler from url_json_handlers and call the handler with  
 the result of `make_http_json_reponse_handler()` and the `http_json_reponse_handler` should do the json encoding.

plugins/http_plugin/include/eosio/http_plugin/common.hpp
```diff
@@ -212,36 +217,58 @@ auto make_in_flight(T&& object, std::shared_ptr<http_plugin_state> plugin_state)
...
+                                 }
+                              } catch (...) {
+                                 session_ptr->handle_exception();
                               }
-                           } catch(...) {
-                              session_ptr->handle_exception();
-                           }
-                        });
-   };// end lambda
+                           });
```
huangminghuang, line 267: I would just use something like 
```
boost::asio::post(plugin_state->thread_pool->get_executor(),
                           [plugin_state, session_ptr, code, deadline, start,
                                 tracked_response = std::move(tracked_response), content_type]() {
                              try {
                                 if (tracked_response->obj().has_value()) {
                                       std::string str = (content_type == http_content_type::plaintext) ? tracked_response->obj()->as_string() : c::json::to_string(*tracked_response->obj(),
                                                                              deadline +
                                                                              (fc::time_point::now() - start));
                                       in_flight<std::string> tracked_text(std::move(str), plugin_state);
                                       session_ptr->send_response(std::move(tracked_text.obj()), code);
                                 } else {
                                    session_ptr->send_response("{}", code);
                                 }
                              } catch (...) {
                                 session_ptr->handle_exception();
                              }
                           });
```

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -13,7 +13,7 @@
           try { \
              body = parse_params<std::string, http_params_types::no_params>(body); \
              INVOKE \
-             cb(http_response_code, fc::time_point::maximum(), fc::variant(result)); \
+             cb(http_response_code, fc::time_point::maximum(), std::move(fc::variant(result))); \
```
heifner, line 16: Should be `fc::variant(std::move(result))`

huangminghuang (approved):

Issue #478, opened 2022-11-16
Clean up compute_transaction implementation


linh2931:
While https://github.com/AntelopeIO/leap/issues/440 is being worked on, it is discovered previous `compute_transaction` implementation has several shortcomings that conflicts or make it harder to implement new functionalities.  They should be cleaned up so that https://github.com/AntelopeIO/leap/issues/440 and https://github.com/AntelopeIO/leap/issues/364 have a clean baseline code.

Those shortcomings are

1. `compute_transaction` is invoked by `cleos push transaction <transaction> --read-only`. The `read-only` is misleading and conflicts with real read only transactions. A new option `--compute` should be used.
2. Throughout the implementation, `read_only` flag is used to indicate `compute_transaction`. This should be changed.
3. `transaction_metadata` has an enum class `trx_type` which includes `implicit`, `scheduled`, and `read_only`. It also has 3 separate fields `implicit`, `scheduled`, and `read_only`. A transaction type field should be used to support future type extension.
4. `transaction_context` has a field `read_only` for `compute_transaction`. This should be replaced with a transaction type field.

PR #492, opened 2022-11-18
clean up compute_transaction: replace read_only flag with dry_run, refactor trx_type in transaction_metadata and transaction_context

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/478.

The PR resolved the following issues:
1. Throughout the implementation, `read_only` flag was used to indicate `compute_transaction`, which was not true read-only transaction. It has been changed to `dry_run`.
2. `transaction_metadata` had an enum class `trx_type` which included `implicit`, `scheduled`, and `read_only`. It also had 3 separate fields `implicit`, `scheduled`, and `read_only`. `read_only` in `trx_type` has been replaced with `dry_run`. A `trx_type` field  has been added to support future type extension.
3. `transaction_context` had a field `read_only` for `compute_transaction`. This has been replaced with a `trx_type` field.


libraries/chain/include/eosio/chain/transaction_context.hpp
```diff
@@ -83,6 +83,8 @@ namespace eosio { namespace chain {
 
          void validate_referenced_accounts( const transaction& trx, bool enforce_actor_whitelist_blacklist )const;
 
+         const bool is_dry_run()const { return trx_type == transaction_metadata::trx_type::dry_run; };
```
heifner, line 86: Is there a reason to return a `const bool` instead of just `bool`?

libraries/chain/include/eosio/chain/transaction_metadata.hpp
```diff
@@ -24,18 +24,16 @@ class transaction_metadata {
...
          scheduled,
-         read_only
+         dry_run
       };
 
    private:
       const packed_transaction_ptr                               _packed_trx;
       const fc::microseconds                                     _sig_cpu_usage;
       const flat_set<public_key_type>                            _recovered_pub_keys;
+      const trx_type                                           _trx_type;
```
heifner, line 34: Minor indentation issue.

libraries/chain/include/eosio/chain/transaction_metadata.hpp
```diff
@@ -73,6 +69,10 @@ class transaction_metadata {
       fc::microseconds signature_cpu_usage()const { return _sig_cpu_usage; }
       const flat_set<public_key_type>& recovered_keys()const { return _recovered_pub_keys; }
       size_t get_estimated_size() const;
+      const trx_type get_trx_type() const { return _trx_type; };
+      const bool implicit() const { return _trx_type == trx_type::implicit; };
+      const bool scheduled() const { return _trx_type == trx_type::scheduled; };
+      const bool is_dry_run() const { return _trx_type == trx_type::dry_run; };
```
heifner, line 75: Same question, why the `const` on return types?

heifner (changes requested):

libraries/chain/include/eosio/chain/transaction_context.hpp
linh2931: Thanks for your point. As it returns by a value of a built-in type, no need for `const`. Will remove it.

libraries/chain/include/eosio/chain/transaction_metadata.hpp
linh2931: Thanks. Will correct the indentation.

libraries/chain/include/eosio/chain/transaction_metadata.hpp
linh2931: Will remove `const`.

heifner (approved):

Issue #494, opened 2022-11-18
Valgrind - undefined behavior - low priorities


mikelik:
When running:

1.` valgrind /home/mikel/repo/leap/debug/unittests/unit_test --run_test=block_tests --report_level=detailed --color_output --catch_system_errors=no -- --eos-vm`

following error is displayed:
```
==5826== Syscall param msync(start) points to uninitialised byte(s)
==5826==    at 0x5DA8A0F: msync (msync.c:25)
==5826==    by 0xE71A62: boost::interprocess::mapped_region::flush(unsigned long, unsigned long, bool) (mapped_region.hpp:747)
==5826==    by 0xE6F083: chainbase::pinnable_mapped_file::~pinnable_mapped_file() (pinnable_mapped_file.cpp:321)
==5826==    by 0xE6A5C7: chainbase::database::~database() (chainbase.cpp:19)
==5826==    by 0x10C4C57: eosio::chain::controller_impl::~controller_impl() (controller.cpp:742)
==5826==    by 0x111D575: std::default_delete<eosio::chain::controller_impl>::operator()(eosio::chain::controller_impl*) const (unique_ptr.h:81)
==5826==    by 0x1102D3B: std::unique_ptr<eosio::chain::controller_impl, std::default_delete<eosio::chain::controller_impl> >::~unique_ptr() (unique_ptr.h:292)
==5826==    by 0x108D6AF: eosio::chain::controller::~controller() (controller.cpp:2670)
==5826==    by 0x5BD323: std::default_delete<eosio::chain::controller>::operator()(eosio::chain::controller*) const (unique_ptr.h:81)
==5826==    by 0x599597: std::unique_ptr<eosio::chain::controller, std::default_delete<eosio::chain::controller> >::~unique_ptr() (unique_ptr.h:292)
==5826==    by 0x58EBCD: eosio::testing::base_tester::~base_tester() (tester.hpp:155)
==5826==    by 0x650133: eosio::testing::tester::~tester() (tester.hpp:447)
==5826==  Address 0x9f82504 is in a rw- mapped file /tmp/8e56-bbaa-10a6-99aa/state/shared_memory.bin segment
```

I think issue might be in boost in mapped_region.hpp:747 where `addr` is uninitialized.
This should be fixed in boost library.
Temporary it might be possible to check if mapped_region is uninitialized before flushing.

2. `valgrind /home/mikel/repo/leap/debug/plugins/chain_plugin/test/test_account_query_db`
following issues appear
[...]

Issue #493, opened 2022-11-18
Undefined behavior - uninitialized variables


mikelik:
When running valgrind on nodeos and unit tests I hit following issues which might be fixed:

1. Scenario: Running nodeos and stopping it.

Command:
```
valgrind nodeos -e -p eosio --data-dir /home/mikel/app/nodeos --plugin eosio::producer_plugin --plugin eosio::producer_api_plugin --plugin eosio::chain_plugin --plugin eosio::chain_api_plugin --plugin eosio::http_plugin --plugin eosio::state_history_plugin --contracts-console --disable-replay-opts --access-control-allow-origin=* --http-validate-host=false --verbose-http-errors --trace-history --chain-state-history --delete-all-blocks

==15446== Syscall param rt_sigaction(act->sa_mask) points to uninitialised byte(s)
==15446==    at 0x5A524F6: __libc_sigaction (sigaction.c:58)
==15446==    by 0x1BC3604: eosio::chain::eosvmoc::launch_compile_monitor(int) (compile_monitor.cpp:253)
==15446==    by 0x1BC5A56: eosio::chain::eosvmoc::compile_monitor_trampoline::start() (compile_monitor.cpp:288)
==15446==    by 0x1BC3820: __wrap_main (compile_monitor.cpp:302)
==15446==    by 0x5E15082: (below main) (libc-start.c:308)
==15446==  Address 0x1ffefff708 is on thread 1's stack
==15446==  in frame #0, created by __libc_sigaction (sigaction.c:43)
```

2. Unit test test_account_query_db

Command:
valgrind /home/mikel/repo/leap/debug/plugins/chain_plugin/test/test_account_query_db

```
==12171== Conditional jump or move depends on uninitialised value(s)
==12171==    at 0x2B02083: void WASM::serialize<Serialization::MemoryInputStream>(Serialization::MemoryInputStream&, IR::ControlStructureImm&, IR::FunctionDef const&) (WASMSerialization.cpp:234)
==12171==    by 0x2AE5E1F: WASM::serializeFunctionBody(Serialization::InputStream&, IR::Module&, IR::FunctionDef&) (WASMSerialization.cpp:531)
==12171==    by 0x2B041F9: void WASM::serializeCodeSection<Serialization::InputStream>(Serialization::InputStream&, IR::Module&)::{lambda(Serialization::InputStream&)#1}::operator()(Serialization::InputStream&) const (WASMSerialization.cpp:768)
[...]

PR #495, opened 2022-11-18
Fix undefined behavior - uninitialized variables revealed by valgrind

mikelik:
Fixing https://github.com/AntelopeIO/leap/issues/493.

Fixed undefined behavior revealed by valgrind. After fixes all mentioned issues are cleared out:
1.
```
valgrind nodeos -e -p eosio --data-dir /home/mikel/app/nodeos --plugin eosio::producer_plugin --plugin eosio::producer_api_plugin --plugin eosio::chain_plugin --plugin eosio::chain_api_plugin --plugin eosio::http_plugin --plugin eosio::state_history_plugin --contracts-console --disable-replay-opts --access-control-allow-origin=* --http-validate-host=false --verbose-http-errors --trace-history --chain-state-history --delete-all-blocks

(...)
==46508== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```
2.
`valgrind /home/mikel/repo/leap/debug/plugins/chain_plugin/test/test_account_query_db`

Valgrind reports only issues which are created as a separate issue (https://github.com/AntelopeIO/leap/issues/494).

3.
Original warning is gone:

```
valgrind /home/mikel/repo/leap/debug/unittests/unit_test "--run_test=chain_tests" "--report_level=detailed" "--color_output" "--catch_system_errors=no" "--" "--eos-vm-oc"
(...)
==46794== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

4. 
```
valgrind /home/mikel/repo/leap/debug/libraries/libfc/test/network/test_message_buffer
(...)
==46567== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

### DIFF
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/ipc_helpers.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/ipc_helpers.cpp
@@ -89,7 +89,7 @@ bool write_message_with_fds(int fd_to_send_to, const eosvmoc_message& message, c
       .iov_len = sz
    };
    union {
-      char buf[CMSG_SPACE(max_num_fds * sizeof(int))];
+      char buf[CMSG_SPACE(max_num_fds * sizeof(int))] = {};
       struct cmsghdr align;
    } u;
 
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/ipc_helpers.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/ipc_helpers.cpp
@@ -108,7 +108,6 @@ bool write_message_with_fds(int fd_to_send_to, const eosvmoc_message& message, c
          memcpy(p, &thisfd, sizeof(thisfd));
          p += sizeof(thisfd);
       }
-      msg.msg_controllen = cmsg->cmsg_len;
    }
 
    int wrote;
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/ipc_helpers.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/ipc_helpers.cpp
@@ -108,6 +108,7 @@ bool write_message_with_fds(int fd_to_send_to, const eosvmoc_message& message, c
          memcpy(p, &thisfd, sizeof(thisfd));
          p += sizeof(thisfd);
       }
+      msg.msg_controllen = cmsg->cmsg_len;
    }
 
    int wrote;
--- a/libraries/libfc/test/network/test_message_buffer.cpp
+++ b/libraries/libfc/test/network/test_message_buffer.cpp
@@ -79,7 +79,7 @@ BOOST_AUTO_TEST_CASE(message_buffer_growth)
     BOOST_CHECK_EQUAL(mb.bytes_to_read(), 100u);
     BOOST_CHECK_NE(mb.read_ptr(), nullptr);
     BOOST_CHECK_NE(mb.write_ptr(), nullptr);
-    BOOST_CHECK_EQUAL((mb.read_ptr() + 100), mb.write_ptr());
+    BOOST_CHECK_EQUAL(static_cast<void*>(mb.read_ptr() + 100), static_cast<void*>(mb.write_ptr()));
 
     {
       auto mbs = mb.get_buffer_sequence_for_boost_async_read();
--- a/libraries/wasm-jit/Include/IR/Operators.h
+++ b/libraries/wasm-jit/Include/IR/Operators.h
@@ -14,60 +14,60 @@ namespace IR
 
 	struct ControlStructureImm
 	{
-		ResultType resultType;
+		ResultType resultType{};
 	};
 
 	struct BranchImm
 	{
-		U32 targetDepth;
+		U32 targetDepth{};
 	};
 
 	struct BranchTableImm
 	{
-		Uptr defaultTargetDepth;
-		Uptr branchTableIndex; // An index into the FunctionDef's branchTables array.
+		Uptr defaultTargetDepth{};
+		Uptr branchTableIndex{}; // An index into the FunctionDef's branchTables array.
 	};
 
 	template<typename Value>
 	struct LiteralImm
 	{
-		Value value;
+		Value value{};
 	};
 
 	template<bool isGlobal>
 	struct GetOrSetVariableImm
 	{
-		U32 variableIndex;
+		U32 variableIndex{};
 	};
 
 	struct CallImm
 	{
-		U32 functionIndex;
+		U32 functionIndex{};
 	};
 
 	struct CallIndirectImm
 	{
-		IndexedFunctionType type;
+		IndexedFunctionType type{};
 	};
 
 	template<Uptr naturalAlignmentLog2>
 	struct LoadOrStoreImm
 	{
-		U8 alignmentLog2;
-		U32 offset;
+		U8 alignmentLog2{};
+		U32 offset{};
 	};
 
 	#if ENABLE_SIMD_PROTOTYPE
 	template<Uptr numLanes>
 	struct LaneIndexImm
 	{
-		U8 laneIndex;
+		U8 laneIndex{};
 	};
 	
 	template<Uptr numLanes>
 	struct ShuffleImm
 	{
-		U8 laneIndices[numLanes];
+		U8 laneIndices[numLanes] = {};
 	};
 	#endif
 
--- a/libraries/wasm-jit/Include/IR/Operators.h
+++ b/libraries/wasm-jit/Include/IR/Operators.h
@@ -19,55 +19,55 @@ namespace IR
 
 	struct BranchImm
 	{
-		U32 targetDepth{};
+		U32 targetDepth;
 	};
 
 	struct BranchTableImm
 	{
-		Uptr defaultTargetDepth{};
-		Uptr branchTableIndex{}; // An index into the FunctionDef's branchTables array.
+		Uptr defaultTargetDepth;
+		Uptr branchTableIndex; // An index into the FunctionDef's branchTables array.
 	};
 
 	template<typename Value>
 	struct LiteralImm
 	{
-		Value value{};
+		Value value;
 	};
 
 	template<bool isGlobal>
 	struct GetOrSetVariableImm
 	{
-		U32 variableIndex{};
+		U32 variableIndex;
 	};
 
 	struct CallImm
 	{
-		U32 functionIndex{};
+		U32 functionIndex;
 	};
 
 	struct CallIndirectImm
 	{
-		IndexedFunctionType type{};
+		IndexedFunctionType type;
 	};
 
 	template<Uptr naturalAlignmentLog2>
 	struct LoadOrStoreImm
 	{
-		U8 alignmentLog2{};
-		U32 offset{};
+		U8 alignmentLog2;
+		U32 offset;
 	};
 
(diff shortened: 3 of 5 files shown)


libraries/libfc/test/network/test_message_buffer.cpp
```diff
@@ -79,7 +79,7 @@ BOOST_AUTO_TEST_CASE(message_buffer_growth)
     BOOST_CHECK_EQUAL(mb.bytes_to_read(), 100u);
     BOOST_CHECK_NE(mb.read_ptr(), nullptr);
     BOOST_CHECK_NE(mb.write_ptr(), nullptr);
-    BOOST_CHECK_EQUAL((mb.read_ptr() + 100), mb.write_ptr());
+    BOOST_CHECK_EQUAL(static_cast<void*>(mb.read_ptr() + 100), static_cast<void*>(mb.write_ptr()));
```
mikelik, line 82: That one is interesting, because BOOST_CHECK_EQUAL was using strncmp (and compare uninitialized memory) while we were interested in raw pointers comparison.

libraries/chain/webassembly/runtimes/eos-vm-oc/ipc_helpers.cpp
```diff
@@ -89,7 +89,7 @@ bool write_message_with_fds(int fd_to_send_to, const eosvmoc_message& message, c
       .iov_len = sz
    };
    union {
-      char buf[CMSG_SPACE(max_num_fds * sizeof(int))];
+      char buf[CMSG_SPACE(max_num_fds * sizeof(int))] = {};
```
mikelik, line 92: This was uninitialized when called from ipc_helpers.cpp:115.

In line 115 `wrote = sendmsg(fd_to_send_to, &msg, 0);` msg is used.
`msg` has member: `msg_control` which is initialized in line 99: `msg.msg_control = u.buf;`.
However memeber of union `u.buf` is unitialized (declaration in line 94).

This happens in every scenario where function `write_message_with_fds` is being used.

libraries/wasm-jit/Include/IR/Operators.h
```diff
@@ -14,60 +14,60 @@ namespace IR
 
 	struct ControlStructureImm
 	{
-		ResultType resultType;
+		ResultType resultType{};
```
mikelik, line 17: In this file valgrind found only issue with uninitialized resultType.
However I initialized by default all other structures just in case.

Stack trace for this issue:
```
==12171==    at 0x2B02083: void WASM::serialize<Serialization::MemoryInputStream>(Serialization::MemoryInputStream&, IR::ControlStructureImm&, IR::FunctionDef const&) (WASMSerialization.cpp:234)
==12171==    by 0x2AE5E1F: WASM::serializeFunctionBody(Serialization::InputStream&, IR::Module&, IR::FunctionDef&) (WASMSerialization.cpp:531)
```

resultType is being checked in WASMSerialization.cpp:234:
`I8 encodedResultType = imm.resultType == ResultType::none ? -64 : -(I8)imm.resultType;`
However the argument `ControlStructureImm& imm` is being passed up from function above: WASMSerialization.cpp:531.
The declaration of `ConstrolStructureImm& imm` I believe is in WASMSerialization.cpp:525 - which (as a good practice) could also be initialized.
However fixing the issue at the very bottom in file Operators.h will work for any usage of ControlStructureImm.

Above scenario happens for every use of `WASM::serializeFunctionBody`. One example is unit test:
`plugins/chain_plugin/test/test_account_query_db`

libraries/wasm-jit/Source/WASM/WASMSerialization.cpp
```diff
@@ -406,8 +406,8 @@ namespace WASM
 
 	struct LocalSet
 	{
-		Uptr num;
-		ValueType type;
+		Uptr num{};
```
mikelik, line 409: Stack trace for it
```
WASMSerialization.cpp:28
WASMSerialization.cpp:417
WASMSerialization.cpp:504
```

Starting from top in line `WASMSerialization.cpp:504` we define: `LocalSet localSet` which is unitialized. It is passed down to `serialize(Stream& stream,ValueType& type)` in WASMSerialization.cpp:26 where LocalSet::type is read (uninitialized).

It happens for every usage of `serializeFunctionBody(InputStream& sectionStream,Module& module,FunctionDef& functionDef)` in WASMSerialization.cpp.
Unfortunately I lost an example unit test which uses this function.

libraries/libfc/test/network/test_message_buffer.cpp
greg7mdp: I'm surprised that the test didn't fail!

libraries/wasm-jit/Include/IR/Operators.h
greg7mdp: There might be a slight performance cost if we initialize structs needlessly (if they are being overwritten in all cases). Is this a tradeoff we want?

libraries/wasm-jit/Source/WASM/WASMSerialization.cpp
greg7mdp: I thing that another way to address this umr would be to change the serialize function to:

```
	template<typename Stream>
	void serialize(Stream& stream,ValueType& type)
	{
            if(Stream::isInput) {
                serializeVarInt7(stream,encodedValueType);
                type = (ValueType)-encodedValueType;
            } else {
		I8 encodedValueType = -(I8)type;
		serializeVarInt7(stream,encodedValueType);
            }
	}
```

libraries/chain/webassembly/runtimes/eos-vm-oc/ipc_helpers.cpp
greg7mdp: I am not seeing why uninitialized memory was accessed? At line 104, we set `cmsg->cmsg_len = CMSG_LEN(sizeof(int) * fds.size());`, which is the same size we initialize in the loop at line 106. So I assume that the `sendmsg` function should only access that area of `u.buf` that has been initialized. Zeroing the whole buffer might be overkill.

greg7mdp (commented):
Just leaving some comments.

libraries/chain/webassembly/runtimes/eos-vm-oc/ipc_helpers.cpp
spoonincode: Right. My guess is valgrind just isn't smart enough to fully reason on the behavior of this syscall. But it's probably best practice to initialize it anyways. It's small and not performance sensitive.

mikelik: I confirm (sizeof(int) * fds.size()) == 8 in my case.
Also from debugger: CMSG_LEN(sizeof(int) * fds.size()) == 24.
Apparently last 8 bytes of u.msg are not initialized and they are being looked at by sendmsg.
I don't have sendmsg code in debugger, but probably alternative / better way would be to set msg.msg_controllen to 24 from 32 in our case? However the logic would be harder to read i.e. something like this after end of the loop
` msg.msg_controllen = (void*) p - (void*) cmsg;`

Details from debugger, in my case:
u.buf -> points to i.e. 0x1ffeffa4b0 - 32 bytes
cmsg -> 0x1ffeffa4b0 - 32 bytes
cmsg->cmsg_len -> 0x1ffeffa4b0 - initialized 8 bytes
cmsg->cmsg_level -> 0x1ffeffa4b8 - initialized 4 bytes
cmsg->cmsg_type -> 0x1ffeffa4bc - initialized 4 bytes
CMSG_DATA(cmsg) -> 0x1ffeffa4c0 - initialized 8 bytes total in the loop = 2x sizeof(thisfd)

greg7mdp: I couldn't find where `CMSG_LEN` is defined, but `CMSG_LEN(sizeof(int) * fds.size()) == 24` seems wrong.
Also just noticed that `max_num_fds == 4` so yes pre-zeroing the buffer of 4 ints should be fast enough.

mikelik: CMSG_LEN definition: https://github.com/openSUSE/kernel/blob/master/include/linux/socket.h#L122
And help is here: https://linux.die.net/man/3/cmsg_space
"first initialize the msg_controllen member of the msghdr with the length of the control message buffer"
so msg.msg_controllen looks good.
The same for CMSG_LEN: "In each control message, initialize cmsg_len (with CMSG_LEN()), the other cmsghdr header fields, and the data portion using CMSG_DATA().".

So I don't see any other way than just initialize whole u.buf (or at least 8 unused bytes).

greg7mdp: Thanks @mikelik , I didn't realize that was a system macro. And yes, it does make sense to initialize the whole `u.buf`. Thanks for the explanation!

greg7mdp: Actually, after looking at the doc you referred to, I do think it would be more correct to set:

`msg.msg_controllen = p - CMSG_DATA(cmsg);`

After the loop (so that we correctly only send the descriptors we initialized, and not extra descriptors set to 0).

mikelik: The proper way is shown in the example at the bottom: https://linux.die.net/man/3/cmsg_space

libraries/wasm-jit/Source/WASM/WASMSerialization.cpp
mikelik: Yes, it seems both are good solutions.

libraries/wasm-jit/Include/IR/Operators.h
mikelik: I don't think this is performance critical code.

libraries/libfc/test/network/test_message_buffer.cpp
mikelik: Both pointers were pointing to the same memory, so it was comparing some garbage to the same garbage :)

mikelik (commented):
Thanks Greg for the review

libraries/libfc/test/network/test_message_buffer.cpp
greg7mdp: Of course, makes sense, thanks @mikelik

mikelik:
Now the tests are passing, however I had to remove most of the changes from Operators.h.
It seems that build is not triggered when you change review from "Draft" to "Ready for review".
Build: https://github.com/AntelopeIO/leap/actions/runs/3524072249

libraries/wasm-jit/Source/WASM/WASMSerialization.cpp
greg7mdp: I slightly prefer my solution and I'll say why. The `serialize` function can either output the parameter `type` to the stream, or read from the stream and set it. When reading from the stream, it makes no sense to access the variable `type` just before overwriting what it contains.

So I think it is better to remove the useless read, rather than initializing the memory, so that the useless read we do does not read uninitialized memory.

But this is me being a little of a perfectionist... so I guess not a big deal we you leave it as is. Thanks!

greg7mdp (approved):

ScottBailey (approved):
lgtm

Issue #286, opened 2022-10-05
dfuseeos issue with accepting trxs


NatPDeveloper:
I keep getting this when trying on 3.1.1 or 3.1.0.

This happens only when I set `api-accept-transactions = true`. When it's false it works as expected.

```
info  2022-10-05T03:13:58.065 nodeos    producer_plugin.cpp:504       on_incoming_block    ] Received block 6baa9329efacae6d... #252803482 @ 2022-10-05T03:13:58.000 signed by eosiosg11111 [trxs: 0, lib: 252803151, conf: 0, latency: 65 ms]
reading from console logs (mindreader/mindreader.go:230) {"error": "START_BLOCK: abi decoder: start block for block #252803474 received while already processing block #252803474 (line \"START_BLOCK 252803474\")"}

################################################################
Fatal error in app mindreader-stdin:

START_BLOCK: abi decoder: start block for block #252803474 received while already processing block #252803474 (line "START_BLOCK 252803474")
################################################################
 (launcher/launcher.go:174) 
Application mindreader-stdin shutdown unexpectedly, quitting 
Error: unable to launch: START_BLOCK: abi decoder: start block for block #252803474 received while already processing block #252803474 (line "START_BLOCK 252803474")
dfuse (derr/cli.go:25) {"error": "unable to launch: START_BLOCK: abi decoder: start block for block #252803474 received while already processing block #252803474 (line \"START_BLOCK 252803474\")"}
DMLOG FPRINTF_FAILED failed written=0 remaining=28 1 Broken pipe
DMLOG FPRINTF_FAILURE_TERMINATED
```

Command:

`dfuseeos purge -f; nodeos --disable-replay-opts | dfuseeos start -c ./mindreader.yaml`

Steps to reproduce: https://medium.com/@NatPDeveloper/antelopes-magic-api-firehose-14dec52b5670 using leap 3.1.1 and using `api-accept-transactions = true`.

Regarding priority, this requires us to run a second nodeos instance in production for DSPs to handle transactions with the retry handler enabled.

---

Also I am running the mindreader node locally with zeus with no issues so on a single node local network this works fine.

heifner:
The START_BLOCK is logged on all `start_block()` calls including speculative blocks for trx processing. When `api-accept-transactions = false` & `p2p-accept-transactions=false`, `producer_plugin_impl::start_block()` does not create a speculative block. A speculative block is also not created when running in `irreversible` or `read_only` modes.

[...]

PR #500, opened 2022-11-19
Added validation of plugin configuration for deep-mind

systemzax:
This PR adds validation of plugin configuration for deep-mind and will throw an exception if the plugin is not configured properly

See : https://github.com/AntelopeIO/leap/issues/286

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -1124,6 +1124,17 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
          //
          // For the time being, when `deep-mind = true` is activated, we set `stdout` here to
          // be an unbuffered I/O stream.
+
+         //verify configuration is correct
+         EOS_ASSERT( my->chain_config->read_mode == db_read_mode::HEAD, plugin_config_exception, 
+            "read-mode must be set to head in order to enable deep-mind logging.");
+
+         EOS_ASSERT( options.at("api-accept-transactions").as<bool>() == false, plugin_config_exception,
+            "api-accept-transactions must be set to false in order to enable deep-mind logging.");
+
+         EOS_ASSERT( options.at("p2p-accept-transactions").as<bool>() == false, plugin_config_exception,
+            "p2p-accept-transactions must be set to false in order to enable deep-mind logging.");
+
          setbuf(stdout, NULL);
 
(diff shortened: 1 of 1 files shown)


systemzax:
I mistakenly created a new repo for that fork, and couldn't deleted. Apologies!

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1124,6 +1124,17 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
          //
          // For the time being, when `deep-mind = true` is activated, we set `stdout` here to
          // be an unbuffered I/O stream.
+
+         //verify configuration is correct
+         EOS_ASSERT( my->chain_config->read_mode == db_read_mode::HEAD, plugin_config_exception, 
+            "read-mode must be set to head in order to enable deep-mind logging.");
```
heifner, line 1130: This disallows running deep-mind in irreversible mode. Unless we know it doesn't work in irreversible mode, this should be removed. In `main` the only two modes are `HEAD` and `IRREVERSIBLE`.

heifner (changes requested):

plugins/chain_plugin/chain_plugin.cpp
systemzax: Dfuse docs suggests using read-mode = HEAD, which is what we have been using in all of our production setups, but we haven't tried running it with IRREVERSIBLE.

194128707-d6c1173e-738e-488a-8af1-48d8d4d382c6

Will remove the read_mode check, we can revisit this if someone runs into trouble running it in irreversible mode.

heifner (approved):

heifner:
@spoonincode thoughts on if this should be back-ported to 3.2 and 3.1?

spoonincode:
Given the simplicity of the fix, and that users did encounter grief with this, my vote would be a yes.

heifner:
@systemzax would you be willing to create a PR for 3.1 for this change. Our procedure is to put it into 3.1 and then merge it into 3.2 and then merge it into main.

systemzax:
> @systemzax would you be willing to create a PR for 3.1 for this change. Our procedure is to put it into 3.1 and then merge it into 3.2 and then merge it into main.

No problem, will do

heifner:
Replaced by #503

Issue #511, opened 2022-11-23
cleos is broken on main branch by newly introduced libcurl


linh2931:
`cleos` recently switched from openssl to libcurl. Any command requiring keys stopped working due to that.
```
cleos create account eosio hello EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV -p eosio@active
Error 3110001: Missing Chain API Plugin
Ensure that you have eosio::chain_api_plugin added to your node's configuration!
Error Details:
Chain API plugin is not enabled
```

After much digging, I tracked to `determine_required_keys` in `cleos/main.cpp` which has two curl calls
```
...
const auto& public_keys = call(wallet_url, wallet_public_keys);
...
const auto& required_keys = call(get_required_keys, get_arg);
```
Those two calls go to the same local host but with different ports (80 and 8888). But somehow libcurl uses the same port for both calls. The investigation eventually led to `do_http_post_libcurl.cpp`'s `handle`, which was declared as a `static` and not completely reset between calls.

I will create a PR shortly to fix this.

I will create another for issue to add cleos tests to capture future regressions.

PR #512, opened 2022-11-24
Fixed broken cleos due to reuse of curl handle

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/511

`cleos` recently switched from openssl to libcurl. Any command requiring keys stopped working due to that. For example,

```
cleos create account eosio hello EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV -p eosio@active
Error 3110001: Missing Chain API Plugin
Ensure that you have eosio::chain_api_plugin added to your node's configuration!
Error Details:
Chain API plugin is not enabled
```

In `determine_required_keys` in `cleos/main.cpp` which has two curl calls
```
...
const auto& public_keys = call(wallet_url, wallet_public_keys);
...
const auto& required_keys = call(get_required_keys, get_arg);
```
Those two calls go to the same local host but with different ports (80 and 8888). But somehow libcurl uses the same port for both calls. 

The root cause was `do_http_post_libcurl.cpp`'s `handle`, which was declared as a `static` and not completely reset between calls.

### DIFF
--- a/programs/cleos/do_http_post_libcurl.cpp
+++ b/programs/cleos/do_http_post_libcurl.cpp
@@ -78,7 +78,7 @@ namespace eosio { namespace client { namespace http {
          initialized = true;
       }
 
-      std::unique_ptr<CURL, decltype(&curl_easy_cleanup)> handle(nullptr, &curl_easy_cleanup);
+      static std::unique_ptr<CURL, decltype(&curl_easy_cleanup)> handle(nullptr, &curl_easy_cleanup);
       if (!handle) handle.reset(curl_easy_init());
       auto curl = handle.get();
       EOS_ASSERT(curl != 0, chain::http_exception, "curl_easy_init failed");
@@ -92,6 +92,8 @@ namespace eosio { namespace client { namespace http {
          curl_easy_setopt(curl, CURLOPT_UNIX_SOCKET_PATH, base_uri.c_str() + unix_socket_prefix_len);
          uri = "http://localhost" + path;
       } else {
+         // Disable use of unix domain in case it was enabled in the previous call
+         curl_easy_setopt(curl, CURLOPT_UNIX_SOCKET_PATH, nullptr);
          uri = base_uri + path;
       }
 
--- a/programs/cleos/do_http_post_libcurl.cpp
+++ b/programs/cleos/do_http_post_libcurl.cpp
@@ -78,7 +78,7 @@ namespace eosio { namespace client { namespace http {
          initialized = true;
       }
 
-      static std::unique_ptr<CURL, decltype(&curl_easy_cleanup)> handle(nullptr, &curl_easy_cleanup);
+      std::unique_ptr<CURL, decltype(&curl_easy_cleanup)> handle(nullptr, &curl_easy_cleanup);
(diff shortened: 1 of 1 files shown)


spoonincode (changes requested):
libcurl's documentation seems pretty clear that both a) [the handle is meant to be reused](https://everything.curl.dev/libcurl/easyhandle) and b) [port number is part of its heuristics to determine whether it can reuse an existing connection](https://everything.curl.dev/libcurl/connectionreuse).

I also tried locally, and something like this works
```
$ programs/cleos/cleos --wallet-url http://127.0.0.1:8080 -u http://127.0.0.1:8888 create account eosio b EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV
executed transaction: b4f8a1ec599c0dbe9f3f02696b808fd6f78886cf251d53b63c3da4eed5683a48  200 bytes  570 us
#         eosio <= eosio::newaccount            {"creator":"eosio","name":"b","owner":{"threshold":1,"keys":[{"key":"EOS6MRyAjQq8ud7hVNYcfnVPJqcVpsc...
warning: transaction executed locally, but may not be confirmed by the network yet         ]
```

What appears to be the problem is that
> All options are "sticky"

And critically one of the options is `CURLOPT_UNIX_SOCKET_PATH` which means that once a unix socket is used for a request it'll always be used for a request. I suspect the proper fix here is to unset `CURLOPT_UNIX_SOCKET_PATH` on each request until determining it really is a unix socket.

Reusing the curl handle provides a performance improvement we want to keep.

linh2931:
> libcurl's documentation seems pretty clear that both a) [the handle is meant to be reused](https://everything.curl.dev/libcurl/easyhandle) and b) [port number is part of its heuristics to determine whether it can reuse an existing connection](https://everything.curl.dev/libcurl/connectionreuse).
> 
> I also tried locally, and something like this works
> 
> ```
> $ programs/cleos/cleos --wallet-url http://127.0.0.1:8080 -u http://127.0.0.1:8888 create account eosio b EOS6MRyAjQq8ud7hVNYcfnVPJqcVpscN5So8BhtHuGYqET5GDW5CV
> executed transaction: b4f8a1ec599c0dbe9f3f02696b808fd6f78886cf251d53b63c3da4eed5683a48  200 bytes  570 us
> #         eosio <= eosio::newaccount            {"creator":"eosio","name":"b","owner":{"threshold":1,"keys":[{"key":"EOS6MRyAjQq8ud7hVNYcfnVPJqcVpsc...
> warning: transaction executed locally, but may not be confirmed by the network yet         ]
> ```
> 
> What appears to be the problem is that
> 
> > All options are "sticky"
> 
> And critically one of the options is `CURLOPT_UNIX_SOCKET_PATH` which means that once a unix socket is used for a request it'll always be used for a request. I suspect the proper fix here is to unset `CURLOPT_UNIX_SOCKET_PATH` on each request until determining it really is a unix socket.
> 
> Reusing the curl handle provides a performance improvement we want to keep.

Thanks! Unset `CURLOPT_UNIX_SOCKET_PATH` worked.

spoonincode (approved):

PR #532, opened 2022-12-02
Add block log partition functionality

huangminghuang:
This PR backport https://github.com/EOSIO/eos/pull/9184 to support splitting block logs automatically based on block numbers. It also allows user to specify how many split block log files to retains. Once the limit is reached, the older block files can be move to another directory or deleted.

It also support recovering from the head block is not completely written to disk automatically which would eliminate most use cases for using eosio-blocklog to manually trim the end of block log file.

Some new options for chain plugin are added:

- blocks-log-stride: split the block log file when the head block number is the multiple of the stride. When the stride is reached, the current blog log and index will be renamed 'blocks-<start num>-<end num>.log/index' and a new current block log and index will be created with the most recent block. All files following this format will be used to construct an extended block log.
- max-retained-block-files: the maximum number of blocks files to retain so that the blocks in those files can be queried. When the number is reached, the oldest block file would be move to archive dir or deleted if the archive dir is empty. The retained block log files should not be manipulated by users.
- blocks-archive-dir: the location of the blocks archive directory (absolute path or relative to blocks dir). If the value is empty, blocks files beyond the retained limit will be deleted. All files in the archive directory are completely under user's control, i.e. they won't be accessed by nodeos anymore.
- blocks-retained-dir: the location of the blocks retained directory (absolute path or relative to blocks dir). If the value is empty, it is set to the value of blocks dir.
- fix-irreversible-blocks: When the existing block log is inconsistent with the index, allows fixing the log file automatically based on the index - that is, it will take the highest indexed block if it is valid; otherwise it will repair the block log and reconstruct the index.

Notice that all the newly added options cannot be specified together with `--block-log-retain-blocks` option.


libraries/chain/block_log.cpp
```diff
@@ -31,1039 +21,1291 @@ namespace eosio { namespace chain {
...
-         return;
-      }
+            block_file.open(fc::cfile::truncate_rw_mode);
+            index_file.open(fc::cfile::truncate_rw_mode);
+            preamble.ver             = block_log::max_supported_version;
+            preamble.chain_context   = preamble.chain_id();
+            preamble.first_block_num = this->head->block_num() + 1;
+            preamble.write_to(block_file);
+            flush();
+         }
```
spoonincode, line 982: Can `split_log()` be made safer so that it first tries to open a new file, and if that fails, continues logging to the current file?

libraries/chain/include/eosio/chain/log_data_base.hpp
```diff
@@ -0,0 +1,49 @@
...
+T read_buffer(const char* buf) {
+   T result;
+   memcpy(&result, buf, sizeof(T));
+   return result;
+}
+
+template <typename Derived>
+class log_data_base {
+ protected:
+   boost::iostreams::mapped_file file;
```
spoonincode, line 17: Using a mapped file to supply log data makes it significantly harder to recover from IO errors. As is, should a log entry fail to read, it will SIGBUS the process I believe -- effectively SIGKILLing it. That seems rather extreme failure mode to me, particularly for something like the block log where it's easy to imagine such a failure could nominally be swallowed.

linh2931 (commented):
It is a major rewrite of block_log.cpp. Please make sure test coverage is sufficient.

libraries/chain/block_log.cpp
linh2931: Agreed. Had similar comments to #628.

libraries/chain/include/eosio/chain/log_catalog.hpp
```diff
@@ -0,0 +1,283 @@
...
+            return {};
+
+         auto it = --collection.upper_bound(block_num);
+
+         if (block_num <= it->second.last_block_num) {
+            auto name = it->second.filename_base;
+            log_data.open(name.replace_extension("log"), mode);
+            log_index.open(name.replace_extension("index"));
+            this->active_index = collection.index_of(it);
+            return log_index.nth_block_position(block_num - log_data.first_block_num());
```
spoonincode, line 159: It seems like, if I'm understanding this correctly, only a single partition is ever "loaded" at a time, and if a block is requested from a partition other then the one that is loaded, it closes the loaded partition and opens the needed partition.

For a node serving significant number of syncing peers that seems like it could result in a lot of thrashing. Tearing down a memory mapping and creating a new one is not cheap, and historically has involved actions like TLB flushes which need to halt all threads (I admit to not being up to date on the state of the art here, so maybe it's not as bad now as I remember).

But even forgetting the memory mapping behavior, I'm not sure how to resolve the thrashing of file descriptors: we can't keep them all open. A cache with simple LRU eviction? Or maybe just file descriptor thrashing isn't enough of a grind to worry about.

Because it appears we need to re-aquire resources, it seems very difficult to get the partitioned block log performance to be as good as the flat block log performance. A flat unpruned block log can serve multiple threads without any locking and with only a single resource acquisition at startup, for example.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -229,6 +229,25 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
...
+          "When the number is reached, the oldest block file would be moved to archive dir or deleted if the archive dir is empty.\n"
+          "The retained block log files should not be manipulated by users." )
+         ("blocks-retained-dir", bpo::value<bfs::path>(),
+          "the location of the blocks retained directory (absolute path or relative to blocks dir).\n"
+          "If the value is empty, it is set to the value of blocks dir.")
+         ("blocks-archive-dir", bpo::value<bfs::path>(),
+          "the location of the blocks archive directory (absolute path or relative to blocks dir).\n"
+          "If the value is empty, blocks files beyond the retained limit will be deleted.\n"
+          "All files in the archive directory are completely under user's control, i.e. they won't be accessed by nodeos anymore.")
+         ("fix-irreversible-blocks",
```
spoonincode, line 248: This doesn't seem like something that should be a nodeos option, this looks like a leap-util thing

libraries/chain/block_log.cpp
```diff
@@ -1,28 +1,18 @@
...
-#define LOG_WRITE_C "ab+"
-#define LOG_RW_C "rb+"
-
-#ifndef _WIN32
-#define FC_FOPEN(p, m) fopen(p, m)
-#else
-#define FC_CAT(s1, s2) s1 ## s2
-#define FC_PREL(s) FC_CAT(L, s)
-#define FC_FOPEN(p, m) _wfopen(p, FC_PREL(m))
+#if !(defined(__BYTE_ORDER__) && __BYTE_ORDER__ == 1234)
```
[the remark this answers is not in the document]
huangminghuang: It won't compile.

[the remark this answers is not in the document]
spoonincode: `__BYTE_ORDER__` is not defined to be 1234 on little endian. The proper way to check is via something like `#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__` (consider a `static_assert` instead too)

plugins/chain_plugin/chain_plugin.cpp
huangminghuang: leap-util has the option to fix corrupted block log content already. This is here for the operator who doesn't care about why the log is corrupted and just want to fix it and restart the nodeos anyway. Without the flag, the operator may have to write an extra shell script to detect the situation and use leap-util to correct the problem then restart the nodeos. The option provides the functionality without the complication of shell script.

libraries/chain/include/eosio/chain/log_catalog.hpp
huangminghuang: I am not exactly sure what you want me to do here. Increase or the number of cached file descriptor? Or we can wait until the PR is merged and then have some performance evaluation to decide whether it needs to be addressed?

libraries/chain/include/eosio/chain/block_log.hpp
```diff
@@ -54,23 +50,23 @@ namespace eosio { namespace chain {
...
          signed_block_ptr read_block_by_id(const block_id_type& id)const {
             return read_block_by_num(block_header::num_from_id(id));
          }
 
          /**
           * Return offset of block in file, or block_log::npos if it does not exist.
           */
-         uint64_t get_block_pos(uint32_t block_num) const;
-         signed_block_ptr        read_head()const;
+         uint64_t get_block_pos(uint32_t block_num) const; //used by test only
```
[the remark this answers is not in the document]
huangminghuang: done

libraries/chain/include/eosio/chain/log_data_base.hpp
huangminghuang: the code has been converted to regular file IO instead memory mapped file IO.

programs/eosio-blocklog/main.cpp
```diff
@@ -203,6 +205,11 @@ void blocklog::set_program_options(options_description& cli)
           "Quick test that blocks.log and blocks.index are well formed and agree with each other.")
          ("vacuum", bpo::bool_switch(&vacuum)->default_value(false),
           "Vacuum a pruned blocks.log in to an un-pruned blocks.log")
+         ("split-blocklog", bpo::value<uint32_t>(&blocklog_split_stride)->default_value(0),
+          "split the block log file based on the stride and store the result in the specified 'output-dir'.")
+         ("merge-blocklogs", bpo::bool_switch(&merge_blocklogs)->default_value(false),
+          "Merge block log files in 'blocks-dir' with the file pattern 'blocks-\\d+-\\d+.[log,index]' to 'output-dir' whenever possible."
+          "The files in 'blocks-dir' will be kept without change. Must give 'blocks-dir' and 'output-dir'.")
```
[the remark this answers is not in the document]
huangminghuang: done

tests/block_log.cpp
```diff
@@ -575,11 +589,11 @@ BOOST_DATA_TEST_CASE(no_block_log_basic_nongenesis, bdata::xrange(2) * bdata::xr
 
 void no_block_log_public_functions_test( block_log_fixture& t) {
    BOOST_REQUIRE_NO_THROW(t.log->flush());
-   BOOST_REQUIRE(t.log->read_block(1) == nullptr);
-   BOOST_REQUIRE_NO_THROW(
-      eosio::chain::block_header bh;
-      t.log->read_block_header(bh, 1);
-   );
+   // BOOST_REQUIRE(t.log->read_block(1) == nullptr);
```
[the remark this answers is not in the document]
huangminghuang: done

linh2931 (approved):

plugins/chain_plugin/chain_plugin.cpp
spoonincode: I'd like to get another opinion from at least one of @heifner @arhag or @766C6164 on this. 

While this isn't an option that performs an operation and then quits (_clearly_ leap-util), it's still an operation that sounds like some sort of offline recovery of a nodeos data file, which is in the ballpark of items we're been moving out of nodeos to leap-util.

It's also a problematic option from the standpoint of a user trying to decide whether or not to enable it.. why wouldn't I _always_ have this enabled? I searched through the entire diff and didn't find any error message that told the user when they should use the option or not.

It feels like if we detect an inconsistency in the block log, we should exit with an error message telling the user something is wrong and they should perform a recovery via leap-util. At least then they can decide what kind of recovery to perform. As is, it seems like it'll decide the recovery strategy based on whatever the option is -- an option the user probably didn't even know existed -- hope it's what they wanted!

libraries/chain/include/eosio/chain/log_catalog.hpp
spoonincode: It's not necessarily an actionable item, just something to get everyone thinking about the performance and reliability side effects of this feature.

Can you describe what happens here if we fail to open the file (because, say, we've exhausted all file descriptors)? Clearly it `return {}` but then what happens? Will it get any peers in to a weird state or such?

heifner: Not sure it is part of this effort, but would be nice to have a test that runs nodeos out of file descriptors and verifies it exits cleanly. Such a test would not necessarily run out of file descriptors right here which is why it likely should be a separate task.

plugins/chain_plugin/chain_plugin.cpp
heifner: > It feels like if we detect an inconsistency in the block log, we should exit with an error message telling the user something is wrong and they should perform a recovery via leap-util. At least then they can decide what kind of recovery to perform. As is, it seems like it'll decide the recovery strategy based on whatever the option is -- an option the user probably didn't even know existed -- hope it's what they wanted!

I agree. I think we should exit with an error message and let the user decide recovery strategy via `leap-util` or just restore from their own backup. In many cases the user is likely to restore from there own backup as a faster way to get running again instead of `nodeos` doing some long running recovery step.

libraries/chain/include/eosio/chain/log_catalog.hpp
greg7mdp: Also, maybe we should try to avoid this problem, by checking how many file descriptors we are allowed at startup, and not accepting connections if we get close to that number being in use.

libraries/chain/block_log.cpp
```diff
@@ -31,1490 +21,1619 @@ namespace eosio { namespace chain {
...
+         log_bundle.log_data.light_validate_block_entry_at(log_bundle.log_index.nth_block_position(pos),
+                                                           expected_block_num);
       }
+   }
 
-      return true;
+   std::pair<fc::path, fc::path> blocklog_files(const fc::path& dir, uint32_t start_block_num, uint32_t num_blocks) {
+      const int bufsize = 64;
+      char      buf[bufsize];
+      snprintf(buf, bufsize, "blocks-%d-%d.log", start_block_num, start_block_num + num_blocks - 1);
```
spoonincode, line 1514: These should be `%u` -- the arguments are unsigned ints.

libraries/chain/include/eosio/chain/log_catalog.hpp
huangminghuang: When it returns `{}`, it means the current nodeos instance does not have the requested block, hence the net plugin should be able to response appropriately. The current main brnach can handle this behavior when a nodeos does not have the full block log. This PR does not alter it at all.

plugins/chain_plugin/chain_plugin.cpp
huangminghuang: option removed

libraries/chain/block_log.cpp
```diff
@@ -31,1490 +21,1610 @@ namespace eosio { namespace chain {
...
+         EOS_ASSERT(version() > 0, block_log_exception, "Block log was not setup properly");
+         EOS_ASSERT(block_log::is_supported_version(version()), block_log_unsupported_version,
+                    "Unsupported version of block log. Block log version is ${version} while code supports version(s) "
+                    "[${min},${max}], log file: ${log}",
+                    ("version", version())("min", block_log::min_supported_version)(
+                          "max", block_log::max_supported_version)("log", log_path.generic_string()));
+
+         first_block_num = 1;
+         if (version() != initial_version) {
+            ds.read((char*)&first_block_num, sizeof(first_block_num));
```
ScottBailey, line 83: Is C++ casting preferable to C?
`ds.read( reinterpret_cast<char*>(&first_block_num), sizeof(first_block_num));`

huangminghuang: we have byte order checking in line 11

libraries/chain/block_log.cpp
```diff
@@ -31,1490 +21,1610 @@ namespace eosio { namespace chain {
...
+               std::tie(block_num, block_id) = full_validate_block_entry(block_num, block_id, entry);
+               if (block_num % 1000 == 0)
+                  ilog("Verified block ${num}", ("num", block_num));
+               pos = file.tellp();
+            }
+         } catch (const bad_block_exception& e) {
+            detail::basic_block_log::write_incomplete_block_data(blocks_dir, now, block_num, ro_stream_at(pos));
+            std::rethrow_exception(e.inner);
+         }
+      } catch (const fc::exception& e) { error_msg = e.what(); } catch (const std::exception& e) {
```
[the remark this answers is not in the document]
greg7mdp: Hum, I'm pretty sure this is wrong advice and that `fc::exception` should be caught, as its `what()` implementation is different from the `std::exception` one.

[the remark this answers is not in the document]
ScottBailey: > Hum, I'm pretty sure this is wrong advice and that `fc::exception` should be caught, as its `what()` implementation is different from the `std::exception` one.

Of course they are different. That's how virtual functions work, right? `fc::exception` is derived from `std::exception` and `what()` is a virtual function marked `override` in the derived class. Have we done something special bad here to make this not work?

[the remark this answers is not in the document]
greg7mdp: My bad, you are right @ScottBailey , I missed the fact that it was a virtual function.

libraries/chain/block_log.cpp
ScottBailey: This is corrected now.

ScottBailey (approved):
lgtm

PR #537, opened 2022-12-05
Instrument producer_plugin for Prometheus

ndcgundlach:
Add a base set of metrics for to publish to Prometheus to the producer_plugin.

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -154,6 +154,7 @@ public:
    fc::microseconds                  abi_serializer_max_time_us;
    std::optional<bfs::path>          snapshot_path;
 
+   std::shared_ptr<chain_plugin_metrics> chain_metrics;
 
    // retained references to channels for easy publication
    channels::pre_accepted_block::channel_type&     pre_accepted_block_channel;
@@ -758,6 +759,7 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
       my->chain_config->blocks_dir = my->blocks_dir;
       my->chain_config->state_dir = app().data_dir() / config::default_state_dir_name;
       my->chain_config->read_only = my->readonly;
+      my->chain_metrics = std::make_shared<chain_plugin_metrics>();
 
       if (auto resmon_plugin = app().find_plugin<resource_monitor_plugin>()) {
         resmon_plugin->monitor_directory(my->chain_config->blocks_dir);
@@ -2796,6 +2798,10 @@ fc::variant chain_plugin::get_log_trx(const transaction& trx) const {
         pretty_output = trx;
     }
     return pretty_output;
+}
+
+std::shared_ptr<chain_plugin_metrics> chain_plugin::metrics() {
+   return my->chain_metrics;
 }
 } // namespace eosio
 
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -154,7 +154,7 @@ public:
    fc::microseconds                  abi_serializer_max_time_us;
    std::optional<bfs::path>          snapshot_path;
 
-   std::shared_ptr<chain_plugin_metrics> chain_metrics;
+   chain_plugin_metrics              chain_metrics;
 
    // retained references to channels for easy publication
    channels::pre_accepted_block::channel_type&     pre_accepted_block_channel;
@@ -759,7 +759,6 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
       my->chain_config->blocks_dir = my->blocks_dir;
       my->chain_config->state_dir = app().data_dir() / config::default_state_dir_name;
       my->chain_config->read_only = my->readonly;
-      my->chain_metrics = std::make_shared<chain_plugin_metrics>();
 
       if (auto resmon_plugin = app().find_plugin<resource_monitor_plugin>()) {
         resmon_plugin->monitor_directory(my->chain_config->blocks_dir);
@@ -2800,7 +2799,7 @@ fc::variant chain_plugin::get_log_trx(const transaction& trx) const {
     return pretty_output;
 }
 
-std::shared_ptr<chain_plugin_metrics> chain_plugin::metrics() {
+const chain_plugin_metrics& chain_plugin::metrics() const {
    return my->chain_metrics;
 }
 } // namespace eosio
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -882,7 +882,7 @@ public:
    // return variant of trx for logging, trace is modified to minimize log output
    fc::variant get_log_trx(const transaction& trx) const;
 
-   std::shared_ptr<chain_plugin_metrics> metrics();
+   const chain_plugin_metrics& metrics() const;
 
 private:
    static void log_guard_exception(const chain::guard_exception& e);
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -832,6 +832,10 @@ public:
 
 } // namespace chain_apis
 
+struct chain_plugin_metrics {
+   //runtime_metric
+};
+
 class chain_plugin : public plugin<chain_plugin> {
 public:
    APPBASE_PLUGIN_REQUIRES()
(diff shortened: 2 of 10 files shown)


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -154,6 +154,7 @@ class chain_plugin_impl {
    fc::microseconds                  abi_serializer_max_time_us;
    std::optional<bfs::path>          snapshot_path;
 
+   std::shared_ptr<chain_plugin_metrics> chain_metrics;
```
spoonincode, line 157: What necessitates this being a `shared_ptr` over a `unique_ptr`, or even just returning a reference to a `const chain_plugin_metrics&`?

heifner:
See https://github.com/eosnetworkfoundation/mandel/issues/67

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1817,6 +1821,8 @@ bool producer_plugin_impl::remove_expired_trxs( const fc::time_point& deadline )
             ++num_expired;
    });
 
+   _metrics->unapplied_transactions.value = orig_count - num_expired;
```
heifner, line 1824: I think it is a good plan to only update this at the start of each block instead of every time the queue is modified. I would move all these metric updates into a method and then call that method from `start_block()`.

Also, this could just be `_metrics->unapplied_transactions.value = _unapplied_transactions.size();`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2059,6 +2067,9 @@ producer_plugin_impl::push_transaction( const fc::time_point& block_deadline,
       if( next ) next( trace );
    }
 
+   _metrics->subjective_bill_account_size.value = _subjective_billing.get_account_cache_size();
+   _metrics->subjective_bill_block_size.value = _subjective_billing.get_block_cache_size();
```
heifner, line 2071: Updating this every trx seems a bit too often, I would do it once per block.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -87,11 +119,28 @@ namespace eosio {
...
+         void add_plugin_metrics(std::shared_ptr<producer_plugin_metrics> metrics) {
+            add_plugin_metric(metrics->unapplied_transactions);
+            add_plugin_metric(metrics->blacklisted_transactions);
+            add_plugin_metric(metrics->blocks_produced);
+            add_plugin_metric(metrics->trxs_produced);
+            add_plugin_metric(metrics->last_irreversible);
+            add_plugin_metric(metrics->block_num);
+            add_plugin_metric(metrics->subjective_bill_account_size);
+            add_plugin_metric(metrics->subjective_bill_block_size);
+            add_plugin_metric(metrics->scheduled_trxs);
```
heifner, line 131: Right now you have to modify both `prometheus_plugin` and `producer_plugin` every time there is a new metric. Instead seems like you could do something like:

```
struct producer_plugin_metrics {
   runtime_metric unapplied_transactions{metric_type::gauge, "unapplied_transactions", "unapplied_transactions", 0};
   runtime_metric blacklisted_transactions{metric_type::gauge, "blacklisted_transactions", "blacklisted_transactions", 0};


   std::vector<std::reference_wrapper<runtime_metric>> metrics;

   producer_plugin_metrics() { 
      metrics.emplace_back(std::ref(unapplied_transactions)); 
      // etc.
   }
};
```

Then you could just loop in the `prometheus_plugin`.

plugins/prometheus_plugin/include/eosio/prometheus_plugin/prometheus_plugin.hpp
```diff
@@ -14,7 +15,7 @@ namespace eosio {
       prometheus_plugin();
       ~prometheus_plugin();
 
-      APPBASE_PLUGIN_REQUIRES((http_plugin) (chain_plugin) (net_plugin))
+      APPBASE_PLUGIN_REQUIRES((http_plugin) (chain_plugin) (net_plugin) (producer_plugin))
```
heifner, line 18: `chain_plugin`, `net_plugin`, `producer_plugin` are all required plugins so doubt it matters, but in general I don't think you need to require any plugin except `http_plugin`.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -61,24 +61,56 @@ namespace eosio {
...
 
+         void add_gauge_metric(runtime_metric& plugin_metric) {
+               auto &gauge_family = BuildGauge()
+                     .Name(plugin_metric.family)
+                     .Help("")
+                     .Register(*_registry);
+            auto &gauge = gauge_family.Add({});
+
+            _gauges.push_back(
+                  std::tuple<Family<Gauge> &, Gauge &, runtime_metric &>(gauge_family, gauge, plugin_metric));
```
heifner, line 85: style: Remove space before `&`. Also for the `auto &` should be `auto&`.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1817,6 +1835,7 @@ bool producer_plugin_impl::remove_expired_trxs( const fc::time_point& deadline )
             ++num_expired;
    });
 
+
```
heifner, line 1838: Can you clean up these unneeded extra newlines that were added.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -410,6 +411,21 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+         _metrics.blacklisted_transactions.value = _blacklisted_transactions.size();
+         _metrics.unapplied_transactions.value = _unapplied_transactions.size();
+
+         auto& chain = chain_plug->chain();
+         _metrics.last_irreversible.value = chain.last_irreversible_block_num();
+         _metrics.block_num.value = chain.head_block_num();
+
+         const auto& sch_idx = chain.db().get_index<generated_transaction_multi_index,by_delay>();
+         _metrics.scheduled_trxs.value = sch_idx.size();
+      }
```
heifner, line 427: In cases were we have a group like this probably worth while to have an enabled flag that is true only if the the prometheus plugin is enabled.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -61,37 +61,72 @@ namespace eosio {
...
-         void add_plugin_metric(runtime_metric& plugin_metric) {
-            auto& gauge_family = BuildGauge()
-                  .Name(plugin_metric.family)
-                  .Help("")
-                  .Register(*_registry);
+         void add_gauge_metric(const runtime_metric& plugin_metric) {
+               auto &gauge_family = BuildGauge()
+                     .Name(plugin_metric.family)
+                     .Help("")
+                     .Register(*_registry);
```
heifner, line 78: This is an odd re-format of the code. The before looks more like what we would want.

plugins/net_plugin/include/eosio/net_plugin/net_plugin.hpp
```diff
@@ -16,9 +16,22 @@ namespace eosio {
...
-      runtime_metric num_clients{"net_plugin", "num_clients", 0};
-      runtime_metric dropped_trxs{"net_plugin", "dropped_trxs", 0};
+      runtime_metric num_peers{metric_type::gauge, "num_peers", "num_peers", 0};
+      runtime_metric num_clients{metric_type::gauge, "num_clients", "num_clients", 0};
+      runtime_metric dropped_trxs{metric_type::counter, "dropped_trxs", "dropped_trxs", 0};
+
+      std::vector<std::reference_wrapper<runtime_metric>> metrics;
+
+      void enable(bool enabled)  {_enabled = enabled; }
+      bool enabled() {return _enabled;}
```
heifner, line 26: These methods and `metrics` could be moved to a base class so that any *_metrics struct could derive and get it without duplicating it for all *_metrics structs.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -61,37 +61,73 @@ namespace eosio {
...
          void update_plugin_metrics() {
-            for (auto& rtm : _runtime_metrics) {
+            for (auto& rtm : _gauges) {
                auto new_val = static_cast<double>(std::get<2>(rtm).value);
                std::get<1>(rtm).Set(new_val);
             }
+
+            for (auto& rtm : _counters) {
+               auto new_val = static_cast<double>(std::get<2>(rtm).value);
+               std::get<1>(rtm).Increment(new_val-std::get<1>(rtm).Value());
```
heifner, line 129: Did you decide this was not needed?

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -832,6 +832,10 @@ class read_write {
 
 } // namespace chain_apis
 
+struct chain_plugin_metrics {
+   //runtime_metric
```
heifner, line 836: I assume this will be cleaned up with the chain_plugin PR to the feature branch.

heifner (approved):

PR #538, opened 2022-12-05
Launch nodeos instances per configuration, unsharing the network first.

jgiszczak:
Also adds all of the unimplemented features required to use the Python launcher in place of the C++ launcher.

### Documentation for developers
To debug a cluster which has been left running, use:
```bash
sudo nsenter -n -t <nodeos-process-id>
```
Once in a shell running in the network namespace, cleos commands to access the cluster will work as usual.
An easy way to find the appropriate process id is:
```bash
lsns -t net
```


spoonincode:
What tests currently use this code path? I would not have expected it to work at all without migrating the container to be `--privileged`, nor for many other tests without (effectively) `ip link set lo0 up`, neither of which I see here on a cursory look.

Also, once getting to main, an appropriate fallback needs to be in place for when the ability to unshare fails (either due to lack of it, or due to permissions).

tests/launcher.py
```diff
@@ -479,6 +488,58 @@ def make_mesh(self):
...
+        eosdcmd.append(str(instance.config_dir_name))
+        eosdcmd.append('--data-dir')
+        eosdcmd.append(str(instance.data_dir_name))
+        eosdcmd.append('--genesis-json')
+        eosdcmd.append(f'{instance.config_dir_name}/genesis.json')
+        if self.args.timestamp:
+            eosdcmd.append('--genesis-timestamp')
+            eosdcmd.append(self.args.timestamp)
+
+        if 'eosio::history_api_plugin' in eosdcmd and 'eosio::trace_api_plugin' in eosdcmd:
```
ClaytonCalabrese, line 517: Is this for backward compatibility, or why are we referencing the history plugin?

jgiszczak: Yes, that's backwards compatibility code preserved from `main`.

jgiszczak:
@spoonincode No tests currently use this method of launching a cluster.  This is a step along the path toward that end.  I don't expect that `--privileged` will be necessary, as it is remapping the unprivileged user by default, so it runs without root.  It does still need to enable local loopback.

ClaytonCalabrese (approved):

spoonincode:
All tests are run in a docker container. Docker does not allow userns by default.
```
$ docker run --rm -it ubuntu:focal
root@88014cbb1b2b:/# unshare -n -r
unshare: unshare failed: Operation not permitted
```

tests/launcher.py
```diff
@@ -538,6 +540,9 @@ def launch(self, instance: nodeDefinition):
 
     def start_all(self):
         unshare(CLONE_NEWNET)
+        for index, name in socket.if_nameindex():
+            if getInterfaceFlags(name) & IFF_LOOPBACK:
+                os.system(f'ip link set {name} up')
```
spoonincode, line 545: This adds a new dependency to run the tests. Need to update the documentation or (preferably) just do the ioctl here to up the interface.

jgiszczak: The `iproute` or `iproute2` package containing `ip` has been a dependency of `ubuntu-minimal` for over a decade and Ubuntu doesn't run without it.  It's not a dependency that needs to be explicit in application packages.

spoonincode: If the code in this PR was actually used you'd see it won't even work in CI right now since `iproute2` is _not_ included in ubuntu's official docker image, nor is it indirectly included after installing 300+ dep packages we use for building.

ex, ubuntu22 image used on main:
```
$ docker run --rm -it ghcr.io/antelopeio/builders:75cd0fb7364b148e5d0a663ca8f4aed434571489a69dd228628732f8a8ef2a49
Unable to find image 'ghcr.io/antelopeio/builders:75cd0fb7364b148e5d0a663ca8f4aed434571489a69dd228628732f8a8ef2a49' locally
75cd0fb7364b148e5d0a663ca8f4aed434571489a69dd228628732f8a8ef2a49: Pulling from antelopeio/builders
e96e057aae67: Pull complete
faf78f8066b7: Pull complete
Digest: sha256:fa291238c1cb85d37c6f8c374de55e05a1766332f0af182204cb1b7007b05809
Status: Downloaded newer image for ghcr.io/antelopeio/builders:75cd0fb7364b148e5d0a663ca8f4aed434571489a69dd228628732f8a8ef2a49
root@a2f548020b32:/# ip
bash: ip: command not found
```

But really you should just do the ioctl yourself instead of adding a new dependency since it's straightforward to do the ioctl.

jgiszczak: And if you try `/sbin/ip`?

jgiszczak: Unbelievable.  It doesn't.

Whatever.  I already pushed a change to directly calling ioctl.

spoonincode:
```
$ docker run --rm -it ghcr.io/antelopeio/builders:75cd0fb7364b148e5d0a663ca8f4aed434571489a69dd228628732f8a8ef2a49
root@ca9038f4d6ea:/# /sbin/ip
bash: /sbin/ip: No such file or directory
```

.github/workflows/build.yaml
```diff
@@ -121,6 +121,7 @@ jobs:
         platform: [ubuntu20, ubuntu22]
     runs-on: ["self-hosted", "enf-x86-hightier"]
     container: ${{fromJSON(needs.d.outputs.p)[matrix.platform].image}}
+      options: --privileged
```
spoonincode, line 124: The documentation and example on how to do this is at https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idcontainer

It's something like
```yaml
container:
  image: ${{fromJSON(needs.d.outputs.p)[matrix.platform].image}}
  options: --privileged
```

.github/workflows/build.yaml
```diff
@@ -157,6 +158,7 @@ jobs:
         uses: ./.github/actions/parallel-ctest-containers
         with:
           container: ${{fromJSON(needs.d.outputs.p)[matrix.platform].image}}
+          options: --privileged
```
spoonincode, line 161: This is a noop. `parallel-ctest-containers` is a custom action that runs all these non-parallel tests in parallel by launching a bunch of docker containers. I'm not sure what the status of the current support for running the tests in parallel now is.

**If tests can be run in parallel:**
You should remove usage of `parallel-ctest-containers` and fold together Tests & NP Tests. You will need to do some other upkeep to ensure logs continue to be uploaded properly on failed tests, etc.

**If tests cannot be run in parallel:**
You need to add `--privileged` to the `parallel-ctest-containers` impl where it launches docker containers
https://github.com/AntelopeIO/leap/blob/c36f68d00bc397b55a8d9a2479b5290d513d19f0/.github/actions/parallel-ctest-containers/main.mjs#L31

.github/workflows/build.yaml
jgiszczak: It seems to be accepting this format, then not honoring it.

.github/actions/parallel-ctest-containers/main.mjs
```diff
@@ -12,7 +12,7 @@ const log_tarball_prefix = core.getInput('log-tarball-prefix', {required: true})
 const tests_label = core.getInput('tests-label', {required: true});
 
 try {
-   if(child_process.spawnSync("docker", ["run", "--name", "base", "-v", `${process.cwd()}/build.tar.zst:/build.tar.zst`, "--workdir", "/__w/leap/leap", container, "sh", "-c", "zstdcat /build.tar.zst | tar x"], {stdio:"inherit"}).status)
+   if(child_process.spawnSync("docker", ["run", "--privileged", "--name", "base", "-v", `${process.cwd()}/build.tar.zst:/build.tar.zst`, "--workdir", "/__w/leap/leap", container, "sh", "-c", "zstdcat /build.tar.zst | tar x"], {stdio:"inherit"}).status)
```
spoonincode, line 15: you'll need to regenerate the [dist/index.mjs](https://github.com/AntelopeIO/leap/blob/main/.github/actions/parallel-ctest-containers/dist/index.mjs) it seems I failed to document this well anywhere; probably because I didn't think we'd still be using it still.

```
cd .github/actions/parallel-ctest-containers
npm ci
# stay in parallel-ctest-containers directory
./node_modules/.bin/ncc build main.mjs --minify
```

I _hope_ this ends up with an dist/index.mjs that has just this minimal change in it for you -- I really hope it regenerates this minified version reproducibly/deterministically for you. If it's a huge change maybe don't commit it: I kinda hate to continue to bloat the repo with this large file  which ultimately we'll be removing. Worst case we can manually modify the dist/index.mjs with this ~dozen character change if it comes to it.

jgiszczak: Due to the minification, the diff is very large.  Something more than half the file.  Should I commit it or try to unravel the minification?

spoonincode: oof, that's a major bummer. I'm no JS dep expert, but I was hopeful that by having the package lock committed others would get the same versions of everything ergo the recreation of the amalgamation (minified or not) would be reproducible.

You can either
* just commit the big blob diff (kinda yucky), or
* just manually modify the one line in the big blob diff to add that `"--privileged", ` change. (not really so bad to just fine that one line in the file and change it, just a wee bit of work

oschwaldp-oci (approved):

oschwaldp-oci (approved):

ClaytonCalabrese (changes requested):

tests/TestHarness/Cluster.py
```diff
@@ -439,6 +473,7 @@ def connectGroup(group, producerNodes, bridgeNodes) :
             self.unstartedNodes=self.discoverUnstartedLocalNodes(unstartedNodes, totalNodes)
 
         biosNode=self.discoverBiosNode(timeout=Utils.systemWaitTimeout)
+
```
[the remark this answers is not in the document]
jgiszczak: Looks like an unintentional leak.  I'll remove it.

tests/TestHarness/Cluster.py
```diff
@@ -262,17 +259,15 @@ def launch(self, pnodes=1, unstartedNodes=0, totalNodes=1, prodCount=1, topo="me
...
+        loggingLevelDictString = json.dumps(self.loggingLevelDict, separators=(',', ':'))
+        cmd="%s -p %s -n %s -d %s -i %s -f %s --unstarted-nodes %s --logging-level %s --logging-level-map %s" % (
+            "python3 tests/launcher.py", pnodes, totalNodes, delay, datetime.datetime.utcnow().strftime("%Y-%m-%dT%H:%M:%S.%f")[:-3],
+            producerFlag, unstartedNodes, self.loggingLevel, loggingLevelDictString)
         cmdArr=cmd.split()
         if self.staging:
             cmdArr.append("--nogen")
 
-        nodeosArgs="--max-transaction-time -1 --abi-serializer-max-time-ms 990000 --p2p-max-nodes-per-host %d --max-clients %d" % (maximumP2pPerHost, maximumClients)
+        nodeosArgs="--resource-monitor-not-shutdown-on-threshold-exceeded --max-transaction-time -1 --abi-serializer-max-time-ms 990000 --p2p-max-nodes-per-host %d --max-clients %d" % (maximumP2pPerHost, maximumClients)
```
ClaytonCalabrese, line 270: @heifner  If I remember right I added this at some point and we removed it. Is that correct, or are we okay with it now?

tests/TestHarness/Cluster.py
```diff
@@ -486,6 +481,7 @@ def connectGroup(group, producerNodes, bridgeNodes) :
             return False
 
         startedNodes=totalNodes-unstartedNodes
+        self.nodes=list(range(startedNodes)) # placeholder for cleanup purposes only
```
ClaytonCalabrese, line 484: Is this a TODO for this PR or for later down the line?

ClaytonCalabrese (changes requested):

tests/TestHarness/Cluster.py
jgiszczak: That's a merge error.  The line was removed and the merge put it back.  It's nonsense that doesn't work.

tests/launcher.py
```diff
@@ -40,7 +60,7 @@ class nodeDefinition:
     base_p2p_port: ClassVar[int] = 9876
     base_http_port: ClassVar[int] = 8888
     #file_size: int
-    instance_name: str = field(init=False, repr=False)
+    #instance_name: str = field(init=False, repr=False)
```
oschwaldp-oci, line 63: Is there a future plan to reintroduce this, or should it just be deleted?

tests/launcher.py
```diff
@@ -478,7 +539,139 @@ def make_mesh(self):
...
+                        self.terminate_wait_pid(pid, raise_if_missing=False)
+                    self.launch(node)
+
+    def down(self, nodeNumbers):
+        for num in nodeNumbers:
+            for node in self.network.nodes.values():
+                if self.network.name + num == node.name:
+                    with open(node.data_dir_name / f'{Utils.EosServerName}.pid', 'r') as f:
+                        pid = int(f.readline())
+                        self.terminate_wait_pid(pid)
```
oschwaldp-oci, line 641: Would it be useful to have some debug logging in `bounce` and `down`?

tests/launcher.py
```diff
@@ -478,7 +539,139 @@ def make_mesh(self):
...
+                        f.write(f'biosport={biosport}\n')
+                    elif 'prodkeys' in line:
+                        for node in self.network.nodes.values():
+                            f.write(f'wcmd import -n ignition --private-key {node.keys[0].privkey}\n')
+                    elif 'cacmd' in line:
+                        for node in self.network.nodes.values():
+                            if node.producers and node.producers[0] == 'eosio':
+                                continue
+                            for producer in node.producers:
+                                f.write(f'cacmd {producer} {node.keys[0].pubkey} {node.keys[0].pubkey}\n')
```
oschwaldp-oci, line 672: are both of these supposed to be `pubkey` or is one supposed to be `privkey`?

oschwaldp-oci (changes requested):

tests/TestHarness/Cluster.py
ClaytonCalabrese: https://github.com/AntelopeIO/leap/commit/25768770c3f56484ee3fe34c1d13739da95408da

ClaytonCalabrese: Following a discussion, this should be put behind a version check but is otherwise fine.

tests/launcher.py
jgiszczak: I imitated what the C++ code was doing.  Turns out the second key was never used in the script, so I eliminated it.

tests/launcher.py
jgiszczak: Added.

tests/launcher.py
jgiszczak: Deleted this and numerous other bits of commented code and cruft.

oschwaldp-oci (changes requested):

tests/performance_tests/performance_test_basic.py
```diff
@@ -458,7 +458,7 @@ def main():
     extraNodeosHttpPluginArgs = PerformanceBasicTest.ClusterConfig.ExtraNodeosArgs.ExtraNodeosHttpPluginArgs(httpMaxResponseTimeMs=args.http_max_response_time_ms)
     extraNodeosArgs = PerformanceBasicTest.ClusterConfig.ExtraNodeosArgs(chainPluginArgs=extraNodeosChainPluginArgs, httpPluginArgs=extraNodeosHttpPluginArgs, producerPluginArgs=extraNodeosProducerPluginArgs)
 
-    testClusterConfig = PerformanceBasicTest.ClusterConfig(pnodes=args.p, totalNodes=args.n, topo=args.s, genesisPath=args.genesis, prodsEnableTraceApi=args.prods_enable_trace_api, extraNodeosArgs=extraNodeosArgs, nodeosVers=Utils.getNodeosVersion().split('.')[0])
+    testClusterConfig = PerformanceBasicTest.ClusterConfig(pnodes=args.p, totalNodes=args.n, topo=args.s, genesisPath=args.genesis, prodsEnableTraceApi=args.prods_enable_trace_api, extraNodeosArgs=extraNodeosArgs)
```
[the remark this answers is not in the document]
jgiszczak: Restored in [837a5761](https://github.com/AntelopeIO/leap/pull/538/commits/837a5761f15d8af38cfb8bd85618a6406f05a13a), though with modifications.

tests/TestHarness/Node.py
```diff
@@ -47,21 +43,38 @@ def __init__(self, host, port, nodeId, pid=None, cmd=None, walletMgr=None, nodeo
...
+        while info == None:
+            time.sleep(0.5)
+            info = self.getInfo(silentErrors=True)
+            if datetime.now() > timeout:
+                break
+        else:
+            self.configureVersion(info)
+
+    def configureVersion(self, info):
+        self.nodeosVers=info["server_version_string"]
```
oschwaldp-oci, line 60: We should probably have some exceptional or graceful handling of the timeout scenario so that either we have a way to handle in `info` is still `None` and just default to the `else` in `configureVersion` or else raise an exception from `pollVersion` instead of waiting to try to access `info` in `configureVersion`.

tests/TestHarness/Cluster.py
jgiszczak: Removed again, due to the overhead of creating a test.

tests/TestHarness/Node.py
jgiszczak: Added in [27d1430a](https://github.com/AntelopeIO/leap/pull/538/commits/27d1430af550ff6c3d66672baf7d4917368a30e9).

tests/CMakeLists.txt
```diff
@@ -226,7 +222,6 @@ set_tests_properties(http_plugin_test PROPERTIES TIMEOUT 100)
 set_property(TEST http_plugin_test PROPERTY LABELS nonparallelizable_tests)
 
 add_test(NAME plugin_http_api_test COMMAND tests/plugin_http_api_test.py WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
-set_tests_properties(plugin_http_api_test PROPERTIES TIMEOUT 40)
```
oschwaldp-oci, line 229: I'm not sure the `TIMEOUT` should be completely removed.  I'd be okay with increasing the timeout a bit if it is requiring more time after some changes, but prefer not giving the test carte blanche to run forever.

oschwaldp-oci (changes requested):

oschwaldp-oci (approved):

oschwaldp-oci (approved):

ClaytonCalabrese (approved):

PR #539, opened 2022-12-05
Update management of `in_flight` bytes in http_plugin

### LINKED ISSUE #488: http_plugin track bytes in flight decrements before http processing complete
Currently a RAII object `in_flight` is created to track bytes in flight in the http_plugin. This object is destroyed and it decrements the bytes in flight when the `send_response` is called. However, `send_response` does an `async_write`, so the decrement is likely before the data is actually sent and the http session is complete. 
The corresponding `requests_in_flight` is tracked in the construction/destruction of the `beast_http_session`.
Seems like we could move the tracking of the bytes to the lifetime of the `beast_http_session` OR modify `send_response` so it can be given a lambda to call when the actual `async_write` is complete and destroy the `in_flight` at that time.

greg7mdp:
This PR resolves #488.

Management of maximum `in_flight` bytes is updated:
- we remove the overly complicated `in_flight` struct which was not necessary. As a consequence we also don't allocate memory for it via make_shared.
- we make the `verify_max_bytes_in_flight` and `verify_max_requests_in_flight` pure verification functions, returning an error string in case of excess, and add a `send_busy_response` function to send the error response.
- when `max_bytes_in_flight` is exceeded, it means that we are processing incoming requests and enqueuing them via `http::async_write` faster than they can be written. As a result we want to stop enqueuing new writes using `send_response`. 

I have checked that the two http unit tests pass (see below), and also some basic calls using curl:

```
ctest -R http                                                                                       26s 
Test project /home/greg/github/enf/leap/build_debug
    Start 2043: http_plugin_test
1/2 Test #2043: http_plugin_test .................   Passed    5.28 sec
    Start 2044: plugin_http_api_test
2/2 Test #2044: plugin_http_api_test .............   Passed   21.45 sec

100% tests passed, 0 tests failed out of 2
```

```
~/github/enf/leap/build_debug greg7mdp/GH-488_bytes_in_flight !3 ?5 ❯ curl http://127.0.0.1:8888/v1/chain/                                                                    
{"code":404,"message":"Not Found","error":{"code":0,"name":"exception","what":"unspecified","details":[{"message":"Unknown Endpoint","file":"beast_http_session.hpp","line_number":160,"method":"handle_request"}]}}%  

~/github/enf/leap/build_debug greg7mdp/GH-488_bytes_in_flight !3 ?5 ❯ curl http://127.0.0.1:8888/v1/chain/get_info                                                            
{"server_version":"013bbea3","chain_id":"8a34ec7df1b8cd06ff4a8abbaa7cc50300823350cadc59ab296cb00d104d2b8f","head_block_num":38,"last_irreversible_block_num":37,"last_irreversible_block_id":"00000025b42cb1d0b3c146b4534779f269b75cd00ab1f624b16296d7f7ef1cd9","head_block_id":"0000002695635bab48a6d6c778e3d5892150b32b231b814a8c187d5dbeef22eb","head_block_time":"2022-12-05T18:49:20.500","head_block_producer":"eosio","virtual_block_cpu_limit":207526,"virtual_block_net_limit":1088100,"block_cpu_limit":199900,"block_net_limit":1048576,"server_version_string":"v4.0.0-dev","fork_db_head_block_num":38,"fork_db_head_block_id":"0000002695635bab48a6d6c778e3d5892150b32b231b814a8c187d5dbeef22eb","server_full_version_string":"v4.0.0-dev-013bbea3d918382edda60292f6c5cca8952d7a14-dirty","total_cpu_weight":0,"total_net_weight":0,"earliest_available_block_num":1,"last_irreversible_block_time":"2022-12-05T18:49:20.000"}%
```

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -88,22 +88,22 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
             auto next_ptr = std::make_shared<url_handler>(std::move(next));
             return [my=std::move(my), priority, next_ptr=std::move(next_ptr)]
                        ( detail::abstract_conn_ptr conn, string r, string b, url_response_callback then ) {
-               auto tracked_b = make_in_flight(b, my->plugin_state);
+               auto tracked_b = make_in_flight(std::move(b), my->plugin_state);
                if (!conn->verify_max_bytes_in_flight()) {
                   return;
                }
 
-               url_response_callback wrapped_then = [tracked_b = std::move(tracked_b), then=std::move(then)](int code, const fc::time_point& deadline, std::optional<fc::variant> resp) {
+               url_response_callback wrapped_then = [tracked_b, then=std::move(then)](int code, const fc::time_point& deadline, std::optional<fc::variant> resp) {
                   then(code, deadline, std::move(resp));
                };
 
                // post to the app thread taking shared ownership of next (via std::shared_ptr),
                // sole ownership of the tracked body and the passed in parameters
-               app().post( priority, [next_ptr, conn=std::move(conn), r=std::move(r), b=std::move(b), wrapped_then=std::move(wrapped_then)]() mutable {
+               app().post( priority, [next_ptr, conn=std::move(conn), r=std::move(r), tracked_b, wrapped_then=std::move(wrapped_then)]() mutable {
                   try {
                      if( app().is_quiting() ) return; // http_plugin shutting down, do not call callback
                      // call the `next` url_handler and wrap the response handler
-                     (*next_ptr)( std::move(r), std::move(b), std::move(wrapped_then)) ;
+                     (*next_ptr)( std::move( r ), std::move(tracked_b->obj()), std::move(wrapped_then)) ;
                   } catch( ... ) {
                      conn->handle_exception();
                   }
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -88,22 +88,22 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
             auto next_ptr = std::make_shared<url_handler>(std::move(next));
             return [my=std::move(my), priority, next_ptr=std::move(next_ptr)]
                        ( detail::abstract_conn_ptr conn, string r, string b, url_response_callback then ) {
-               auto tracked_b = make_in_flight<string>(std::move(b), my->plugin_state);
+               auto tracked_b = make_in_flight(b, my->plugin_state);
                if (!conn->verify_max_bytes_in_flight()) {
                   return;
                }
 
-               url_response_callback wrapped_then = [tracked_b, then=std::move(then)](int code, const fc::time_point& deadline, std::optional<fc::variant> resp) {
+               url_response_callback wrapped_then = [tracked_b = std::move(tracked_b), then=std::move(then)](int code, const fc::time_point& deadline, std::optional<fc::variant> resp) {
                   then(code, deadline, std::move(resp));
                };
 
                // post to the app thread taking shared ownership of next (via std::shared_ptr),
                // sole ownership of the tracked body and the passed in parameters
-               app().post( priority, [next_ptr, conn=std::move(conn), r=std::move(r), tracked_b, wrapped_then=std::move(wrapped_then)]() mutable {
+               app().post( priority, [next_ptr, conn=std::move(conn), r=std::move(r), b=std::move(b), wrapped_then=std::move(wrapped_then)]() mutable {
                   try {
                      if( app().is_quiting() ) return; // http_plugin shutting down, do not call callback
                      // call the `next` url_handler and wrap the response handler
-                     (*next_ptr)( std::move( r ), std::move(tracked_b->obj()), std::move(wrapped_then)) ;
+                     (*next_ptr)( std::move(r), std::move(b), std::move(wrapped_then)) ;
                   } catch( ... ) {
                      conn->handle_exception();
                   }
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -88,22 +88,21 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
             auto next_ptr = std::make_shared<url_handler>(std::move(next));
             return [my=std::move(my), priority, next_ptr=std::move(next_ptr)]
                        ( detail::abstract_conn_ptr conn, string r, string b, url_response_callback then ) {
-               auto tracked_b = make_in_flight(std::move(b), my->plugin_state);
-               if (!conn->verify_max_bytes_in_flight()) {
+               if (!conn->verify_max_bytes_in_flight(b.size())) {
                   return;
                }
 
-               url_response_callback wrapped_then = [tracked_b, then=std::move(then)](int code, const fc::time_point& deadline, std::optional<fc::variant> resp) {
+               url_response_callback wrapped_then = [then=std::move(then)](int code, const fc::time_point& deadline, std::optional<fc::variant> resp) {
                   then(code, deadline, std::move(resp));
                };
 
                // post to the app thread taking shared ownership of next (via std::shared_ptr),
                // sole ownership of the tracked body and the passed in parameters
-               app().post( priority, [next_ptr, conn=std::move(conn), r=std::move(r), tracked_b, wrapped_then=std::move(wrapped_then)]() mutable {
+               app().post( priority, [next_ptr, conn=std::move(conn), r=std::move(r), b = std::move(b), wrapped_then=std::move(wrapped_then)]() mutable {
                   try {
                      if( app().is_quiting() ) return; // http_plugin shutting down, do not call callback
                      // call the `next` url_handler and wrap the response handler
-                     (*next_ptr)( std::move( r ), std::move(tracked_b->obj()), std::move(wrapped_then)) ;
+                     (*next_ptr)( std::move( r ), std::move(b), std::move(wrapped_then)) ;
                   } catch( ... ) {
                      conn->handle_exception();
                   }
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -88,7 +88,8 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
             auto next_ptr = std::make_shared<url_handler>(std::move(next));
             return [my=std::move(my), priority, next_ptr=std::move(next_ptr)]
                        ( detail::abstract_conn_ptr conn, string r, string b, url_response_callback then ) {
-               if (!conn->verify_max_bytes_in_flight(b.size())) {
+               if (auto error_str = conn->verify_max_bytes_in_flight(b.size()); !error_str.empty()) {
+                  conn->send_busy_response(std::move(error_str));
                   return;
                }
 
--- a/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
@@ -112,8 +112,8 @@ protected:
       // Request path must be absolute and not contain "..".
       if(req.target().empty() || req.target()[0] != '/' || req.target().find("..") != beast::string_view::npos) {
          error_results results{static_cast<uint16_t>(http::status::bad_request), "Illegal request-target"};
-         auto tracked_json = make_in_flight(fc::json::to_string(results, fc::time_point::maximum()), plugin_state_);
-         send_response(std::move(tracked_json), static_cast<unsigned int>(http::status::bad_request) );
+         send_response(fc::json::to_string(results, fc::time_point::maximum()),
+                       static_cast<unsigned int>(http::status::bad_request) );
          return;
       }
 
@@ -136,12 +136,9 @@ protected:
 
          // Respond to options request
          if(req.method() == http::verb::options) {
-            send_response(make_in_flight(std::string("{}"), plugin_state_), static_cast<unsigned int>(http::status::ok));
+            send_response("{}", static_cast<unsigned int>(http::status::ok));
             return;
          }
-
-         // verfiy bytes in flight/requests in flight
-         if(!verify_max_bytes_in_flight()) return;
 
          std::string resource = std::string(req.target());
          // look for the URL handler to handle this resource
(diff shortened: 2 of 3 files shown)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -88,21 +88,21 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
             auto next_ptr = std::make_shared<url_handler>(std::move(next));
             return [my=std::move(my), priority, next_ptr=std::move(next_ptr)]
                        ( detail::abstract_conn_ptr conn, string r, string b, url_response_callback then ) {
-               auto tracked_b = make_in_flight<string>(std::move(b), my->plugin_state);
+               auto tracked_b = make_in_flight(b, my->plugin_state);
```
heifner, line 91: We should avoid this string copy.
#488 mentioned tracking bytes in flight in `beast_http_session` instead. That would allow for the removal of `make_in_flight` and the `in_flight` struct. The whole `in_flight` struct seems rather convoluted. It might have been needed at some point, but now it seems like overkill.

Can you try moving this into `beast_http_session` and see if it is cleaner. If that doesn't work see if a simplified `in_flight` is possible. Something as simple as 
`fc::make_scoped_exit([&](){atomic_in_flight -= count;});` might be possible.

heifner (changes requested):

plugins/http_plugin/http_plugin.cpp
greg7mdp: Kevin, thanks indeed my change was making an extra copy. I fixed it and reverted this code to its previous state. I'll look into the other changes you suggest.

greg7mdp: Hi Kevin, I don't think we can move the `bytes_in_flight` to the `beast_http_session`, because we want to keep track of the bytes in flight across all the sessions (see [here](https://github.com/AntelopeIO/leap/blob/03235ed3d2349c9b454212e78491883651a91b08/plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp#L113-L116)).
However, I think we can keep the count in `http_plugin_state`, and still remove the `in_flight` struct. Let me look into it.

heifner: We increment `requests_in_flight` in `beast_http_session` constructor and decrement it in the destructor. Could you not decrement `bytes_in_flight` in the `beast_http_session` destructor?

greg7mdp: I don't think it would be a good idea, because the `beast_http_session` can remain as long as the socket is connected, serving multiple requests, and we want to decrement `bytes_in_flight` as soon as we have written the response to the socket.

Give me a little time to finish and test my changes, we can talk more then.

greg7mdp: @heifner I did remove the `make_in_flight` and the `in_flight` struct, and update the count of in_flight bytes directly. I think the code is simpler and easier to follow, and also it better keeps track of all in_flight bytes (including for error reports). It also removes the `shared_ptr` memory allocation and free to hold the payload.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -195,7 +197,8 @@ class beast_http_session : public detail::abstract_conn {
          ei.name = "Busy";
          ei.what = "Too many requests in flight: " + std::to_string( requests_in_flight_num );
          error_results results{static_cast<uint16_t>(http::status::too_many_requests), "Busy", ei};
-         send_response( fc::json::to_string( results, fc::time_point::maximum() ), static_cast<unsigned int>(http::status::too_many_requests) );
+         send_response(fc::json::to_string(results, fc::time_point::maximum()),
+                       static_cast<unsigned int>(http::status::too_many_requests) );
```
heifner, line 201: Please avoid reformat changes outside code actually being modified.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -140,9 +140,6 @@ class beast_http_session : public detail::abstract_conn {
             return;
          }
 
-         // verfiy bytes in flight/requests in flight
-         if(!verify_max_bytes_in_flight()) return;
```
heifner, line 144: I think I prefer this to be left as is without the addition of a passed in size. The idea here is to stop new requests after the threshold has been exceeded. Once you have done the work to create the large string you might as well send it.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -159,31 +156,36 @@ class beast_http_session : public detail::abstract_conn {
...
       if(bytes_in_flight_size > plugin_state_->max_bytes_in_flight) {
          fc_dlog(plugin_state_->logger, "429 - too many bytes in flight: ${bytes}", ("bytes", bytes_in_flight_size));
          error_results::error_info ei;
          ei.code = static_cast<int64_t>(http::status::too_many_requests);
          ei.name = "Busy";
          ei.what = "Too many bytes in flight: " + std::to_string( bytes_in_flight_size );
          error_results results{static_cast<uint16_t>(http::status::too_many_requests), "Busy", ei};
-         send_response( fc::json::to_string( results, fc::time_point::maximum() ), static_cast<unsigned int>(http::status::too_many_requests) );
+         send_response(fc::json::to_string(results, fc::time_point::maximum()),
+                       static_cast<unsigned int>(http::status::too_many_requests) );
```
heifner, line 182: Please avoid reformat changes outside code actually being modified.

plugins/http_plugin/include/eosio/http_plugin/common.hpp
```diff
@@ -215,8 +147,9 @@ auto make_in_flight(T&& object, std::shared_ptr<http_plugin_state> plugin_state)
 auto make_http_response_handler(std::shared_ptr<http_plugin_state> plugin_state, detail::abstract_conn_ptr session_ptr) {
    return [plugin_state{std::move(plugin_state)},
            session_ptr{std::move(session_ptr)}](int code, fc::time_point deadline, std::optional<fc::variant> response) {
-      auto tracked_response = make_in_flight(std::move(response), plugin_state);
-      if(!session_ptr->verify_max_bytes_in_flight()) {
+      //auto tracked_response = make_in_flight(std::move(response), plugin_state);
```
heifner, line 150: Remove commented out code.

plugins/http_plugin/include/eosio/http_plugin/common.hpp
```diff
@@ -225,15 +158,17 @@ auto make_http_response_handler(std::shared_ptr<http_plugin_state> plugin_state,
...
+                        [plugin_state, session_ptr, code, deadline, start, payload_size, response = std::move(response)]() {
                            try {
-                              if(tracked_response->obj().has_value()) {
-                                 std::string json = fc::json::to_string(*tracked_response->obj(), deadline + (fc::time_point::now() - start));
-                                 auto tracked_json = make_in_flight(std::move(json), plugin_state);
-                                 session_ptr->send_response(std::move(tracked_json->obj()), code);
+                              plugin_state->bytes_in_flight -= payload_size;
+                              if (response.has_value()) {
+                                 std::string json = fc::json::to_string(*response, deadline + (fc::time_point::now() - start));
+                                 if (session_ptr->verify_max_bytes_in_flight(json.size()))
```
heifner, line 170: Remove this verify call.

plugins/http_plugin/include/eosio/http_plugin/common.hpp
```diff
@@ -225,15 +158,17 @@ auto make_http_response_handler(std::shared_ptr<http_plugin_state> plugin_state,
...
       boost::asio::post(plugin_state->thread_pool->get_executor(),
-                        [plugin_state, session_ptr, code, deadline, start,
-                         tracked_response = std::move(tracked_response)]() {
+                        [plugin_state, session_ptr, code, deadline, start, payload_size, response = std::move(response)]() {
                            try {
-                              if(tracked_response->obj().has_value()) {
-                                 std::string json = fc::json::to_string(*tracked_response->obj(), deadline + (fc::time_point::now() - start));
-                                 auto tracked_json = make_in_flight(std::move(json), plugin_state);
-                                 session_ptr->send_response(std::move(tracked_json->obj()), code);
+                              plugin_state->bytes_in_flight -= payload_size;
```
heifner, line 167: Not that it matters much, but I think it would look cleaner to move this to the end of this scope.

plugins/http_plugin/http_plugin.cpp
heifner: Nice!

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -370,8 +375,8 @@ class beast_http_session : public detail::abstract_conn {
       http::async_write(
             derived().stream(),
             *res_,
-            [self, close](beast::error_code ec, std::size_t bytes_transferred) {
-               self->on_write(ec, bytes_transferred, close);
+            [self, close, payload_size](beast::error_code ec, std::size_t bytes_transferred) {
+               self->on_write(ec, payload_size, close);
```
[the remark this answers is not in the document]
greg7mdp: done!

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
greg7mdp: OK

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
greg7mdp: The check is done in both branches of the following `if` statement, either before calling `send_response()` or in `make_http_response_handler()`

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
greg7mdp: OK

plugins/http_plugin/include/eosio/http_plugin/common.hpp
greg7mdp: Sure, my mistake!

plugins/http_plugin/include/eosio/http_plugin/common.hpp
greg7mdp: Why? It is needed to take into account the actual bytes in flight after converting to json.

plugins/http_plugin/include/eosio/http_plugin/common.hpp
greg7mdp: I think it has to be there for the following verify() call to be correct.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
heifner: We had a nice discussion and Greg convinced me the his approach here is better.

plugins/http_plugin/include/eosio/http_plugin/common.hpp
heifner: We discussed and going with the it as is.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -112,8 +112,8 @@ class beast_http_session : public detail::abstract_conn {
       // Request path must be absolute and not contain "..".
       if(req.target().empty() || req.target()[0] != '/' || req.target().find("..") != beast::string_view::npos) {
          error_results results{static_cast<uint16_t>(http::status::bad_request), "Illegal request-target"};
-         send_response( fc::json::to_string( results, fc::time_point::maximum() ),
-                        static_cast<unsigned int>(http::status::bad_request) );
+         send_response(fc::json::to_string(results, fc::time_point::maximum()),
+                       static_cast<unsigned int>(http::status::bad_request) );
```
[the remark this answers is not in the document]
greg7mdp: updated!

heifner (approved):

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -159,46 +156,45 @@ class beast_http_session : public detail::abstract_conn {
             error_results results{static_cast<uint16_t>(http::status::not_found), "Not Found",
                                   error_results::error_info( fc::exception( FC_LOG_MESSAGE( error, "Unknown Endpoint" ) ),
                                                              http_plugin::verbose_errors() )};
-            send_response( fc::json::to_string( results, fc::time_point::now() + plugin_state_->max_response_time ),
-                           static_cast<unsigned int>(http::status::not_found) );
+            auto json = fc::json::to_string(results, fc::time_point::now() + plugin_state_->max_response_time);
```
linh2931, line 159: Usually we would not make format changes if not related to the problem being resolved.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -346,19 +342,29 @@ class beast_http_session : public detail::abstract_conn {
          res_->set(http::field::content_type, "application/json");
          res_->keep_alive(false);
          res_->set(http::field::server, BOOST_BEAST_VERSION_STRING);
-
+
```
linh2931, line 345: Remove empty spaces here.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -346,19 +342,29 @@ class beast_http_session : public detail::abstract_conn {
...
       }
    }
 
-   virtual void send_response(std::string json_body, unsigned int code) override {
+   void increment_bytes_in_flight(size_t sz) {
+      plugin_state_->bytes_in_flight += sz;
+   }
+
+   void decrement_bytes_in_flight(size_t sz) {
+      plugin_state_->bytes_in_flight -= sz;
```
linh2931, line 356: I guess `bytes_in_flight` would never be smaller than `sz` before subtract.

linh2931 (commented):
Not sure if it is too much to add unit tests to count in-flight bytes.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
greg7mdp: That is correct, we always decrement after incrementing with the same value.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
greg7mdp: Done, thanks @linh2931

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
greg7mdp: Done, thanks @linh2931

linh2931 (approved):

Issue #378, opened 2022-10-25
(3.2+) consider setting a more specific HTTP Server header


spoonincode:
After the beast migration, the HTTP headers in the response includes `Server: Boost.Beast/330` (or possibly some other number based on boost version being used). This isn't any weirder than pre 3.2, which has `Server: WebSocket++/0.7.0`, but maybe we can consider modifying this to `nodeos/3.2.0` or something.

PR #542, opened 2022-12-06
[3.2] use consistent HTTP server header on both success and failure responses

spoonincode:
#378 sets a more specific HTTP Host header for replies. Well, apparently it only did on successful replies because the code overwrites that header on error replies, which I missed. There appears to be no reason to set these header fields again here -- they're set when the `res_` object is created.

Also, forcing the connection close on an error reply here seems overkill. There is no reason that needs to happen: an error response doesn't prevent a new request from being sent on the same connection. Not touching that in 3.2 though.


spoonincode:
will need to be reworked going forward

PR #543, opened 2022-12-06
[3.1] Interrupt speculative start_block when a block is received

### LINKED ISSUE #517: Invalid warn message logged
```
Nov 26 11:57:00 xxx nodeos[4485]: info  2022-11-26T11:57:00.142 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block b10d9866ab2232ea... #215999358 @ 2022-11-26T11:57:00.000 signed by teamgreymass [trxs: 24, lib: 215999034, conf: 240, latency: 142 ms]
Nov 26 11:57:00 xxx nodeos[4485]: info  2022-11-26T11:57:00.774 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block eb3262600a640fc5... #215999359 @ 2022-11-26T11:57:00.500 signed by teamgreymass [trxs: 79, lib: 215999034, conf: 0, latency: 274 ms]
Nov 26 11:57:00 xxx nodeos[4485]: warn  2022-11-26T11:57:00.774 nodeos    producer_plugin.cpp:1745      remove_expired_persi ] Unable to process all expired transactions in unapplied queue before deadline, Persistent expired 0, Other expired 0
Nov 26 11:57:01 xxx nodeos[4485]: info  2022-11-26T11:57:01.669 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block 1733b406305c29d8... #215999360 @ 2022-11-26T11:57:01.000 signed by teamgreymass [trxs: 461, lib: 215999034, conf: 0, latency: 669 ms]
Nov 26 11:57:01 xxx nodeos[4485]: info  2022-11-26T11:57:01.847 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block 64c6e197dd1e45dc... #215999361 @ 2022-11-26T11:57:01.500 signed by teamgreymass [trxs: 317, lib: 215999034, conf: 0, latency: 347 ms]
Nov 26 11:57:02 xxx nodeos[4485]: info  2022-11-26T11:57:02.214 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block 4946de59c09667e5... #215999362 @ 2022-11-26T11:57:02.000 signed by teamgreymass [trxs: 127, lib: 215999034, conf: 0, latency: 214 ms]
Nov 26 11:57:02 xxx nodeos[4485]: info  2022-11-26T11:57:02.794 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block 729e10524e2208aa... #215999363 @ 2022-11-26T11:57:02.500 signed by teamgreymass [trxs: 175, lib: 215999034, conf: 0, latency: 294 ms]
Nov 26 11:57:02 xxx nodeos[4485]: warn  2022-11-26T11:57:02.795 nodeos    producer_plugin.cpp:1745      remove_expired_persi ] Unable to process all expired transactions in unapplied queue before deadline, Persistent expired 0, Other expired 0
Nov 26 11:57:03 xxx nodeos[4485]: info  2022-11-26T11:57:03.245 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block 997db8c680097ce5... #215999364 @ 2022-11-26T11:57:03.000 signed by teamgreymass [trxs: 127, lib: 215999034, conf: 0, latency: 245 ms]
Nov 26 11:57:03 xxx nodeos[4485]: warn  2022-11-26T11:57:03.246 nodeos    producer_plugin.cpp:1745      remove_expired_persi ] Unable to process all expired transactions in unapplied queue before deadline, Persistent expired 0, Other expired 0
Nov 26 11:57:03 xxx nodeos[4485]: info  2022-11-26T11:57:03.803 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block f05604f33c9065f8... #215999365 @ 2022-11-26T11:57:03.500 signed by teamgreymass [trxs: 138, lib: 215999034, conf: 0, latency: 303 ms]
Nov 26 11:57:03 xxx nodeos[4485]: warn  2022-11-26T11:57:03.804 nodeos    producer_plugin.cpp:1745      remove_expired_persi ] Unable to process all expired transactions in unapplied queue before deadline, Persistent expired 0, Other expired 0
Nov 26 11:57:04 xxx nodeos[4485]: info  2022-11-26T11:57:04.303 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block 6cf8ff2299ed863e... #215999366 @ 2022-11-26T11:57:04.000 signed by teamgreymass [trxs: 138, lib: 215999034, conf: 0, latency: 303 ms]
Nov 26 11:57:04 xxx nodeos[4485]: warn  2022-11-26T11:57:04.304 nodeos    producer_plugin.cpp:1745      remove_expired_persi ] Unable to process all expired transactions in unapplied queue before deadline, Persistent expired 0, Other expired 0
Nov 26 11:57:04 xxx nodeos[4485]: info  2022-11-26T11:57:04.824 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block 1abbd09fa4c7656e... #215999367 @ 2022-11-26T11:57:04.500 signed by teamgreymass [trxs: 127, lib: 215999034, conf: 0, latency: 324 ms]
Nov 26 11:57:04 xxx nodeos[4485]: warn  2022-11-26T11:57:04.825 nodeos    producer_plugin.cpp:1745      remove_expired_persi ] Unable to process all expired transactions in unapplied queue before deadline, Persistent expired 0, Other expired 0
Nov 26 11:57:05 xxx nodeos[4485]: info  2022-11-26T11:57:05.214 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block 69fbccb3a6052d00... #215999368 @ 2022-11-26T11:57:05.000 signed by teamgreymass [trxs: 121, lib: 215999034, conf: 0, latency: 214 ms]
Nov 26 11:57:05 xxx nodeos[4485]: warn  2022-11-26T11:57:05.215 nodeos    producer_plugin.cpp:1745      remove_expired_persi ] Unable to process all expired transactions in unapplied queue before deadline, Persistent expired 0, Other expired 0
Nov 26 11:57:05 xxx nodeos[4485]: info  2022-11-26T11:57:05.469 nodeos    producer_plugin.cpp:379       on_incoming_block    ] Received block 4869950285134d61... #215999369 @ 2022-11-26T11:57:05.500 signed by teamgreymass [trxs: 99, lib: 215999034, conf: 0, latency: -30 ms]
```

These `Unable to process all expired transactions in unapplied queue before deadline` are not valid messages. They look to be a side-effect of an "invalid" deadline passed to `remove_expired_trxs`.

### LINKED ISSUE #525: Block propagation delay under load
During `start_block` a deadline is set so that trx processing does not exceed the configured block production window according to `produce-time-offset-us` and `cpu-effort-percent`. Pending transactions are processed in a tight loop on the main thread until all pending are processed or the deadline is hit.

For a production block, if a block comes in during the block production window it is dropped.

For a speculative block, the incoming block is not processed until the main thread is released for processing the next task. Currently this means that the incoming block is not processed until after the completion of `start_block`. This can delay the incoming block processing by as much as 500ms.

Block propagation can be improved during high load by exiting speculative block `start_block` when a block comes into the node so that it can be processed immediately after the execution of the current trx.

heifner:
Interrupt speculative `start_block` when a block is received, after the block header has been validated, so it can be immediately processed even when there is a backlog of pending transactions.

- Forkdb made thread-safe.
- Block header validation now done on net threads instead of producer threads when previous block already processed.

Resolves #525 
Resolves #517


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3143,6 +3143,7 @@ namespace eosio {
    // called from connection strand
    void connection::handle_message( const block_id_type& id, signed_block_ptr ptr ) {
       peer_dlog( this, "received signed_block ${num}, id ${id}", ("num", ptr->block_num())("id", id) );
+      my_impl->producer_plug->received_block();
```
spoonincode, line 3146: I can't remember where we settled on how much validation to perform on a received block before interruption. Should there at least be some sort of check so we don't interrupt on (duplicate) reception of the existing head block?

heifner: This only interrupts speculative start_block. Any validation would require interacting with the main thread; at least for now. Possibly we could make forkdb thread safe and begin validation before interruption in the future. I see this as a step in the right direction and minimum changes for targeting 3.1 & 3.2.

spoonincode: A node may be connected to a handful of nodes & reception of the same block from all those nodes could easily be spread across a wide time window; for each reception it's going to potentially interrupt the speculative start block. On the surface seems like there could be a lot of spurious interruptions?

heifner: The `received_block()` call is after the de-dup in `net_plugin` `process_next_block_message()`. So only one interruption per unique block_id.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1854,16 +1855,23 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
...
          if( !remove_expired_trxs( preprocess_deadline ) )
             return start_block_result::exhausted;
+         if( _pending_block_mode == pending_block_mode::speculating && _received_block )
+            return start_block_result::exhausted;
          if( !remove_expired_blacklisted_trxs( preprocess_deadline ) )
             return start_block_result::exhausted;
+         if( _pending_block_mode == pending_block_mode::speculating && _received_block )
+            return start_block_result::exhausted;
          if( !_subjective_billing.remove_expired( _log, chain.pending_block_time(), fc::time_point::now(), preprocess_deadline ) )
             return start_block_result::exhausted;
```
greg7mdp, line 1871:
```suggestion
         if(should_stop_speculating() || !remove_expired_trxs( preprocess_deadline ) )
            return start_block_result::exhausted;
         if(should_stop_speculating() || ! !remove_expired_blacklisted_trxs( preprocess_deadline ) )
            return start_block_result::exhausted;
         if(should_stop_speculating() || ! !_subjective_billing.remove_expired( _log, chain.pending_block_time(), fc::time_point::now(), preprocess_deadline ) )
            return start_block_result::exhausted;
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2009,7 +2017,7 @@ bool producer_plugin_impl::process_unapplied_trxs( const fc::time_point& deadlin
       auto end_itr = (_pending_block_mode == pending_block_mode::producing) ?
                      _unapplied_transactions.unapplied_end()   : _unapplied_transactions.persisted_end();
       while( itr != end_itr ) {
-         if( deadline <= fc::time_point::now() ) {
+         if( deadline <= fc::time_point::now() || (_pending_block_mode == pending_block_mode::speculating && _received_block) ) {
```
greg7mdp, line 2020:
```suggestion
         if( deadline <= fc::time_point::now() || should_stop_speculating() ) {
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2210,7 +2218,7 @@ bool producer_plugin_impl::process_incoming_trxs( const fc::time_point& deadline
       auto itr = _unapplied_transactions.incoming_begin();
       auto end = _unapplied_transactions.incoming_end();
       while( pending_incoming_process_limit && itr != end ) {
-         if (deadline <= fc::time_point::now()) {
+         if (deadline <= fc::time_point::now() || (_pending_block_mode == pending_block_mode::speculating && _received_block) ) {
```
greg7mdp, line 2221:
```suggestion
         if (deadline <= fc::time_point::now() || should_stop_speculating()) {
```

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: Super sorry, I commited my suggestion by mistake! I didn't realize that was what "Commit suggestion" did!.
Should I try to undo it?

plugins/producer_plugin/producer_plugin.cpp
heifner: Added a `start_block_interrupted` method. This was much cleaner and handles all the cases. The only deadline still in play for speculative block is the block deadline passed to `push_transaction`. This really should be `fc::time_point::maximum()` for speculative (it will use max transaction time then), but I thought that a bit too much of a change for 3.1 & 3.2. I'll make that change for `main`.

plugins/producer_plugin/producer_plugin.cpp
heifner: I reverted this commit.

libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
```diff
@@ -96,15 +96,15 @@ class unapplied_transaction_queue {
       return itr->trx_meta;
    }
 
-   template <typename Func>
-   bool clear_expired( const time_point& pending_block_time, const time_point& deadline, Func&& callback ) {
+   template <typename Yeild, typename Callback>
```
greg7mdp, line 99: Probably should be `Yield` (both type and parameter name)

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1680,6 +1681,14 @@ fc::time_point producer_plugin_impl::calculate_block_deadline( const fc::time_po
...
 }
 
+bool producer_plugin_impl::start_block_interrupted( const fc::time_point& deadline ) {
+   if( _pending_block_mode == pending_block_mode::speculating ) {
+      return _received_block;
+   } else {
+      return deadline <= fc::time_point::now();
+   }
+}
+
```
greg7mdp, line 1691: I don't think this is correct since the deadline if considered only when `_pending_block_mode` is *not* speculating.

libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
heifner: If only I could spell. Thanks!

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: maybe:
```
return (_pending_block_mode == pending_block_mode::speculating && _received_block) ||
        deadline <= fc::time_point::now();
```

heifner: That is what we want. A speculative block doesn't have a deadline. It was given one according to scheduled block, but that is really not meaningful for a speculative block.

heifner: We could argue that this is too much of a change for 3.1 & 3.2 and still go with the meaningless deadline and only remove the deadline for `main`.

greg7mdp: OK, thanks Kevin. Just a detail, which probably doesn't matter here, but personally I try to avoid some paths with no return, even if they are unreachable, so I would do either:

```
   if( _pending_block_mode == pending_block_mode::speculating ) {
      return _received_block;
   } 
   return deadline <= fc::time_point::now();
```

or 

```
return _pending_block_mode == pending_block_mode::speculating ? _received_block : deadline <= fc::time_point::now();
```

heifner: Done.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: Nice!

unittests/unapplied_transaction_queue_tests.cpp
```diff
@@ -305,7 +305,7 @@ BOOST_AUTO_TEST_CASE( unapplied_transaction_queue_test ) try {
    q.add_aborted( { trx20, trx22 } );
    q.add_persisted( trx21 );
    q.add_persisted( trx23 );
-   q.clear_expired( fc::time_point::now(), fc::time_point::now() + fc::seconds( 300 ), [](auto, auto){} );
+   q.clear_expired( fc::time_point::now(), [](){ return false; }, [](auto, auto){} );
```
greg7mdp, line 308: I assume it is not dangerous to lose this timeout, right?

greg7mdp (approved):

unittests/unapplied_transaction_queue_tests.cpp
heifner: Correct, this is test code.

plugins/net_plugin/net_plugin.cpp
heifner: Discussed and we want to look at adding block header validation before interrupting speculative `start_block`.

heifner:
Example of it working:
```
debug 2022-12-15T02:33:00.553 nodeos    producer_plugin.cpp:2289      process_incoming_trx ] Processing 6995 pending transactions
info  2022-12-15T02:33:00.629 net-1     net_plugin.cpp:3145           handle_message       ] ["localhost:9876 - 19961c6" - 5 127.0.0.1:9876] received signed_block 649, id 00000289dedc8390a50b989642d19d1a790816ab780671c47935f349b3d8f777
info  2022-12-15T02:33:00.648 net-1     net_plugin.cpp:3174           handle_message       ] ["localhost:9876 - 19961c6" - 5 127.0.0.1:9876] posting signed_block 649, id 00000289dedc8390a50b989642d19d1a790816ab780671c47935f349b3d8f777
debug 2022-12-15T02:33:00.650 nodeos    producer_plugin.cpp:2314      process_incoming_trx ] Processed 2903 pending transactions, 4092 left
debug 2022-12-15T02:33:00.650 nodeos    producer_plugin.cpp:2373      schedule_production_ ] Speculative Block Created; Scheduling Speculative/Production Change
debug 2022-12-15T02:33:00.650 nodeos    producer_plugin.cpp:2441      schedule_delayed_pro ] Scheduling Speculative/Production Change at 2022-12-15T02:33:17.500
```

plugins/net_plugin/net_plugin.cpp
heifner: Block header validation now performed before interrupting `start_block`.

libraries/chain/fork_database.cpp
```diff
@@ -456,16 +519,21 @@ namespace eosio { namespace chain {
       }
 
       for( const auto& block_id : remove_queue ) {
-         auto itr = my->index.find( block_id );
-         if( itr != my->index.end() )
-            my->index.erase(itr);
+         auto itr = index.find( block_id );
+         if( itr != index.end() )
+            index.erase(itr);
```
greg7mdp, line 524: Can't we replace these 3 lines with `index.erase(block_id);`?

libraries/chain/fork_database.cpp
```diff
@@ -436,13 +494,18 @@ namespace eosio { namespace chain {
...
    void fork_database::remove( const block_id_type& id ) {
+      std::lock_guard g( my->mtx );
+      return my->remove_impl( id );
+   }
+
+   void fork_database_impl::remove_impl( const block_id_type& id ) {
       deque<block_id_type> remove_queue{id};
-      const auto& previdx = my->index.get<by_prev>();
-      const auto& head_id = my->head->id;
+      const auto& previdx = index.get<by_prev>();
```
greg7mdp, line 507: for the `remove_queue`, is there a benefit to use `std::deque` vs `std::vector`?

unittests/block_tests.cpp
```diff
@@ -44,9 +44,9 @@ BOOST_AUTO_TEST_CASE(block_with_invalid_tx_test)
 
    // Push block with invalid transaction to other chain
    tester validator;
-   auto bs = validator.control->create_block_state_future( copy_b->calculate_id(), copy_b );
+   auto bsf = validator.control->create_block_state_future( copy_b->calculate_id(), copy_b );
    validator.control->abort_block();
-   BOOST_REQUIRE_EXCEPTION(validator.control->push_block( bs, forked_branch_callback{}, trx_meta_cache_lookup{} ), fc::exception ,
+   BOOST_REQUIRE_EXCEPTION(validator.control->push_block( bsf.get(), forked_branch_callback{}, trx_meta_cache_lookup{} ), fc::exception ,
```
greg7mdp, line 49: Just wondering, before the change, the `future.get()` occured within a try block inside `push_block`. Now the `get()` is done in the caller. Could this be an issue?

libraries/chain/fork_database.cpp
```diff
@@ -262,48 +291,53 @@ namespace eosio { namespace chain {
...
 
       return block_header_state_ptr();
    }
 
-   void fork_database_impl::add( const block_state_ptr& n,
-                                 bool ignore_duplicate, bool validate,
-                                 const std::function<void( block_timestamp_type,
-                                                           const flat_set<digest_type>&,
-                                                           const vector<digest_type>& )>& validator )
+   void fork_database_impl::add_impl( const block_state_ptr& n,
```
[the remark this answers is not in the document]
heifner: Funny enough, boost `insert` that takes an rvalue reference just passes it to `final_insert_rv_` which takes it by `const&`. I agree according to the interface it should be more efficient, I was just a little surprised it is not in practice. 

All uses of `fork_database::add` can't move it as it is used after. So going to leave it a `const&`.

libraries/chain/fork_database.cpp
```diff
@@ -357,8 +399,13 @@ namespace eosio { namespace chain {
...
 
    branch_type fork_database::fetch_branch( const block_id_type& h, uint32_t trim_after_block_num )const {
+      std::shared_lock g( my->mtx );
+      return my->fetch_branch_impl( h, trim_after_block_num );
+   }
+
+   branch_type fork_database_impl::fetch_branch_impl( const block_id_type& h, uint32_t trim_after_block_num )const {
       branch_type result;
-      for( auto s = get_block(h); s; s = get_block( s->header.previous ) ) {
+      for( auto s = get_block_impl(h); s; s = get_block_impl( s->header.previous ) ) {
```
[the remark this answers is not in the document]
heifner: It is a `deque` because we found in testing that it is more efficient as a `deque`.

libraries/chain/fork_database.cpp
```diff
@@ -381,9 +433,15 @@ namespace eosio { namespace chain {
...
+      return my->fetch_branch_from_impl( first, second );
+   }
+
+   pair< branch_type, branch_type >  fork_database_impl::fetch_branch_from_impl( const block_id_type& first,
+                                                                                 const block_id_type& second )const {
       pair<branch_type,branch_type> result;
-      auto first_branch = (first == my->root->id) ? my->root : get_block(first);
-      auto second_branch = (second == my->root->id) ? my->root : get_block(second);
+      auto first_branch = (first == root->id) ? root : get_block_impl(first);
+      auto second_branch = (second == root->id) ? root : get_block_impl(second);
```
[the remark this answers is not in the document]
heifner: Because `root` is not a complete block.

libraries/chain/fork_database.cpp
```diff
@@ -392,7 +450,7 @@ namespace eosio { namespace chain {
       {
          result.first.push_back(first_branch);
          const auto& prev = first_branch->header.previous;
-         first_branch = (prev == my->root->id) ? my->root : get_block( prev );
+         first_branch = (prev == root->id) ? root : get_block_impl( prev );
```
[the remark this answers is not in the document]
heifner: Because `root` is not a complete block.

libraries/chain/fork_database.cpp
```diff
@@ -403,7 +461,7 @@ namespace eosio { namespace chain {
       {
          result.second.push_back( second_branch );
          const auto& prev = second_branch->header.previous;
-         second_branch = (prev == my->root->id) ? my->root : get_block( prev );
+         second_branch = (prev == root->id) ? root : get_block_impl( prev );
```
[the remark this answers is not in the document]
heifner: Because `root` is not a complete block.

libraries/chain/fork_database.cpp
heifner: Yes, this was a performance improvement awhile back. Significant performance improvement using `deque` over `vector` for large number of transactions. Note it is `deque` not `std::deque`, we actually use a `boost` `deque` for modern `boost` versions...
Or at least we are suppose to be. I guess that never got ported over to leap.
https://github.com/EOSIO/eos/blob/master/libraries/chain/include/eosio/chain/types.hpp#L85
I created https://github.com/AntelopeIO/leap/issues/576 for this.

unittests/block_tests.cpp
heifner: They try-catch in `push_block` just logs and re-throws.

libraries/chain/fork_database.cpp
heifner: Yes.

libraries/chain/controller.cpp
```diff
@@ -2056,43 +2056,59 @@ struct controller_impl {
       }
    } FC_CAPTURE_AND_RETHROW() } /// apply_block
 
+
+   block_state_ptr create_block_state( const block_id_type& id, const signed_block_ptr& b, const block_header_state_ptr& prev ) {
```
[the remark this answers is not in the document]

libraries/chain/controller.cpp
```diff
@@ -2056,43 +2056,59 @@ struct controller_impl {
...
-                        *prev,
-                        move( b ),
-                        control->protocol_features.get_protocol_feature_set(),
-                        [control]( block_timestamp_type timestamp,
-                                   const flat_set<digest_type>& cur_features,
-                                   const vector<digest_type>& new_features )
-                        { control->check_protocol_features( timestamp, cur_features, new_features ); },
-                        skip_validate_signee
-         );
+      block_state_ptr bsp;
```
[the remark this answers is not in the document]
heifner: Removed.

libraries/chain/fork_database.cpp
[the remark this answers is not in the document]
greg7mdp: Hum, I don't quite understand the problem. I was suggesting adding the first two lines in the following function:

```
   block_state_ptr   fork_database_impl::get_block_impl(const block_id_type& id)const {
      if (id == root->id)
         return root;
      auto itr = index.find( id );
      if( itr != index.end() )
         return *itr;
      return block_state_ptr();
   }
```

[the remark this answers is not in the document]
heifner: We can't do that because `root` is not a complete block. See comment on `root` declaration and see `reset_impl`.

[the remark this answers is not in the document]
greg7mdp: But that is the same `root` that is assigned to `first_branch`, isn't it? Unless I am missing something, we are accessing the same member of the same class.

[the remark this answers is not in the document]
heifner: Yes, but `get_block` is used externally as well. We could create a `get_block_or_root_impl` and use it for these cases, but not sure it is worth it.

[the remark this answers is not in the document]
greg7mdp: OK, I think I get it, there are some call locations where it would be OK to do this, and some other locations where it wouldn't. Maybe we should have two different `get_block_impl`? Still not 100% clear to me but that's OK.

libraries/chain/fork_database.cpp
```diff
@@ -60,34 +61,46 @@ namespace eosio { namespace chain {
...
-   :my( new fork_database_impl( *this, data_dir ) )
+   :my( new fork_database_impl( data_dir ) )
    {}
 
 
    void fork_database::open( const std::function<void( block_timestamp_type,
                                                        const flat_set<digest_type>&,
                                                        const vector<digest_type>& )>& validator )
    {
+      std::lock_guard g( my->mtx );
```
greg7mdp, line 103: I think it would be better to follow the pattern you established and have a fork_database_impl::open_impl(). The benefit being:
1. removing the `my->` 
2. that within open_impl(), we can only call the methods from `fork_database_impl`, so there is no risk of trying to lock the mutex again

libraries/chain/fork_database.cpp
```diff
@@ -163,6 +176,7 @@ namespace eosio { namespace chain {
    }
 
    void fork_database::close() {
```
greg7mdp, line 178: Same suggestion as for `open_impl`

libraries/chain/fork_database.cpp
heifner: Done.

libraries/chain/fork_database.cpp
heifner: Done.

plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
```diff
@@ -38,7 +38,6 @@ namespace eosio { namespace chain { namespace plugin_interface {
 
    namespace incoming {
       namespace channels {
-         using block                 = channel_decl<struct block_tag, signed_block_ptr>;
          using transaction           = channel_decl<struct transaction_tag, packed_transaction_ptr>;
```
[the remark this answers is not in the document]
heifner: Sounds good to me.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3143,14 +3143,44 @@ namespace eosio {
...
+         peer_elog(this, "bad block: #${n} ${id}...: unknown exception",
+                   ("n", ptr->block_num())("id", id.str().substr(8,16)));
+      }
+      if( exception ) {
+         my_impl->sync_master->rejected_block( shared_from_this(), ptr->block_num() );
+         my_impl->dispatcher->rejected_block( id );
+         return;
+      }
+
+      bool signal_producer = !!bsp; // ready to process immediately, so signal producer to interrupt start_block
```
linh2931, line 3173: why double `!!` here?

greg7mdp: > why double `!!` here?

Compact way to convert to a bool without a compiler warning. Here it tests that the shared_ptr is not empty. I like using this pattern as well :-)

heifner: Exactly, `bool signal_producer = bsp;` is a compiler error.

libraries/chain/controller.cpp
```diff
@@ -2056,43 +2056,58 @@ struct controller_impl {
       }
    } FC_CAPTURE_AND_RETHROW() } /// apply_block
 
+
+   block_state_ptr create_block_state( const block_id_type& id, const signed_block_ptr& b, const block_header_state& prev ) {
+      auto trx_mroot = calculate_trx_merkle( b->transactions );
```
linh2931, line 2061: Will `calculate_trx_merkle` take long time on the `main` thread when called by `create_block_state`?

libraries/chain/controller.cpp
```diff
@@ -2056,43 +2056,58 @@ struct controller_impl {
...
-         const bool skip_validate_signee = false;
+         return control->create_block_state( id, b, *prev );
+      } );
+   }
 
-         auto trx_mroot = calculate_trx_merkle( b->transactions );
-         EOS_ASSERT( b->transaction_mroot == trx_mroot, block_validate_exception,
-                     "invalid block transaction merkle root ${b} != ${c}", ("b", b->transaction_mroot)("c", trx_mroot) );
+   block_state_ptr create_block_state( const block_id_type& id, const signed_block_ptr& b ) {
+      EOS_ASSERT( b, block_validate_exception, "null block" );
```
linh2931, line 2099: Should the `existing` block check in `create_block_state_future` be done here too?

libraries/chain/controller.cpp
```diff
@@ -2101,7 +2116,7 @@ struct controller_impl {
          trusted_producer_light_validation = old_value;
       });
       try {
-         block_state_ptr bsp = block_state_future.get();
+         EOS_ASSERT( bsp, block_validate_exception, "null block" );
```
linh2931, line 2119: Will this break consensus as there was not check before? Or should `bsp` is checked before `push_block` is called?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1680,6 +1669,14 @@ fc::time_point producer_plugin_impl::calculate_block_deadline( const fc::time_po
    }
 }
 
+bool producer_plugin_impl::start_block_interrupted( const fc::time_point& deadline ) const {
+   if( _pending_block_mode == pending_block_mode::producing ) {
+      return deadline <= fc::time_point::now();
+   }
+   // if we can produce then honor deadline so production starts on time
+   return (!_producers.empty() && deadline <= fc::time_point::now()) || _received_block;
```
linh2931, line 1677: Why `!_producers.empty()` here? Shouldn't we interrupt whenever the deadline is passed?

`_received_block` will trigger interruption in both `producing` and `speculating` modes?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1680,6 +1669,14 @@ fc::time_point producer_plugin_impl::calculate_block_deadline( const fc::time_po
    }
 }
 
+bool producer_plugin_impl::start_block_interrupted( const fc::time_point& deadline ) const {
```
linh2931, line 1672: Not sure if the name `start_block_interrupted` can be changed. The method actually determines whether to interrupt `start_block` or not. But it sounds like `start_block` has already been interrupted. This causes a little confusion when you read the code below when `start_block_interrupted` is used.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2251,6 +2249,7 @@ bool producer_plugin_impl::block_is_exhausted() const {
 // -> Idle
 // --> Start block B (block time y.000) at time x.500
 void producer_plugin_impl::schedule_production_loop() {
+   _received_block = false;
```
linh2931, line 2252: Is it possible a race condition that a block is received after `start_block_interrupted` is called but before `schedule_production_loop`?

libraries/chain/controller.cpp
heifner: `create_block_state` is never called from the main thread. I'll add a comment.

libraries/chain/controller.cpp
heifner: This is one of those cases for @greg7mdp abort. It should never happen. If b is null then core dump should happen soon if this `EOS_ASSERT` was removed.

plugins/producer_plugin/producer_plugin.cpp
heifner: Yes, :(. Not really anything we can do about that.

plugins/producer_plugin/producer_plugin.cpp
heifner: In speculative mode (HEAD mode) you just keep using the same block until a block is received. There is no reason to start a new block until another block is received. 

Yes, `_received_block` will trigger interruption in either case.

plugins/producer_plugin/producer_plugin.cpp
heifner: `is_start_block_interrupted` ?

linh2931: Or `can_interrupt_start_block` ? Hard to come up a good name.

heifner: We use `should` in some places, how about `should_interrupt_start_block`.

linh2931: `should_interrupt_start_block` is good. Thanks.

linh2931 (approved):

larryk85 (approved):

PR #556, opened 2022-12-08
Set keosd max http response time to unlimited

### LINKED ISSUE #230: keosd http-max-response-time-ms setting
Users often hit the default `30ms` issue with timeouts when using `keosd` due to the default `http-max-response-time-ms` 30ms. Setting contracts or ABI can often take longer than 30ms to parse in keosd. This error is confusing to users and often hard to track down what is going wrong. Also `keosd` is meant to only be run locally and not provided as a service. 

Consider making `http-max-response-time-ms` unlimited for `keosd`.

heifner:
Users sometimes hit the default 30ms issue with timeouts when using `keosd` due to the default `http-max-response-time-ms` 30ms. Setting contracts or ABI can often take longer than 30ms to parse in `keosd`. This error is confusing to users and often hard to track down what is going wrong. Also `keosd` is meant to only be run locally and not provided as a service.

This PR makes `http-max-response-time-ms` unlimited for `keosd`.

Also includes a bit of cleanup around exception handling in `beast_http_session`.

Resolves #230

### DIFF
--- a/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
@@ -159,7 +159,7 @@ protected:
             error_results results{static_cast<uint16_t>(http::status::not_found), "Not Found",
                                   error_results::error_info( fc::exception( FC_LOG_MESSAGE( error, "Unknown Endpoint" ) ),
                                                              http_plugin::verbose_errors() )};
-            send_response( fc::json::to_string( results, fc::time_point::now() + plugin_state_->max_response_time ),
+            send_response( fc::json::to_string( results, fc::time_point::maximum() ),
                            static_cast<unsigned int>(http::status::not_found) );
          }
       } catch(...) {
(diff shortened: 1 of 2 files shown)


plugins/wallet_api_plugin/wallet_api_plugin.cpp
```diff
@@ -25,7 +25,7 @@ using namespace eosio;
    [&api_handle](string, string body, url_response_callback cb) mutable { \
           try { \
              INVOKE \
-             cb(http_response_code, fc::time_point::maximum(), fc::variant(result)); \
+             cb(http_response_code, fc::time_point::now() + fc::days(365), fc::variant(result)); \
```
linh2931, line 28: Original 30ms can be too short, but 365 days might be too long. Do we want to set it to some reasonable number like 10 seconds so feedback can be returned in case keosd is really down?

linh2931: Got it.

linh2931 (approved):

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -159,7 +159,7 @@ class beast_http_session : public detail::abstract_conn {
             error_results results{static_cast<uint16_t>(http::status::not_found), "Not Found",
                                   error_results::error_info( fc::exception( FC_LOG_MESSAGE( error, "Unknown Endpoint" ) ),
                                                              http_plugin::verbose_errors() )};
-            send_response( fc::json::to_string( results, fc::time_point::now() + plugin_state_->max_response_time ),
+            send_response( fc::json::to_string( results, fc::time_point::maximum() ),
```
greg7mdp, line 162: Why this change?

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -311,27 +311,33 @@ class beast_http_session : public detail::abstract_conn {
          } catch(const fc::exception& e) {
             err_str = e.to_detail_string();
             fc_elog(plugin_state_->logger, "fc::exception: ${w}", ("w", err_str));
-            error_results results{static_cast<uint16_t>(http::status::internal_server_error),
-                                  "Internal Service Error",
-                                  error_results::error_info( e, http_plugin::verbose_errors() )};
-            err_str = fc::json::to_string( results, fc::time_point::now() + plugin_state_->max_response_time );
+            if( is_send_exception_response_ ) {
```
greg7mdp, line 314: Not related to this PR, but I don't think the name `is_send_exception_response_` is meaningful, I would prefer `should_send_exception_response_`.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
heifner: Because the input to `to_string` is small and static so no possibility of json bomb.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
heifner: We do somewhat consistently use `is` for bools, but I don't disagree.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -311,27 +311,33 @@ class beast_http_session : public detail::abstract_conn {
...
+               error_results results{static_cast<uint16_t>(http::status::internal_server_error),
+                                     "Internal Service Error",
+                                     error_results::error_info(
+                                           fc::exception( FC_LOG_MESSAGE( error, err_str ) ),
+                                           http_plugin::verbose_errors() )};
+               err_str = fc::json::to_string( results, fc::time_point::maximum() );
+            }
          }
       } catch (fc::timeout_exception& e) {
```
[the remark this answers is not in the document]
heifner: No because we log something diff than we return.

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
greg7mdp: OK, thanks!

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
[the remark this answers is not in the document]
heifner: No, it will have `err_str = e.to_detail_string();` or `err_str = e.what();` or `err_str = "Unknown exception";`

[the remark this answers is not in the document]
greg7mdp: Oh, man I get it now, thanks for explaining!

greg7mdp (approved):

Issue #320, opened 2022-10-13
Further enhancements to leap-util framework


arhag:
Enhancements to CLI11 or leap-util framework:

- [ ] Show common options for sub-commands higher in the hierarchy of the target sub-command that the `--help` is displayed for.
- [ ] Disallow framework users from allowing non-leaf sub-command from being actionable.
- [ ] Re-use shared options for several sub-commands (by constructing option object once and adding it to the various sub-commands).

766C6164:
PR is in cli11 repo for submodule:
https://github.com/AntelopeIO/CLI11/pull/5

arhag:
The item "Disallow framework users from allowing non-leaf sub-command from being actionable." was resolved not by changing the CLI11 framework to do this automatically but by using the existing framework functionality of `require_subcommand` on the non-leaf subcommands. It is the responsibility of the developer to add that command to each non-leaf subcommand.

PR #557, opened 2022-12-08
CLI11: Use inherited option vs duplicating it in leap-util (blocklog)

766C6164:
This PR takes an advantage of ->fallthrough() subcommand method of CLI11 to avoid duplication of same option in multiple subcommands by inheriting it from parent.
In relation to https://github.com/AntelopeIO/leap/issues/320


programs/leap-util/actions/generic.cpp
```diff
@@ -4,7 +4,14 @@
...
 void generic_actions::setup(CLI::App& app) {
    auto* sub = app.add_subcommand("version", "retrieve version information");
-   sub->require_subcommand(1);
+
+   /// makes  subcommand "version" non-actionable, actions expected in child subcommands
+   sub->require_subcommand(); 
+   
+   /// this will make options defined on parent subcommand to be actionable for child subcommands and to be displayed in --help call for them
+   /// it's not required in this handler, but set here for documentation purposes to showcase that in most cases for custom action handlers it is a desirable behavior
+   sub->fallthrough();
```
arhag, line 13: This should be added for the other top-level commands too, right?

766C6164: Great suggestion, added!

linh2931 (approved):

arhag (approved):

Issue #440, opened 2022-11-03
New RPC endpoint send_read_only_transaction


linh2931:
This is the first of the three phases of implementing parallel readonly transaction execution. The task is to make sure readonly transactions to be handled correctly without multithreading involved. After this is done, move to parallelizing readonly transactions and necessary configuration changes.

1. Create a new RPC endpoint `send_readonly_transaction`.
2. Do not set contract account usage with default CPU and NET usage initially.  This is to prevent `resource_limits_config_object` from updated.
3. Do not create `transaction_object`, as duplicate transaction checking is not needed for readonly transactions.
4. Do not increment action sequence (dynamic_global_property_object), receiver sequence (`account_metadata_object`), auth sequence (`account_metadata_object`) at the end of transaction execution. Readonly transactions should be a no-op on those sequences. 
5. Do not update transaction usage (`resource_usage_object`) at the end of transaction execution.
6. Do not allow authority checks so that `permission_usage_object` is not updated.
7. Assert when a user (a smart contract) tries to modify DB states. Specifically, add asserts to `db_store_i64`, `db_update_i64`, and `db_remove_i64` in `chain/apply_context.cpp`, `store`, `update`, and `remove` in `generic_index` class in `apply_context.hpp`.
8. Assert when native functions are called to change database. Specifically, add assert to any method in `eosio_contract.cpp` which modifies DB states. 
9. Set `chainbase` mode to `assert_on_wirte`. Assert if any of `create`, `modify`, `remove` is called. This is intended to be a safety net. Any assert in `assert_on_wirte` mode is a bug which requires a fix.
10. Only enforce block CPU deadline, and in next phase enforce parallel readonly transactions deadline.
11. Do not perform deepmind logging for readonly transactions.

See also
https://github.com/eosnetworkfoundation/product/issues/95
https://github.com/AntelopeIO/leap/issues/364

PR #558, opened 2022-12-08
introduce new read-only transaction RPC end point

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/440

This PR introduces a new read-only transaction RPC end point. Major features are

- Provide a new endpoint `send_read_only_transaction`
- Provide a cleos `push transaction` option `--read` for pushing read-only transaction
- Provide a cleos `push action` option `--read` for pushing read-only action
- Do not sign any read-only transactions by `cleos`
- Do not allow any permissions for read-only transactions
- Assert any attempt to modify DB tables using DB API host functions
- Assert any attempt to modify DB tables using native functions
- Do not modify global_action_sequence and recv_sequence
- Do not do deepmind logging
- Add a catch-all flag in Chainbase to prevent any unlikely missed DB modification attempts.
- Add integration tests for cleos/chain_plugin/producer_plugin.
- Add unit tests for internal logic

Please note, the Chainbase submodule commit will be updated after corresponding Chainbase PR https://github.com/AntelopeIO/chainbase/pull/4 is merged.


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2521,6 +2521,42 @@ void read_only::compute_transaction(const compute_transaction_params& params, ne
...
+
+void read_only::send_read_only_transaction(const send_read_only_transaction_params& params, next_function<send_read_only_transaction_results> next) const {
+   try {
+      auto pretty_input = std::make_shared<packed_transaction>();
+      auto resolver = make_resolver(db, abi_serializer::create_yield_function( abi_serializer_max_time ));
+      try {
+         abi_serializer::from_variant(params.transaction, *pretty_input, resolver, abi_serializer::create_yield_function( abi_serializer_max_time ));
+      } EOS_RETHROW_EXCEPTIONS(chain::packed_transaction_type_exception, "Invalid packed transaction")
+
+      app().get_method<incoming::methods::transaction_async>()(pretty_input, false /* api_trx */, transaction_metadata::trx_type::read_only, true /* return_failure_trace */,
```
heifner, line 2533: Both here and compute_transaction, `api_trx` should be `true`. It is only used (in `main`) to determine which subjective billing configuration to honor. Subjective billing is not applicable to read_only and compute_transaction, but these are both technically api trxs (and not p2p transactions which is what `false` means).

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2006,7 +2008,13 @@ producer_plugin_impl::push_transaction( const fc::time_point& block_deadline,
       }
    }
 
+   if( trx->is_read_only() ) {
+      chain.set_db_read_only_mode();
+   }
    auto trace = chain.push_transaction( trx, block_deadline, max_trx_time, prev_billed_cpu_time_us, false, sub_bill );
```
heifner, line 2014: An exception thrown here leaves chain in read_only mode. Also not sure how this will work when you make it multi-threaded. I would have expected the `set_db_read_only_mode` to be set elsewhere according to a timer.

programs/cleos/main.cpp
```diff
@@ -492,6 +495,16 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
...
+         } else if( tx_read ) {
+            EOSC_ASSERT( !tx_retry_lib, "ERROR: --retry-irreversible can not be used with --read" );
+            EOSC_ASSERT( !tx_retry_num_blocks, "ERROR: --retry-num-blocks can not be used with --read" );
+            try {
+               auto args = fc::mutable_variant_object ("transaction", packed_transaction(trx,compression));
+               return call( send_read_only_txn_func, args );
+            } catch( chain::missing_chain_api_plugin_exception& ) {
+               std::cerr << "New RPC readonly_transaction may not be supported. Submit to a different node." << std::endl;
+               throw;
+            }
```
heifner, line 507: Seems like this could be combined with the above `if` and remove some duplicate code.

heifner (changes requested):

libraries/chain/eosio_contract.cpp
```diff
@@ -417,6 +425,8 @@ void apply_eosio_linkauth(apply_context& context) {
 void apply_eosio_unlinkauth(apply_context& context) {
 //   context.require_write_lock( config::eosio_auth_scope );
 
+   EOS_ASSERT( !context.trx_context.is_read_only(), action_validate_exception, "unlinkauth not allowed in read-only transaction" );
+
    auto& db = context.db;
    auto unlink = context.get_action().data_as<unlinkauth>();
```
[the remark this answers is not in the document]
linh2931: Thanks. I missed it. canceldelay modifies `generated_transaction_object`. Will add the check.

libraries/chain/apply_context.cpp
```diff
@@ -363,7 +363,7 @@ void apply_context::execute_inline( action&& a ) {
                                       control.pending_block_time() - trx_context.published,
                                       std::bind(&transaction_context::checktime, &this->trx_context),
                                       false,
-                                      trx_context.is_dry_run(),
+                                      trx_context.is_dry_run() || trx_context.is_read_only(), // check_but_dont_fail
```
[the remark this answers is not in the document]
linh2931: Thanks for your point! I will modify `controller::skip_auth_check()` to return `true` if a transaction is read-only, so all auth checks in the code are skipped.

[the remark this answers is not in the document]
linh2931: Actually I cannot just change `controller::skip_auth_check()`. Will skip auth check on transaction basis.

libraries/chain/apply_context.cpp
```diff
@@ -182,7 +182,7 @@ void apply_context::exec_one()
       print_debug(receiver, trace);
    }
 
-   if (auto dm_logger = control.get_deep_mind_logger())
+   if (auto dm_logger = control.get_deep_mind_logger(); dm_logger && !trx_context.is_read_only())
```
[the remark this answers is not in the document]
linh2931: Currently only in push_transaction of controller.cpp, deepmind is not logged for compute_transactions. I will make make compute_transaction consistent.

libraries/chain/controller.cpp
```diff
@@ -1592,7 +1592,7 @@ struct controller_impl {
                              std::move(trx_context.executed_action_receipt_digests) );
 
             // call the accept signal but only once for this transaction
-            if (!trx->is_dry_run()) {
+            if (!trx->is_dry_run() && !trx->is_read_only()) {
```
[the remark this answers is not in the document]
linh2931: Great idea. Will do that.

libraries/chain/transaction_context.cpp
```diff
@@ -222,6 +222,66 @@ namespace eosio { namespace chain {
       is_initialized = true;
    }
 
+   void transaction_context::init_for_read_only_trx()
+   {
+      // For read-only transactions, net and ram limits are not enforced.
```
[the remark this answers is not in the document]
linh2931: Correct. This function (`init_for_read_only_trx`) does not do any ram checks.

libraries/chain/transaction_context.cpp
```diff
@@ -321,6 +386,8 @@ namespace eosio { namespace chain {
    void transaction_context::finalize() {
       EOS_ASSERT( is_initialized, transaction_exception, "must first initialize" );
 
+      if ( is_read_only() ) return;
```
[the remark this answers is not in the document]
linh2931: Will change. Thanks.

libraries/chain/transaction_context.cpp
```diff
@@ -222,6 +222,66 @@ namespace eosio { namespace chain {
...
+      }
+
+      if(control.skip_trx_checks()) {
+         transaction_timer.start( fc::time_point::maximum() );
+      } else {
+         transaction_timer.start( _deadline );
+         checktime(); // Fail early if deadline has already been exceeded
+      }
+
+      is_initialized = true;
```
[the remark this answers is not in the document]
linh2931: I don't like the duplicate code. But we would add quite a few `if`s in `init` method.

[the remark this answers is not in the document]
heifner: This is complicated code that I think needs to be kept with the rest, I'm concerned about duplicating it due to maintenance.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: I think it should use a RAII object with a static counter, sets it to read_only when the counter is incremented 0 -> 1, and set to read_write  when the last destructor resets the counter to 0.

linh2931: This is only a temporary place to set the DB read_only_mode. It will be moved to where we toggle single threaded and multi-threaded execution modes when multi-threaded is implemented. But I will add a catch here to prevent DB stuck in read_only mode for now. Thanks.

plugins/chain_plugin/chain_plugin.cpp
linh2931: Thanks for the explanation.

programs/cleos/main.cpp
linh2931: Thanks.

libraries/chain/apply_context.cpp
```diff
@@ -289,7 +289,7 @@ void apply_context::require_recipient( account_name recipient ) {
          schedule_action( action_ordinal, recipient, false )
       );
 
-      if (auto dm_logger = control.get_deep_mind_logger()) {
+      if (auto dm_logger = control.get_deep_mind_logger(); dm_logger && !trx_context.is_read_only()) {
```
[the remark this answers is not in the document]
linh2931: Thanks for your suggestion. Since read-only is an exception to deep-mind logging, it helps to understand the code by specifically showing when we don't do deep-mind logging.

plugins/producer_plugin/producer_plugin.cpp
heifner: I agree that RAII (`fc::scoped_exit`) would be much cleaner than the try-catch.

libraries/chain/transaction_context.cpp
```diff
@@ -321,6 +333,14 @@ namespace eosio { namespace chain {
    void transaction_context::finalize() {
       EOS_ASSERT( is_initialized, transaction_exception, "must first initialize" );
 
+      net_usage = ((net_usage + 7)/8)*8; // Round up to nearest multiple of word size (8 bytes)
+      auto now = fc::time_point::now();
+      trace->elapsed = now - start;
```
heifner, line 338: We want `elapsed` to include as much of the transaction time as possible. There are some db lookups and modifications after this line that should be included in the `elapsed` time.

plugins/producer_plugin/producer_plugin.cpp
linh2931: I did not see Greg's RAII suggestion while replying. Thanks @greg7mdp and @heifner. Will use RAII. That's much better.

greg7mdp: I think the `scoped_exit` usage is a bit awkward.

Could we add this class in libfc:

```
#pragma once

#include <utility>

namespace fc {

template<class Set, class Unset>
class scoped_set_unset_if {
public:
   scoped_set_unset_if(Set&& set, Unset &&unset, bool do_it = true) :
      unset_(std::forward<Unset>(unset))
      do_it_(do_it)
   {
      if (do_it_)
         std::forward<Set>(set)();
   }

   ~scoped_set_unset_if() {
      if (do_it_)
         std::forward<Unset>(unset_)();
   }

   scoped_set_unset_if(const scoped_set_unset_if&)            = delete;
   scoped_set_unset_if& operator=(const scoped_set_unset_if&) = delete;

private:
   Unset unset_;
   bool  do_it_;
};

} // namespace fc
``` 

and then we could just write:

```
fc::scoped_set_unset_if readonly([]() { chain.set_db_read_only_mode(); },
                                 []() { chain.unset_db_read_only_mode(); },
                                 trx->is_read_only());
```

linh2931: Thanks Greg for the detailed suggestion. `fc::scoped_exit ` is a common pattern in the code base. Instead of introducing more code, I think direct use of it is OK (PS. Setting and unsetting DB read only mode will be moved soon when multi-threading is implemented.

libraries/chain/apply_context.cpp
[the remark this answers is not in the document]
heifner: I thought these would be changed to `!is_transient()`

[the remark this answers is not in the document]
linh2931: Thanks! I missed it.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1971,7 +1972,8 @@ producer_plugin_impl::push_transaction( const fc::time_point& block_deadline,
 
    bool disable_subjective_enforcement = (api_trx && _disable_subjective_api_billing)
                                          || (!api_trx && _disable_subjective_p2p_billing)
-                                         || trx->is_dry_run();
+                                         || trx->is_dry_run()
+                                         || trx->is_read_only();
```
heifner, line 1976: This can be `trx->is_transient()`

heifner (approved):

heifner (changes requested):

libraries/chain/transaction_context.cpp
```diff
@@ -137,72 +137,84 @@ namespace eosio { namespace chain {
...
-         if( subjective_cpu_bill_us > 0 && (start + fc::microseconds(validate_account_cpu_limit) < _deadline) ) {
-            _deadline = start + fc::microseconds(validate_account_cpu_limit);
-            billing_timer_exception_code = tx_cpu_usage_exceeded::code_value;
+         // Explicit billed_cpu_time_us should be used, block_deadline will be maximum unless in test code
+         if( explicit_billed_cpu_time ) {
+            _deadline = block_deadline;
+            deadline_exception_code = deadline_exception::code_value;
+         } else {
+            deadline_exception_code = billing_timer_exception_code;
          }
```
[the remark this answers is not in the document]
linh2931: Thanks @heifner for your careful review! Please ignore the above emoji. It was clicked by mistake.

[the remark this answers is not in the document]
linh2931: Yes, we will add a new node operator limit for read-only transactions when multi-threaded is implemented shortly. That is why I leave on-chain `max_transaction_cpu_usage` and block cpu limit for the worst case protection for now.

On another note, do you think I should go back to the original separate `init_for_read_only_trx`, not polluting the main `init`?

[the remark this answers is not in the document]
heifner: Should this PR be targeted to a feature branch then? Not sure about merging this into `main` until it is complete. Maybe we should put a hard-coded limit for now. 150ms is a rather large amount of time.

No, I still think having this all in `init` is easier to follow.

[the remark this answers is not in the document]
linh2931: Read-only transactions can be a useful separate feature. I will add a hard-coded limit. It would be good to get into `main` to soak it.

I will keep all initializations in `init`.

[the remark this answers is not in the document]
heifner: The code is still duplicated.

libraries/chain/transaction_context.cpp
```diff
@@ -101,17 +106,21 @@ namespace eosio { namespace chain {
 
       net_limit = rl.get_block_net_limit();
 
-      objective_duration_limit = fc::microseconds( rl.get_block_cpu_limit() );
+      if( is_read_only() ) {
+         objective_duration_limit = read_only_trx_cpu_usage_limit;
```
heifner, line 110: Not sure we should classify this as an objective limit. It certainly will not be once it is made configurable.
Probably clearer to add another section like:
```
      if( is_read_only() ) {
         // Possibly limit deadline to node configurable read_only_trx_cpu_usage_limit 
         if( start + read_only_trx_cpu_usage_limit ) <= _deadline ) {
            _deadline = start + read_only_trx_cpu_usage_limit ;
            billing_timer_exception_code = tx_cpu_usage_exceeded::code_value;
         }
```
Down below.
Will need to also report that differently in https://github.com/AntelopeIO/leap/pull/581

libraries/chain/transaction_context.cpp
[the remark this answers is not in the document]
linh2931: I am so sorry for this merge error.

libraries/chain/transaction_context.cpp
linh2931: I will add a new cpu exceeded type after #581 is merged.

libraries/chain/transaction_context.cpp
```diff
@@ -125,92 +137,104 @@ namespace eosio { namespace chain {
...
-         if( trx_specified_cpu_usage_limit <= objective_duration_limit ) {
-            objective_duration_limit = trx_specified_cpu_usage_limit;
-            billing_timer_exception_code = tx_cpu_usage_exceeded::code_value;
-            _deadline = start + objective_duration_limit;
+         if( is_read_only() ) {
+            // Possibly limit deadline to optional limit set in transaction header
+            if( trx_specified_cpu_usage_limit <= read_only_trx_cpu_usage_limit ) {
+               billing_timer_exception_code = tx_cpu_usage_exceeded::code_value;
+               _deadline = start + trx_specified_cpu_usage_limit;
+            }
```
heifner, line 147: If trx specified `max_cpu_usage_ms` is lower than `read_only_trx_cpu_usage_limit` then it would be an objective limit. I think the code should be the same for read_only and normal for `max_cpu_usage_ms`.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: OK, but I still think that what I suggested is nicer.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -765,6 +768,8 @@ void producer_plugin::set_program_options(
           "Number of worker threads in producer thread pool")
          ("snapshots-dir", bpo::value<bfs::path>()->default_value("snapshots"),
           "the location of the snapshots directory (absolute path or relative to application data dir)")
+         ("max-read-only-transaction-time", bpo::value<int32_t>()->default_value(150),
```
heifner, line 771: Use `_max_read_only_transaction_time_ms` instead of `150`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -525,15 +528,15 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                                          bool return_failure_traces,
                                          next_function<transaction_trace_ptr> next) {
          chain::controller& chain = chain_plug->chain();
-         const auto max_trx_time_ms = _max_transaction_time_ms.load();
+         const auto max_trx_time_ms = ( trx_type == transaction_metadata::trx_type::read_only ) ? _max_read_only_transaction_time_ms : _max_transaction_time_ms.load();
```
heifner, line 531: `_max_read_only_transaction_time_ms` is not thread safe. It would need to be made `atomic` to be used here as this is called from the net_plugin thread pool. However, you need to access it here as you can use `max_trx_time_ms` of `-1` for read only transactions since `start_recover_keys` doesn't do anything for read only because read only do not have signatures.

heifner: `transaction::get_signature_keys` should not call `sig_digest` when `signatures.empty()`. This will be a slight optimization for read only transactions.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -765,6 +768,8 @@ void producer_plugin::set_program_options(
           "Number of worker threads in producer thread pool")
          ("snapshots-dir", bpo::value<bfs::path>()->default_value("snapshots"),
           "the location of the snapshots directory (absolute path or relative to application data dir)")
+         ("max-read-only-transaction-time", bpo::value<int32_t>()->default_value(150),
+          "Limits the maximum time (in milliseconds) that is allowed a pushed read-only transaction's code to execute before being considered invalid")
```
heifner, line 772: `Limits the maximum time (in milliseconds) a read-only transaction can execute before being considered invalid`

libraries/chain/transaction_context.cpp
```diff
@@ -138,53 +138,60 @@ namespace eosio { namespace chain {
...
       // Possibly limit deadline to subjective max_transaction_time
       if( max_transaction_time_subjective != fc::microseconds::maximum() && (start + max_transaction_time_subjective) <= _deadline ) {
          _deadline = start + max_transaction_time_subjective;
-         tx_cpu_usage_reason = billed_cpu_time_us > 0 ? tx_cpu_usage_exceeded_reason::speculative_executed_adjusted_max_transaction_time :
-                                                        tx_cpu_usage_exceeded_reason::node_configured_max_transaction_time;
+         tx_cpu_usage_reason = billed_cpu_time_us > 0 ?
+            tx_cpu_usage_exceeded_reason::speculative_executed_adjusted_max_transaction_time :
+            ( is_read_only() ?
+               tx_cpu_usage_exceeded_reason::node_configured_max_read_only_transaction_time :
+               tx_cpu_usage_exceeded_reason::node_configured_max_transaction_time );
```
heifner, line 194: Can you add a new test case to cover this. There are tests for all the others.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2007,7 +2020,22 @@ producer_plugin_impl::push_transaction( const fc::time_point& block_deadline,
...
+      chain.set_execution_mode(trx_execution_mode::DRY_RUN);
    auto trace = chain.push_transaction( trx, block_deadline, max_trx_time, prev_billed_cpu_time_us, false, sub_bill );
+   if( trx->is_transient() ) {
+      chain.set_execution_mode( trx_execution_mode::REGULAR );
+      set_execution_mode_scoped_exit.cancel();
+   }
+
    auto end = fc::time_point::now();
    push_result pr;
    if( trace->except ) {
```
heifner, line 2041: I wonder if we should track read-only/dry-run trx fail-time and success-time separately than regular trxs. See `account_failures::report`.

heifner (changes requested):

libraries/chain/controller.cpp
```diff
@@ -232,6 +232,7 @@ struct controller_impl {
    const chain_id_type             chain_id; // read by thread_pool threads, value will not be changed
    bool                            replaying = false;
    db_read_mode                    read_mode = db_read_mode::HEAD;
+   trx_execution_mode              execution_mode = trx_execution_mode::REGULAR;
```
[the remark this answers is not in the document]
linh2931: I was trying to streamline the deep-mind logging code for both dry-run and read-only transaction. Your suggestion is better, especially for upcoming multithreaded support. Will revert back. Thanks.

plugins/producer_plugin/producer_plugin.cpp
linh2931: Thanks for those two suggestions. Will change.

plugins/producer_plugin/producer_plugin.cpp
linh2931: I think we should separate them from regular transactions. Separate stats will be useful. Will add `dry_run_trx_success_time, dry_run_trx_success_num, dry_run_trx_fail_time, dry_run_trx_fail_num; read_only_trx_success_time, read_only_trx_success_num, read_only_trx_fail_time, read_only_trx_fail_num`.

heifner: I'm not convinced this is needed. If you do add it, seems like a separate `fc_dlog` with these details if any are `!= 0` would be good. Many nodes will not have any dry-run or read-only trxs.

linh2931: OK. We won't do it at the moment, not to add unnecessary code.

libraries/chain/authorization_manager.cpp
```diff
@@ -193,14 +194,14 @@ namespace eosio { namespace chain {
          p.last_updated = creation_time;
          p.auth         = std::move(auth);
 
-         if (auto dm_logger = _control.get_deep_mind_logger()) {
+         if (auto dm_logger = _control.get_deep_mind_logger(); dm_logger && !trx_is_transient) {
```
heifner, line 197: There are two `get_deep_mind_logger` methods. One on `controller` and one on `transaction_context`. It took me a bit to realize that `transaction_context::get_deep_mind_logger` returns null for transient where `controller::get_deep_mind_logger` does not. 

I would recommend either renaming one of them or maybe change `controller::get_deep_mind_logger` to take a `bool trx_is_transient` flag so that it is obvious they do something different.

heifner: Would actually prefer no default on that. In general I dislike default unless it causes a lot of code changes. In this case I think you need to pass in the bool on every call.

libraries/chain/authorization_manager.cpp
```diff
@@ -158,7 +158,7 @@ namespace eosio { namespace chain {
          p.last_updated = creation_time;
          p.auth         = auth;
 
-         if (auto dm_logger = _control.get_deep_mind_logger()) {
+         if (auto dm_logger = _control.get_deep_mind_logger(false)) {
```
heifner, line 161: We don't want to log this for dry-run transactions.

libraries/chain/authorization_manager.cpp
```diff
@@ -193,20 +194,20 @@ namespace eosio { namespace chain {
...
 
-   void authorization_manager::modify_permission( const permission_object& permission, const authority& auth ) {
+   void authorization_manager::modify_permission( const permission_object& permission, const authority& auth, bool trx_is_transient ) {
       for(const key_weight& k: auth.keys)
          EOS_ASSERT(k.key.which() < _db.get<protocol_state_object>().num_supported_key_types, unactivated_key_type,
            "Unactivated key type used when modifying permission");
 
       _db.modify( permission, [&](permission_object& po) {
-         auto dm_logger = _control.get_deep_mind_logger();
+         auto dm_logger = _control.get_deep_mind_logger(false);
```
heifner, line 210: This should use the passed in `trx_is_transient`.

libraries/chain/controller.cpp
```diff
@@ -295,9 +295,9 @@ struct controller_impl {
     blog( cfg.blocks_dir, cfg.prune_config ),
     fork_db( cfg.blocks_dir / config::reversible_blocks_dir_name ),
     wasmif( cfg.wasm_runtime, cfg.eosvmoc_tierup, db, cfg.state_dir, cfg.eosvmoc_config, !cfg.profile_accounts.empty() ),
-    resource_limits( db, [&s]() { return s.get_deep_mind_logger(); }),
+    resource_limits( db, [&s]() { return s.get_deep_mind_logger(false); }),
```
heifner, line 298: Seems like this should be a lambda that take `bool trx_is_transient` and pass that on to `controller.get_deep_mind_logger()`.

libraries/chain/controller.cpp
```diff
@@ -295,9 +295,9 @@ struct controller_impl {
...
     authorization( s, db ),
-    protocol_features( std::move(pfs), [&s]() { return s.get_deep_mind_logger(); } ),
+    protocol_features( std::move(pfs), [&s]() { return s.get_deep_mind_logger(false); } ),
```
heifner, line 300: Seems like this should be a lambda that take `bool trx_is_transient` and pass that on to `controller.get_deep_mind_logger()`.

libraries/chain/controller.cpp
```diff
@@ -378,7 +378,7 @@ struct controller_impl {
    }
 
    void dmlog_applied_transaction(const transaction_trace_ptr& t) {
-      if (auto dm_logger = get_deep_mind_logger()) {
+      if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 381: We do not want to log read-only or dry-run transaction traces. This likely points that we should include the transaction type in transaction_trace. The type would be useful here, but it might also be useful for other clients of transaction_trace to know if a transaction was a read-only, dry-run, etc.

libraries/chain/controller.cpp
```diff
@@ -706,7 +706,7 @@ struct controller_impl {
 
       protocol_features.init( db );
 
-      if (auto dm_logger = get_deep_mind_logger()) {
+      if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 709: We don't want to log dry-run transactions. I know `on_startup` does not log anything, but seems cleaner to never call dm_logger if trx is transient.

libraries/chain/controller.cpp
```diff
@@ -1033,7 +1033,7 @@ struct controller_impl {
       ram_delta += owner_permission.auth.get_billable_size();
       ram_delta += active_permission.auth.get_billable_size();
 
-      if (auto dm_logger = get_deep_mind_logger()) {
+      if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 1036: Same comment as above.

libraries/chain/controller.cpp
```diff
@@ -1203,7 +1203,7 @@ struct controller_impl {
    }
 
    int64_t remove_scheduled_transaction( const generated_transaction_object& gto ) {
-      if (auto dm_logger = get_deep_mind_logger()) {
+      if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 1206: We do not want to log this for dry-run transactions.

libraries/chain/controller.cpp
```diff
@@ -1331,7 +1331,7 @@ struct controller_impl {
          trace->except_ptr = std::current_exception();
          trace->elapsed = fc::time_point::now() - start;
 
-         if (auto dm_logger = get_deep_mind_logger()) {
+         if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 1334: We do not want to log this for dry-run transactions.

libraries/chain/controller.cpp
```diff
@@ -1667,7 +1667,7 @@ struct controller_impl {
 
       emit( self.block_start, head->block_num + 1 );
 
-      if (auto dm_logger = get_deep_mind_logger()) {
+      if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 1670: We do not want to log this for dry-run transactions.

libraries/chain/controller.cpp
```diff
@@ -1919,7 +1919,7 @@ struct controller_impl {
             EOS_ASSERT( bsp == head, fork_database_exception, "committed block did not become the new head in fork database");
          }
 
-         if (auto dm_logger = get_deep_mind_logger()) {
+         if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 1922: We do not want to log this for dry-run transactions.

libraries/chain/controller.cpp
```diff
@@ -2796,7 +2796,7 @@ void controller::preactivate_feature( const digest_type& feature_digest ) {
                ("digest", feature_digest)
    );
 
-   if (auto dm_logger = get_deep_mind_logger()) {
+   if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 2799: We do not want to log this for dry-run transactions.

libraries/chain/controller.cpp
```diff
@@ -3454,7 +3454,7 @@ void controller::add_to_ram_correction( account_name account, uint64_t ram_bytes
       } );
    }
 
-   if (auto dm_logger = get_deep_mind_logger()) {
+   if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 3457: We do not want to log this for dry-run transactions.

libraries/chain/controller.cpp
```diff
@@ -3607,7 +3615,7 @@ void controller_impl::on_activation<builtin_protocol_feature_t::replace_deferred
                ("name", itr->name)("adjust", itr->ram_correction)("current", current_ram_usage) );
       }
 
-      if (auto dm_logger = get_deep_mind_logger()) {
+      if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 3618: We do not want to log this for dry-run transactions.

libraries/chain/include/eosio/chain/authorization_manager.hpp
```diff
@@ -38,12 +38,13 @@ namespace eosio { namespace chain {
...
-                                                     time_point initial_creation_time = time_point()
+                                                     time_point initial_creation_time = time_point(),
+                                                     bool trx_is_transient = false
                                                    );
 
-         void modify_permission( const permission_object& permission, const authority& auth );
+         void modify_permission( const permission_object& permission, const authority& auth, bool trx_is_transient = false );
 
-         void remove_permission( const permission_object& permission );
+         void remove_permission( const permission_object& permission, bool trx_is_transient = false );
```
heifner, line 47: Can we remove the default `= false` on these two without too much churn?

libraries/chain/include/eosio/chain/resource_limits.hpp
```diff
@@ -78,7 +78,7 @@ namespace eosio { namespace chain {
          void set_block_parameters( const elastic_limit_parameters& cpu_limit_parameters, const elastic_limit_parameters& net_limit_parameters );
 
          void update_account_usage( const flat_set<account_name>& accounts, uint32_t ordinal );
-         void add_transaction_usage( const flat_set<account_name>& accounts, uint64_t cpu_usage, uint64_t net_usage, uint32_t ordinal );
+         void add_transaction_usage( const flat_set<account_name>& accounts, uint64_t cpu_usage, uint64_t net_usage, uint32_t ordinal, bool trx_is_transient = false );
```
heifner, line 82: Can we remove the default `= false` without too much churn?

libraries/chain/controller.cpp
```diff
@@ -2293,7 +2293,7 @@ struct controller_impl {
          ilog("switching forks from ${current_head_id} (block number ${current_head_num}) to ${new_head_id} (block number ${new_head_num})",
               ("current_head_id", head->id)("current_head_num", head->block_num)("new_head_id", new_head->id)("new_head_num", new_head->block_num) );
 
-         if (auto dm_logger = get_deep_mind_logger()) {
+         if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 2296: We do not want to log this for dry-run transactions.

libraries/chain/controller.cpp
```diff
@@ -2612,21 +2612,21 @@ struct controller_impl {
          trx.set_reference_block( self.head_block_id() );
       }
 
-      if (auto dm_logger = get_deep_mind_logger()) {
+      if (auto dm_logger = get_deep_mind_logger(false)) {
```
heifner, line 2615: We do not want to log this for dry-run transactions.

heifner (changes requested):

libraries/chain/authorization_manager.cpp
heifner: Note how the non-default argument made it much easier to see wrong use. I doubt I would have caught these new issues below if it was defaulted.

linh2931: Thanks! That is a good practice.

libraries/chain/authorization_manager.cpp
linh2931: I did not notice this version of create_permission is used in apply_eosio_updateauth. Thanks. Will change.

libraries/chain/authorization_manager.cpp
linh2931: Sorry, I missed this one.

libraries/chain/controller.cpp
linh2931: `dmlog_applied_transaction` is called in `push_scheduled_transaction` and `push_transaction`. read-only or dry-run transaction would not be a deferred transaction. In `push_transaction`, `dmlog_applied_transaction` is only called inside `if (!trx->is_transient())`. That's why I did not change `dmlog_applied_transaction`.

heifner: Ok, I see that. You might want to add a comment that this is not called for transient transactions.

libraries/chain/controller.cpp
linh2931: At startup, we don't know transaction types.

libraries/chain/controller.cpp
linh2931: `create_native_account` is only called during startup, at which time we don't know transaction types.

libraries/chain/controller.cpp
linh2931: `remove_scheduled_transaction` is not called for read-only or dry-run transactions.

libraries/chain/controller.cpp
linh2931: This is inside `push_scheduled_transaction`, which is not called for transient transactions.

libraries/chain/controller.cpp
linh2931: At `start_block`, we don't have transaction type information.

libraries/chain/controller.cpp
linh2931: At `commit_block`, we don't know transaction types.

libraries/chain/controller.cpp
linh2931: This is only called by `apply_context::schedule_deferred_transaction`

libraries/chain/controller.cpp
heifner: Sorry, I thought I was looking at transaction_context here.

libraries/chain/controller.cpp
linh2931: This is called inside `controller_impl::on_activation<builtin_protocol_feature_t::                    replace_deferred>()`, which is for deferred transactions.

libraries/chain/controller.cpp
heifner: Right, only called during database initialization. Nevermind.

libraries/chain/controller.cpp
heifner: Please add a comment that says this is only called from `push_scheduled_transaction`

libraries/chain/controller.cpp
linh2931: This is inside `maybe_switch_forks`, which is not transaction specific.

libraries/chain/controller.cpp
heifner: Yep, I got a bit copy/paste happy on my comments.

libraries/chain/controller.cpp
linh2931: This is inside `get_on_block_transaction`, which cannot be a transient transaction.

libraries/chain/controller.cpp
linh2931: I will add a comment to any `get_deep_mind_logger(false`).

libraries/chain/controller.cpp
heifner: Please add a comment indicating that.

heifner (approved):

libraries/chain/apply_context.cpp
```diff
@@ -1029,17 +1033,25 @@ int apply_context::db_end_i64( name code, name scope, name table ) {
...
+   }
 }
 
 uint64_t apply_context::next_recv_sequence( const account_metadata_object& receiver_account ) {
-   db.modify( receiver_account, [&]( auto& ra ) {
-      ++ra.recv_sequence;
-   });
-   return receiver_account.recv_sequence;
+   if ( trx_context.is_read_only() ) {
+      return receiver_account.recv_sequence + 1;
```
arhag, line 1048: Similar to above, we could just keep this as 0.

libraries/chain/resource_limits.cpp
```diff
@@ -220,7 +223,9 @@ void resource_limits_manager::add_pending_ram_usage( const account_name account,
    _db.modify( usage, [&]( auto& u ) {
       u.ram_usage += ram_delta;
 
-      if (auto dm_logger = _get_deep_mind_logger()) {
+      // add_pending_ram_usage is only called by transaction_context::schedule_transaction,
```
arhag, line 226: This comment isn't accurate. There are other ways `add_pending_ram_usage` can be called. It needs to have a `bool is_trx_transient` parameter plumbed in as well. In most places it is called, `false` can just be passed in as an argument for that new parameter. The only exception is from within `transaction_context::add_ram_usage` where `is_transient()` should be used instead.

Note that `transaction_context::add_ram_usage` is only directly called from `transaction_context::schedule_transaction` and `apply_context::add_ram_usage`. But that second usage of it is what allows it to be indirectly called from a bunch of places within `apply_context.cpp` (DB operations, and scheduling/cancelling deferred transactions) and `eosio_context.cpp` (native actions). And while DB operations and native operations are disallowed in read-only transactions, they can still be used by dry-run transactions which are considered transient.

Also note that scheduling/canceling deferred transactions also needs to be restricted from within a read-only transaction context.

libraries/chain/webassembly/privileged.cpp
```diff
@@ -17,14 +17,14 @@ namespace eosio { namespace chain { namespace webassembly {
    }
 
    void interface::preactivate_feature( legacy_ptr<const digest_type> feature_digest ) {
```
arhag, line 19: Read-only functions must not be allowed to call the `preactivate_feature` host function, along with all of the following ones:

- `set_resource_limits`
- `set_wasm_parameters_packed`
- `set_proposed_producers`
- `set_proposed_producers_ex`
- `set_blockchain_parameters_packed`
- `set_parameters_packed`
- `set_privileged`

arhag (changes requested):

libraries/chain/apply_context.cpp
```diff
@@ -419,13 +419,14 @@ void apply_context::execute_context_free_inline( action&& a ) {
...
 
-   if (auto dm_logger = control.get_deep_mind_logger()) {
+   if (auto dm_logger = control.get_deep_mind_logger(trx_context.is_transient())) {
       dm_logger->on_send_context_free_inline();
    }
 }
 
 
 void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, account_name payer, transaction&& trx, bool replace_existing ) {
+   EOS_ASSERT( !trx_context.is_read_only(), transaction_exception, "cannot schedule a readonly deferred transaction" );
```
arhag, line 429: A better error message would be `"cannot schedule a deferred transaction from within a readonly transaction"`.

libraries/chain/apply_context.cpp
```diff
@@ -626,10 +627,11 @@ void apply_context::schedule_deferred_transaction( const uint128_t& sender_id, a
 }
 
 bool apply_context::cancel_deferred_transaction( const uint128_t& sender_id, account_name sender ) {
+   EOS_ASSERT( !trx_context.is_read_only(), transaction_exception, "cannot cancel a readonly deferred transaction" );
```
arhag, line 630: A better error message would be `"cannot cancel a deferred transaction from within a readonly transaction"`.

libraries/chain/controller.cpp
```diff
@@ -1016,22 +1020,23 @@ struct controller_impl {
...
       ram_delta += active_permission.auth.get_billable_size();
 
-      if (auto dm_logger = get_deep_mind_logger()) {
+      // This is only called at startup, no transaction specific logging is possible
+      if (auto dm_logger = get_deep_mind_logger(false)) {
          dm_logger->on_ram_trace(RAM_EVENT_ID("${name}", ("name", name)), "account", "add", "newaccount");
       }
 
-      resource_limits.add_pending_ram_usage(name, ram_delta);
+      resource_limits.add_pending_ram_usage(name, ram_delta, false); // false for not doing dm logging
```
arhag, line 1039: This comment and the other similar ones you added for the other places `add_pending_ram_usage` does not seem correct.

Passing in `true` would avoid doing dm logging. But passing in `false` would do dm logging. I think we want to do dm logging in this (and the other) cases since that is what was done before. So the code isn't wrong, only the comment is.

arhag (approved):
I approved. But please fix the three remaining issues (two error strings, and the comments repeated at four different lines) before merging.

libraries/chain/controller.cpp
linh2931: Thank you @arhag so much for careful review! It occurred to me the comments were wrong late last night and was to correct them the first thing in the morning.

PR #563, opened 2022-12-12
Adds support for a 'state-dir' command line option in chain_plugin.

### LINKED ISSUE #433: Addition of a `state-dir` configuration option.
Currently all data output from nodeos uses the `data-dir`. This includes state, blocks, snapshots, state history, etc. Most of these also have specific parameters you can specify to split them out into their own locations, like `blocks-dir` or `state-history-dir `. 

If your goal is to create an isolated area for state (e.g. loading into tmpfs), you currently need to specify locations for each other data type (blocks, etc) to turn the `data-dir` into a state-exclusive folder. 

This request is for the creation of a `state-dir` parameter, that when specified, allows you to specify the file location of where the `state` folder is placed on the system.

greg7mdp:
Resolves #433 .

Adds support for a `state-dir` command line option, specifying the directory in which to store the state db.

### DIFF
--- a/plugins/producer_plugin/test/test_options.cpp
+++ b/plugins/producer_plugin/test/test_options.cpp
@@ -1,0 +1,57 @@
+#define BOOST_TEST_MODULE full_producer_trxs
+#include <boost/test/included/unit_test.hpp>
+
+#include <eosio/producer_plugin/producer_plugin.hpp>
+
+#include <eosio/testing/tester.hpp>
+
+#include <eosio/chain/genesis_state.hpp>
+#include <eosio/chain/thread_utils.hpp>
+#include <eosio/chain/trace.hpp>
+#include <eosio/chain/name.hpp>
+
+#include <appbase/application.hpp>
+
+using namespace eosio;
+using namespace eosio::chain;
+
+BOOST_AUTO_TEST_SUITE(program_options)
+
+BOOST_AUTO_TEST_CASE(state_dir) {
+   boost::filesystem::path temp = boost::filesystem::temp_directory_path() / boost::filesystem::unique_path();
+   boost::filesystem::path state_dir = temp / "state";
+   boost::filesystem::path custom_state_dir = temp / "custom_state_dir";
+      
+   try {
+      std::promise<std::tuple<producer_plugin*, chain_plugin*>> plugin_promise;
+      std::future<std::tuple<producer_plugin*, chain_plugin*>> plugin_fut = plugin_promise.get_future();
+      std::thread app_thread( [&]() {
+         fc::logger::get(DEFAULT_LOGGER).set_log_level(fc::log_level::debug);
+         std::vector<const char*> argv =
+               {"test", "--data-dir", temp.c_str(),  "--state-dir", custom_state_dir.c_str(), "--config-dir", temp.c_str(),
+                "-p", "eosio", "-e", "--max-transaction-time", "475", "--disable-subjective-billing=true" };
+         appbase::app().initialize<chain_plugin, producer_plugin>( argv.size(), (char**) &argv[0] );
+         appbase::app().startup();
+         plugin_promise.set_value(
+               {appbase::app().find_plugin<producer_plugin>(), appbase::app().find_plugin<chain_plugin>()} );
+         appbase::app().exec();
+      } );
+
+      auto[prod_plug, chain_plug] = plugin_fut.get();
+      [[maybe_unused]] auto chain_id = chain_plug->get_chain_id();
+
+      // check that "--state-dir" option was taken into account
+      BOOST_CHECK(  exists( custom_state_dir ));
+      BOOST_CHECK( !exists( state_dir ));
+      
+      appbase::app().quit();
+      app_thread.join();
+
+   } catch ( ... ) {
+      bfs::remove_all( temp );
+      throw;
+   }
+   bfs::remove_all( temp );
+}
+
+BOOST_AUTO_TEST_SUITE_END()
--- a/plugins/producer_plugin/test/test_options.cpp
+++ b/plugins/producer_plugin/test/test_options.cpp
@@ -18,40 +18,40 @@ using namespace eosio::chain;
 BOOST_AUTO_TEST_SUITE(program_options)
 
 BOOST_AUTO_TEST_CASE(state_dir) {
-   boost::filesystem::path temp = boost::filesystem::temp_directory_path() / boost::filesystem::unique_path();
-   boost::filesystem::path state_dir = temp / "state";
-   boost::filesystem::path custom_state_dir = temp / "custom_state_dir";
+   fc::temp_directory temp;
+   auto temp_dir = temp.path();
+   auto state_dir = temp.path() / "state";
+   auto custom_state_dir = temp.path() / "custom_state_dir";
+
+   auto temp_dir_str = temp_dir.string();
+   auto custom_state_dir_str = custom_state_dir.string();
       
-   try {
-      std::promise<std::tuple<producer_plugin*, chain_plugin*>> plugin_promise;
-      std::future<std::tuple<producer_plugin*, chain_plugin*>> plugin_fut = plugin_promise.get_future();
-      std::thread app_thread( [&]() {
-         fc::logger::get(DEFAULT_LOGGER).set_log_level(fc::log_level::debug);
-         std::vector<const char*> argv =
-               {"test", "--data-dir", temp.c_str(),  "--state-dir", custom_state_dir.c_str(), "--config-dir", temp.c_str(),
-                "-p", "eosio", "-e", "--max-transaction-time", "475", "--disable-subjective-billing=true" };
-         appbase::app().initialize<chain_plugin, producer_plugin>( argv.size(), (char**) &argv[0] );
-         appbase::app().startup();
-         plugin_promise.set_value(
-               {appbase::app().find_plugin<producer_plugin>(), appbase::app().find_plugin<chain_plugin>()} );
-         appbase::app().exec();
-      } );
+   std::promise<std::tuple<producer_plugin*, chain_plugin*>> plugin_promise;
+   std::future<std::tuple<producer_plugin*, chain_plugin*>> plugin_fut = plugin_promise.get_future();
+   std::thread app_thread( [&]() {
+      fc::logger::get(DEFAULT_LOGGER).set_log_level(fc::log_level::debug);
+      std::vector<const char*> argv =
+         {"test",
+          "--data-dir",   temp_dir_str.c_str(),
+          "--state-dir",  custom_state_dir_str.c_str(),
+          "--config-dir", temp_dir_str.c_str(),
+          "-p", "eosio", "-e", "--max-transaction-time", "475", "--disable-subjective-billing=true" };
+      appbase::app().initialize<chain_plugin, producer_plugin>( argv.size(), (char**) &argv[0] );
+      appbase::app().startup();
+      plugin_promise.set_value(
+         {appbase::app().find_plugin<producer_plugin>(), appbase::app().find_plugin<chain_plugin>()} );
+      appbase::app().exec();
+   } );
 
-      auto[prod_plug, chain_plug] = plugin_fut.get();
-      [[maybe_unused]] auto chain_id = chain_plug->get_chain_id();
+   auto[prod_plug, chain_plug] = plugin_fut.get();
+   [[maybe_unused]] auto chain_id = chain_plug->get_chain_id();
 
-      // check that "--state-dir" option was taken into account
-      BOOST_CHECK(  exists( custom_state_dir ));
-      BOOST_CHECK( !exists( state_dir ));
+   // check that "--state-dir" option was taken into account
+   BOOST_CHECK(  exists( custom_state_dir ));
+   BOOST_CHECK( !exists( state_dir ));
       
-      appbase::app().quit();
-      app_thread.join();
-
-   } catch ( ... ) {
-      bfs::remove_all( temp );
-      throw;
-   }
-   bfs::remove_all( temp );
+   appbase::app().quit();
+   app_thread.join();
 }
 
 BOOST_AUTO_TEST_SUITE_END()
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -519,7 +519,7 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
          }
       } else {
          EOSC_ASSERT( !tx_unpack_data, "ERROR: --unpack-action-data not supported with --return-packed" );
-         return fc::variant(packed_transaction(trx, compression));
+        return fc::variant(packed_transaction(trx, compression));
       }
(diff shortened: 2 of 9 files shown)


programs/cleos/main.cpp
```diff
@@ -519,9 +519,11 @@ fc::variant push_transaction( signed_transaction& trx, const std::vector<public_
          }
       } else {
          EOSC_ASSERT( !tx_unpack_data, "ERROR: --unpack-action-data not supported with --return-packed" );
-        return fc::variant(packed_transaction(trx, compression));
+         return fc::variant(packed_transaction(trx, compression));
       }
    }
+   assert(0);
+   return {}; // fix warning: control reaches end of non-void function [-Wreturn-type]
```
spoonincode, line 526: What compiler is complaining here? It doesn't seem any of the builds in CI nor a local gcc12 complains, and it seems like `Wreturn-type` is enabled. Is this the best way to silence a warning such as this? Is `__builtin_unreachable()` any better or worse?

spoonincode:
I don't think appbase needs to be aware about `state-dir`, just like it is not aware of `block-dir`, `state-history-dir`, `trace-dir`, etc today. It seems like this should be handled as another configuration local to `chain_plugin`.

programs/cleos/main.cpp
greg7mdp: My compiler (on ubuntu 22.04.1) is gcc 11.3.0.
I don't think `__builtin_unreachable()`  is a good alternative. First it is an optimization pragma, not an assertion. And if some future code change allows the code to reach here, we should have a valid return statement rather that returning some random value from the stack.

spoonincode: oh, I see, you don't get the warning until adding the `assert(0)`

greg7mdp:
> I don't think appbase needs to be aware about `state-dir`, just like it is not aware of `block-dir`, `state-history-dir`, `trace-dir`, etc today. It seems like this should be handled as another configuration local to `chain_plugin`.

The issue has the following comment from Steven:
`Make sure to update resource monitor as part of the scope of this work.`
So it seems to imply that the resource monitor plugin should also be aware of `state_dir`.

I think the change in appbase ensures that this is defined in one location consistently. This could become important if we added an API `set_default_state_dir` similar to the `set_default_data_dir` in appbase. 

What do you think is the best way to go?

programs/cleos/main.cpp
greg7mdp: No, I did get the warning, which was why I assed the `assert()` and the `return`

spoonincode:
appbase (especially while it continues to be a submodule) is not intended to be leap specific; something like state-dir _is_ leap specific -- really it's even just nodeos specific (keosd wouldn't use it, but it uses appbase).

For resource monitoring,
https://github.com/AntelopeIO/leap/blob/4c7c07c13e88849effb1c2a8e70942fb053925bc/plugins/chain_plugin/chain_plugin.cpp#L565-L568
So no matter how `my->chain_config->state_dir` gets populated it'll be monitored.

programs/cleos/main.cpp
greg7mdp: When removing both lines (the assert and the return):

```
[2/3] /usr/bin/c++ -DBOOST_ALL_NO_LIB -DEOSIO_EOS_VM_JIT_RUNTIME_ENABLED -DEOSIO_EOS_VM_OC_RUNTIME_ENABLED -DEOSIO_EOS_VM_RUNTIME_ENABLED -DEOS_VM_SOFTFLOAT -DOPENSSL_API_COMPAT=0x10100000L -DOPENSSL_NO_DEPRECATED -I/home/greg/github/enf/leap/build_debug/programs/cleos -I/home/greg/github/enf/leap/programs/cleos -I/home/greg/github/enf/leap/libraries/appbase/include -I/home/greg/github/enf/leap/libraries/version/include -I/home/greg/github/enf/leap/libraries/cli11/include -I/home/greg/github/enf/leap/plugins/chain_api_plugin/include -I/home/greg/github/enf/leap/plugins/chain_plugin/include -I/home/greg/github/enf/leap/plugins/chain_plugin/../chain_interface/include -I/home/greg/github/enf/leap/plugins/chain_plugin/../../libraries/appbase/include -I/home/greg/github/enf/leap/plugins/chain_plugin/../producer_plugin/include -I/home/greg/github/enf/leap/plugins/chain_plugin/../resource_monitor_plugin/include -I/home/greg/github/enf/leap/plugins/producer_plugin/include -I/home/greg/github/enf/leap/plugins/producer_plugin/../chain_interface/include -I/home/greg/github/enf/leap/plugins/http_client_plugin/include -I/home/greg/github/enf/leap/libraries/chain/include -I/home/greg/github/enf/leap/build_debug/libraries/chain/include -I/home/greg/github/enf/leap/libraries/chain/../wasm-jit/Include -I/home/greg/github/enf/leap/libraries/libfc/include -I/home/greg/github/enf/leap/libraries/libfc/libraries/ff/libff/.. -I/home/greg/github/enf/leap/libraries/libfc/secp256k1/secp256k1 -I/home/greg/github/enf/leap/libraries/libfc/secp256k1/secp256k1/include -I/home/greg/github/enf/leap/libraries/chainbase/include -I/home/greg/github/enf/leap/libraries/chain/../rapidjson/include -I/home/greg/github/enf/leap/libraries/wasm-jit/Source/Runtime/../../../chain/include -I/home/greg/github/enf/leap/libraries/softfloat/source/include -I/home/greg/github/enf/leap/libraries/softfloat/source/8086-SSE -I/home/greg/github/enf/leap/libraries/softfloat/build/Linux-x86_64-GCC -I/home/greg/github/enf/leap/libraries/builtins -I/home/greg/github/enf/leap/libraries/builtins../softfloat/source/include -I/home/greg/github/enf/leap/libraries/eos-vm/include -I/home/greg/github/enf/leap/libraries/eos-vm/external/softfloat/source/include -I/home/greg/github/enf/leap/plugins/signature_provider_plugin/include -I/home/greg/github/enf/leap/plugins/resource_monitor_plugin/include -I/home/greg/github/enf/leap/plugins/http_plugin/include -Wall -g -fdiagnostics-color=always -Wno-register -std=gnu++17 -MD -MT programs/cleos/CMakeFiles/cleos.dir/main.cpp.o -MF programs/cleos/CMakeFiles/cleos.dir/main.cpp.o.d -o programs/cleos/CMakeFiles/cleos.dir/main.cpp.o -c /home/greg/github/enf/leap/programs/cleos/main.cpp
/home/greg/github/enf/leap/programs/cleos/main.cpp: In function ‘fc::variant push_transaction(eosio::chain::signed_transaction&, const std::vector<fc::crypto::public_key>&)’:
/home/greg/github/enf/leap/programs/cleos/main.cpp:525:1: warning: control reaches end of non-void function [-Wreturn-type]
  525 | }
      | ^

❯ /usr/bin/c++ --version                                                                                                   20s 
c++ (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
Copyright (C) 2021 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

greg7mdp:
Sounds good, than you @spoonincode , I'll update my change.

programs/cleos/main.cpp
spoonincode: The latest ubuntu22 build on master doesn't seem to have a warning there, so not sure what is causing the difference, it's possibly Debug vs Release related
https://github.com/AntelopeIO/leap/actions/runs/3664388649/jobs/6194769231

greg7mdp: Oh yes, that is probably the reason, I am building in debug mode.
But regardless of  why the warning is obtained, I don't think it hurts to address it.

linh2931: `assert` is only enabled in debug. If control ever reaches here, it is a bug. Why not just print out the error and exit?

heifner: In general we use `FC_ASSERT` or `EOS_ASSERT`

greg7mdp: I can make the change to `FC_ASSERT` or `EOS_ASSERT`, but I am a bit surprised that these throw an exception.

I believe throwing exception is appropriate for conditions that are expected to happen in exceptional conditions (for example memory allocation failing). In these cases we throw an exception to 'abort' the current action, with the hope that the exception will be caught and that the stack unrolling will free some of the resource crunch causing the exception. It may be viewed as a normal program occurence.

However sometimes there are conditions that logically should never happen (loop invariants for example, preconditions, etc...), and you feel so strongly about that that you assert that condition in the code. You effectively state, this will **always** be true, and if you are wrong you want to be immediately notified in the loudest and most immediate way. 
In those cases you do *not* want to throw an exception, that can be gracefully caught at a higher level and silenced.

Anyways, that is how I understand asserts(). Just thought I'd mention it.

spoonincode: `assert()` is a noop in release builds (`NDEBUG`), that is pretty sneaky and not desirable for many places we'd use `FC_ASSERT` as a "this should never happen"

greg7mdp: imho `FC_ASSERT` has a better chance of being ignored as the throw can easily be caught and be invisible. That is why for a condition that "should never happen" a hard termination is better. 
And because it should never happen, it should have no cost in release mode, not even a test and code branch.

heifner: We currently do not build/run debug mode in CI/CD.

greg7mdp: But developers do run the code under the debugger, don't they? And they run the unit tests with the debug build as well (I do). That should be enough to trigger the asserts if they fail I think.
If not running CI/CD on a debug build is the only objection, then we definitely should indeed change this. The benefit is that we free the release build from testing conditions that will never occur, and empower the developer to add as many asserts he/she wants, knowing there is no cost in release builds.

greg7mdp: @spoonincode @heifner do you still think I should change the `assert()` to `FC_ASSERT()`?

spoonincode: The
```c++
   assert(0);
   return {};
```
isn't an established pattern in Leap, so it doesn't seem appropriate here imo. `FC_ASSERT()` is more in line with common patterns. (I also generally don't like adding cruft to silence spurious warnings such as this, however I can sympathize with the angle that this is a long function that's hard to reason through if all branches return so it ought to have the clear assertion)

linh2931: I vote with FC_ASSERT.

greg7mdp: Thank you @spoonincode and @linh2931 . I had a long discussion yesterday evening with Areg and we arrived at a different conclusion. Areg suggested we talk about this in the engineering meeting on Tuesday, and I have added this topic to the agenda.
So for now, I might just remove this change (it was only to address a compiler warning) until we come to an agreement.

plugins/producer_plugin/test/test_options.cpp
```diff
@@ -0,0 +1,57 @@
...
+      BOOST_CHECK( !exists( state_dir ));
+      
+      appbase::app().quit();
+      app_thread.join();
+
+   } catch ( ... ) {
+      bfs::remove_all( temp );
+      throw;
+   }
+   bfs::remove_all( temp );
```
spoonincode, line 54: Using `fc::temp_directory` will alleviate the need to do manual cleanup

plugins/producer_plugin/test/test_options.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <eosio/chain/name.hpp>
+
+#include <appbase/application.hpp>
+
+using namespace eosio;
+using namespace eosio::chain;
+
+BOOST_AUTO_TEST_SUITE(program_options)
+
+BOOST_AUTO_TEST_CASE(state_dir) {
```
heifner, line 20: Instead of adding a complete new test executable for this. Can we not just add a `--state-dir` option to the `test_trx_full` test and verify it is honored there. Seems like a large overhead for such a small test.

arhag: I think we eventually really need to reorganize a lot of our tests (though not for this PR). For example, some of these tests (and perhaps the test in this PR is also the case) are really trying to bring up a minimal Leap environment (with appbase and plugins) for testing but without having to resort to a Python integration test that actually leverages the Leap binaries. Those kinds of tests don't seem to belong as unit tests of any particular plugin (often these tests will need to bring up the minimum required plugins of the chain plugin and producer plugin), so I am not sure why we should put the tests in the directory of one of the plugins.

Even for tests isolated to unit testing the functionality of the just one plugin, why do we need multiple executables rather than combining them together into one executable as we do with our other unit tests.

Anyway, I'm not suggesting such major reorganization to be done as part of this PR. But to me it seems like what we really should do is move this test to the chain_plugin directory for now. Then later either integrate those chain_plugin tests together into a collection of test suites built into one executable (similar to the unit_test executable), or if appropriate move them out of the plugins and put them somewhere else entirely (again probably as one executable).

What are your thoughts on that approach @heifner?

heifner: Discussed in our meeting.

The current issue with these minimal Leap environment (with appbase and plugins) is that appbase does not currently support teardown/startup which is why there is only one test per executable.

For this PR:
Move Leap env tests that include appbase into `tests/xxx` where `xxx` name is up for suggestions.

Future work:
- Enhance appbase to allow for teardown/startup. Currently you can only run one test per executable that does a plugin startup/shutdown because of appbase limitation.
- Merge existing plugin/tests into one executable to reduce link time and disk usage.

arhag: I created an issue to track the limitation in appbase that prevents having more than one test per executable: https://github.com/AntelopeIO/leap/issues/585.

We discussed how we cannot just move to the chain_plugin directory, so are options are either keeping it here or moving it into `tests`. Ultimately, it should be moved into `tests`. But we have more things to figure out regarding how we wish to organize our tests and we don't want that to hold up this PR indefinitely. So we agreed that for now it can exist in the producer_plugin directory and will need to be moved later (along with many other tests) as part of that reorganization effort (tracked in https://github.com/AntelopeIO/leap/issues/586).

greg7mdp: After discussing with Areg, we'll leave it as is for now, but we plan on creating a new directory for `system` level tests, and address the appbase 1 test instance per executable limitation.

programs/cleos/main.cpp
greg7mdp: After talking with Areg, I removed the change, there may be a better resolution for this warning by rearranging the code.

plugins/producer_plugin/test/test_options.cpp
greg7mdp: Thanks @spoonincode , I have made this change (and after checking with Areg I also did the same update in `test_trx_full.cpp`).

heifner (approved):

arhag (approved):

PR #575, opened 2022-12-15
Small Misc python test improvements

heifner:
- Adds better error output on failures.
- Adds `-f` force option to `pushMessage`
- Sets `keosd` to have very large timeout so when `--keep-running` is used the wallet doesn't auto-lock.

### DIFF
--- a/tests/TestHarness/Node.py
+++ b/tests/TestHarness/Node.py
@@ -821,8 +821,9 @@ class Node(object):
         except subprocess.CalledProcessError as ex:
             if not shouldFail:
                 end=time.perf_counter()
+                out=ex.output.decode("utf-8")
                 msg=ex.stderr.decode("utf-8")
-                Utils.Print("ERROR: Exception during set contract.  cmd Duration: %.3f sec.  %s" % (end-start, msg))
+                Utils.Print("ERROR: Exception during set contract. stderr: %s.  stdout: %s.  cmd Duration: %.3f sec." % (msg, out, end-start))
                 return None
             else:
                 retMap={}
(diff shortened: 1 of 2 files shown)


oschwaldp-oci (approved):

linh2931 (approved):

tests/TestHarness/Node.py
```diff
@@ -821,8 +821,9 @@ def publishContract(self, account, contractDir, wasmFile, abiFile, waitForTransB
         except subprocess.CalledProcessError as ex:
             if not shouldFail:
                 end=time.perf_counter()
+                out=ex.output.decode("utf-8")
                 msg=ex.stderr.decode("utf-8")
-                Utils.Print("ERROR: Exception during set contract.  cmd Duration: %.3f sec.  %s" % (end-start, msg))
+                Utils.Print("ERROR: Exception during set contract. stderr: %s.  stdout: %s.  cmd Duration: %.3f sec." % (msg, out, end-start))
```
greg7mdp, line 826: Would it be clearer to add quotes for the error strings?
```
Utils.Print("ERROR: Exception during set contract. stderr: \"%s\". stdout: \"%s.\" cmd Duration: %.3f sec." % (msg, out, end-start))
```

heifner: They are often multi-line. Might make sense to put them on separate lines with `----` or some such distinction. This follows existing pattern of printing them out. Maybe can get that cleaned up in a future PR.

greg7mdp: When they were printed at the end, it may have been clearer than printing these before the `command duration`.

PR #580, opened 2022-12-16
revert to previous appbase version

greg7mdp:


arhag (approved):
https://github.com/AntelopeIO/leap/pull/539 contained an erroneous submodule update of appbase. This PR fixes that by reverting it back. Because of that, the CICD submodule regression check is expected to fail.

PR #599, opened 2022-12-26
Split ship log

huangminghuang:
This PR backports https://github.com/EOSIO/eos/pull/9277 and https://github.com/EOSIO/eos/pull/9821 to support supports the trace and chain state log spliting in state history plugin. 

Some new options for state history plugin are added:

- state-history-stride: split the state history log files when the block number is the multiple of the stride. When the stride is reached, the current history log and index will be renamed '*-history--.log/index' and a new current history log and index will be created with the most recent blocks. All files following this format will be used to construct an extended history log.
- max-retained-history-files: the maximum number of state history log file groups to retain so that the blocks in those files can be queried. When the number is reached, the oldest log files would be moved to archive dir or deleted if the archive dir is empty. The retained log files should not be manipulated by users.
- state-history-retained-dir: the location of the state history retained directory (absolute path or relative to state-history dir). 
- state-history-archive-dir: the location of the state history archive directory (absolute path or relative to state-history dir). If the value is empty string, log files beyond the retained limit will be deleted. All files in the archive directory are completely under user's control, i.e. they won't be accessed by nodeos anymore.

### DIFF
--- a/libraries/chain/include/eosio/chain/log_catalog.hpp
+++ b/libraries/chain/include/eosio/chain/log_catalog.hpp
@@ -1,0 +1,283 @@
+#pragma once
+#include <boost/container/flat_map.hpp>
+#include <boost/filesystem/path.hpp>
+#include <boost/iostreams/device/mapped_file.hpp>
+#include <fc/io/cfile.hpp>
+#include <fc/io/datastream.hpp>
+#include <regex>
+
+namespace eosio {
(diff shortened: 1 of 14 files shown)


heifner:
Conflicts resolved in #628 which is probably the best place to start/continue review of these changes.

huangminghuang:
Superseded by PR #628

PR #603, opened 2022-12-29
Add timer on http listener to avoid fast looping on accept

### LINKED ISSUE #471: http_plugin needs cleaner behavior when `accept()` fails with `too_many_files_open`
When http_plugin's `accept()` fails with `too_many_files_open` the http threads enter an event storm consuming 100% CPU until the condition clears (i.e. the `accept()` completes or the "overflowing" pending connection is terminated). This behavior needs to be improved so these threads don't run away in this condition. I would recommend using some sort of backoff for retrying the `async_accept()` when it completes with `too_many_files_open`. I would also consider modifying the code to not call `async_accept()` again unless the prior call completed with either no error or `too_many_files_open` (there may be other ways the acceptor can enter a failed state).

Since the main thread appears unaffected, and the event storm is cleared properly when the condition is no longer present, I'm not classifying this a security defect. But due to its ugliness I'm initially classifying it as a defect.

greg7mdp:
Resolves #471.

The issue is that when we reach the limit of open files, the `accept()` call returns `boost::system::errc::too_many_files_open`.

When this happens, we log an error and immediately loop on `self->do_accept();` which will call `accept()` again and fails with the same error.

This creates a tight loop using 100% of the CPU, until a file handle is released and the `accept()` call succeds.

This proposed fix is to add a short timer, which gives a chance for the accumulation of requests to be processed, so the next time we call `accept()` it is likely to succeed.

This was not easy to test. I tried modifying the `http_plugin_test.py` to create many requests simultaneously (using the `get_info` request), but I think the requests were processed by nodeos faster than the python program sent them, to I could never reach the `too_many_files_open` error.

In order to slow down the processing, I introduced a delay in chain_plugin.cpp:

```c++
 #include <fc/variant.hpp>
 #include <signal.h>
 #include <cstdlib>
+#include <thread>
 
 // reflect chainbase::environment for --print-build-info option
 FC_REFLECT_ENUM( chainbase::environment::os_t,
@@ -1237,7 +1238,7 @@ std::string itoh(I n, size_t hlen = sizeof(I)<<1) {
 
 read_only::get_info_results read_only::get_info(const read_only::get_info_params&, const fc::time_point&) const {
    const auto& rm = db.get_resource_limits_manager();
-
+   std::this_thread::sleep_for(std::chrono::milliseconds(50));
    return {
       itoh(static_cast<uint32_t>(app().version())),
       db.get_chain_id(),
```

With that change, I was able to use the `http_plugin_test.py` with the addition of a timer:

```
@@ -44,12 +44,10 @@ try:
         cmdError("launcher")
         errorExit("Failed to stand up eos cluster.")
 
-    Print("Getting cluster info")
-    cluster.getInfos()
+    time.sleep(90)
     testSuccessful = True
 finally:
     TestHelper.shutdown(cluster, None, testSuccessful, killEosInstances, True, keepLogs, killAll, dumpErrorDetails)
 ```

As well as a new python script to send a bunch of http requests at once (Note: I couldn't use the python `requests` or `url3` libraries as they cache the http connections for performance reasons):

```
#!/usr/bin/env python3

from multiprocessing.dummy import Pool as ThreadPool
#import urllib3
import resource
import socket
import http.client

###############################################################
# request_storm.py
#
###############################################################

def getInfos():
    try:
        conn = http.client.HTTPConnection("localhost:8888")
        conn.request("GET", "/v1/chain/get_info")
        response = conn.getresponse()
        conn.close()
        return response
    except:
        return []

try:
    if True:
        new_limit = 4096
        print("Default allowance: %s, %s open files" % resource.getrlimit(resource.RLIMIT_NOFILE))
        resource.setrlimit(resource.RLIMIT_NOFILE, (new_limit, new_limit))
        print("Updated allowance: %s, %s open files" % resource.getrlimit(resource.RLIMIT_NOFILE))
    
    num_conn = 1200
    pool = ThreadPool(num_conn)
    
    futures = []
    for x in range(num_conn):
        futures.append(pool.apply_async(getInfos, []))
        
    result = []
    for future in futures[0:num_conn]:
        result.append(future.get())
        
    print(result[0])
    print("Received %s responses" % len(futures))
finally:
    print("test Done")
    
exit(0)
```

Using these modifications and scripts, I was able to confirm the 100% cpu usage before this fix, and a much lesser (1-3%) CPU usage after the fix. The total time for processing connections was the same in both cases.

### CODE REVIEW

File: plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp [hunk 1]
```diff
@@ -28,6 +28,8 @@ class beast_http_listener : public std::enable_shared_from_this<beast_http_liste
    typename protocol_type::acceptor acceptor_;
    socket_type socket_;
 
+   boost::asio::deadline_timer timer_;
```
heifner, line 31: I think a more descriptive name would be good.

plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp
```diff
@@ -107,7 +109,13 @@ class beast_http_listener : public std::enable_shared_from_this<beast_http_liste
       auto self = this->shared_from_this();
       acceptor_.async_accept(socket_, [self](beast::error_code ec) {
          if(ec) {
-            fail(ec, "accept", self->plugin_state_->logger, "closing connection");
+            if (ec == boost::system::errc::too_many_files_open) {
+               fail(ec, "accept", self->plugin_state_->logger, "too many files open - waiting 500ms");
+               self->timer_.expires_from_now(boost::posix_time::milliseconds(500));
```
heifner, line 114: `boost::asio::deadline_timer` is not thread safe. The `async_accept` handler is called from the thread pool executor.

heifner: I wonder if a simple escalating sleep amount would be fine here with a max of say 500ms.

heifner (changes requested):

plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp
greg7mdp: > `boost::asio::deadline_timer` is not thread safe. The `async_accept` handler is called from the thread pool executor.

I was thinking, instead of `boost::asio::deadline_timer`, maybe I could just use `std::this_thread::sleep_for(std::chrono::milliseconds(500));`

About the thread safety remark, I think we have only one listener object (for each type) waiting on accept(), right?

greg7mdp: > I wonder if a simple escalating sleep amount would be fine here with a max of say 500ms.

I don't see a good benefit if adding complexity with an escalating delay. I think because we reach a state where all file handles are in use, it means that we are accepting connections faster than we are able to process the requests from these connections. So pausing the accept for 500ms to give nodeos a chance to clear the backlog a bit doesn't seem excessive to me, and valid flow control. But maybe I'm missing a subtelty here?

heifner: Ah, I think you are correct about the safety remark. Never mind.

plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp
```diff
@@ -36,7 +38,7 @@ class beast_http_listener : public std::enable_shared_from_this<beast_http_liste
    beast_http_listener& operator=(const beast_http_listener&) = delete;
    beast_http_listener& operator=(beast_http_listener&&) = delete;
 
-   beast_http_listener(std::shared_ptr<http_plugin_state> plugin_state) : is_listening_(false), plugin_state_(std::move(plugin_state)), acceptor_(plugin_state_->thread_pool->get_executor()), socket_(plugin_state_->thread_pool->get_executor()) {}
+   beast_http_listener(std::shared_ptr<http_plugin_state> plugin_state) : is_listening_(false), plugin_state_(std::move(plugin_state)), acceptor_(plugin_state_->thread_pool->get_executor()), socket_(plugin_state_->thread_pool->get_executor()), timer_(plugin_state_->thread_pool->get_executor(), boost::posix_time::seconds(0)) {}
```
spoonincode, line 41: Can just use the single argument ctor

plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp
spoonincode: Use std::chrono instead of boost

plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp
```diff
@@ -107,7 +109,13 @@ class beast_http_listener : public std::enable_shared_from_this<beast_http_liste
       auto self = this->shared_from_this();
       acceptor_.async_accept(socket_, [self](beast::error_code ec) {
          if(ec) {
-            fail(ec, "accept", self->plugin_state_->logger, "closing connection");
+            if (ec == boost::system::errc::too_many_files_open) {
+               fail(ec, "accept", self->plugin_state_->logger, "too many files open - waiting 500ms");
+               self->timer_.expires_from_now(boost::posix_time::milliseconds(500));
+               self->timer_.wait();
```
spoonincode, line 115: This needs to be an `async_wait()` -- the synchronous `wait()` is going to block the thread for the duration of the timeout which is a problem especially if running only a single http thread

spoonincode (changes requested):
As the issue mentions, I'd really recommend changing the logic to
```
if(ec == too_many_files_open)
  //do the delay
else if(ec)
  // do nothing
else
  // create session & call do_accept() again
```

I realize this is not what the current code does (it still calls `do_accept()` even on error), but general asio best practice is to avoid reentering if there is an error.

plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp
greg7mdp: Actually, what we want to do is block the thread that does call `accept()`, so that it doesn't accept any new connections for 500ms. Can you elaborate on why doing a synchronous `wait()` would be an issue?

greg7mdp:
> As the issue mentions, I'd really recommend changing the logic to
> 
> ```
> if(ec == too_many_files_open)
>   //do the delay
> else if(ec)
>   // do nothing
> else
>   // create session & call do_accept() again
> ```
> 
> I realize this is not what the current code does (it still calls `do_accept()` even on error), but general asio best practice is to avoid reentering if there is an error.

Does this mean that we want to give up on accepting new connections in http_plugin the first time that we have an error that is not `too_many_files_open`?

plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp
spoonincode: Blocking the thread blocks the thread from performing other work. In the case of there only being a single HTTP thread, that means blocking the HTTP code from performing _any_ HTTP work: writing replies, reading requests, handling closed connections, etc

greg7mdp: Thanks Matt, I guess I missed the fact that async_accept() was nonblocking. I'll update to an async wait.

greg7mdp:
> That then leaves us with the non-POSIX Linux specific grab bag. Fair enough -- that may be compelling enough reason to leave the code as it is originally. I'd be curious to see realistically how those errors can occur and whether they can be bubbled up remotely. Maybe sending a SYN with a source address that isn't routeable by the box running nodeos? Regardless, after looking closer, indeed the number of errors that could send us in to a busy loop seem vanishingly small which was the source of the concern.

Thanks @spoonincode . What do you think of the current version of the code in the PR? If you think lines 121 to 125 are unneeded I'm happy to remove them.

plugins/http_plugin/include/eosio/http_plugin/beast_http_listener.hpp
greg7mdp: @spoonincode I replaced the synchronous `wait()` with `async_wait()`

spoonincode:
I think it's probably unnecessary to keep those cases; as long as we retry as we previously did. If you elect to keep them I think we will need to make some changes: they need to be `ifdef`ed so the code continues to compile on systems that don't have those (alternatively maybe just use `boost::system::errc::` if those exist?), it also probably elevates the complexity enough to migrate this in to a helper class of some sort (we don't want to copy paste this jumble of error codes to 3+ places).

greg7mdp:
> I think it's probably unnecessary to keep those cases; as long as we retry as we previously did. If you elect to keep them I think we will need to make some changes: they need to be `ifdef`ed so the code continues to compile on systems that don't have those (alternatively maybe just use `boost::system::errc::` if those exist?), it also probably elevates the complexity enough to migrate this in to a helper class of some sort (we don't want to copy paste this jumble of error codes to 3+ places).

So you changed your mind and you think we should retry (as we did before this PR) in case of error? Before you suggested we do nothing (essentially stop accepting connections).

spoonincode:
Yes, since we can't quantify how those grab bag of errors occur (can they be generated via a remote connection? or is just local misconfiguration type stuff?) , I think it's fine to either
1. delay retry on `too_many_files_open` & immediately retry on other errors, or
2. delay retry on _any_ error, or
3. delay retry on `too_many_files_open`, immediately retry on "Linux Grab Bag", and then treat other errors as non-retryable

But if you choose number 3 I think it should be refactored in to a common reusable acceptor so we don't have to copy paste this logic all over 3+ places, and it should be `ifdef`ed just to Linux.

greg7mdp:
Thanks @spoonincode , I'll do #1 because it is the least change from the preexisting behavior, it fixes exactly the issue and no more, and I don't have certainty that the other solutions are actually better.

greg7mdp:
@spoonincode @heifner can I get your approval on this PR?

spoonincode (approved):

PR #606, opened 2022-12-30
[3.1] Attempt connection retry for duplicate connections

### LINKED ISSUE #605: Syncing sometimes stops on sync wait timeout
When syncing from one peer a sync wait timeout can cause the connected peer to determine the reconnect is a duplicate. This prevents reconnection and requires a manual disconnect/reconnect to reestablish the connection.

This is likely aggravated by the `close(true)` added by: https://github.com/eosnetworkfoundation/mandel/pull/381

heifner:
The fix https://github.com/eosnetworkfoundation/mandel/pull/756 did not address that `no_retry` is also redundantly checked in `connect`. This PR removes the duplicate check since `connect` is only called from `resolve_and_connect` where the `no_retry` is already verified.

Resolves #605


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2341,14 +2341,6 @@ namespace eosio {
 
    // called from connection strand
    void connection::connect( const std::shared_ptr<tcp::resolver>& resolver, tcp::resolver::results_type endpoints ) {
-      switch ( no_retry ) {
-         case no_reason:
-         case wrong_version:
-         case benign_other:
-            break;
```
linh2931, line 2348: The `no_retry` check in `resolve_and_connect` has an extra case `duplicate`. Before the removal of the check here, `duplicate` will just `return`. Now `duplicate` will continue. Is this the behavior intended?

heifner: Yes.

linh2931 (approved):

766C6164 (approved):

greg7mdp (approved):

PR #612, opened 2023-01-03
Add test contract code for existing abi/wasm and clean up libraries/testing & unittests dependencies

### LINKED ISSUE #54: Test Contracts Source Files
There are a handful of test contracts in [unittests/contracts](https://github.com/AntelopeIO/leap/tree/main/unittests/contracts) that are not accompanied by source files for easy indication of versioning or reproduction. Especially with `eos-system-contracts` living outside the AntelopIO organization, would it be helpful to bring the source across for reference?

They should be clearly marked as example contracts for testing and not for production use and should live in the `unittests/*` directory.

They can be pulled from: https://github.com/EOSIO/eos/tree/develop/contracts/contracts

Once example contracts' source files are included, they should be compiled when `EOSIO_COMPILE_TEST_CONTRACTS` is indicated.

heifner:
Updated `unittests/contracts/eosio.msig, unittests/contracts/eosio.system, unittests/contracts/eosio.token, unittests/contracts/eosio.wrap` and `libraries/testing/contracts/eosio.bios` with contract code as close as possible to existing abi/wasm files. Before only the abi/wasm existed in the leap repo.

Rebuilt all contracts mentioned above with the latest CDT (3.2.0-dev ea6d926c3d2e9fa37dd513b8a946d8283b1aa233) and replaced the abi/wasm with the newly generated abi/wasm. Now all these contracts can be built using `-DEOSIO_COMPILE_TEST_CONTRACTS=true`.

`libraries/testing` no longer depends on `unittests`.
Moved `snapshot_suites.hpp` out of `libraries/testing` and back into `unittests` where it belongs.

Resolves #54


spoonincode:
> Note: `eosio.token` is embedded, like `eosio.bios`, in `eosio_testing_contracts` library

hm, fixes eosnetworkfoundation/mandel#681 I think

spoonincode:
The libtester created here seems to have a problem building contract unit tests
https://github.com/eosnetworkfoundation/eos-system-contracts/actions/runs/3843842909

heifner:
> The libtester created here seems to have a problem building contract unit tests https://github.com/eosnetworkfoundation/eos-system-contracts/actions/runs/3843842909

This is probably because of these changes:
https://github.com/AntelopeIO/leap/pull/612/files#diff-3fe8325df2bcbadd6d0031003362dac51ecb2035e5342e871de92ba5f0123501L4
I'm not sure why that would cause an issue as it should be linked into `eosio_testing`. At this point I think we should just remove `txn_test_gen_plugin` and revert this section of the `CMakeList.txt` changes. Also move `eosio.token` back down into `unittests` where it belongs.

spoonincode:
I haven't really looked very closely, but is it possible the new static library `eosio_testing_contracts` needs to be part of EosioTester.cmake (both of them)? In this area
https://github.com/AntelopeIO/leap/blob/b7e18846c0c18b4b75ff33553b98c02bbbb4fc71/CMakeModules/EosioTester.cmake.in#L55-L59

Also that means `eosio_testing_contracts` would need to be cmake `install()`ed

heifner:
> but is it possible the new static library `eosio_testing_contracts` needs to be part of EosioTester.cmake (both of them)

I was assuming it was something like that. I really think the best approach is to go back to `eosio_testing_contracts` being an `INTERFACE` and not a library. That all can be put back like it was once `txn_test_gen_plugin` is removed.

heifner:
> The libtester created here seems to have a problem building contract unit tests https://github.com/eosnetworkfoundation/eos-system-contracts/actions/runs/3843842909

Fixed with the latest: https://github.com/eosnetworkfoundation/eos-system-contracts/actions/runs/3913111725

spoonincode:
> with contract code as close as possible to existing abi/wasm files

I had a hard time finding where this contract source originated from. Especially unittests/contracts/eosio.system doesn't seem to match anything I can find in reference-contracts. How sensitive are the tests to using the latest reference contracts? Is it possible (or silly) to use the reference contracts as a submodule?

heifner:
> > with contract code as close as possible to existing abi/wasm files
> 
> I had a hard time finding where this contract source originated from. Especially unittests/contracts/eosio.system doesn't seem to match anything I can find in reference-contracts. How sensitive are the tests to using the latest reference contracts? Is it possible (or silly) to use the reference contracts as a submodule?

I had a hard time finding the source for these contracts. I dug around in github history, but for some of them the contract code corresponding to the time of the wasm change does not match the wasm. Some of it was trial and error to get to something that works.

I think ideally we would not use any reference-contracts in our tests. Something like them are needed to test protocol features and other host functions. We want the contracts to remain static for our tests. This is likely why we didn't before bother including the source code for them in leap. However, I do believe we should be able to recreate the wasms from source if desired, hence this PR.

spoonincode (approved):
I had a follow up question on whether it made sense (not now in this PR, but in the future) to have a CI job ensure the contracts are still compiling 1:1 with checked in binaries. I feel like you may be indirectly answered that already though..

heifner:
> I had a follow up question on whether it made sense (not now in this PR, but in the future) to have a CI job ensure the contracts are still compiling 1:1 with checked in binaries. I feel like you may be indirectly answered that already though..

Ideally it would be nice to have a job that makes sure the contracts still compile with the latest CDT and all tests pass with the latest CDT compiled contracts.

larryk85 (approved):

PR #618, opened 2023-01-05
[PH] Performance harness Log Number of Forks and Forked Blocks

ClaytonCalabrese:
PH now uses nodeos logs to report forks and forked blocks during performance testing.

### DIFF
--- a/tests/performance_tests/log_reader.py
+++ b/tests/performance_tests/log_reader.py
@@ -152,6 +152,7 @@ class chainData():
         self.totalElapsed = 0
         self.totalTime = 0
         self.totalLatency = 0
+        self.forkedBlocks = []
     def __eq__(self, other):
         return self.startBlock == other.startBlock and\
          self.ceaseBlock == other.ceaseBlock and\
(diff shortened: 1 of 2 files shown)


tests/performance_tests/log_reader.py
```diff
@@ -200,6 +202,9 @@ def scrapeLog(data, path):
                         data.updateTotal(int(value[2]), 0, 0, 0, 0, int(v2Logging[0]))
                 else:
                     print("Error: Unknown log format")
+        forks = re.findall(r'switching forks from ([0-9a-fA-F]+) \(block number (\d+)\) to ([0-9a-fA-F]+) \(block number (\d+)\)', line)
+        for fork in forks:
+            data.forkedBlocks.append(int(fork[1]) - int(fork[3]))
```
heifner, line 208: I think this needs a `+1`. Example:
info  2022-12-28T18:06:18.000 nodeos    controller.cpp:2309           maybe_switch_forks   ] switching forks from 000000ddca004dc9992db186880efea4908850901e21999ae0861d2dc9916402 (block number 221) to 000000dd705ab2c8b1a3d527c691ae98ad904e3b49b1f3d45ae9ae09fd22e6be (block number 221)

Also want to track # of times a fork happens separately from total fork distance.

ClaytonCalabrese: # of times a fork happens is done using len(data.forkedBlocks) so it doesn't need to be saved here unless I'm missing osmething. I'll add the +1 because what you said makes sense.

heifner: Thanks, missed the `len(data.forkedBlocks)`

oschwaldp-oci (approved):

Issue #592, opened 2022-12-22
3.2 state_history_plugin crashes under heavy load


heifner:
[Information reported by community]

3.1 does not have this issue. Observed with both v3.2.0-rc1 and v3.2.0.

Reproducible by indexing with Hyperion with high scale values.

![image](https://user-images.githubusercontent.com/3933649/209135001-ee8c7ae4-78d6-493d-83cb-1bf11cec1aff.png)

Double these values to speed up the indexing and will reproduce the state_history_plugin crash.

![image](https://user-images.githubusercontent.com/3933649/209135023-c8c2097a-bcac-468c-a427-d8f8be30a59b.png)

```
OC enabled:
##############
info  2022-12-13T16:24:31.615 nodeos    state_history_plugin.c:266    operator()           ] received get_blocks_ack_request_v0 = {"num_messages":1}
info  2022-12-13T16:24:31.620 nodeos    state_history_plugin.c:266    operator()           ] received get_blocks_ack_request_v0 = {"num_messages":1}
info  2022-12-13T16:24:31.622 nodeos    state_history_plugin.c:266    operator()           ] received get_blocks_ack_request_v0 = {"num_messages":1}
info  2022-12-13T16:24:31.625 nodeos    state_history_plugin.c:266    operator()           ] received get_blocks_ack_request_v0 = {"num_messages":1}
info  2022-12-13T16:24:31.627 nodeos    state_history_plugin.c:266    operator()           ] received get_blocks_ack_request_v0 = {"num_messages":1}
info  2022-12-13T16:24:31.628 nodeos    state_history_plugin.c:266    operator()           ] received get_blocks_ack_request_v0 = {"num_messages":1}
info  2022-12-13T16:24:31.629 nodeos    state_history_plugin.c:266    operator()           ] received get_blocks_ack_request_v0 = {"num_messages":1}
info  2022-12-13T16:24:31.632 nodeos    state_history_plugin.c:266    operator()           ] received get_blocks_ack_request_v0 = {"num_messages":1}
info  2022-12-13T16:24:31.632 nodeos    state_history_plugin.c:266    operator()           ] received get_blocks_ack_request_v0 = {"num_messages":1}
info  2022-12-13T16:24:31.632 nodeos    state_history_plugin.c:266    operator()           ] received get_blocks_ack_request_v0 = {"num_messages":1}
[...]

PR #632, opened 2023-01-10
Add handling of exceptions to named_thread_pool

heifner:
Instead of exceptions leaking out and terminating thread in `named_thread_pool` provide ability to specify an exception handler. In all our non-test cases we want the application to shutdown.

Pulled in changes from #592 that allows for delayed start of the thread pool.
Added some tests for `named_thread_pool`.

As discussed here: https://github.com/AntelopeIO/leap/pull/665#discussion_r1086080750 Moved to always using delayed start with number of threads provided in `start()`. Also moved specification of exception handler to `start()` as that is cleaner than requiring it in the constructor.


libraries/chain/thread_utils.cpp
```diff
@@ -23,10 +24,33 @@ void named_thread_pool::start() {
    _ioc_work.emplace( boost::asio::make_work_guard( _ioc ) );
    _ioc.restart();
    for( size_t i = 0; i < _num_threads; ++i ) {
-      _thread_pool.emplace_back( [&ioc = _ioc, &name_prefix = _name_prefix, i]() {
+      _thread_pool.emplace_back( [&ioc = _ioc, &name_prefix = _name_prefix, on_except = _on_except, i]() {
```
greg7mdp, line 27: Why not capture `&_on_except` instead of making copies `on_except = _on_except`?

libraries/chain/thread_utils.cpp
```diff
@@ -23,10 +24,33 @@ void named_thread_pool::start() {
...
+         try {
+            fc::set_os_thread_name( tn );
+            ioc.run();
+         } catch( const fc::exception& e ) {
+            if( on_except ) {
+               on_except( e );
+            } else {
+               elog( "Exiting thread ${t} on exception: ${e}", ("t", tn)("e", e.to_detail_string()) );
+            }
+         } catch( const std::exception& e ) {
```
greg7mdp, line 38: Is it on purpose that you change the exiting behavior by catching the exceptions and not rethrowing them (even when `on_except` is empty).
When `on_except` is not empty, I guess the way you use it by calling `app().quit()` makes it OK to not rethrow, but should we assume that will always be the case? and if that is always the case, maybe instead of passing a lambda, it would be simpler to pass an "origin" string and call `app().quit()` here?
Just some thoughts!

plugins/chain_plugin/test/test_account_query_db.cpp
```diff
@@ -115,7 +115,7 @@ BOOST_FIXTURE_TEST_CASE(updateauth_test_multi_threaded, TESTER) { try {
    produce_block();
    create_account(tester_account);
 
-   named_thread_pool thread_pool( "test", 5 );
+   named_thread_pool thread_pool( "test", 5, nullptr );
```
greg7mdp, line 118: I think I'd prefer `{}` since `std::function` is not a pointer.

libraries/chain/include/eosio/chain/thread_utils.hpp
```diff
@@ -41,14 +47,15 @@ namespace eosio { namespace chain {
...
       std::optional<ioc_work_t>      _ioc_work;
+      on_except_t                    _on_except;
    };
 
 
-   // async on thread_pool and return future
+   // async on io_context and return future
    template<typename F>
-   auto async_thread_pool( boost::asio::io_context& thread_pool, F&& f ) {
+   auto async_thread_pool( boost::asio::io_context& ioc, F&& f ) {
```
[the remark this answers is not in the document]
heifner: I agree, when the function was first created it did use the `thread_pool` instead of `io_context`. I'll update the name to better reflect what it does now.

libraries/chain/thread_utils.cpp
heifner: Hmm, yeah, maybe re-throw is better when not provided an `on_except`. If provided an `on_except` the user can throw or not depending on what they want to do.

`app()` is not available everywhere `thread_pool` is used.

greg7mdp (approved):
Nice changes Kevin!

libraries/chain/thread_utils.cpp
```diff
@@ -4,29 +4,51 @@
...
 
 named_thread_pool::~named_thread_pool() {
    stop();
 }
 
-void named_thread_pool::start() {
+void named_thread_pool::start( size_t num_threads, on_except_t on_except ) {
    FC_ASSERT( !_ioc_work, "Thread pool already started" );
    _ioc_work.emplace( boost::asio::make_work_guard( _ioc ) );
    _ioc.restart();
```
greg7mdp, line 20:
```suggestion
   _ioc.restart();
   _thread_pool.reserve(num_threads);
```

greg7mdp (approved):

spoonincode (approved):

PR #635, opened 2023-01-12
First cut at a more message based interface between prometheus and instrumented plugins

ndcgundlach:
New interface between prometheus and instrumented plugins.  

Now, prometheus will register a listener with instrumented plugins.  Instrumented plugins will 'batch' updates and call back the prometheus listener.  

The listener that prometheus passes to the instrumented plugins is a lambda that posts another lambda to a prometheus strand.  The lambda that is posted to the prometheus strand will update the prometheus data model.  The updating of the data model is not implemented in this PR.

One open issue is where can we use 'move' rather than 'copy' semantics.

### DIFF
--- a/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
+++ b/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
@@ -30,8 +30,7 @@ namespace eosio { namespace chain { namespace plugin_interface {
    struct plugin_metrics {
       virtual vector<runtime_metric> metrics()=0;
       bool should_post() {
-         fc::time_point now = fc::time_point::now();
-         return ((_listener) && now > (_last_post + fc::milliseconds(_min_post_interval_ms)));
+         return ((_listener) && fc::time_point::now() > (_last_post + fc::milliseconds(_min_post_interval_ms)));
       }
 
       bool post_metrics() {
--- a/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
+++ b/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
@@ -18,20 +18,15 @@ namespace eosio { namespace chain { namespace plugin_interface {
    };
 
    struct runtime_metric {
-      const metric_type type;
-      const std::string family;
-      const std::string label;
-      std::atomic<int64_t> value;
+      const metric_type type = metric_type::gauge;
+      const std::string family = "";
+      const std::string label = "";
+      int64_t value = 0;
+
    };
 
-   struct plugin_metrics {
-      std::vector<std::reference_wrapper<runtime_metric>> metrics;
-      void enable(bool enabled)  {_enabled = enabled; }
-      bool enabled() {return _enabled;}
+   using metrics_listener = std::function<void(std::vector<runtime_metric>)>;
 
-   private:
-      bool _enabled = false;
-   };
    template<typename T>
    using next_function = std::function<void(const std::variant<fc::exception_ptr, T>&)>;
 
--- a/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
+++ b/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
@@ -36,7 +36,7 @@ namespace eosio { namespace chain { namespace plugin_interface {
 
       bool post_metrics() {
          if (should_post()){
-            _listener(std::move(metrics()));
+            _listener(metrics());
             _last_post = fc::time_point::now();
             return true;
          }
--- a/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
+++ b/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
@@ -18,9 +18,9 @@ namespace eosio { namespace chain { namespace plugin_interface {
    };
 
    struct runtime_metric {
-      const metric_type type = metric_type::gauge;
-      const std::string family;
-      const std::string label;
+      metric_type type = metric_type::gauge;
+      std::string family;
+      std::string label;
       int64_t value = 0;
 
    };
--- a/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
+++ b/plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
@@ -19,13 +19,42 @@ namespace eosio { namespace chain { namespace plugin_interface {
 
    struct runtime_metric {
       const metric_type type = metric_type::gauge;
-      const std::string family = "";
-      const std::string label = "";
+      const std::string family;
+      const std::string label;
       int64_t value = 0;
 
    };
 
    using metrics_listener = std::function<void(std::vector<runtime_metric>)>;
+
+   struct plugin_metrics {
+      virtual vector<runtime_metric> metrics()=0;
+      bool should_post() {
+         fc::time_point now = fc::time_point::now();
+         return ((_listener) && now > (_last_post + fc::milliseconds(_min_post_interval_ms)));
+      }
+
+      bool post_metrics() {
+         if (should_post()){
+            _listener(std::move(metrics()));
+            _last_post = fc::time_point::now();
+            return true;
+         }
+
+         return false;
+      }
+
+      void register_listener(metrics_listener listener) {
+         _listener = listener;
+      }
+
+      plugin_metrics(int64_t min_post_interval_ms=250) : _min_post_interval_ms(min_post_interval_ms), _listener(nullptr) {}
+
+   private:
+      int64_t _min_post_interval_ms;
+      metrics_listener _listener;
+      fc::time_point _last_post;
+   };
 
    template<typename T>
    using next_function = std::function<void(const std::variant<fc::exception_ptr, T>&)>;
--- a/plugins/net_plugin/include/eosio/net_plugin/net_plugin.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/net_plugin.hpp
@@ -15,15 +15,18 @@ namespace eosio {
 
    using namespace plugin_interface;
 
-   struct net_plugin_metrics : plugin_metrics {
+   struct net_plugin_metrics {
       runtime_metric num_peers{metric_type::gauge, "num_peers", "num_peers", 0};
       runtime_metric num_clients{metric_type::gauge, "num_clients", "num_clients", 0};
       runtime_metric dropped_trxs{metric_type::counter, "dropped_trxs", "dropped_trxs", 0};
 
-      net_plugin_metrics() {
-         metrics.emplace_back(std::ref(num_peers));
-         metrics.emplace_back(std::ref(num_clients));
-         metrics.emplace_back(std::ref(dropped_trxs));
+      vector<runtime_metric> metrics() {
+         vector<runtime_metric> metrics;
+         metrics.emplace_back(num_peers);
+         metrics.emplace_back(num_clients);
+         metrics.emplace_back(dropped_trxs);
+
+         return metrics;
       }
    };
 
@@ -46,7 +49,7 @@ namespace eosio {
         std::optional<connection_status>  status( const string& endpoint )const;
         vector<connection_status>         connections()const;
 
-        net_plugin_metrics& metrics();
+        void register_metrics_listener(metrics_listener listener);
 
       private:
         std::shared_ptr<class net_plugin_impl> my;
--- a/plugins/net_plugin/include/eosio/net_plugin/net_plugin.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/net_plugin.hpp
@@ -20,12 +20,12 @@ namespace eosio {
       runtime_metric num_clients{metric_type::gauge, "num_clients", "num_clients", 0};
       runtime_metric dropped_trxs{metric_type::counter, "dropped_trxs", "dropped_trxs", 0};
 
-      virtual vector<runtime_metric> metrics() {
-         vector<runtime_metric> metrics;
-         metrics.reserve(3);
-         metrics.emplace_back(num_peers);
-         metrics.emplace_back(num_clients);
-         metrics.emplace_back(dropped_trxs);
+      vector<runtime_metric> metrics() final {
+         vector<runtime_metric> metrics {
+            num_peers,
+            num_clients,
+            dropped_trxs
+         };
 
          return metrics;
       }
--- a/plugins/net_plugin/include/eosio/net_plugin/net_plugin.hpp
+++ b/plugins/net_plugin/include/eosio/net_plugin/net_plugin.hpp
@@ -15,13 +15,14 @@ namespace eosio {
 
    using namespace plugin_interface;
 
-   struct net_plugin_metrics {
+   struct net_plugin_metrics : plugin_metrics {
       runtime_metric num_peers{metric_type::gauge, "num_peers", "num_peers", 0};
       runtime_metric num_clients{metric_type::gauge, "num_clients", "num_clients", 0};
       runtime_metric dropped_trxs{metric_type::counter, "dropped_trxs", "dropped_trxs", 0};
 
-      vector<runtime_metric> metrics() {
+      virtual vector<runtime_metric> metrics() {
          vector<runtime_metric> metrics;
+         metrics.reserve(3);
          metrics.emplace_back(num_peers);
          metrics.emplace_back(num_clients);
          metrics.emplace_back(dropped_trxs);
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -287,7 +287,9 @@ namespace eosio {
 
       uint16_t                                       thread_pool_size = 2;
       std::optional<eosio::chain::named_thread_pool> thread_pool;
-      net_plugin_metrics metrics;
+
+      net_plugin_metrics   metrics;
+      metrics_listener     net_metrics_listener;
 
    private:
       mutable std::mutex            chain_info_mtx; // protects chain_*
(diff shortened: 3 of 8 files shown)


plugins/net_plugin/include/eosio/net_plugin/net_plugin.hpp
```diff
@@ -15,15 +15,18 @@ namespace eosio {
...
       runtime_metric num_peers{metric_type::gauge, "num_peers", "num_peers", 0};
       runtime_metric num_clients{metric_type::gauge, "num_clients", "num_clients", 0};
       runtime_metric dropped_trxs{metric_type::counter, "dropped_trxs", "dropped_trxs", 0};
 
-      net_plugin_metrics() {
-         metrics.emplace_back(std::ref(num_peers));
-         metrics.emplace_back(std::ref(num_clients));
-         metrics.emplace_back(std::ref(dropped_trxs));
+      vector<runtime_metric> metrics() {
+         vector<runtime_metric> metrics;
```
heifner, line 24: Since we know the size, should `reserve` it.

plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
```diff
@@ -18,20 +18,15 @@ namespace eosio { namespace chain { namespace plugin_interface {
    };
 
    struct runtime_metric {
-      const metric_type type;
-      const std::string family;
-      const std::string label;
-      std::atomic<int64_t> value;
+      const metric_type type = metric_type::gauge;
+      const std::string family = "";
+      const std::string label = "";
```
heifner, line 23: These strings are copied over and over, but small string optimization should avoid dynamic memory allocation. We should strive to always keep these strings under 24 chars.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3348,9 +3350,12 @@ namespace eosio {
       }
       g.unlock();
 
-      metrics.num_clients.value = num_clients;
-      metrics.num_peers.value = num_peers;
-      
+      if (net_metrics_listener) {
+         metrics.num_clients.value = num_clients;
+         metrics.num_peers.value = num_peers;
```
heifner, line 3355: What about `dropped_trxs`?

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
```diff
@@ -24,16 +24,19 @@ struct producer_plugin_metrics : public plugin_metrics {
...
-      metrics.emplace_back(std::ref(blacklisted_transactions));
-      metrics.emplace_back(std::ref(blocks_produced));
-      metrics.emplace_back(std::ref(trxs_produced));
-      metrics.emplace_back(std::ref(last_irreversible));
-      metrics.emplace_back(std::ref(block_num));
-      metrics.emplace_back(std::ref(subjective_bill_account_size));
-      metrics.emplace_back(std::ref(subjective_bill_block_size));
-      metrics.emplace_back(std::ref(scheduled_trxs));
+   vector<runtime_metric> metrics() {
+      vector<runtime_metric> metrics;
```
heifner, line 28: Since we know the size, should `reserve` it.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -343,16 +343,18 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+   /*
+    * HACK ALERT
+    * Boost timers can be in a state where a handler has not yet executed but is not abortable.
+    * As this method needs to mutate state handlers depend on for proper functioning to maintain
+    * invariants for other code (namely accepting incoming transactions in a nearly full block)
+    * the handlers capture a corelation ID at the time they are set.  When they are executed
+    * they must check that correlation_id against the global ordinal.  If it does not match that
+    * implies that this method has been called with the handler in the state where it should be
+    * cancelled but wasn't able to be.
+    */
```
heifner, line 357: Please undo this indent change.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -412,18 +414,22 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+            _metrics.unapplied_transactions.value = _unapplied_transactions.size();
 
-         auto& chain = chain_plug->chain();
-         _metrics.last_irreversible.value = chain.last_irreversible_block_num();
-         _metrics.block_num.value = chain.head_block_num();
+            auto &chain = chain_plug->chain();
+            _metrics.last_irreversible.value = chain.last_irreversible_block_num();
+            _metrics.block_num.value = chain.head_block_num();
+
+            const auto &sch_idx = chain.db().get_index<generated_transaction_multi_index, by_delay>();
```
heifner, line 428: Minor thing, but use `auto&` instead of `auto &var`

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -130,6 +123,20 @@ namespace eosio {
...
          }
 
+         void update_metrics(vector<runtime_metric> metrics) {
+
+         }
+
+         metrics_listener create_metrics_listener() {
+            return  [self=this](vector<runtime_metric> metrics) {
+               self->_metrics_updater.post(
+                     [self, metrics]() {
```
heifner, line 133: `metrics{std::move(metrics)}` and make the lambda `mutable`

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -130,6 +123,20 @@ namespace eosio {
...
+                        self->update_metrics(metrics);
```
heifner, line 134: `self->update_metrics(std::move(metrics));`

heifner (changes requested):

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -60,7 +60,8 @@ namespace eosio {
          std::vector<std::shared_ptr<Collectable>> _collectables;
          const prometheus::TextSerializer _serializer;
          std::shared_ptr<Registry> _registry;
-
+         boost::asio::io_context _context;
+         boost::asio::io_context::strand _metrics_updater{_context};
```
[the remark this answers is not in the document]
spoonincode: Just remember though that a strand guarantees ordering while just using named_thread_pool's io_context, even with a single thread, does not. I can't quite tell if the recommendation is to remove the strand usage or just the io_context and stuff (you can still make a strand using named_thread_pool's io_context, of course).

spoonincode:
hm, I'm realizing now that a downside of this approach (unless adding more complexity), is that during sync it's going to cause lots of message flow between producer_plugin->prometheus_plugin

heifner:
> hm, I'm realizing now that a downside of this approach (unless adding more complexity), is that during sync it's going to cause lots of message flow between producer_plugin->prometheus_plugin

I think we could add to `plugin_metrics` 
```
   bool should_post() {
     return listener && (fc::time_point::now() > last_post + fc::milliseconds(250));
  }
```
And then each plugin can use: `if (metrics.should_post()) { ` instead of the current `if (net_metrics_listener) {`.

plugins/prometheus_plugin/prometheus_plugin.cpp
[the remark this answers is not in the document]
heifner: I was thinking order was not important for this. But yeah easy enough to use a strand with the `thread_pool.get_executor()`.

ndcgundlach:
> > hm, I'm realizing now that a downside of this approach (unless adding more complexity), is that during sync it's going to cause lots of message flow between producer_plugin->prometheus_plugin
> 
> I think we could add to `plugin_metrics`
> 
> ```
>    bool should_post() {
>      return listener && (fc::time_point::now() > last_post + fc::milliseconds(250));
>   }
> 
>   void post(std::vector<runtime_metrics> m) {
>      last_post = fc::time_point::now();
>      listener( std::move(m) );
>    }
> ```
> 
> And then each plugin can use: `if (metrics.should_post()) { ` instead of the current `if (net_metrics_listener) {`.

The producer plugin is once per block, which would be 500ms.  Another option would be some sort of hysteresis (ie. don't send unless value changes outside of a range), which would be a little more complicated.

heifner:
> The producer plugin is once per block, which would be 500ms. Another option would be some sort of hysteresis (ie. don't send unless value changes outside of a range), which would be a little more complicated.

During sync you can process 100s of blocks per 500ms.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1594,9 +1600,7 @@ fc::time_point producer_plugin_impl::calculate_block_deadline( const fc::time_po
 
 producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
    chain::controller& chain = chain_plug->chain();
-   if (_metrics.enabled()) {
-      update_block_metrics();
-   }
+   update_block_metrics();
```
heifner, line 1603: This also needs called from `on_incoming_block`.

plugins/net_plugin/net_plugin.cpp
ndcgundlach: This is incremented when the trx is dropped - line 2689

plugins/chain_interface/include/eosio/chain/plugin_interface.hpp
```diff
@@ -18,20 +18,44 @@ namespace eosio { namespace chain { namespace plugin_interface {
...
-      bool enabled() {return _enabled;}
+      virtual vector<runtime_metric> metrics()=0;
+      bool should_post() {
+         fc::time_point now = fc::time_point::now();
+         return ((_listener) && now > (_last_post + fc::milliseconds(_min_post_interval_ms)));
+      }
+
+      bool post_metrics() {
+         if (should_post()){
+            _listener(std::move(metrics()));
```
heifner, line 39: Remove `std::move` here.

plugins/net_plugin/include/eosio/net_plugin/net_plugin.hpp
```diff
@@ -20,10 +20,14 @@ namespace eosio {
       runtime_metric num_clients{metric_type::gauge, "num_clients", "num_clients", 0};
       runtime_metric dropped_trxs{metric_type::counter, "dropped_trxs", "dropped_trxs", 0};
 
-      net_plugin_metrics() {
-         metrics.emplace_back(std::ref(num_peers));
-         metrics.emplace_back(std::ref(num_clients));
-         metrics.emplace_back(std::ref(dropped_trxs));
+      virtual vector<runtime_metric> metrics() {
+         vector<runtime_metric> metrics;
+         metrics.reserve(3);
```
heifner, line 25: Actually what is better so we don't have to maintain the reserve count is to do this:

```
         vector<runtime_metric> metrics{ 
            num_peers, 
            num_clients, 
            dropped_trxs 
         };
```

plugins/net_plugin/net_plugin.cpp
heifner: This should be `post_metrics()` and also call `post_metrics()` inside `if (fc::time_point::now() - fc::seconds(1) >= last_dropped_trx_msg_time)`.

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
```diff
@@ -24,16 +25,20 @@ struct producer_plugin_metrics : public plugin_metrics {
...
-      metrics.emplace_back(std::ref(blocks_produced));
-      metrics.emplace_back(std::ref(trxs_produced));
-      metrics.emplace_back(std::ref(last_irreversible));
-      metrics.emplace_back(std::ref(block_num));
-      metrics.emplace_back(std::ref(subjective_bill_account_size));
-      metrics.emplace_back(std::ref(subjective_bill_block_size));
-      metrics.emplace_back(std::ref(scheduled_trxs));
+   virtual vector<runtime_metric> metrics() {
+      vector<runtime_metric> metrics;
+      metrics.reserve(9);
```
heifner, line 30: Use vector initialization list instead.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -6,6 +6,32 @@
...
+                  cb(http_response_code, fc::time_point::maximum(), std::visit(async_result_visitor(), result));\
+               }\
+            });\
+      } catch (...) { \
+         http_plugin::handle_exception(#api_name, #call_name, body, cb); \
+      } \
+   }\
+}
 
 #define CALL_WITH_400(api_name, api_handle, call_name, INVOKE, http_response_code) \
```
heifner, line 36: This can be removed.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -6,6 +6,32 @@
...
+                  }\
+               } else {\
+                  cb(http_response_code, fc::time_point::maximum(), std::visit(async_result_visitor(), result));\
+               }\
+            });\
+      } catch (...) { \
+         http_plugin::handle_exception(#api_name, #call_name, body, cb); \
+      } \
+   }\
+}
```
heifner, line 34: How about moving this into a http_plugin header called `macros.hpp` and using it here and in `chain_api_plugin.cpp` to avoid the duplication.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -55,12 +81,21 @@ namespace eosio {
             {}, Summary::Quantiles{{0.5, 0.05}, {0.9, 0.01}, {0.99, 0.001}})) {}
       };
 
+      struct async_result_visitor : public fc::visitor<fc::variant> {
+         template<typename T>
+         fc::variant operator()(const T& v) const {
+            return fc::variant(v);
+         }
+      };
```
heifner, line 89: This could also be moved into the common header.

plugins/prometheus_plugin/prometheus_plugin.cpp
heifner: Also make the lambda `mutable`.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -193,11 +237,28 @@ namespace eosio {
...
+         _pp.metrics_async(results);
+      }
+
+      prometheus_api(prometheus_plugin_impl& plugin) : _pp(plugin){
+      }
+
+   };
+   prometheus_plugin::prometheus_plugin() {
+      prometheus_api handle(*this->my);
+      app().get_plugin<http_plugin>().add_api({
```
heifner, line 260: This should be `add_async_api`

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -69,7 +104,7 @@ namespace eosio {
          std::unique_ptr<prometheus_plugin_metrics> _metrics;
 
          // hold onto other plugin metrics
-         prometheus_plugin_impl() { }
+         prometheus_plugin_impl(boost::asio::io_context& ctx): _prometheus_thread_pool("prometheus", 1), _prometheus_strand(_prometheus_thread_pool.get_executor()){ }
```
heifner, line 107: Remove `boost::asio::io_context& ctx`

heifner, line 107: "prometheus" -> "prom" as it should be 6 or less chars.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -55,12 +81,21 @@ namespace eosio {
...
+      };
+
+
       struct prometheus_plugin_impl {
-         std::mutex _collectables_mutex;
          std::vector<std::shared_ptr<Collectable>> _collectables;
          const prometheus::TextSerializer _serializer;
          std::shared_ptr<Registry> _registry;
-
+         // boost::asio::io_context _context;
```
heifner, line 96: Remove comment

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -177,6 +215,12 @@ namespace eosio {
             return body;
          }
 
+         void metrics_async(chain::plugin_interface::next_function<std::string> results) {
+            _prometheus_strand.post([self=this, results]() {
```
heifner, line 219: `results` can be moved into the lambda.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -193,11 +237,28 @@ namespace eosio {
...
+      }
+
+   };
+   prometheus_plugin::prometheus_plugin() {
+      prometheus_api handle(*this->my);
+      app().get_plugin<http_plugin>().add_api({
+        CALL_ASYNC_WITH_400(prometheus, handle, eosio, metrics, std::string, 200, http_params_types::no_params)});
    }
 
    prometheus_plugin::~prometheus_plugin() {}
```
heifner, line 264: Should `stop()` the thread pool in `plugin_shutdown()`

heifner (changes requested):

heifner (changes requested):

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -6,198 +6,200 @@
...
-               case metric_type::counter:
-                  add_counter_metric(plugin_metric);
-                  break;
-
-               default:
-                  break;
+         void update_metrics(std::string plugin_name, vector<runtime_metric> metrics) {
+            auto plugin_metrics = _plugin_metrics.find(plugin_name);
+            if (plugin_metrics != _plugin_metrics.end()) {
+               plugin_metrics->second = metrics;
```
heifner, line 117: You can now `std::move` `metrics` to avoid the copy.

heifner (approved):

PR #636, opened 2023-01-12
Update appbase/plugins to support multiple instances/exe

### LINKED ISSUE #585: Update appbase to allow multiple appbase Boost test cases to exist in the same test executable
It would not be possible with the current appbase to, for example, add another test case to the test suite in https://github.com/AntelopeIO/leap/blob/main/plugins/producer_plugin/test/test_trx_full.cpp.

We want to fix this limitation to allow multiple tests that depend on running appbase to coexist in the same executable and run in parallel.

greg7mdp:
Resolves #585.

This change allows multiple boost test cases within one executable, where each  test case can create and destroy an appbase instance, as in:

```
      appbase::app().initialize<chain_plugin, producer_plugin>( argv.size(), (char**) &argv[0] );
      appbase::app().startup();
      // do something
      appbase::app().exec();
```

The main issues resolved are:
* signal catching issues in second instance, because signals were not unblocked in `~application_impl()`
* delay plugin registration in a vector of lambdas, so that plugins can be registered again in each new `application` instance.

Other changes:
* in leap, most plugins are automatically registered by declaring a static variable which is initialized by a call to `register_plugin()`. This PR changes the type of these variables to `auto`, and the call to the static member `register_plugin()` is made via the namespace, instead of `app()`, to avoid the creation of `application` objects before `main()` is executed.
* in executables and tests, use `appbase::scoped_app app;` instead of `app()` to create a scoped object which cleans up the static application instance when the variable gets out of scope.


heifner (approved):
Can we add some tests for this new capability.

greg7mdp:
> Can we add some tests for this new capability.

Sure I'll add a test.

spoonincode:
I guess it's not really well documented anywhere, but the ENF runners have a relatively strict maximum runtime (they aren't cheap!). All ENF runners other then `lowtier` have a maximum runtime of 10 minutes currently. So the Tests job that is failing after ~10 minutes with `Error: Process completed with exit code 143.` probably has a test getting wedged due to these changes.

greg7mdp:
> I guess it's not really well documented anywhere, but the ENF runners have a relatively strict maximum runtime (they aren't cheap!). All ENF runners other then `lowtier` have a maximum runtime of 10 minutes currently. So the Tests job that is failing after ~10 minutes with `Error: Process completed with exit code 143.` probably has a test getting wedged due to these changes.

Oh, thanks @spoonincode I had missed this... will investigate!

greg7mdp:
> > I guess it's not really well documented anywhere, but the ENF runners have a relatively strict maximum runtime (they aren't cheap!). All ENF runners other then `lowtier` have a maximum runtime of 10 minutes currently. So the Tests job that is failing after ~10 minutes with `Error: Process completed with exit code 143.` probably has a test getting wedged due to these changes.

The test was hanging on a mutex. It is fixed now.

greg7mdp:
@heifner @spoonincode I'm done with my changes in both appbase PRs (this one and https://github.com/AntelopeIO/appbase/pull/4), waiting for final approval!

heifner (approved):

linh2931 (approved):

Issue #637, opened 2023-01-13
Test failure: nodeos_snapshot_diff_test


heifner:
Likely related to #634 

```
Traceback (most recent call last):
  File "/__w/leap/leap/build/tests/nodeos_snapshot_diff_test.py", line 151, in <module>
    assert steadyStateAvg>=minRequiredTransactions, "Expected to at least receive %s transactions per block, but only getting %s" % (minRequiredTransactions, steadyStateAvg)
AssertionError: Expected to at least receive 400.2 transactions per block, but only getting 394.5
```
400 is a lot of trxs to expect in our github test env. 50 per block is more than enough for all the tests that use the trx_generator.

heifner:
Another example: https://github.com/AntelopeIO/leap/actions/runs/3914498862/jobs/6691719468

PR #639, opened 2023-01-16
Working to fix nodeos_snapshot_diff_test by sync on trx gen startup

oschwaldp-oci:
Refactor - move `launchTrxGenerators` from `Node.py` to `Cluster.py` where it fits better.

Add `waitForTrxGeneratorsSpinup` function which allows tests to wait for initial trx from each generator to appear in a block.  This allows the tests to sync on the trx generators spinning up and knowing that trxs are flowing before proceeding with tests.

Update tests that used `launchTrxGenerators` for the refactor and to use the `waitForTrxGeneratorsSpinup` function.

`trx_generator` now logs first transaction it is preparing to send right before sending it to allow synchronization by tests without influencing its generation rate since it occurs prior only on initial trx and prior to initial trx send.

Resolves: #637

### DIFF
--- a/tests/nodeos_contrl_c_test.py
+++ b/tests/nodeos_contrl_c_test.py
@@ -106,7 +106,7 @@ try:
                                 numGenerators=trxGeneratorCnt, durationSec=testTrxGenDurationSec, waitToComplete=False)
 
     Print("Give txn generator time to spin up and begin producing trxs")
-    cluster.waitForTrxGeneratorsSpinup(nodeId=nonProdNode.nodeId, numGenerators=trxGeneratorCnt)
+    time.sleep(10)
 
     Print("Kill non-producer bridge node")
     testSuccessful = nonProdNode.kill(signal.SIGTERM)
--- a/tests/nodeos_contrl_c_test.py
+++ b/tests/nodeos_contrl_c_test.py
@@ -116,8 +116,6 @@ try:
         errorExit("Failed to kill the seed node")
 
 finally:
-    if trxGenLauncher is not None:
-        trxGenLauncher.killAll()
     TestHelper.shutdown(cluster, walletMgr, testSuccessful=testSuccessful, killEosInstances=True, killWallet=True, keepLogs=True, cleanRun=True, dumpErrorDetails=True)
 
 errorCode = 0 if testSuccessful else 1
--- a/tests/nodeos_contrl_c_test.py
+++ b/tests/nodeos_contrl_c_test.py
@@ -98,7 +98,7 @@ try:
     testSuccessful=False
 
     Print("Configure and launch txn generators")
-    targetTpsPerGenerator = 500
+    targetTpsPerGenerator = 100
     testTrxGenDurationSec=60
     trxGeneratorCnt=1
     cluster.launchTrxGenerators(contractOwnerAcctName=cluster.eosioAccount.name, acctNamesList=[accounts[0].name,accounts[1].name],
@@ -107,7 +107,6 @@ try:
 
     Print("Give txn generator time to spin up and begin producing trxs")
     cluster.waitForTrxGeneratorsSpinup(nodeId=nonProdNode.nodeId, numGenerators=trxGeneratorCnt)
-    time.sleep(5)
 
     Print("Kill non-producer bridge node")
     testSuccessful = nonProdNode.kill(signal.SIGTERM)
--- a/tests/nodeos_contrl_c_test.py
+++ b/tests/nodeos_contrl_c_test.py
@@ -98,16 +98,16 @@ try:
     testSuccessful=False
 
     Print("Configure and launch txn generators")
-    nonProducerP2pPort = cluster.getNodeP2pPort(nonProdNode.nodeId)
     targetTpsPerGenerator = 500
     testTrxGenDurationSec=60
     trxGeneratorCnt=1
-    nonProdNode.launchTrxGenerators(tpsPerGenerator=targetTpsPerGenerator, numGenerators=trxGeneratorCnt, durationSec=testTrxGenDurationSec,
-                                    contractOwnerAcctName=cluster.eosioAccount.name, acctNamesList=[accounts[0].name,accounts[1].name],
-                                    acctPrivKeysList=[account1PrivKey,account2PrivKey], p2pListenPort=nonProducerP2pPort, waitToComplete=False)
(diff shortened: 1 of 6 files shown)


tests/nodeos_contrl_c_test.py
```diff
@@ -98,16 +98,16 @@
...
-                                    contractOwnerAcctName=cluster.eosioAccount.name, acctNamesList=[accounts[0].name,accounts[1].name],
-                                    acctPrivKeysList=[account1PrivKey,account2PrivKey], p2pListenPort=nonProducerP2pPort, waitToComplete=False)
+    cluster.launchTrxGenerators(contractOwnerAcctName=cluster.eosioAccount.name, acctNamesList=[accounts[0].name,accounts[1].name],
+                                acctPrivKeysList=[account1PrivKey,account2PrivKey], nodeId=nonProdNode.nodeId, tpsPerGenerator=targetTpsPerGenerator,
+                                numGenerators=trxGeneratorCnt, durationSec=testTrxGenDurationSec, waitToComplete=False)
 
     Print("Give txn generator time to spin up and begin producing trxs")
-    time.sleep(10)
+    cluster.waitForTrxGeneratorsSpinup(nodeId=nonProdNode.nodeId, numGenerators=trxGeneratorCnt)
+    time.sleep(5)
```
heifner, line 110: This sleep is no longer needed, right?

tests/nodeos_snapshot_diff_test.py
```diff
@@ -120,12 +120,13 @@ def waitForBlock(node, blockNum, blockType=BlockType.head, timeout=None, reportI
     waitForBlock(node0, blockNum, blockType=BlockType.lib)
 
     Print("Configure and launch txn generators")
-    producerP2pPort = cluster.getNodeP2pPort(snapshotNodeId)
     targetTpsPerGenerator = 667
```
heifner, line 123: This is way overkill, 10 should be more than enough. Just need a few.

tests/nodeos_contrl_c_test.py
```diff
@@ -98,16 +98,16 @@
     testSuccessful=False
 
     Print("Configure and launch txn generators")
-    nonProducerP2pPort = cluster.getNodeP2pPort(nonProdNode.nodeId)
     targetTpsPerGenerator = 500
```
heifner, line 101: We can reduce this. Maybe 100 would be better.

tests/trx_generator/trx_generator.cpp
```diff
@@ -97,8 +97,21 @@ namespace eosio::testing {
...
+      std::ofstream out(fileName.str());
+
+      out << fc::string(trx.id()) << "\n";
+      out.close();
+   }
+
    void transfer_trx_generator::push_transaction(p2p_trx_provider& provider, signed_transaction_w_signer& trx, uint64_t& nonce_prefix, uint64_t& nonce, const fc::microseconds& trx_expiration, const chain_id_type& chain_id, const block_id_type& last_irr_block_id) {
       update_resign_transaction(trx._trx, trx._signer, ++nonce_prefix, nonce, trx_expiration, chain_id, last_irr_block_id);
+      if (_txcount == 0)
+      {
```
heifner, line 112: Odd, change of style for this `if`.

heifner (changes requested):

tests/nodeos_snapshot_diff_test.py
oschwaldp-oci: Happy to reduce it. Original aim was to maintain generation rates original attempted when using the trx_test_gen_plugin.

tests/nodeos_contrl_c_test.py
oschwaldp-oci: Happy to reduce it. Original aim was to maintain generation rates original attempted when using the trx_test_gen_plugin.

tests/nodeos_snapshot_diff_test.py
heifner: I created this test, no idea why I had it that large. Probably a copy/paste from somewhere else. Only need a couple of trx per block.

tests/nodeos_contrl_c_test.py
heifner: Understood. Might as well reduce the load of this test while you are in here. The idea of the test is to have some transactions running during shutdown. 50 per block seems plenty for that requirement.

tests/trx_generator/trx_generator.cpp
oschwaldp-oci: Addressed: [de3a1c8](https://github.com/AntelopeIO/leap/pull/639/commits/de3a1c8392564d6d0ab41412374948c4c8f5359c)

tests/nodeos_snapshot_diff_test.py
oschwaldp-oci: Addressed: [de3a1c8](https://github.com/AntelopeIO/leap/pull/639/commits/de3a1c8392564d6d0ab41412374948c4c8f5359c)

tests/nodeos_contrl_c_test.py
oschwaldp-oci: Addressed: [de3a1c8](https://github.com/AntelopeIO/leap/pull/639/commits/de3a1c8392564d6d0ab41412374948c4c8f5359c)

heifner: It was just waiting before to make sure trx were flowing. So I think you can remove the sleep with your better wait for first trx.

heifner (approved):

heifner:
Test failure:
```
2023-01-16T16:13:16.774983  Found first trx record: var/lib/first_trx_1338.txt
2023-01-16T16:13:16.775182  first transactions: ['7297b739c41a14c6ba115913d5bd63b0e690825060450b8e31e0ff6f81573756']
2023-01-16T16:14:14.449017             FAILURE - Exception during "get transaction_trace". Exception message: Error 3200005: http request fail
Error Details:
Error code 404
: {"code":404,"message":"Trace API: transaction id missing in the transaction id log files","error":{"code":0,"name":"","what":"","details":[]}}
```

tests/nodeos_contrl_c_test.py
oschwaldp-oci: Actually can't do the wait here due to all production nodes being shut down.

heifner (approved):

linh2931 (approved):

PR #661, opened 2023-01-21
Feature bls signatures narrow interface

systemzax:
This pull request introduces aggregate signatures (BLS signature scheme) as a protocol feature. It uses Chia's BLS-signatures library, which in turn depends on the Relic toolkit.

It defines the following new host functions :

bls_verify
bls_aggregate_pubkeys
bls_aggregate_sigs
bls_aggregate_verify

as well as new built-in types :

bls_public_key
bls_signature

Unit tests are included in libraries/libfc/test/test_bls.cpp

Corresponding pull request on cdt is available here : https://github.com/AntelopeIO/cdt/pull/92

A smart contract that can be used to test the new functionalities can be found here : https://github.com/systemzax/agg-sig-tests-contract

### DIFF
--- a/libraries/chain/webassembly/crypto.cpp
+++ b/libraries/chain/webassembly/crypto.cpp
@@ -7,6 +7,7 @@
 #include <fc/crypto/sha3.hpp>
 #include <fc/crypto/k1_recover.hpp>
 #include <bn256/bn256.h>
+#include <fc/crypto/bls_utils.hpp>
 
 namespace {
     uint32_t ceil_log2(uint32_t n)
@@ -19,6 +20,8 @@ namespace {
 }
 
 namespace eosio { namespace chain { namespace webassembly {
+
+   using namespace bn256;
 
    void interface::assert_recover_key( legacy_ptr<const fc::sha256> digest,
                                        legacy_span<const char> sig,
@@ -118,7 +121,7 @@ namespace eosio { namespace chain { namespace webassembly {
 
    int32_t interface::alt_bn128_add(span<const char> op1, span<const char> op2, span<char> result ) const {
       if (op1.size() != 64 ||  op2.size() != 64 ||  result.size() < 64 ||
-         bn256::g1_add({(const uint8_t*)op1.data(), 64}, {(const uint8_t*)op2.data(), 64}, { (uint8_t*)result.data(), 64}) == -1)
+         bn256::bn256_g1_add({(const uint8_t*)op1.data(), 64}, {(const uint8_t*)op2.data(), 64}, { (uint8_t*)result.data(), 64}) == -1)
          return return_code::failure;
       return return_code::success;
    }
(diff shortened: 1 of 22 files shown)


systemzax:
The build test for Ubuntu 18 fails because the minimum CMake version required by the Chia bls-signatures library is 3.14, however the build action is set to use 3.10.2. Is it possible to use 3.14+ on Ubuntu 18?

matthewdarwin:
Why do we want to support Ubuntu 18 here?  Since 3.1, leap only supports 20 and up.

libraries/libfc/src/crypto/bls_utils.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+     char* r = (char*) malloc(32);
+
+     rand_bytes(r, 32);
+
+     std::vector<uint8_t> v(r, r+32);
+
+     return bls_private_key(v);
+
+   }
```
greg7mdp, line 20:
```suggestion
   bls_private_key generate() {
     std::vector<uint8_t> v(32);
     rand_bytes((char *)v.begin(), v.size());
     return bls_private_key(std::move(v));
   }
```

greg7mdp: previous code has memory leak

libraries/libfc/test/test_bls.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include <fc/crypto/bls_public_key.hpp>
+#include <fc/crypto/bls_signature.hpp>
+
+#include <fc/crypto/bls_utils.hpp>
+
+#include <fc/exception/exception.hpp>
+
+using std::cout;
+
+//using namespace fc;
```
greg7mdp, line 15: remove comment

greg7mdp (changes requested):

spoonincode:
> Why do we want to support Ubuntu 18 here? Since 3.1, leap only supports 20 and up.

Ubuntu 18 is still supported in leap 3.1 and 3.2.

libraries/libfc/include/fc/crypto/bls_signature.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+} } }  // fc::crypto::blslib
+
+namespace fc {
+   void to_variant(const crypto::blslib::bls_signature& var, variant& vo, const fc::yield_function_t& yield = fc::yield_function_t());
+
+   void from_variant(const variant& var, crypto::blslib::bls_signature& vo);
+} // namespace fc
+
+FC_REFLECT(fc::crypto::blslib::bls_signature, (_sig) )
```
spoonincode, line 53: The encoding of `bls_signature`, `bls_pubic_key` and `bls_private_key` are all a variable length `vector` but that doesn't seem appropriate for these as they're always fixed 96, 48 and 32 bytes respectfully, from my understanding. The storage should probably be replaced with the appropriate sized `fc::array<char, x>` so they are encoded without the needless length prefix.

libraries/chain/webassembly/crypto.cpp
```diff
@@ -246,4 +249,88 @@ namespace eosio { namespace chain { namespace webassembly {
       return return_code::success;
    }
 
+   bool interface::bls_verify( span<const char> signature, span<const char> digest, span<const char> pub) const {
```
spoonincode, line 252: From what I can tell `digest` here is a misnomer: it's actually a message.

It's not clear to me if there would be utility in a separate host function for mapping the message to the group, so that this host function would already take only a mapped point.

As a message, a couple things:

There is no need to stuff the message in a `vector` and then unpack that, the `span` type already defines the length of the message because the interface deconstructs to a `char* message, size_t message_size` effectively.

Any host function that processes an arbitrary length of bytes needs to be mindful about taking too long because host functions aren't interruptible. You'll see in, for example, the `sha256` host function how it periodically checks to see if it should abort due to time constraints. Since this function is hashing an arbitrary length message, you will need to do one of:
* periodically check for time. This is ideal, but implies changes down to Relic I assume. Or,
* Add some sort of subjective check to the size of the message (see the `mod_exp` host function as a good recent example)

libraries/chain/webassembly/crypto.cpp
```diff
@@ -246,4 +249,88 @@ namespace eosio { namespace chain { namespace webassembly {
...
+      std::cerr << u_pub.to_string() << "\n";
+      std::cerr << u_sig.to_string() << "\n";
+
+      bool result = fc::crypto::blslib::bls_utils::verify(u_pub, u_digest, u_sig);
+
+      return result;
+
+   }
+
+   int32_t interface::bls_aggregate_pubkeys( span<const char> pubkeys, span<char> aggregate) const {
```
spoonincode, line 276: Similar comment for all these aggregate host functions; unless we can handle ~1 million inputs (33MiB worth of inputs) in 1-2 milliseconds, we need to checktime (or add a subjective limit).

libraries/libfc/include/fc/crypto/bls_signature.hpp
greg7mdp: I agree, but why `fc::array` and not `std::array` ?

spoonincode: I was probably thinking `fc::array<>` just because it'd be consistent with the other public_key/private_key/signature types, but `std::array<>` likely works too

libraries/libfc/src/crypto/bls_utils.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+   bool verify( const bls_public_key &pubkey,
+                  const std::vector<uint8_t> &message,
+                  const bls_signature &signature){
+
+      return bls::PopSchemeMPL().Verify(pubkey._pkey, message, signature._sig);
```
spoonincode, line 26: A big mistake/oversight we made in 1.0 with `recover_key` host function was that given bad input (bad signature) it would throw an exception which fails the transaction: it isn't possible for the contract to handle that failure somehow. Devs have asked for that to be fixed for years and it still isn't (entirely).

It appears this implementation has the same kind of flaw. This will at least do a `G1Eelement::FromBytes(pubkey)` and `G2Element::FromBytes(signature)`. Both of those will throw an exception on various conditions of the input. I didn't look deeper in to the actual verification.

Admittedly, I'm not familiar enough with the kind of use cases for this host function to know if there are use cases for it to be "non throwing"

systemzax: @arhag suggested we broaden the scope of this feature to expose lower level cryptographic primitives (g1/g2 add, multiply, exponentiation and pairing functions). I believe this would offer a satisfactory solution to this issue, since it would allow the implementation of custom verification or aggregation functions.

spoonincode: yep I was going to mention that -- it seems like we'd want the exposed operations to be as granular as the proposed EIP-2537 so that Antelope's EVM can easily make use of them too

PR #663, opened 2023-01-23
SHiP: add delay when `accept()` fails with `too_many_files_open`

### LINKED ISSUE #470: state_history_plugin needs cleaner behavior when `accept()` fails with `too_many_files_open`
When state_history_plugin's `accept()` fails with `too_many_files_open` it causes nodeos' main thread to enter an event storm that **does not** seem to clear on its own (nodeos' main thread will remain at 100% CPU until its clean termination even if existing and pending ship connections are closed). Some parts of nodeos' operation appear to be impaired during this storm. This behavior needs to be improved so the main thread doesn't run away in this condition. I would recommend using some sort of backoff for retrying the `async_accept()` when it completes with `too_many_files_open`. I would also consider modifying the code to not call `async_accept()` again unless the prior call completed with either no error or `too_many_files_open` (there may be other ways the acceptor can enter a failed state).

Since state_history_plugin warns that `Caution: only expose this port to your internal network.` I'm not classifying this a security defect. It is a significant local DoS though, and ugly, so I'm classifying it as a defect.

greg7mdp:
Resolves #470 .

This PR add:
* When `async_accept()` runs out of file descriptors (`Too many open files` error), add an `async_timer`  of 200ms before retrying, so we don't have a cpu loop using 100% cpu.
* When `async_accept()` returns an error other than `Too many open files`, log the error and resume doing `accept()`, so we can accept other connections.
* Minor code cleanup.


heifner (approved):

greg7mdp:
@linh2931 do you think you can approve this PR? It is ready to merge. Thank you!

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -84,7 +84,23 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
+   };
+   
+   struct unix_acceptor {
+      using socket_type = unixs::acceptor::protocol_type::socket;
+      unix_acceptor(boost::asio::io_context& ioc) : acceptor_(ioc), socket_(ioc), error_timer_(ioc) {}
+      unixs::acceptor             acceptor_;
+      socket_type                 socket_;
+      boost::asio::deadline_timer error_timer_;
+   };
+
```
spoonincode, line 102: It may be possible to collapse the above in to a templated class; both `ip::tcp` and `local::stream_protocol` have `acceptor` and `socket` typedefs

spoonincode (approved):

plugins/state_history_plugin/state_history_plugin.cpp
greg7mdp: Excellent Matt, I checked in the change!

PR #664, opened 2023-01-23
add missing include statements for gcc13 (libstdc++13)

spoonincode:
Some fixes when compiling with gcc13 (really libstdc++13 I take it). Seems stdint.h is no longer being included in some standard header, and these files missed including it as they should.

Needs AntelopeIO/abieos#17

### DIFF
--- a/libraries/chain/include/eosio/chain/wast_to_wasm.hpp
+++ b/libraries/chain/include/eosio/chain/wast_to_wasm.hpp
@@ -1,6 +1,7 @@
 #pragma once
 #include <vector>
 #include <string>
+#include <stdint.h>
 
 namespace eosio { namespace chain {
 


libraries/chain/include/eosio/chain/wast_to_wasm.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 #include <vector>
 #include <string>
+#include <stdint.h>
```
heifner, line 4: Why not `<cstdint>` ?

spoonincode: I was trying to simply add the missing header: not refactor code in any way. In this case there is no `std` usage in here so `stdint.h` seemed like the more correct choice.

heifner (approved):

linh2931 (approved):

PR #665, opened 2023-01-24
net_plugin: add delay when `accept()` fails with `too_many_files_open`

### LINKED ISSUE #654: Investigate error processing of `async_accept` in net_plugin

greg7mdp:
Resolves #654 .

This PR add a delay (100ms) when async_accept() runs out of file descriptors (Too many open files error), using an async_timer, so we don't have a cpu loop using 100% cpu. The shorter delay (100ms vs 200ms for `SHiP` and 500ms for `http_plugin`) reflects the fact that `net_plugin` responsiveness in accepting connections is more critical than `http_plugin`.

Also small refactor of `plugin_shutdown`, so we can call it within a catch block in a member function of `net_plugin_impl`.


heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2412,8 +2416,29 @@ namespace eosio {
...
+                        try {
+                           accept_error_timer = std::make_unique<boost::asio::deadline_timer>(my_impl->thread_pool->get_executor());
+                        } catch(...) {
+                           this->plugin_shutdown();
+                           throw;
+                        }
+                     }
+                     if (accept_error_timer) {
+                        // using shorter 50ms timer than SHiP or http_plugin as net_pluging is more critical
+                        accept_error_timer->expires_from_now(boost::posix_time::milliseconds(50));
```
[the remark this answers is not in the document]
greg7mdp: will update, thanks!

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2412,8 +2416,29 @@ namespace eosio {
...
                switch (ec.value()) {
+                  case EMFILE: // same as boost::system::errc::too_many_files_open
+                  {
+                     // no file descriptors available to accept the connection. Wait on async_timer
+                     // and retry listening
+                     if (!accept_error_timer) {
+                        try {
+                           accept_error_timer = std::make_unique<boost::asio::deadline_timer>(my_impl->thread_pool->get_executor());
+                        } catch(...) {
+                           this->plugin_shutdown();
```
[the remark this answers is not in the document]
greg7mdp: OK, implemented this.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2412,8 +2416,29 @@ namespace eosio {
...
+                           throw;
+                        }
+                     }
+                     if (accept_error_timer) {
```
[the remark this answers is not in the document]
greg7mdp: needed now since I implemented your suggestion below and don't rethrow.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2412,8 +2416,29 @@ namespace eosio {
...
                // For the listed error codes below, recall start_listen_loop()
                switch (ec.value()) {
+                  case EMFILE: // same as boost::system::errc::too_many_files_open
+                  {
+                     // no file descriptors available to accept the connection. Wait on async_timer
+                     // and retry listening
+                     if (!accept_error_timer) {
+                        try {
+                           accept_error_timer = std::make_unique<boost::asio::deadline_timer>(my_impl->thread_pool->get_executor());
+                        } catch(...) {
```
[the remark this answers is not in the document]
greg7mdp: added a `elog` statement.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2412,8 +2416,29 @@ namespace eosio {
...
+                           elog( "net_plugin_impl::start_listen_loop() - exception when allocating memory" );
```
heifner, line 2427: Should be `fc_elog( logger` in `net_plugin`. Feel free to fix the couple of other wrong uses of `elog` and one use of `wlog` in `net_plugin`.

greg7mdp: how is the `logger` symbol accessible gere since we don't capture `this`?

https://github.com/AntelopeIO/leap/blob/98dea84c0a4ab6319080df48d6d24d7f8561f3a3/plugins/net_plugin/net_plugin.cpp#L3840

heifner: `logger` is static to the file.

heifner (approved):

plugins/net_plugin/net_plugin.cpp
greg7mdp: oh my bad I missed the closing brace and thought it was in `net_plugin_impl`. Thanks!

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2412,8 +2416,29 @@ namespace eosio {
                fc_elog( logger, "Error accepting connection: ${m}", ("m", ec.message()));
                // For the listed error codes below, recall start_listen_loop()
                switch (ec.value()) {
+                  case EMFILE: // same as boost::system::errc::too_many_files_open
+                  {
+                     // no file descriptors available to accept the connection. Wait on async_timer
+                     // and retry listening
+                     if (!accept_error_timer) {
+                        try {
+                           accept_error_timer = std::make_unique<boost::asio::deadline_timer>(my_impl->thread_pool->get_executor());
```
spoonincode, line 2425: It seems like it'd be easier to create the timer at the same time the thread pool is created, so you don't need this try/catch/quit here.

fwiw I think this is something unfortunate about the interface of `named_thread_pool`. I would have liked the ability to use it like:
```c++
struct stuff {
   named_thrad_pool pool{"stuff"};
   boost::asio::deadline_timer timer{pool.get_executor()};
};
```
This way no optionals are needed for the thread pool nor consumers of its io_context.

Bad sadly this isn't doable with the current design; I think we'd need to completely disallow auto start & move the number of threads to `start()`

heifner: This makes that possible: https://github.com/AntelopeIO/leap/pull/632

spoonincode: That still requires the number of threads at ctor time, which means it typically can't be used as a member in a plugin_impl because the number of threads typically isn't known until plugin_startup time after configs have been parsed due to it being configurable. It seems like it'd need to be `start(num_threads)` or such instead.

heifner: Ah, good point, I'll make that change.

greg7mdp: Thanks @spoonincode and @heifner , I made the change to create the timer at the same time the thread pool is created, and it is indeed simpler.

heifner (approved):
Please update PR description to indicate it is 100ms and not 50ms delay.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2413,8 +2417,19 @@ namespace eosio {
                fc_elog( logger, "Error accepting connection: ${m}", ("m", ec.message()));
                // For the listed error codes below, recall start_listen_loop()
                switch (ec.value()) {
+                  case EMFILE: // same as boost::system::errc::too_many_files_open
+                  {
+                     // no file descriptors available to accept the connection. Wait on async_timer
+                     // and retry listening using shorter 100ms timer than SHiP or http_plugin
+                     // as net_pluging is more critical
+                     accept_error_timer->expires_from_now(boost::posix_time::milliseconds(100));
+                     accept_error_timer->async_wait([self = shared_from_this()]( const boost::system::error_code &ec) {
```
spoonincode, line 2426: Not seeing the need for the shared_ptr capture here

plugins/net_plugin/net_plugin.cpp
```diff
@@ -287,6 +291,8 @@ namespace eosio {
       eosio::chain::named_thread_pool       thread_pool{ "net" };
 
 
+      unique_ptr<boost::asio::deadline_timer> accept_error_timer;
```
spoonincode, line 294: Consider modifying to the new pattern in #632 that doesn't need this is be a a unique_ptr

spoonincode (approved):

plugins/net_plugin/net_plugin.cpp
greg7mdp: right. updated, thanks!

plugins/net_plugin/net_plugin.cpp
greg7mdp: I'd have to move the initialization of `net_plugin_impl` into its constructor (right now it has the default constructor and `deadline_timer` doesn't default construct. It would be a good change, but beyond the scope of this issue. I'd be totally game to cleanup `net_plugin.cpp` but I don't think it is on the table for now.

spoonincode: Admittedly I didn't look too closely, what's the problem with
```c++
eosio::chain::named_thread_pool       thread_pool{ "net" };
boost::asio::deadline_timer accept_error_timer{thread_pool.get_executor();};
```

greg7mdp: Right now these are initialized in `net_plugin::plugin_startup()` while `net_plugin_impl` is constructed in `net_plugin::net_plugin()`. So default-constructing these in the class would change when these are constructed. And `net_plugin::set_program_options()` can change `my->thread_pool_size` which is used in the thread pool constructor. 
So I'm not saying that this cannot be changed, just that it is beyond the original issue I think, and if we change this we probably want to do even more of a cleanup for consistency.

spoonincode: Pretty sure it's just replacing
```c++
unique_ptr<boost::asio::deadline_timer> accept_error_timer;
```
with
```c++
boost::asio::deadline_timer accept_error_timer{thread_pool.get_executor()};
```

but it's your call

greg7mdp: You are right of course, I had not realized that the `named_thread_pool` created its `ioc`  in the constructor. My bad, and thanks for insisting! I will make the change.

PR #667, opened 2023-01-26
[3.1] Test fix: nodeos_forked_chain_lr_test

### LINKED ISSUE #631: Test failure: nodeos_forked_chain_lr_test
https://github.com/AntelopeIO/leap/actions/runs/3883806727/jobs/6625633500

heifner:
Fix #533 should have updated `preKillBlockNum` & `preKillBlockProducer` as these are used in the validation calculation.

Ran 20 times without error in CI/CD:
https://github.com/AntelopeIO/leap/actions/runs/4016264242
https://github.com/AntelopeIO/leap/actions/runs/4016267469

Resolves #631

### DIFF
--- a/tests/nodeos_forked_chain_test.py
+++ b/tests/nodeos_forked_chain_test.py
@@ -375,6 +375,8 @@ try:
     if preKillBlockProducer == "defproducerj" or preKillBlockProducer == "defproducerk":
         # wait for defproduceri so there is plenty of time to send kill before defproducerk
         nonProdNode.waitForProducer("defproduceri")
+        preKillBlockNum=nonProdNode.getBlockNum()
+        preKillBlockProducer=nonProdNode.getBlockProducerByNum(preKillBlockNum)
     Print("preKillBlockProducer = {}".format(preKillBlockProducer))
     # kill at last block before defproducerl, since the block it is killed on will get propagated
     killAtProducer="defproducerk"


oschwaldp-oci (approved):

tests/nodeos_forked_chain_test.py
```diff
@@ -375,6 +375,8 @@ def getBlock(self, blockNum):
     if preKillBlockProducer == "defproducerj" or preKillBlockProducer == "defproducerk":
         # wait for defproduceri so there is plenty of time to send kill before defproducerk
         nonProdNode.waitForProducer("defproduceri")
+        preKillBlockNum=nonProdNode.getBlockNum()
+        preKillBlockProducer=nonProdNode.getBlockProducerByNum(preKillBlockNum)
```
linh2931, line 379: Can we move those two assignments (`preKillBlockNum` and `preKillBlockNum`) out of the `if`, and delete the same assignments on line 373 and 374?

oschwaldp-oci: Honestly, could probably get rid of 373-375 and just make the `waitForProducer` unconditional and get the same result.

heifner: I don't see how. `preKillBlockNum` is used in `nonProdNode.getBlockProducerByNum(preKillBlockNum)` call so it is needed in both places as the value has changed.

heifner: If you always wait that adds a very long wait to the test in a lot of cases it is not needed.

oschwaldp-oci: You're right. I was thinking about that a little incorrectly -- forgetting how many producers were enabled here.  Yeah, I think it is probably fine the way it is, as you need the assignments on 373-4 to use in the conditional 375, and only reassign if proper conditions met.

linh2931: Still do the wait inside the `if`. But calculate preKillBlockNum and preKillBlockProducer after the `if`, so the calculation is only done once.

heifner: The `if` uses `preKillBlockProducer` which is obtained from `nonProdNode.getBlockProducerByNum(preKillBlockNum)` which needs `preKillBlockNum`.

oschwaldp-oci: `preKillBlockProducer` is used in the `if`, how are you thinking about getting around that?

linh2931: Oha, I missed preKillBlockNum is used in the  `if` condition. Sorry.

linh2931 (approved):

PR #683, opened 2023-02-03
named_thread_pool description for debugger

heifner:
I found this useful for recent debugging of crashes and deadlocks, decided would be nice to just have it always baked in.

Add `eosio::name` description to `named_thread_pool` lambda to assist in determining which thread pool is active for a given thread stack trace from a `core` file.

I tried capturing `name_prefix` by value in the lambda but the debugger still showed it as invalid.
Another option, #687,  makes `named_thread_pool` a template instantiated with `eosio::name` so the value would be in the symbol name. 

Release builds with clang11 and lldb.
Before:
```
<snip>
[Inlined] boost::asio::io_context::run() io_context.ipp:63
$_0::operator()() const thread_utils.cpp:27
<unknown> 0x00007f2b3dbcede4
start_thread 0x00007f2b3de6d609
__clone 0x00007f2b3da0a133
```
```
0x926dc68 = {int} 153541736
this = {const (unnamed class) *} 0x5133c28 
 ioc = {boost::asio::io_context &} 
 name_prefix = {std::basic_string<> &} 
   = invalid value
 on_except = {eosio::chain::named_thread_pool::on_except_t} 
 i = {size_t} 1
tn = {std::string} error: summary string parsing error
```

After:
```
<snip>
[Inlined] boost::asio::io_context::run() io_context.ipp:63
$_0::operator()() const thread_utils.cpp:30
<unknown> 0x00007fea901e6de4
start_thread 0x00007fea90485609
__clone 0x00007fea90022133```
```
```
0x926dc68 = {int} 153541736
this = {const (unnamed class) *} 0x436ebb8 
 name_prefix = {eosio::chain::name} 
  value = {uint64_t} 4849507634736267264
 ioc = {boost::asio::io_context &} 
 on_except = {eosio::chain::named_thread_pool::on_except_t} 
 i = {size_t} 1
tn = {std::string} error: summary string parsing error
```
```
 ../../tests/abieos/tools/tools/num2name 4849507634736267264
chain
```

### DIFF
--- a/libraries/chain/thread_utils.cpp
+++ b/libraries/chain/thread_utils.cpp
@@ -1,4 +1,5 @@
 #include <eosio/chain/thread_utils.hpp>
+#include <eosio/chain/name.hpp>
 #include <fc/log/logger_config.hpp>
 #include <fc/exception/exception.hpp>
 
@@ -19,8 +20,15 @@ void named_thread_pool::start( size_t num_threads, on_except_t on_except ) {
    _ioc_work.emplace( boost::asio::make_work_guard( _ioc ) );
    _ioc.restart();
    _thread_pool.reserve( num_threads );
+
+   // capture name since name_prefix is invalid in debugger, use abieos/tools/num2name to get name
+   std::string sanitized_name = fc::to_lower(_name_prefix);
+   sanitized_name = sanitized_name.substr(0, 12);
+   name desc(string_to_uint64_t(sanitized_name));
+
    for( size_t i = 0; i < num_threads; ++i ) {
-      _thread_pool.emplace_back( [&ioc = _ioc, &name_prefix = _name_prefix, on_except, i]() {
+      _thread_pool.emplace_back( [desc, &ioc = _ioc, &name_prefix = _name_prefix, on_except, i]() {
+         (void)desc;
          std::string tn = name_prefix + "-" + std::to_string( i );
          try {
             fc::set_os_thread_name( tn );
--- a/libraries/chain/thread_utils.cpp
+++ b/libraries/chain/thread_utils.cpp
@@ -5,8 +5,8 @@
 
 namespace eosio { namespace chain {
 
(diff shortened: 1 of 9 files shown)


libraries/chain/thread_utils.cpp
```diff
@@ -19,8 +20,15 @@ void named_thread_pool::start( size_t num_threads, on_except_t on_except ) {
...
+
+   // capture name since name_prefix is invalid in debugger, use abieos/tools/num2name to get name
+   std::string sanitized_name = fc::to_lower(_name_prefix);
+   sanitized_name = sanitized_name.substr(0, 12);
+   name desc(string_to_uint64_t(sanitized_name));
+
    for( size_t i = 0; i < num_threads; ++i ) {
-      _thread_pool.emplace_back( [&ioc = _ioc, &name_prefix = _name_prefix, on_except, i]() {
+      _thread_pool.emplace_back( [desc, &ioc = _ioc, &name_prefix = _name_prefix, on_except, i]() {
+         (void)desc;
```
linh2931, line 31: It would be add a comment here that `desc` is just for putting on the stack for debug.

linh2931 (approved):

libraries/chain/thread_utils.cpp
heifner: Added comment

greg7mdp (approved):

libraries/chain/thread_utils.cpp
linh2931: Thanks.

spoonincode:
> I tried capturing `name_prefix` by value in the lambda but the debugger still showed it as invalid. Another option, I think would work, would be to make `named_thread_pool` a template instantiated with `eosio::name` so the value would be in the symbol name. If that would be preferred I could give it a try to see if it works.

That sounds really nice, I think it's worth exploring that.

heifner:
Replaced by #687

PR #687, opened 2023-02-04
named_thread_pool template description for debugger

heifner:
I found this useful for recent debugging of crashes and deadlocks, thought it would be nice to have it baked in.

Add a tag type to `named_thread_pool` to assist in determining which thread pool is active for a given thread stack trace from a core file.

I tried capturing `std::string` `name_prefix` by value in the lambda but the debugger still showed it as invalid. This PR uses a tag type as a template param which makes it part of the symbol type. As seen below this makes it visible in a simple backtrace of the `run_thread` method. 

Alternative to #683

Release builds with clang11 and lldb.

Before:
```
<snip>
[Inlined] boost::asio::io_context::run() io_context.ipp:63
$_0::operator()() const thread_utils.cpp:27
<unknown> 0x00007f2b3dbcede4
start_thread 0x00007f2b3de6d609
__clone 0x00007f2b3da0a133
```
```
0x926dc68 = {int} 153541736
this = {const (unnamed class) *} 0x5133c28 
 ioc = {boost::asio::io_context &} 
 name_prefix = {std::basic_string<> &} 
   = invalid value
 on_except = {eosio::chain::named_thread_pool::on_except_t} 
 i = {size_t} 1
tn = {std::string} error: summary string parsing error
```

After:
```
<snip>

[Inlined] boost::asio::io_context::run() io_context.ipp:63
eosio::chain::named_thread_pool<eosio::chain::controller_impl::chain>::run_thread(unsigned long, const std::function<void (const fc::exception &)> &) thread_utils.hpp:72
<unknown> 0x00007fb9dd0aede4
start_thread 0x00007fb9dd34d609
__clone 0x00007fb9dceea133
```
```
0x926dc68 = {int} 153541736
this = register rax is not available
i = variable not available
on_except = {const on_except_t &} 
tn = {std::string} error: summary string parsing error
offset = variable not available
```
```
 ../../tests/abieos/tools/tools/num2name 4849507634736267264
chain
```


spoonincode:
I think what you need to do is refactor the lambda out in to a member function. This kind of got me what I think you're going for...

```c++
template<uint64_t thread_name = 0>
struct tpool {
   tpool() {}
   ~tpool() {
      if(t.joinable())
         t.join();
   }

   void dothedew() {
      t = std::thread(&tpool::runthread, this);
   }

private:
   void runthread() {
      std::cout << "I'm in the thread!" << std::endl;

      char* p = 0x00;
      std::cout << *p << std::endl;
   }

   std::thread t;
};

int main() {
   tpool<67> t;
   t.dothedew();

   return 0;
}
```

Compiled with gcc -O3,
```
(lldb) bt
* thread #2, name = 'm', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
  * frame #0: 0x0000555555555413 m`tpool<67ul>::runthread() + 83
    frame #1: 0x00007ffff7cd7283 libstdc++.so.6`execute_native_thread_routine at thread.cc:82:18
    frame #2: 0x00007ffff7a9f8fd libc.so.6`___lldb_unnamed_symbol3500 + 717
    frame #3: 0x00007ffff7b21d20 libc.so.6`___lldb_unnamed_symbol3926 + 11
```

heifner:
> I think what you need to do is refactor the lambda out in to a member function. This kind of got me what I think you're going for...

That does produce what I was looking for. However, I don't think it is better than just using `eosio::chain::name` like what is currently in #683.

spoonincode:
I think the main benefit is that it's easy to see with a simple `bt`, where, if I'm understanding #683 right, you have to go fishing a little more with the debugger? Asking someone to just show the output of `bt` is easier than asking them to go fishing, if they don't want to divulge a coredump because it may contain something sensitive (like producing keys). But maybe going fishing in #683 isn't too hard... I'm not entirely sure from what you've written.

heifner:
> I think the main benefit is that it's easy to see with a simple `bt`

That is a good point. I'll update this PR then.

libraries/chain/include/eosio/chain/thread_utils.hpp
```diff
@@ -34,15 +46,64 @@ namespace eosio { namespace chain {
...
+         _ioc.stop();
+         for( auto& t : _thread_pool ) {
+            t.join();
+         }
+         _thread_pool.clear();
+      }
+
+   private:
+      void run_thread( size_t i, const on_except_t& on_except ) {
+         // copying the name_prefix and i into lambda makes it available to be seen on the stack in the debugger
```
spoonincode, line 71: comment could be updated

heifner: Removed

greg7mdp:
Maybe I am missing something, but what about something like this:

```
namespace tp_type {   class chain;   class producer;   class net;   class ship;  };

template<class TP_TYPE>
class named_thread_pool {
    ...
};
```

In the debugger I see:

```
gdb) s
named_thread_pool<tp_type::producer>::set_x (this=0x7fffffffdc94, val=2) at enum_tp.cpp:11
11	   void set_x(int val) { x = val; }
```

heifner:
> Maybe I am missing something, but what about something like this:
> namespace tp_type {   class chain;   class producer;   class net;   class ship;  };

Same idea, just using an `enum` (int) instead of `name`. I prefer `name` as it doesn't require declaring a new `enum` value each time.

greg7mdp:
You can also set the name of the thread like this:

```
#include <thread>
#include <iostream>
#include <pthread.h>
#include <typeinfo>

namespace tp_type {
   class chain;
   class producer;
   class net;
   class ship;
};

template<class TP_TYPE>
struct tpool {
   tpool() {}
   
   ~tpool() {
      if(t.joinable())
         t.join();
   }

   void dothedew() {
      t = std::thread(&tpool::runthread, this);
   }

private:
   void runthread() {
      pthread_setname_np( pthread_self(), typeid(*this).name());

      char* p = 0x00;
      std::cout << *p << std::endl;
   }

   std::thread t;
};

int main() {
   tpool<tp_type::producer> t;
   t.dothedew();
   return 0;
}
```

and in gdb:

```
(gdb) info threads
  Id   Target Id                                  Frame 
  1    Thread 0x7ffff7a4b3c0 (LWP 538266) "a.out" __futex_abstimed_wait_common64 (private=128, cancel=true, abstime=0x0, op=265, expected=538269, 
    futex_word=0x7ffff7a47910) at ./nptl/futex-internal.c:57
* 2    Thread 0x7ffff7a47640 (LWP 538269) "a.out" tpool<tp_type::producer>::runthread (this=0x7fffffffdc80) at enum_tp.cpp:28
```

heifner:
> You can also set the name of the thread like this:

It already does that via `fc::set_os_thread_name`.

greg7mdp:
> It already does that via fc::set_os_thread_name.

Yes, I saw that. I meant tou don't need to pass and save _name_prefix (`_name_prefix(NamePrefix)`)

heifner:
> I'm not sure I understand? What I suggested was not an enum, but a class type, and the name is printed in clear without requiring running the num2name command.

Oh, right, sorry. I read that wrong.

greg7mdp:
With storing `_name_prefix` in clear:

```
#include <thread>
#include <iostream>
#include <string>
#include <pthread.h>
#include <typeinfo>
#include <cxxabi.h>

namespace tp_type { class chain {}; class prod {}; class net {}; class ship {}; };

template<class TP_TYPE>
struct tpool {
   tpool()
   {
      _name_prefix = abi::__cxa_demangle(typeid(TP_TYPE).name(), nullptr, nullptr, nullptr);
      auto offset = _name_prefix.find("::");
      if (offset != std::string::npos)
         _name_prefix.erase(0, offset+2);
   }
   
   ~tpool() {
      t.join();
   }

   void do_it() {
      t = std::thread(&tpool::runthread, this);
   }

private:
   void runthread() {
      pthread_setname_np( pthread_self(), _name_prefix.c_str());
      const char* p = "I'm in the thread!\n";
      std::cout << p;
   }

   std::thread t;
   std::string _name_prefix;
};

int main() {
   tpool<tp_type::prod> t;
   t.do_it();
   return 0;
}

```

spoonincode:
You don't even need to use a completely defined type.

```c++
template<typename T>
struct tpool {
   tpool() {}
   ~tpool() {
      if(t.joinable())
         t.join();
   }

   void dothedew() {
      t = std::thread(&tpool::runthread, this);
   }

private:
   void runthread() {
      const std::string mangled_name = boost::core::demangle(typeid(tpool<T>).name());
      std::cout << "I'm in the thread " << mangled_name.substr(mangled_name.rfind("::")+2, mangled_name.rfind(">") - mangled_name.rfind("::") - 2) << std::endl;

      char* p = 0x00;
      std::cout << *p << std::endl;
   }

   std::thread t;
};

int main() {
   tpool<struct ship> t;
   t.dothedew();

   return 0;
}
```

This will print
```
I'm in the thread ship
```
and `bt`
```
(lldb) bt
* thread #2, name = 'm', stop reason = signal SIGSEGV: invalid address (fault address: 0x0)
  * frame #0: 0x000055555555666c m`tpool<main::ship>::runthread() + 492
    frame #1: 0x00007ffff7cd7283 libstdc++.so.6`execute_native_thread_routine at thread.cc:82:18
    frame #2: 0x00007ffff7a9f8fd libc.so.6`___lldb_unnamed_symbol3500 + 717
    frame #3: 0x00007ffff7b21d20 libc.so.6`___lldb_unnamed_symbol3926 + 11
```

Now whether this is wise or not I'm unsure

spoonincode (approved):

heifner:
> You don't even need to use a completely defined type.

On clang for me the demangle required a defined type, but I have a work around for that. PR updated shortly.

spoonincode:
hm okay what version? The above worked for me on gcc 12.2.1 and clang 15.0.7

heifner:
> hm okay what version? The above worked for me on gcc 12.2.1 and clang 15.0.7

My workaround you already have, which is to use the template type with the embedded tag name and not the tag type directly.

greg7mdp (approved):

PR #688, opened 2023-02-04
Add a named_thread class that includes a strand

heifner:
Create a dedicated class for the use of one named thread. Use a strand for the executor type. Will be useful for the `state_history_plugin` and `prometheus_plugin`.


heifner:
Replaced by #690

PR #690, opened 2023-02-04
Add a named_thread class that includes a strand

heifner:
Create a dedicated class for the use of one named thread. Use a strand for the executor type. Will be useful for the `state_history_plugin` and `prometheus_plugin`.

### DIFF
--- a/libraries/chain/include/eosio/chain/thread_utils.hpp
+++ b/libraries/chain/include/eosio/chain/thread_utils.hpp
@@ -5,6 +5,7 @@
 #include <fc/log/logger_config.hpp>
 #include <boost/asio/io_context.hpp>
 #include <boost/asio/post.hpp>
+#include <boost/asio/strand.hpp>
 #include <future>
 #include <memory>
 #include <optional>
@@ -29,7 +30,7 @@ namespace eosio { namespace chain {
    public:
       using on_except_t = std::function<void(const fc::exception& e)>;
 
-      explicit named_thread_pool()
+      named_thread_pool()
             : _name_prefix(NamePrefix)
             , _ioc()
       {}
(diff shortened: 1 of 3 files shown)


libraries/chain/include/eosio/chain/thread_utils.hpp
```diff
@@ -109,6 +110,33 @@ namespace eosio { namespace chain {
...
+      ///                  if an empty function then logs and rethrows exception on thread which will terminate.
+      /// @throw assert_exception if already started and not stopped.
+      void start( typename named_thread_pool<NamePrefix>::on_except_t on_except ) { _pool.start( 1, std::move( on_except ) ); }
+
+      /// destroy work guard, stop io_context, join thread_pool
+      void stop() { _pool.stop(); }
+
+   private:
+      named_thread_pool<NamePrefix> _pool;
+      decltype(boost::asio::make_strand(_pool.get_executor())) _strand{ boost::asio::make_strand( _pool.get_executor() ) };
```
spoonincode, line 138: I think make_strand() is 1.70+

spoonincode:
I'm not quite sure I see (or understand?) the utility in this class. It seems to be adding ~20 lines of code to save the developer from doing a simple 1 line (that is not really prone to any sort of error or deviation). It seems to be changing
```c++
thread_pool pool;
boost::asio::io_conext::strand strand{pool.get_executor()};
// ...
pool.start(1);
```
to
```c++
named_thread stranded_thread;
// ...
stranded_thread.start();
```

More complex usages will want more flexibility anyways: you use state_history_plugin as an example but that's already one example where more flexibility may be optimal.

heifner:
> I'm not quite sure I see (or understand?) the utility in this class. It seems to be adding ~20 lines of code to save the developer from doing a simple 1 line (that is not really prone to any sort of error or deviation). It seems to be changing

Fair enough, I was thinking it would be helpful to provide the `stand` baked in as it is often not obvious that a `strand` is needed. I'll just close this.

PR #696, opened 2023-02-07
Move get_block abi serialization off the main thread

### LINKED ISSUE #677: http requests fails with timeout on large block (`json.hpp` issue?)
```
$ curl -X POST -d '{"block_num_or_id": 291953778}' http://localhost:8888/v1/chain/get_block
{"code":500,"message":"Internal Service Error","error":{"code":2,"name":"timeout_exception","what":"deadline 2023-01-31T22:14:07.564 exceeded by 914656us ","details":[{"message":"deadline 2023-01-31T22:14:07.564 exceeded by 914656us ","file":"json.hpp","line_number":63,"method":"operator()"}]}}#
```

in [eosq](https://eos.eosq.eosnation.io/block/1166dc72e60565f69a41c668adf227b4bf25988a4f0c36ccd406898ce000b906) block explorer.

The message mentions `json.hpp:63`, which is in a `string to_string()` function. Maybe we have an issue where creating json output takes excessively long for large trees?

heifner:
Move the abi serialization of `signed_block` for `/v1/chain/get_block` off the main thread. Also the abi cache used in this implementation actually provides a nice performance improvement. See comments on #677. 

This PR also uses the min of `--http-max-response-time-ms` &  `--abi-serializer-max-time-ms` for the deadline of `get_block` instead of running for the full `--abi-serializer-max-time-ms` before checking `--http-max-response-time-ms`.

Also includes reduction of string copies in `abi_serializer`.

Resolves #677

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -149,29 +149,29 @@ namespace eosio { namespace chain {
       variants.clear();
       action_results.clear();
 
-      for( auto&& st : abi.structs )
+      for( auto& st : abi.structs )
          structs[st.name] = std::move(st);
 
-      for( auto&& td : abi.types ) {
+      for( auto& td : abi.types ) {
          EOS_ASSERT(!_is_type(td.new_type_name, ctx), duplicate_abi_type_def_exception,
                     "type already exists", ("new_type_name",impl::limit_size(td.new_type_name)));
-         typedefs[td.new_type_name] = std::move(td.type);
+         typedefs[std::move(td.new_type_name)] = std::move(td.type);
       }
 
-      for( auto&& a : abi.actions )
-         actions[a.name] = std::move(a.type);
+      for( auto& a : abi.actions )
+         actions[std::move(a.name)] = std::move(a.type);
 
-      for( auto&& t : abi.tables )
-         tables[t.name] = std::move(t.type);
+      for( auto& t : abi.tables )
+         tables[std::move(t.name)] = std::move(t.type);
 
-      for( auto&& e : abi.error_messages )
-         error_messages[e.error_code] = std::move(e.error_msg);
+      for( auto& e : abi.error_messages )
+         error_messages[std::move(e.error_code)] = std::move(e.error_msg);
 
-      for( auto&& v : abi.variants.value )
-         variants[v.name] = std::move(v);
+      for( auto& v : abi.variants.value )
+         variants[std::move(v.name)] = std::move(v);
 
-      for( auto&& r : abi.action_results.value )
-         action_results[r.name] = std::move(r.result_type);
+      for( auto& r : abi.action_results.value )
+         action_results[std::move(r.name)] = std::move(r.result_type);
 
       /**
        *  The ABI vector may contain duplicates which would make it
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -168,7 +168,7 @@ namespace eosio { namespace chain {
          error_messages[std::move(e.error_code)] = std::move(e.error_msg);
 
       for( auto& v : abi.variants.value )
-         variants[std::move(v.name)] = std::move(v);
+         variants[v.name] = std::move(v);
 
       for( auto& r : abi.action_results.value )
          action_results[std::move(r.name)] = std::move(r.result_type);
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -149,8 +149,11 @@ namespace eosio { namespace chain {
       variants.clear();
       action_results.clear();
 
-      for( auto& st : abi.structs )
-         structs[st.name] = std::move(st);
+      for( auto& st : abi.structs ) {
+         // side effect rules indicate std::move can happen before st.name is accessed.
+         auto n = st.name;
+         structs[std::move(n)] = std::move(st);
+      }
 
       for( auto& td : abi.types ) {
          EOS_ASSERT(!_is_type(td.new_type_name, ctx), duplicate_abi_type_def_exception,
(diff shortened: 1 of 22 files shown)


greg7mdp (approved):

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -343,7 +343,7 @@ namespace eosio { namespace testing {
                   const auto& accnt = control->db().get<account_object, by_name>( name );
                   abi_def abi;
                   if( abi_serializer::to_abi( accnt.abi, abi )) {
```
greg7mdp, line 345:
```suggestion
                  if( abi_def abi; abi_serializer::to_abi( accnt.abi, abi )) {
```

greg7mdp: Same change can be done to other locations as well. It helps enfoce the fact that the moved from variable is not reused.

libraries/chain/abi_serializer.cpp
```diff
@@ -141,47 +149,47 @@ namespace eosio { namespace chain {
...
-         structs[st.name] = st;
+      for( auto&& st : abi.structs )
+         structs[st.name] = std::move(st);
 
-      for( const auto& td : abi.types ) {
+      for( auto&& td : abi.types ) {
          EOS_ASSERT(!_is_type(td.new_type_name, ctx), duplicate_abi_type_def_exception,
                     "type already exists", ("new_type_name",impl::limit_size(td.new_type_name)));
-         typedefs[td.new_type_name] = td.type;
+         typedefs[td.new_type_name] = std::move(td.type);
```
greg7mdp, line 158:
```suggestion
         typedefs[std::move(td.new_type_name)] = std::move(td.type);
```

libraries/chain/abi_serializer.cpp
```diff
@@ -141,47 +149,47 @@ namespace eosio { namespace chain {
...
       }
 
-      for( const auto& a : abi.actions )
-         actions[a.name] = a.type;
+      for( auto&& a : abi.actions )
+         actions[a.name] = std::move(a.type);
```
greg7mdp, line 162:
```suggestion
         actions[std::move(a.name)] = std::move(a.type);
```

libraries/chain/abi_serializer.cpp
```diff
@@ -141,47 +149,47 @@ namespace eosio { namespace chain {
...
 
-      for( const auto& t : abi.tables )
-         tables[t.name] = t.type;
+      for( auto&& t : abi.tables )
+         tables[t.name] = std::move(t.type);
```
greg7mdp, line 165:
```suggestion
         tables[std::move(t.name)] = std::move(t.type);
```

libraries/chain/abi_serializer.cpp
```diff
@@ -141,47 +149,47 @@ namespace eosio { namespace chain {
...
 
-      for( const auto& e : abi.error_messages )
-         error_messages[e.error_code] = e.error_msg;
+      for( auto&& e : abi.error_messages )
+         error_messages[e.error_code] = std::move(e.error_msg);
```
greg7mdp, line 168:
```suggestion
         error_messages[std::move(e.error_code)] = std::move(e.error_msg);
```

libraries/chain/abi_serializer.cpp
```diff
@@ -141,47 +149,47 @@ namespace eosio { namespace chain {
...
 
-      for( const auto& v : abi.variants.value )
-         variants[v.name] = v;
+      for( auto&& v : abi.variants.value )
+         variants[v.name] = std::move(v);
 
-      for( const auto& r : abi.action_results.value )
-         action_results[r.name] = r.result_type;
+      for( auto&& r : abi.action_results.value )
+         action_results[r.name] = std::move(r.result_type);
```
greg7mdp, line 174:
```suggestion
         action_results[std::move(r.name)] = std::move(r.result_type);
```

greg7mdp (changes requested):
delaying approval because more strings should be moved (as indicated above).

libraries/chain/abi_serializer.cpp
```diff
@@ -141,47 +149,47 @@ namespace eosio { namespace chain {
       variants.clear();
       action_results.clear();
 
-      for( const auto& st : abi.structs )
-         structs[st.name] = st;
+      for( auto&& st : abi.structs )
```
greg7mdp, line 152: Not 100% sure, but I don't think `auto&&` is required, `auto&` is probably enough.

heifner: `auto&` is enough, but I think `auto&&` is clearer intent.

greg7mdp: Hum. `st` will be a lvalue anyways. `auto&& st = <expr>` would be useful only if <expr> was (or could be) a rvalue.

In this case `abi.structs` is a lvalue and I don't believe that just declaring the loop variable `auto&&` does much, but then again it doesn't hurt either so imho it is fine to leave as is.

heifner: On second thought, `auto&` is likely clearer intent here.

libraries/chain/abi_serializer.cpp
```diff
@@ -141,47 +149,47 @@ namespace eosio { namespace chain {
...
 
-      for( const auto& e : abi.error_messages )
-         error_messages[e.error_code] = e.error_msg;
+      for( auto& e : abi.error_messages )
+         error_messages[std::move(e.error_code)] = std::move(e.error_msg);
 
-      for( const auto& v : abi.variants.value )
-         variants[v.name] = v;
+      for( auto& v : abi.variants.value )
+         variants[std::move(v.name)] = std::move(v);
```
greg7mdp, line 171: `v.name` should not be moved here because v is moved

unittests/abi_tests.cpp
```diff
@@ -551,8 +551,9 @@ BOOST_AUTO_TEST_CASE(general)
 { try {
 
    auto abi = eosio_contract_abi(fc::json::from_string(my_abi).as<abi_def>());
+   auto abi2 = abi;
 
-   abi_serializer abis(abi, abi_serializer::create_yield_function( max_serialization_time ));
+   abi_serializer abis(std::move(abi), abi_serializer::create_yield_function( max_serialization_time ));
```
greg7mdp, line 556: Instead of introducing abi2, why not:

`abi_serializer abis(abi_def(abi), abi_serializer::create_yield_function( max_serialization_time ));`

libraries/chain/abi_serializer.cpp
greg7mdp: weird - the diff shown is old. The current code doesn't match.

greg7mdp (approved):

libraries/chain/abi_serializer.cpp
heifner: GitHub is bad about that. Often you have to hit refresh.

greg7mdp (approved):

linh2931 (approved):

PR #628, opened 2023-01-09
Merge SHiP stability fixes & log splitting & memory usage improvements

heifner:
Merges #638 on top of #577 & #599. Also includes some additional misc. cleanup.

Addresses the double memory used when starting from a snapshot via merge of #577.

This PR backports https://github.com/EOSIO/eos/pull/9277 and https://github.com/EOSIO/eos/pull/9821 to support supports the trace and chain state log spliting in state history plugin. 

Some new options for state history plugin are added:

- `state-history-stride`: split the state history log files when the block number is the multiple of the stride. When the stride is reached, the current history log and index will be renamed '*-history--.log/index' and a new current history log and index will be created with the most recent blocks. All files following this format will be used to construct an extended history log.
- `max-retained-history-files`: the maximum number of state history log file groups to retain so that the blocks in those files can be queried. When the number is reached, the oldest log files would be moved to archive dir or deleted if the archive dir is empty. The retained log files should not be manipulated by users.
- `state-history-retained-dir`: the location of the state history retained directory (absolute path or relative to state-history dir). 
- `state-history-archive-dir`: the location of the state history archive directory (absolute path or relative to state-history dir). If the value is empty string, log files beyond the retained limit will be deleted. All files in the archive directory are completely under user's control, i.e. they won't be accessed by nodeos anymore.

Verified this branch also fixes #592.

PR #727, opened 2023-02-21
SHiP move client handling off the main thread

### LINKED ISSUE #627: refactor thread and asio usage in state history plugin
Some usage of asio in state history plugin looks at odds with typical asio patterns. For example: needing a connection to call an explicit `close()` function after a failure occurs (normally you'd just let the shared_ptr fall out of scope to dtor it), and some sort of `stopping` variable (normally you'd just `stop()` the io_context which then ensures no callbacks are running, and then just let everything dtor itself). To be fair there may be yet-undiscovered-to-me reasons these are required, but so far I haven't seen it.

state history plugin is also very loose and haphazard with how it handles its thread, oftentimes punting back and forth between main thread. Yet, it seems like state history plugin's thread usage could be awfully simplistic: build up blocks' traces & deltas on the main thread and then when `on_accepted_block()` simply `post()` over to ship's thread which then _does everything it needs to entirely on the ship thread without ever needing to go back to main thread_.

Resolving the above issues will make it significantly easier to reason about state history plugin's flow and will be less prone to introducing failures.

heifner:
Moves client processing of `state_history_plugin` requests off the main application thread. 
- `block_log` has been made thread safe.
- `net_plugin` updated to use thread safe `block_log` from net threads instead of having to post onto the main thread for processing.
- SHiP writes to the state history logs on the main thread, but reads from them on the SHiP thread. The state history logs have been made thread safe.

Verified this branch does not crash under heavy load #592 
Resolves #627


libraries/chain/block_log.cpp
```diff
@@ -1184,46 +1186,79 @@ namespace eosio { namespace chain {
...
    void     block_log::set_initial_version(uint32_t ver) { detail::block_log_impl::default_initial_version = ver; }
-   uint32_t block_log::version() const { return my->version(); }
+   uint32_t block_log::version() const {
+      std::lock_guard g(my->mtx);
+      return my->version();
+   }
 
-   void block_log::append(const signed_block_ptr& b, const block_id_type& id) { my->append(b, id, fc::raw::pack(*b)); }
+   void block_log::append(const signed_block_ptr& b, const block_id_type& id) {
+      std::vector<char> packed_block = fc::raw::pack(*b);
```
linh2931, line 1200: Should this line be protected too? Would someone modify `b` while `pack` is running?

libraries/chain/controller.cpp
```diff
@@ -392,14 +392,15 @@ struct controller_impl {
...
       EOS_ASSERT( fork_db.root(), fork_database_exception, "fork database not properly initialized" );
 
-      const auto& log_head = blog.head();
+      block_id_type log_head_id = blog.head_id();
+      const bool valid_log_head = !log_head_id.empty();
 
-      auto lib_num = log_head ? log_head->block_num() : (blog.first_block_num() - 1);
+      auto lib_num = valid_log_head ? block_header::num_from_id(log_head_id) : (blog.first_block_num() - 1);
 
       auto root_id = fork_db.root()->id;
```
linh2931, line 400: You can add `const` to other local variables in this function as they are not modified after initialized.

libraries/chain/controller.cpp
```diff
@@ -3120,11 +3121,6 @@ block_state_ptr controller::fetch_block_state_by_number( uint32_t block_num )con
 } FC_CAPTURE_AND_RETHROW( (block_num) ) }
```
linh2931, line 3122: Maybe add a comment saying this is thread-safe.

libraries/chain/block_log.cpp
heifner: We would have to protect all access to signed_block to guard against that. We already assume that signed_block is not modified after finalization. We already use signed_block on other threads after finalization.

libraries/chain/controller.cpp
heifner: There is a comment in the header that says that.

libraries/chain/controller.cpp
heifner: Done.

linh2931 (approved):

766C6164 (approved):

greg7mdp:
@heifner I've not started reviewing this yet, and I see you have 2 approvals, so feel free to merge it if you'd like. I don't mind having a look either if you'd like.

heifner:
> @heifner I've not started reviewing this yet, and I see you have 2 approvals, so feel free to merge it if you'd like. I don't mind having a look either if you'd like.

This has to wait on PR #628 which is waiting on PR #532. You can review if you like, I'm good either way.

PR #731, opened 2023-02-21
[PH] Add TransactionGeneratorLauncher to TestHelper Shutdown

ClaytonCalabrese:
Utilizes the TransactionGeneratorLauncher's `killAll` in TestHelper `shutdown` assuming a TransactionGeneratorLauncher is passed in. This allows the removal of a now unused `CompletedRun` local variable.

### DIFF
--- a/tests/TestHarness/TestHelper.py
+++ b/tests/TestHarness/TestHelper.py
@@ -149,7 +149,7 @@ class TestHelper(object):
     
     @staticmethod
     # pylint: disable=too-many-arguments
-    def shutdown(cluster, walletMgr, trxGenLauncher=None, testSuccessful=True, killEosInstances=True, killWallet=True, keepLogs=False, cleanRun=True, dumpErrorDetails=False):
+    def shutdown(cluster, walletMgr, testSuccessful=True, killEosInstances=True, killWallet=True, keepLogs=False, cleanRun=True, dumpErrorDetails=False, trxGenLauncher=None):
         """Cluster and WalletMgr shutdown and cleanup."""
         assert(cluster)
         assert(isinstance(cluster, Cluster))
--- a/tests/TestHarness/TestHelper.py
+++ b/tests/TestHarness/TestHelper.py
@@ -1,6 +1,7 @@
 from .testUtils import Utils
 from .Cluster import Cluster
 from .WalletMgr import WalletMgr
+from .launch_transaction_generators import TransactionGeneratorsLauncher
 from datetime import datetime
 import platform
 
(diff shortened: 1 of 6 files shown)


tests/performance_tests/performance_test_basic.py
```diff
@@ -493,16 +492,13 @@ def runTest(self) -> bool:
                 self.testHelperConfig._killWallet,
                 self.testHelperConfig.keepLogs,
                 self.testHelperConfig.killAll,
-                self.testHelperConfig.dumpErrorDetails
+                self.testHelperConfig.dumpErrorDetails,
+                trxGenLauncher=self.cluster.trxGenLauncher
```
oschwaldp-oci, line 496: `TestHelper` already knows about `cluster`, and now that you've switched over to using the `self.cluster.trxGenLauncher` instance, it would seem to follow the pattern here to add a `self.testHelperConfig.killAllTrxGen` flag instead of passing in the `trxGenLauncher` itself.

Although, now that I'm looking at it more closely, is adding this to `TestHelper.shutdown` even necessary since `cluster.killall` will clean up the `self.trxGenLauncher` if the `self.testHelperConfig.killAll` flag is set?

tests/performance_tests/log_reader.py
```diff
@@ -10,7 +10,7 @@
 from pathlib import Path, PurePath
 sys.path.append(str(PurePath(PurePath(Path(__file__).absolute()).parent).parent))
 
-from TestHarness import Utils
+from TestHarness import Utils, TransactionGeneratorsLauncher
```
oschwaldp-oci, line 13: This appears to be unused.

tests/TestHarness/TestHelper.py
```diff
@@ -199,3 +202,6 @@ def reportProductionAnalysis(thresholdMs):
                 Utils.Print("Cleanup wallet data.")
                 walletMgr.cleanup()
 
+        if trxGenLauncher is not None:
+            Utils.Print("Shut down the TransactionGeneratorsLauncher and TransactionGenerators")
+            trxGenLauncher.killAll()
```
oschwaldp-oci, line 207: Is this still necessary, now that `trxGenLauncher` is being put on the `cluster` instance, and `cluster.killAll` will clean up the `self.trxGenLauncher`?

oschwaldp-oci (changes requested):

tests/performance_tests/performance_test_basic.py
ClaytonCalabrese: This appears correct. Now that it is set in the cluster, it should be handled since we are using killall.

oschwaldp-oci (approved):

PR #732, opened 2023-02-21
Update leap to use new appbase version.

greg7mdp:
The latest appbase version allows to customize the appbase execution engine by providing an `executor` class.
Besides updating the appbase submodule tip, this PR has a couple small changes to support the new architecture.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -2303,7 +2303,7 @@ void producer_plugin_impl::schedule_production_loop() {
       _timer.expires_from_now( boost::posix_time::microseconds( config::block_interval_us  / 10 ));
 
       // we failed to start a block, so try again later?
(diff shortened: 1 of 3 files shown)


linh2931 (approved):

heifner (approved):
I assume the move away from the deprecated `app().post` will happen in the PR to move to the new executor.

greg7mdp:
> I assume the move away from the deprecated `app().post` will happen in the PR to move to the new executor.

Yes, I think that would be the way to go. @linh2931 let's make sure to replace the `app().post()` that may remain in leap with `app().executor().post()` when you implement your new executor in leap. Thanks!

linh2931:
> Yes, I think that would be the way to go. @linh2931 let's make sure to replace the `app().post()` that may remain in leap with `app().executor().post()` when you implement your new executor in leap. Thanks!

Yes, will do.

PR #736, opened 2023-02-22
SHiP do not truncate logs when starting from genesis

heifner:
Verify when truncating SHiP logs that truncation is not allowed when starting from genesis.

```
warn  2023-02-22T20:51:14.867 nodeos    controller.cpp:606            startup              ] No existing chain state. Initializing fresh blockchain state.
warn  2023-02-22T20:51:14.867 nodeos    controller.cpp:464            initialize_blockchai ] Initializing new blockchain with genesis state
info  2023-02-22T20:51:14.888 nodeos    controller.cpp:501            replay               ] existing block log, attempting to replay from 2 to 214011 blocks
error 2023-02-22T20:51:14.889 nodeos    state_history_plugin.c:284    on_accepted_block    ] fc::exception: 3110000 plugin_exception: Plugin exception
Existing ship log with 214013 blocks when starting from genesis block 2
    {"eb":214013,"b":2}
    nodeos  log.hpp:513 write_entry

warn  2023-02-22T20:51:14.889 nodeos    controller.cpp:375            emit                 ] controller_emit_signal_exception: 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:292 on_accepted_block

error 2023-02-22T20:51:14.890 nodeos    controller.cpp:2127           apply_block          ] e.to_detail_string(): 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:292 on_accepted_block

warn  2023-02-22T20:51:14.890 nodeos    controller.cpp:2287           replay_push_block    ] 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:292 on_accepted_block

    {}
    nodeos  controller.cpp:2135 apply_block

error 2023-02-22T20:51:14.898 nodeos    main.cpp:162                  main                 ] 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:292 on_accepted_block

    {}
    nodeos  controller.cpp:2135 apply_block
rethrow
    {}
    nodeos  controller.cpp:2287 replay_push_block

    {}
[...]

PR #737, opened 2023-02-22
[3.2] SHiP do not truncate logs when starting from genesis

### LINKED ISSUE #659: make it difficult to accidentally perform a large rollback of state history logs
Consider a nodeos set up with state history,
```
plugin = eosio::state_history_plugin
trace-history = true
chain-state-history = true
```
and this nodeos has been operational for months -- years -- simply by just stopping nodeos with ^C and then restarting the same as always
```
nodeos --disable-replay-opts
```
Then, say there is a need to remove nodeos' state. That sounds pretty extreme, why would anyone need to do that?! Somehow we've become so complacent in the fact that Leap requires users to _manually_ remove their state files at certain upgrade inflection points, that I don't even see it mentioned in release notes that upgrading to 3.1/3.2 from 2.x requires this step. At this point it is important to restart nodeos via something like
```
nodeos --disable-replay-opts --snapshot recent_snapshot.bin
```
but what if instead.. oops.. nodeos is just restarted like it always is,
```
nodeos --disable-replay-opts
```

:boom: just erased entire state history logs. Months -- years -- instantly gone.

This is really quite hostile behavior. The hope is that automatic database migration (Coming Soon) will help stop the need for manually removing state, but this is so hostile maybe something additional needs to be considered outside of that feature. It would be curious to understand what trace_api does too.

heifner:
Verify when truncating SHiP logs that truncation is not allowed when starting from genesis.

```
warn  2023-02-22T20:51:14.867 nodeos    controller.cpp:606            startup              ] No existing chain state. Initializing fresh blockchain state.
warn  2023-02-22T20:51:14.867 nodeos    controller.cpp:464            initialize_blockchai ] Initializing new blockchain with genesis state
info  2023-02-22T20:51:14.888 nodeos    controller.cpp:501            replay               ] existing block log, attempting to replay from 2 to 214011 blocks
error 2023-02-22T20:51:14.889 nodeos    state_history_plugin.c:284    on_accepted_block    ] fc::exception: 3110000 plugin_exception: Plugin exception
Existing ship log with 214013 blocks when starting from genesis block 2
    {"eb":214013,"b":2}
    nodeos  log.hpp:513 write_entry

warn  2023-02-22T20:51:14.889 nodeos    controller.cpp:375            emit                 ] controller_emit_signal_exception: 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:292 on_accepted_block

error 2023-02-22T20:51:14.890 nodeos    controller.cpp:2127           apply_block          ] e.to_detail_string(): 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:292 on_accepted_block

warn  2023-02-22T20:51:14.890 nodeos    controller.cpp:2287           replay_push_block    ] 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:292 on_accepted_block

    {}
    nodeos  controller.cpp:2135 apply_block

error 2023-02-22T20:51:14.898 nodeos    main.cpp:162                  main                 ] 3140002 state_history_write_exception: State history write error
State history encountered an Error which it cannot recover from.  Please resolve the error and relaunch the process
    {}
    nodeos  state_history_plugin.cpp:292 on_accepted_block

    {}
    nodeos  controller.cpp:2135 apply_block
rethrow
    {}
    nodeos  controller.cpp:2287 replay_push_block

    {}
    nodeos  chain_plugin.cpp:1141 plugin_startup
```

Starting from snapshot is allowed:
```
./nodeos --data-dir telos-d --config-dir telos --disable-replay-opts --plugin eosio::chain_api_plugin --plugin eosio::producer_api_plugin --snapshot /home/heifner/ext/leap/cmake-build-debug/bin/telos-d/snapshots/snapshot-0003068e2e7b2030a54466f36af0df3eb47c75d5a7ef4ffe45e702fe07b6884e.bin
```
```
info  2023-02-22T20:53:27.566 nodeos    controller.cpp:571            startup              ] Starting initialization from snapshot, this may take a significant amount of time
info  2023-02-22T20:53:34.702 nodeos    controller.cpp:501            replay               ] existing block log, attempting to replay from 198287 to 214011 blocks
info  2023-02-22T20:53:34.769 nodeos    log.hpp:793                   truncate             ] fork or replay: removed 15726 blocks from trace_history.log
info  2023-02-22T20:53:34.770 nodeos    log.hpp:793                   truncate             ] fork or replay: removed 15726 blocks from chain_state_history.log
info  2023-02-22T20:53:34.943 nodeos    controller.cpp:507            replay               ] 198500 of 214011
```

Test added to `main` branch, see #736 

Resolves #659

### DIFF
--- a/libraries/state_history/include/eosio/state_history/log.hpp
+++ b/libraries/state_history/include/eosio/state_history/log.hpp
@@ -216,12 +216,15 @@ class state_history_log {
          }
       }
 
-      if (block_num < _end_block)
+      if (block_num < _end_block) {
+         EOS_ASSERT( block_num > 2, chain::plugin_exception, "Existing ship log with ${eb} blocks when starting from genesis block ${b}",
+                     ("eb", _end_block)("b", block_num) );
          truncate(block_num); //truncate is expected to always leave file pointer at the end
-      else if (!prune_config)
+      } else if (!prune_config) {
          log.seek_end(0);
-      else if (prune_config && _begin_block != _end_block)
+      } else if (prune_config && _begin_block != _end_block) {
          log.seek_end(-sizeof(uint32_t));  //overwrite the trailing block count marker on this write
+      }
 
       //if we're operating on a pruned block log and this is the first entry in the log, make note of the feature in the header
       if(prune_config && _begin_block == _end_block)
--- a/libraries/state_history/include/eosio/state_history/log.hpp
+++ b/libraries/state_history/include/eosio/state_history/log.hpp
@@ -217,9 +217,16 @@ class state_history_log {
       }
 
       if (block_num < _end_block) {
+         static uint32_t start_block_num = block_num;
          EOS_ASSERT( block_num > 2, chain::plugin_exception, "Existing ship log with ${eb} blocks when starting from genesis block ${b}",
                      ("eb", _end_block)("b", block_num) );
-         truncate(block_num); //truncate is expected to always leave file pointer at the end
+         if ( get_block_id_i(block_num) != header.block_id ) {
+            truncate(block_num); //truncate is expected to always leave file pointer at the end
+         } else {
+            if (start_block_num == block_num || block_num % 1000 == 0 )
+               ilog("log ${name}.log already contains block ${b}, end block ${eb}", ("name", name)("b", block_num)("eb", _end_block));
+            return;
+         }
       } else if (!prune_config) {
          log.seek_end(0);
       } else if (prune_config && _begin_block != _end_block) {
@@ -256,22 +263,31 @@ class state_history_log {
 
    // returns cfile positioned at payload
    fc::cfile& get_entry(uint32_t block_num, state_history_log_header& header) {
+      std::lock_guard lock(mx);
+      return get_entry_i(block_num, header);
+   }
+
+   chain::block_id_type get_block_id(uint32_t block_num) {
+      std::lock_guard lock(mx);
+      return get_block_id_i(block_num);
+   }
+
+ private:
+
+   fc::cfile& get_entry_i(uint32_t block_num, state_history_log_header& header) {
       EOS_ASSERT(block_num >= _begin_block && block_num < _end_block, chain::plugin_exception,
                  "read non-existing block in ${name}.log", ("name", name));
-      std::lock_guard lock(mx);
       log.seek(get_pos(block_num));
       read_header(header);
       return log;
    }
 
-   chain::block_id_type get_block_id(uint32_t block_num) {
-      std::lock_guard          lock(mx);
+   chain::block_id_type get_block_id_i(uint32_t block_num) {
       state_history_log_header header;
-      get_entry(block_num, header);
+      get_entry_i(block_num, header);
       return header.block_id;
    }
 
- private:
    //file position must be at start of last block's suffix (back pointer)
    //called from open_log / ctor 
    bool get_last_block() {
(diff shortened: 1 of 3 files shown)


spoonincode:
I don't think this solution is as generic as the solution that was discussed, agreed on, and placed in the issue
> Compare snapshot block id to block id of appropriate entry within SHiP logs to determine whether nodeos should automatically truncate the SHiP logs

More specifically, I thought we decided that before ship truncates the log it should compare blockids for the "new" block to the existing block in the log, at the given block num. If they match: do nothing. Such a check would be more robust against accidental rollback at any point, and not just at genesis. (though, rollback to genesis is likely the easier one for a user to stumble upon)

heifner:
> More specifically, I thought we decided that before ship truncates the log it should compare blockids for the "new" block to the existing block in the log, at the given block num. If they match: do nothing. Such a check would be more robust against accidental rollback at any point, and not just at genesis. (though, rollback to genesis is likely the easier one for a user to stumble upon)

Thanks. Added in: 4b7e8274c7724ef6c46fec19a982b6772c3c3422

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -216,12 +216,22 @@ class state_history_log {
...
       }
 
-      if (block_num < _end_block)
-         truncate(block_num); //truncate is expected to always leave file pointer at the end
-      else if (!prune_config)
+      if (block_num < _end_block) {
+         static uint32_t start_block_num = block_num;
+         EOS_ASSERT( block_num > 2, chain::plugin_exception, "Existing ship log with ${eb} blocks when starting from genesis block ${b}",
+                     ("eb", _end_block)("b", block_num) );
+         if ( block_num < _begin_block || get_block_id_i(block_num) != header.block_id ) {
```
heifner, line 223: Add comments 
- about why assert - start from genesis, don't want to truncate large ship logs
- `block_num < _begin_block` - this a pruned log and we need to call truncate to allow appending
- `get_block_id_i` - if we already have the block then no need to remove it from log only to re-write it to the log

libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -216,12 +216,22 @@ class state_history_log {
          }
       }
 
-      if (block_num < _end_block)
-         truncate(block_num); //truncate is expected to always leave file pointer at the end
-      else if (!prune_config)
+      if (block_num < _end_block) {
```
heifner, line 219: `// this typically indicates a fork happened`

greg7mdp (approved):

libraries/state_history/include/eosio/state_history/log.hpp
heifner: `get_block_id_i` reads the id from the log. It may not match `header.block_id` if a fork occurred. If they match then we are rewriting the same info. If they differ then a fork occurred and we need to overwrite the old data.

heifner: Updated comments.

linh2931 (approved):

PR #740, opened 2023-02-23
[PH] Expose various Nodeos Options in performance_test_basic

ClaytonCalabrese:
Newly Exposed options:
`wasm-runtime`
`contracts-console`
`eos-vm-oc-cache-size-mb`
`eos-vm-oc-compile-threads`
`eos-vm-oc-enable`
`block-log-retain-blocks`
`http-threads`
Removed Options:
`chain-state-db-size-mb`
A few hardcoded values were set in performance_test_basic using he constructors for *Args:
`maxTransactionTime` (max-transaction-time)
`chainStateDbSizeMb` (chain-state-db-size-mb)
`abiSerializerMaxTimeMs` (abi-serializer-max-time-ms)
`httpMaxBytesInFlightMb` (http-max-bytes-in-flight-mb)
`httpMaxResponseTimeMs` (http-max-response-time-ms)
`maxClients` (max-clients)
`resourceMonitorNotShutdownOnThresholdExceeded` (resource-monitor-not-shutdown-on-threshold-exceeded)

Also as part of this task, a previously hardcoded part of cluster has been changed so that `max-transaction-time`, `abi-serializer-max-time-ms`, `p2p-max-nodes-per-host`, and `max-clients` can be passed to nodeos via cluster without causing failures.

### DIFF
--- a/tests/TestHarness/Cluster.py
+++ b/tests/TestHarness/Cluster.py
@@ -257,8 +257,15 @@ class Cluster(object):
         cmdArr=cmd.split()
         if self.staging:
             cmdArr.append("--nogen")
-
-        nodeosArgs="--max-transaction-time -1 --abi-serializer-max-time-ms 990000 --p2p-max-nodes-per-host %d --max-clients %d" % (maximumP2pPerHost, maximumClients)
+        nodeosArgs=""
+        if extraNodeosArgs.find("--max-transaction-time") == -1:
+            nodeosArgs += " --max-transaction-time -1"
+        if extraNodeosArgs.find("--abi-serializer-max-time-ms") == -1:
+            nodeosArgs += " --abi-serializer-max-time-ms 990000"
+        if extraNodeosArgs.find("--p2p-max-nodes-per-host") == -1:
+            nodeosArgs += f" --p2p-max-nodes-per-host {maximumP2pPerHost}"
+        if extraNodeosArgs.find("--max-clients") == -1:
+            nodeosArgs += f" --max-clients {maximumClients}"
         if not self.walletd:
             nodeosArgs += " --plugin eosio::wallet_api_plugin"
         if Utils.Debug:
--- a/tests/performance_tests/performance_test_basic.py
+++ b/tests/performance_tests/performance_test_basic.py
@@ -527,7 +527,6 @@ class PtbArgumentsHandler(object):
         ptbBaseParserGroup.add_argument("--num-blocks-to-prune", type=int, help=("The number of potentially non-empty blocks, in addition to leading and trailing size 0 blocks, "
                                                                 "to prune from the beginning and end of the range of blocks of interest for evaluation."), default=2)
         ptbBaseParserGroup.add_argument("--signature-cpu-billable-pct", type=int, help="Percentage of actual signature recovery cpu to bill. Whole number percentages, e.g. 50 for 50%%", default=0)
-        ptbBaseParserGroup.add_argument("--chain-state-db-size-mb", type=int, help="Maximum size (in MiB) of the chain state database", default=10*1024)
         ptbBaseParserGroup.add_argument("--chain-threads", type=int, help="Number of worker threads in controller thread pool", default=2)
         ptbBaseParserGroup.add_argument("--database-map-mode", type=str, help="Database map mode (\"mapped\", \"heap\", or \"locked\"). \
                                                                 In \"mapped\" mode database is memory mapped as a file. \
@@ -557,6 +556,18 @@ class PtbArgumentsHandler(object):
         ptbBaseParserGroup.add_argument("--contract-dir", type=str, help="Path to contract dir", default="unittests/contracts/eosio.system")
         ptbBaseParserGroup.add_argument("--wasm-file", type=str, help="WASM file name for contract", default="eosio.system.wasm")
         ptbBaseParserGroup.add_argument("--abi-file", type=str, help="ABI file name for contract", default="eosio.system.abi")
+        ptbBaseParserGroup.add_argument("--wasm-runtime", type=str, help="Override default WASM runtime (\"eos-vm-jit\", \"eos-vm\")\
+                                         \"eos-vm-jit\" : A WebAssembly runtime that compiles WebAssembly code to native x86 code prior to\
+                                         execution. \"eos-vm\" : A WebAssembly interpreter.", default="eos-vm-jit")
+        ptbBaseParserGroup.add_argument("--contracts-console", type=bool, help="print contract's output to console", default=False)
+        ptbBaseParserGroup.add_argument("--eos-vm-oc-cache-size-mb", type=int, help="Maximum size (in MiB) of the EOS VM OC code cache", default=1024)
+        ptbBaseParserGroup.add_argument("--eos-vm-oc-compile-threads", type=int, help="Number of threads to use for EOS VM OC tier-up", default=1)
+        ptbBaseParserGroup.add_argument("--eos-vm-oc-enable", type=bool, help="Enable EOS VM OC tier-up runtime", default=False)
+        ptbBaseParserGroup.add_argument("--block-log-retain-blocks", type=int, help="If set to greater than 0, periodically prune the block log to\
+                                         store only configured number of most recent blocks. If set to 0, no blocks are be written to the block log;\
+                                         block log file is removed after startup.", default=None)
+        ptbBaseParserGroup.add_argument("--http-threads", type=int, help="Number of worker threads in http thread pool", default=2)
+
         return ptbBaseParser
 
     @staticmethod
@@ -580,6 +591,8 @@ class PtbArgumentsHandler(object):
     def parseArgs():
         ptbParser=PtbArgumentsHandler.createArgumentParser()
         args=ptbParser.parse_args()
+        if args.contracts_console:
+            print("Enabling contracts-console will not print anything unless debug level is 'debug' or higher.")
         return args
 
 def main():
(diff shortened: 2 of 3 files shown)


tests/TestHarness/Cluster.py
```diff
@@ -257,8 +257,15 @@ def launch(self, pnodes=1, unstartedNodes=0, totalNodes=1, prodCount=1, topo="me
...
-        nodeosArgs="--max-transaction-time -1 --abi-serializer-max-time-ms 990000 --p2p-max-nodes-per-host %d --max-clients %d" % (maximumP2pPerHost, maximumClients)
+        nodeosArgs=""
+        if extraNodeosArgs.find("--max-transaction-time") == -1:
+            nodeosArgs += " --max-transaction-time -1"
+        if extraNodeosArgs.find("--abi-serializer-max-time-ms") == -1:
+            nodeosArgs += " --abi-serializer-max-time-ms 990000"
+        if extraNodeosArgs.find("--p2p-max-nodes-per-host") == -1:
+            nodeosArgs += f" --p2p-max-nodes-per-host {maximumP2pPerHost}"
+        if extraNodeosArgs.find("--max-clients") == -1:
+            nodeosArgs += f" --max-clients {maximumClients}"
```
oschwaldp-oci, line 268: I believe, since you are not actually needing to know the location of the substring, the pythonic way to do this would be:

```
if '--max-transaction-time' not in extraNodeosArgs:
```

tests/performance_tests/performance_test_basic.py
```diff
@@ -557,6 +556,18 @@ def createBaseArgumentParser():
         ptbBaseParserGroup.add_argument("--contract-dir", type=str, help="Path to contract dir", default="unittests/contracts/eosio.system")
         ptbBaseParserGroup.add_argument("--wasm-file", type=str, help="WASM file name for contract", default="eosio.system.wasm")
         ptbBaseParserGroup.add_argument("--abi-file", type=str, help="ABI file name for contract", default="eosio.system.abi")
+        ptbBaseParserGroup.add_argument("--wasm-runtime", type=str, help="Override default WASM runtime (\"eos-vm-jit\", \"eos-vm\")\
+                                         \"eos-vm-jit\" : A WebAssembly runtime that compiles WebAssembly code to native x86 code prior to\
+                                         execution. \"eos-vm\" : A WebAssembly interpreter.", default="eos-vm-jit")
```
oschwaldp-oci, line 561: This should probably make use of `choices`.  See `--cluster-log-lvl` or `--database-map-mode` for examples.

tests/performance_tests/performance_test_basic.py
```diff
@@ -580,6 +591,8 @@ def createArgumentParser():
     def parseArgs():
         ptbParser=PtbArgumentsHandler.createArgumentParser()
         args=ptbParser.parse_args()
+        if args.contracts_console:
+            print("Enabling contracts-console will not print anything unless debug level is 'debug' or higher.")
```
oschwaldp-oci, line 595: Would it be worth checking and/or printing out the configured debug level in this message and the cli argument to change it?

tests/performance_tests/performance_test_basic.py
```diff
@@ -590,8 +603,7 @@ def main():
     testHelperConfig = PerformanceTestBasic.TestHelperConfig(killAll=args.clean_run, dontKill=args.leave_running, keepLogs=not args.del_perf_logs,
                                                              dumpErrorDetails=args.dump_error_details, delay=args.d, nodesFile=args.nodes_file, verbose=args.v)
 
-    chainPluginArgs = ChainPluginArgs(signatureCpuBillablePct=args.signature_cpu_billable_pct, chainStateDbSizeMb=args.chain_state_db_size_mb,
-                                      chainThreads=args.chain_threads, databaseMapMode=args.database_map_mode)
+    chainPluginArgs = ChainPluginArgs(chainThreads=args.chain_threads, databaseMapMode=args.database_map_mode, abiSerializerMaxTimeMs=990000, chainStateDbSizeMb=256000)
```
oschwaldp-oci, line 606: Looks like `signatureCpuBillablePct=args.signature_cpu_billable_pct` was removed here possibly by accident?

tests/performance_tests/performance_test_basic.py
```diff
@@ -602,11 +614,12 @@ def main():
     producerPluginArgs = ProducerPluginArgs(disableSubjectiveBilling=args.disable_subjective_billing,
                                             lastBlockTimeOffsetUs=lbto, produceTimeOffsetUs=args.produce_time_offset_us,
                                             cpuEffortPercent=args.cpu_effort_percent, lastBlockCpuEffortPercent=lbcep,
-                                            producerThreads=args.producer_threads)
-    httpPluginArgs = HttpPluginArgs(httpMaxResponseTimeMs=args.http_max_response_time_ms)
-    netPluginArgs = NetPluginArgs(netThreads=args.net_threads)
+                                            producerThreads=args.producer_threads, maxTransactionTime=-1)
+    httpPluginArgs = HttpPluginArgs(httpMaxBytesInFlightMb=-1, httpMaxResponseTimeMs=-1)
```
oschwaldp-oci, line 618: `httpMaxResponseTimeMs` already has an cli argument and defaults to the normally used large value.  We could leave this alone and still leave the option to override this via cli option.  Otherwise, if you're going to hardcode here you need to remove the `--http-max-response-time-ms` argument.

tests/performance_tests/performance_test_basic.py
```diff
@@ -557,6 +556,18 @@ def createBaseArgumentParser():
...
+        ptbBaseParserGroup.add_argument("--contracts-console", type=bool, help="print contract's output to console", default=False)
+        ptbBaseParserGroup.add_argument("--eos-vm-oc-cache-size-mb", type=int, help="Maximum size (in MiB) of the EOS VM OC code cache", default=1024)
+        ptbBaseParserGroup.add_argument("--eos-vm-oc-compile-threads", type=int, help="Number of threads to use for EOS VM OC tier-up", default=1)
+        ptbBaseParserGroup.add_argument("--eos-vm-oc-enable", type=bool, help="Enable EOS VM OC tier-up runtime", default=False)
+        ptbBaseParserGroup.add_argument("--block-log-retain-blocks", type=int, help="If set to greater than 0, periodically prune the block log to\
+                                         store only configured number of most recent blocks. If set to 0, no blocks are be written to the block log;\
+                                         block log file is removed after startup.", default=None)
+        ptbBaseParserGroup.add_argument("--http-threads", type=int, help="Number of worker threads in http thread pool", default=2)
```
oschwaldp-oci, line 569: All of these new options need to be added to their respective `*PluginArgs` constructors in `main`.  Most of them are in `ChainPluginArgs` but will need to be pulled out of the argument and passed into the constructure appropriately.

oschwaldp-oci (changes requested):

tests/performance_tests/performance_test_basic.py
ClaytonCalabrese: Accident, nice catch.

tests/performance_tests/performance_test_basic.py
ClaytonCalabrese: Re-examining our notes, I think I want these as CLI options. I'm also setting the default to -1, unless you think it should just remain a large but limited value.

tests/performance_tests/performance_test_basic.py
```diff
@@ -552,6 +553,19 @@ def createBaseArgumentParser():
...
         ptbBaseParserGroup.add_argument("--abi-file", type=str, help="ABI file name for contract", default="eosio.system.abi")
+        ptbBaseParserGroup.add_argument("--wasm-runtime", type=str, help="Override default WASM runtime (\"eos-vm-jit\", \"eos-vm\")\
+                                         \"eos-vm-jit\" : A WebAssembly runtime that compiles WebAssembly code to native x86 code prior to\
+                                         execution. \"eos-vm\" : A WebAssembly interpreter.",
+                                         choices=["eos-vm-jit", "eos-vm", "eos-vm-jit"], default="eos-vm-jit")
+        ptbBaseParserGroup.add_argument("--contracts-console", type=bool, help="print contract's output to console", default=False)
+        ptbBaseParserGroup.add_argument("--eos-vm-oc-cache-size-mb", type=int, help="Maximum size (in MiB) of the EOS VM OC code cache", default=1024)
+        ptbBaseParserGroup.add_argument("--eos-vm-oc-compile-threads", type=int, help="Number of threads to use for EOS VM OC tier-up", default=1)
+        ptbBaseParserGroup.add_argument("--eos-vm-oc-enable", type=bool, help="Enable EOS VM OC tier-up runtime", default=False)
```
oschwaldp-oci, line 563: `eos-vm-oc-enable` would be for validation nodes but not production nodes. We need to handle this on a per node basis something like how we've used `specificExtraNodeosArgs` in the past.

oschwaldp-oci (changes requested):

tests/performance_tests/performance_test_basic.py
```diff
@@ -548,19 +551,19 @@ def createBaseArgumentParser():
...
                                          execution. \"eos-vm\" : A WebAssembly interpreter.",
-                                         choices=["eos-vm-jit", "eos-vm", "eos-vm-jit"], default="eos-vm-jit")
-        ptbBaseParserGroup.add_argument("--contracts-console", type=bool, help="print contract's output to console", default=False)
+                                         choices=["eos-vm-jit", "eos-vm"], default="eos-vm-jit")
+        ptbBaseParserGroup.add_argument("--contracts-console", help="print contract's output to console", action='store_true')
         ptbBaseParserGroup.add_argument("--eos-vm-oc-cache-size-mb", type=int, help="Maximum size (in MiB) of the EOS VM OC code cache", default=1024)
         ptbBaseParserGroup.add_argument("--eos-vm-oc-compile-threads", type=int, help="Number of threads to use for EOS VM OC tier-up", default=1)
-        ptbBaseParserGroup.add_argument("--eos-vm-oc-enable", type=bool, help="Enable EOS VM OC tier-up runtime", default=False)
+        ptbBaseParserGroup.add_argument("--eos-vm-oc-enable", help="Enable EOS VM OC tier-up runtime", action='store_true')
```
oschwaldp-oci, line 566: I think it might be best if we don't use the chain plugin argument here since we're defining additional behavior to how it is implemented.  Maybe update the argument to something like: `non-prods-enable-eos-vm-oc-enable`.

oschwaldp-oci (changes requested):

oschwaldp-oci (approved):

oschwaldp-oci (approved):

PR #742, opened 2023-02-23
Revert interface change of abi_serializer for tester

### LINKED ISSUE #741: change on main's `abi_def`(?) causing native contract unit tests for system contracts to fail compile
On main, there is some change in `abi_def` (or something closely related I didn't track it down) that is causing system contracts' native contract unit tests to fail to compile. See 
https://github.com/eosnetworkfoundation/eos-system-contracts/actions/runs/4248437173 for example. Also encountered it on EVM contract. I wouldn't be surprised if many native contract unit tests out in the wild would be ensnared by this.

Depending on the difficulty of getting this `abi_def` difference resolved, we could also just say it requires everyone downstream to update their test code. If that's the case, need to create issues in eos-system-contracts & reference-contracts.

heifner:
The recent change to `abi_serializer` https://github.com/AntelopeIO/leap/pull/696 Specifically the change https://github.com/AntelopeIO/leap/pull/696/commits/62903c04ab510784e255e4e38d67fd7cb5e27ba6 caused existing test cases that use contract tester to not compile.

This PR reverts the changes to `abi_serializer` deprecated methods and changes `abi_serializer` constructor and `set_abi` to take `abi_def` by value instead of rvalue reference. This should allow existing users of contract tester to not have to update their test code.

Resolves #741

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -1697,7 +1697,7 @@ read_only::get_producers_result
 read_only::get_producers( const read_only::get_producers_params& params, const fc::time_point& deadline ) const try {
    abi_def abi = eosio::chain_apis::get_abi(db, config::system_account_name);
    const auto table_type = get_table_type(abi, "producers"_n);
-   const abi_serializer abis{ std::move(abi), abi_serializer::create_yield_function( abi_serializer_max_time ) };
+   const abi_serializer abis{ abi_def(abi), abi_serializer::create_yield_function( abi_serializer_max_time ) };
    EOS_ASSERT(table_type == KEYi64, chain::contract_table_query_exception, "Invalid table type ${type} for table producers", ("type",table_type));
 
    const auto& d = db.db();
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -70,14 +70,14 @@ namespace eosio { namespace chain {
       );
    }
 
-   abi_serializer::abi_serializer( abi_def&& abi, const yield_function_t& yield ) {
+   abi_serializer::abi_serializer( abi_def abi, const yield_function_t& yield ) {
       configure_built_in_types();
       set_abi(std::move(abi), yield);
    }
 
-   abi_serializer::abi_serializer( abi_def&& abi, const fc::microseconds& max_serialization_time) {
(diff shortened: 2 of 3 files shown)


heifner:
Successful run of 
[eos-system-contracts](https://github.com/eosnetworkfoundation/eos-system-contracts)
https://github.com/eosnetworkfoundation/eos-system-contracts/actions/runs/4253609467/jobs/7398702034

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1697,7 +1697,7 @@ read_only::get_producers_result
 read_only::get_producers( const read_only::get_producers_params& params, const fc::time_point& deadline ) const try {
    abi_def abi = eosio::chain_apis::get_abi(db, config::system_account_name);
    const auto table_type = get_table_type(abi, "producers"_n);
-   const abi_serializer abis{ std::move(abi), abi_serializer::create_yield_function( abi_serializer_max_time ) };
+   const abi_serializer abis{ abi_def(abi), abi_serializer::create_yield_function( abi_serializer_max_time ) };
```
greg7mdp, line 1700: I don't think this change is right, should still be `std::move(abi)` since `abi_serializer::abi_serializer( abi_def abi, const yield_function_t& yield )` constructor now takes a value. With the current version it will make an unnecessary copy.

greg7mdp:
What was the compilation issue with the previous version?

greg7mdp (changes requested):

plugins/chain_plugin/chain_plugin.cpp
greg7mdp: Never mind I see where `abi` is used afterwards.

heifner: `abi` is used on line 1747, so it can't be `std::move` here. I added the explicit copy to make it obvious that a copy is needed.

greg7mdp: Yes, it can't be `std::move(abi)`, but it should be `abi`, not `abi_def(abi)`. My original comment was wrong.

greg7mdp: You are right, the explicit copy is fine after all!

spoonincode (approved):
thanks

PR #746, opened 2023-02-23
Backport to 3.2 of bootstrap new protocol features activation

jolly-fellow:


mikelik:
I think this PR has to be cancelled, and new one should be created:
"Create a branch that merges release/3.1 into release/3.2
Create a PR into release/3.2 from that branch"

PR #752, opened 2023-02-25
Move initialization of logging before plugin_initialize

heifner:
During work on #659 the log of `EOS_ASSERT`:
```
error 2023-02-22T20:51:14.889 nodeos    state_history_plugin.c:284    on_accepted_block    ] fc::exception: 3110000 plugin_exception: Plugin exception
Existing ship log with 214013 blocks when starting from genesis block 2
    {"eb":214013,"b":2}
    nodeos  log.hpp:513 write_entry
```
was not logged because it was called before `state_history_plugin::plugin_startup()` where the `state_history` logger is initialized. Some plugins work around this by using the default logger in `plugin_initialize()`. However, there are some logs that never have worked. 
For example, in `net_plugin`:
```
   void net_plugin::plugin_initialize( const variables_map& options ) {
      fc_ilog( logger, "Initialize net plugin" );
```
This never actually logs because the `net_plugin_impl` `logger` has not been initialized.

To avoid this confusion, this PR moves the `initialize_logging()`  of `main` before `app` initialize which is before `plugin_initialize()`. Also any plugin loggers are now initialized in `plugin_initialize` as the first step so that loggers are configured and ready to be used in `plugin_initailize`.

With this change you actually now get:
```
info  2023-02-25T03:29:47.026 nodeos    net_plugin.cpp:3667           plugin_initialize    ] Initialize net plugin
```

See #737

### DIFF
--- a/libraries/libfc/src/log/logger.cpp
+++ b/libraries/libfc/src/log/logger.cpp
@@ -3,9 +3,8 @@
 #include <fc/log/appender.hpp>
 #include <fc/exception/exception.hpp>
 #include <fc/filesystem.hpp>
-#include <unordered_map>
-#include <string>
 #include <fc/log/logger_config.hpp>
+#include <unordered_map>
 
 namespace fc {
 
@@ -13,7 +12,7 @@ namespace fc {
       public:
          impl()
          :_parent(nullptr),_enabled(true),_additivity(false),_level(log_level::warn){}
-         fc::string       _name;
+         std::string      _name;
          logger           _parent;
          bool             _enabled;
          bool             _additivity;
@@ -28,7 +27,7 @@ namespace fc {
 
     logger::logger(nullptr_t){}
 
-    logger::logger( const string& name, const logger& parent )
+    logger::logger( const std::string& name, const logger& parent )
     :my( new impl() )
     {
        my->_name = name;
@@ -40,7 +39,7 @@ namespace fc {
     :my(l.my){}
 
     logger::logger( logger&& l )
-    :my(fc::move(l.my)){}
+    :my(std::move(l.my)){}
 
     logger::~logger(){}
 
@@ -82,14 +81,14 @@ namespace fc {
        }
     }
 
-    void logger::set_name( const fc::string& n ) { my->_name = n; }
-    const fc::string& logger::name()const { return my->_name; }
+    void logger::set_name( const std::string& n ) { my->_name = n; }
+    std::string logger::get_name()const { return my->_name; }
 
-    logger logger::get( const fc::string& s ) {
+    logger logger::get( const std::string& s ) {
        return log_config::get_logger( s );
     }
 
-    void logger::update( const fc::string& name, logger& log ) {
+    void logger::update( const std::string& name, logger& log ) {
        log_config::update_logger( name, log );
     }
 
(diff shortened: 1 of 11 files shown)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -271,6 +271,7 @@ class http_plugin_impl : public std::enable_shared_from_this<http_plugin_impl> {
 
    void http_plugin::plugin_initialize(const variables_map& options) {
       try {
+         handle_sighup(); // setup logging
```
greg7mdp, line 274: Not related to your change, so maybe ignore, but:

- no log message "Initializing HTTP plugin" here (same for other plugins)?
- why are these functions named `handle_sigup`, when they initialize the logger?
- inside `handle_sigup` we copy and note that a copy is needed while `name` is passed as a const reference to logger::update()

greg7mdp (approved):

linh2931 (approved):

plugins/http_plugin/http_plugin.cpp
heifner: 1. One could argue it is not needed in net_plugin either since no one ever noticed it was not there. Likely it is noise unless you are trying to debug something. I left it in because "why not"?
2. They are named `handle_sigup` because they override the `handle_sigup` of virtual `appbase::plugin::handle_sighup` which is called anytime a `SIGHUP` is caught. Specifically in Leap we wish to update the logging configuration on `SIGHUP`. We have talked about the possibility of updating other configuration params on `SIGHUP` as well. 
We could create a `update_logger` method for all the plugins and have the override `handle_sigup` call that and also call it from `plugin_initialize`. The lazy way was chosen and just call `handle_sighup` because that is what is needed on startup.
3. Yes, the comment is correct. If you instead did `fc::logger::update(logger().name(), logger())` the `const&` bound to the internals of `logger()` can be destroyed by the `update` of the referenced `logger()`. `name` can then be used after `logger()` is updated causing access to delete memory.

greg7mdp: Thanks for the answers Kevin. 
regarding point 3, I think a  better (and safer) alternative would be to have `logger().name()` return a value and not a reference to a string whose lifetime is not clearly known

heifner: I agree. I'll go ahead and fix that.

libraries/libfc/src/log/logger.cpp
```diff
@@ -82,14 +81,14 @@ namespace fc {
        }
     }
 
-    void logger::set_name( const fc::string& n ) { my->_name = n; }
-    const fc::string& logger::name()const { return my->_name; }
+    void logger::set_name( const std::string& n ) { my->_name = n; }
```
greg7mdp, line 84: No need to change, but I think a good convention for methods accepting a potentially moveable object that is to be immediately copied, is to accept a `value` parameter and `std::move` it where it should go.

greg7mdp: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in

heifner: Logger hopefully will be replaced soon by a third-party logging library.

greg7mdp: Sounds good to me!

greg7mdp (approved):

PR #753, opened 2023-02-25
keosd - Exit program if keosd fails to aquire wallet lock

### LINKED ISSUE #513: if keosd fails to aquire wallet lock it lingers forever
If keosd fails to acquire the wallet lock it will print a failure message but just linger forever until ctrl-C. It should just exit in this case.
```
$ programs/keosd/keosd
info  2022-11-24T02:04:41.639 keosd     wallet_plugin.cpp:33          plugin_initialize    ] initializing wallet plugin
warn  2022-11-24T02:04:41.639 keosd     wallet_plugin.cpp:51          plugin_initialize    ] 3120000 wallet_exception: Wallet exception
Failed to lock access to wallet directory; is another keosd running?
    {}
    keosd  wallet_manager.cpp:304 initialize_lock

Failed to initialize
```
Seen on main, unknown if it affects 3.x

greg7mdp:
Resolves #513 

Should we backport to 3.2?

### DIFF
--- a/programs/keosd/main.cpp
+++ b/programs/keosd/main.cpp
@@ -71,6 +71,17 @@ bfs::path determine_home_directory()
    return home;
 }
 
+enum return_codes {
+   OTHER_FAIL        = -2,
+   INITIALIZE_FAIL   = -1,
+   SUCCESS           = 0,
+   BAD_ALLOC         = 1,
+   DATABASE_DIRTY    = 2,
+   FIXED_REVERSIBLE  = SUCCESS,
+   EXTRACTED_GENESIS = SUCCESS,
+   NODE_MANAGEMENT_SUCCESS = 5
+};
+
 int main(int argc, char** argv)
 {
    try {
@@ -93,7 +104,7 @@ int main(int argc, char** argv)
              opts.count("print-default-config")) {
             return 0;
          }
-         return EXIT_FAILURE;
+         return INITIALIZE_FAIL;
       }
       initialize_logging();
       auto& http = app->get_plugin<http_plugin>();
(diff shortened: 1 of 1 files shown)


programs/keosd/main.cpp
```diff
@@ -93,6 +93,7 @@ int main(int argc, char** argv)
              opts.count("print-default-config")) {
             return 0;
          }
+         return EXIT_FAILURE;
```
heifner, line 96: Seems like this should match `nodeos` `main.cpp`, which would be to return `-1` (`INITIALIZE_FAIL`).

heifner (changes requested):
Should be backported to 3.1 & 3.2.

programs/keosd/main.cpp
greg7mdp: Sure, will do, although `EXIT_FAILURE` is a [platform standard]( https://en.cppreference.com/w/c/program/EXIT_status) while `INITIALIZE_FAIL` is a leap invention :-)

greg7mdp: @heifner do you think this should go to 3.2 first?

heifner: I assume it is applicable to 3.1, so it goes there first then merged up to 3.2 and then to main.

greg7mdp: OK, should I duplicate the whole enum in keosd/main.cpp?

```
enum return_codes {
   OTHER_FAIL        = -2,
   INITIALIZE_FAIL   = -1,
   SUCCESS           = 0,
   BAD_ALLOC         = 1,
   DATABASE_DIRTY    = 2,
   FIXED_REVERSIBLE  = SUCCESS,
   EXTRACTED_GENESIS = SUCCESS,
   NODE_MANAGEMENT_SUCCESS = 5
};
```

It it was me I'd have it in a common header, but I don't know which one.

heifner: Maybe `plugins/chain_interface/include/eosio/chain/plugin_interface.hpp` ?

Issue #650, opened 2023-01-19
Gelf logging should use its own io_context


heifner:
Currently `gelf_appender` uses the main `application` `io_context`. This bypasses the application's priority queue and affectively means that gelf logging runs at the highest priority. This is clearly not what is intended. Gelf logging should be modified to use its own thread and its own `io_context`.

In my mind this should not wait for a new logging framework: https://github.com/eosnetworkfoundation/mandel/issues/152

spoonincode:
A couple other things to consider while touching this code:

gelf_appender is using some fc wrappers over asio stuff (and the wrappers are fairly old style, such as usage of the deprecated io_service instead of io_service). This is generally something we're trying to move away from: just use boost directly. Consider refactoring accordingly and removing the wrappers.

A separate thread is, imo, not strictly required. There will be very little back pressure on a UDP socket so a large socket buff (say 1MB or 2MB) would be really surprising to fill up (at which point gelf_appender could just drop the message on the floor instead of `async`ing it). I'm certainly not saying it needs to be done this way, but imo it would be fine to simplify it further like this. Obviously other opinions welcome.

heifner:
Probably more gain from moving the `format_string` and `zlib_compress` to another thread. But something in gelf logging takes a considerable amount of time: https://github.com/AntelopeIO/leap/issues/555

spoonincode:
huh, yes that's really good point and almost certainly what ought to be attempted if this code is being touched.

spoonincode:
resolved in #758

PR #758, opened 2023-02-28
move gelf logging to another thread

huangminghuang:
This PR address issue #650 to make GELF logging use its own io_context and spawn a separate thread to send GELF logging.

### DIFF
--- a/libraries/libfc/src/log/gelf_appender.cpp
+++ b/libraries/libfc/src/log/gelf_appender.cpp
@@ -46,10 +46,11 @@ namespace fc
     using work_guard_t = boost::asio::executor_work_guard<boost::asio::io_context::executor_type>;
     config                                         cfg;
     std::optional<boost::asio::ip::udp::endpoint>  gelf_endpoint;
+    std::thread                                    thread;
     boost::asio::io_context                        io_context;
     work_guard_t work_guard =                      boost::asio::make_work_guard(io_context);
     udp_socket                                     gelf_socket;
-    std::thread                                    thread;
+
 
     impl(const variant& c)
     {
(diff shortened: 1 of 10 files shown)


libraries/libfc/src/log/gelf_appender.cpp
```diff
@@ -144,7 +162,7 @@ namespace fc
     gelf_message["short_message"] = format_string(message.get_format(), message.get_data(), true);
 
     // use now() instead of context.get_timestamp() because log_message construction can include user provided long running calls
-    const auto time_ns = time_point::now().time_since_epoch().count();
+    //const auto time_ns = time_point::now().time_since_epoch().count();
```
heifner, line 165: Remove

libraries/libfc/src/log/gelf_appender.cpp
```diff
@@ -117,9 +130,17 @@ namespace fc
...
-        my->gelf_socket.initialize(io_service);
+        my->gelf_socket.initialize(my->io_context);
         my->gelf_socket.open();
         std::cerr << "opened GELF socket to endpoint " << my->cfg.endpoint << "\n";
+
+        my->thread = std::thread([this] {
+           try {
+              fc::set_os_thread_name("gelf_appender");
+              my->io_context.run();
+           } catch (...) {
```
heifner, line 141: Should `elog` here at the very least.

heifner (changes requested):

libraries/libfc/src/log/gelf_appender.cpp
heifner: Obviously, as @huangminghuang pointed out we don't want to use `elog` as it would be recursive. How about adding a `std::cerr` call.

libraries/libfc/src/log/gelf_appender.cpp
```diff
@@ -117,9 +132,20 @@ namespace fc
 
       if (my->gelf_endpoint)
       {
-        my->gelf_socket.initialize(io_service);
+        my->gelf_socket.initialize(my->io_context);
         my->gelf_socket.open();
         std::cerr << "opened GELF socket to endpoint " << my->cfg.endpoint << "\n";
+
+        my->thread = std::thread([this] {
```
heifner, line 139: I think this calls `terminate()` on `SIGHUP` which calls `initialize()`

huangminghuang: SIGHUP would eventually create a new logger object and won't call `initialize()` on the same object twice.

heifner (approved):

greg7mdp (approved):

Issue #656, opened 2023-01-19
Programmable snapshots for future blocks


stephenpdeos:
See design document: https://github.com/eosnetworkfoundation/product/pull/131

- [X] Implementation new producer_api endpoints with proper validation and changes to ephemeral state (but not yet persisting to JSON database nor actually triggering snapshots).
- [X] Persisting to JSON database and tests to ensure schedule requests persist restart.
- [x] Implement snapshot triggers from snapshot requests with proper fork handling.
- [X] producer_plugin boost tests
- [ ] Python integration test (checks fork handling and garbage collection)

PR #759, opened 2023-02-28
Programmable snapshots API

766C6164:
This PR relates to https://github.com/AntelopeIO/leap/issues/656 issue and introduces new feature - snapshot scheduling API to create, retrieve and delete/modify automated snapshot schedule.

In more details API specification with examples of it's use can be found here:
https://github.com/eosnetworkfoundation/product/blob/main/api-http/proposals/snapshot-api.md

This draft PR includes initial boost tests and will incorporate additional tests currently being finalized.


plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,172 @@
...
+                                                 BOOST_MULTI_INDEX_MEMBER(snapshot_request_information, uint32_t, block_spacing),
+                                                 BOOST_MULTI_INDEX_MEMBER(snapshot_request_information, uint32_t, start_block_num),
+                                                 BOOST_MULTI_INDEX_MEMBER(snapshot_request_information, uint32_t, end_block_num)>>>>;
+   snapshot_requests _snapshot_requests;
+   snapshot_db_json _snapshot_db;
+   uint32_t _snapshot_id = 0;
+   std::function<void(producer_plugin::next_function<producer_plugin::snapshot_information>)> _create_snapshot;
+
+public:
+   snapshot_scheduler() {}
```
heifner, line 64: `= default;`

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,172 @@
...
+         // assume "asap" for snapshot with missed/zero start, it can have spacing
+         if(req.start_block_num == 0) {
+            auto& snapshot_by_id = _snapshot_requests.get<by_snapshot_id>();
+            auto it = snapshot_by_id.find(req.snapshot_request_id);
+            _snapshot_requests.modify(it, [&height](auto& p) { p.start_block_num = height; });
+         }
+         // remove expired request
+         if(req.end_block_num > 0 && height >= req.end_block_num) {
+            auto& snapshot_by_id = _snapshot_requests.get<by_snapshot_id>();
+            _snapshot_requests.erase(snapshot_by_id.find(req.snapshot_request_id));
```
heifner, line 83: I think this can just be `_snapshot_requests.erase(req.snapshot_request_id);`

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,172 @@
...
+   void execute_snapshot() {
+      auto next = [](const std::variant<fc::exception_ptr, producer_plugin::snapshot_information>& result) {
+         if(std::holds_alternative<fc::exception_ptr>(result)) {
+            try {
+               std::get<fc::exception_ptr>(result)->dynamic_rethrow_exception();
+            } catch(const fc::exception& e) {
+               edump((e.to_detail_string()));
+               throw;
+            } catch(const std::exception& e) {
+               edump((e.what()));
```
heifner, line 161: Avoid using `edump` in production code. It is for quick debugging, not for anything committed to the code base.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,172 @@
...
+               throw;
+            }
+         } else {
+            // success
+         }
+      };
+      _create_snapshot(next);
+      //auto theasync=std::async([&,next]{ return _create_snapshot(next);});
```
heifner, line 169: remove

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,172 @@
...
+               throw;
+            } catch(const std::exception& e) {
+               edump((e.what()));
+               throw;
+            }
+         } else {
+            // success
+         }
+      };
+      _create_snapshot(next);
```
heifner, line 168: Would like to see the `create_snapshot` code moved into this class and out of `producer_plugin`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -988,6 +999,9 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
       }
    }
 
+   my->_snapshot_scheduler.set_db_path(my->_snapshots_dir);
+   my->_snapshot_scheduler.set_create_snapshot_fn(std::bind(&producer_plugin::create_snapshot, this, std::placeholders::_1));
```
heifner, line 1003: Use a lambda instead of `std::bind`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1013,6 +1027,7 @@ void producer_plugin::plugin_startup()
    my->_accepted_block_connection.emplace(chain.accepted_block.connect( [this]( const auto& bsp ){ my->on_block( bsp ); } ));
    my->_accepted_block_header_connection.emplace(chain.accepted_block_header.connect( [this]( const auto& bsp ){ my->on_block_header( bsp ); } ));
    my->_irreversible_block_connection.emplace(chain.irreversible_block.connect( [this]( const auto& bsp ){ my->on_irreversible_block( bsp->block ); } ));
+   my->_block_start_connection.emplace(chain.block_start.connect( [this]( uint32_t bs ){ my->_snapshot_scheduler.on_block(bs); } ));
```
heifner, line 1030: Please use `on_start_block` for `block_start` signal to avoid confusion. We use `on_block` for `accepted_block` signal. Although, I think you want to attach to `accepted_block` not `block_start`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -382,6 +390,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          auto before = _unapplied_transactions.size();
          _unapplied_transactions.clear_applied( bsp );
          _subjective_billing.on_block( _log, bsp, fc::time_point::now() );
+         //_snapshot_scheduler.on_block(bsp->block_num);
```
heifner, line 393: Remove. Although, I wonder why you are calling snapshot_scheduler on blocks start instead of on accepted block.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,172 @@
...
+public:
+   snapshot_scheduler() {}
+
+   // snapshot_scheduler_listener
+   virtual void on_block(uint32_t height) {
+      for(const auto& req: _snapshot_requests.get<0>()) {
+         // execute "asap" or if matches spacing
+         if((req.start_block_num == 0) ||
+            (!((height - req.start_block_num) % req.block_spacing))) {
+            execute_snapshot();
```
heifner, line 72: Since `on_block` is called from start block, doesn't this create a snapshot for the previous block and not the current block?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -25,7 +26,10 @@
 #include <boost/multi_index/member.hpp>
 #include <boost/multi_index/hashed_index.hpp>
 #include <boost/multi_index/ordered_index.hpp>
+#include <boost/multi_index/random_access_index.hpp>
 #include <boost/signals2/connection.hpp>
+#include <boost/range/adaptor/map.hpp>
+#include <boost/range/algorithm/copy.hpp>
```
heifner, line 32: None of these new boost includes are needed.

plugins/producer_plugin/producer_plugin.cpp
heifner: It would be nice to move all the existing snapshot generation code out of producer_plugin into the snapshot_scheduler or related classes.

heifner (changes requested):

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
```diff
@@ -50,6 +50,19 @@ class producer_plugin : public appbase::plugin<producer_plugin> {
       std::string          snapshot_name;
    };
 
+   struct snapshot_request_information {
+      uint32_t             snapshot_request_id = 0;
```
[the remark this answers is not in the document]
766C6164: Yes, that would be definitely cleaner, thank you. Done.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,172 @@
...
+         // if "end" is specified, it should be greater then start
+         EOS_ASSERT(sri.start_block_num <= sri.end_block_num, invalid_snapshot_request, "End block number should be greater or equal to start block number");
+         // if also block_spacing specified, check it
+         if(sri.block_spacing > 0) {
+            EOS_ASSERT(sri.start_block_num + sri.block_spacing <= sri.end_block_num, invalid_snapshot_request, "Block spacing exceeds defined by start and end range");
+         }
+      }
+
+      auto request_id = sri.snapshot_request_id ? sri.snapshot_request_id : _snapshot_id++;
+      _snapshot_requests.emplace(producer_plugin::snapshot_request_information{request_id, sri.block_spacing, sri.start_block_num, sri.end_block_num, {}});
```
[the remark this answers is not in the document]
766C6164: done, snapshot_request_id is out from request and will be always auto generated

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
```diff
@@ -105,7 +118,11 @@ class producer_plugin : public appbase::plugin<producer_plugin> {
    void set_whitelist_blacklist(const whitelist_blacklist& params);
 
    integrity_hash_information get_integrity_hash() const;
+
    void create_snapshot(next_function<snapshot_information> next);
+   void schedule_snapshot(const snapshot_request_information& schedule);
+   void unschedule_snapshot(const snapshot_request_information& schedule);
```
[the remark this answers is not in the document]
766C6164: for un-schedule replaced with struct holging just uint32_t. It arrives from http request parser and this seems a right way to do it. Then struct snapshot_schedule_information : public snapshot_request_id_information, public snapshot_request_information inherited from both which is also seems cleaner

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
```diff
@@ -148,7 +165,7 @@ class producer_plugin : public appbase::plugin<producer_plugin> {
    void received_block();
 
  private:
-   std::shared_ptr<class producer_plugin_impl> my;
+   std::shared_ptr<class producer_plugin_impl>  my;
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <eosio/producer_plugin/producer_plugin.hpp>
+
+#include <boost/date_time/posix_time/posix_time.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+
+namespace eosio {
+
+using namespace eosio::chain;
```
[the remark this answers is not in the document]
766C6164: done, removed

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+class snapshot_db_json {
+public:
+   snapshot_db_json() {}
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,77 @@
...
+   snapshot_db_json(bfs::path db_path) : db_path(db_path) {}
```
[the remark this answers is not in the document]
766C6164: removed this ctor as it ends up never used

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,77 @@
...
+   ~snapshot_db_json() {}
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+   void set_path(bfs::path path) {
+      db_path = path;
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include <boost/date_time/posix_time/posix_time.hpp>
+#include <boost/property_tree/json_parser.hpp>
+#include <boost/property_tree/ptree.hpp>
+
+
+namespace eosio {
+
+using namespace eosio::chain;
+
+class snapshot_db_json {
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,77 @@
...
+         node.put("snapshot_description", key.snapshot_description);
+         node.put("block_spacing", key.block_spacing);
+         node.put("start_block_num", key.start_block_num);
+         node.put("end_block_num", key.end_block_num);
+         node_srs.push_back(std::make_pair("", node));
+      }
+      root.push_back(std::make_pair("snapshot_requests", node_srs));
+
+      std::ofstream file(get_json_path().string());
+      boost::property_tree::write_json(file, root);
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,172 @@
...
+#include <eosio/chain/transaction.hpp>
+#include <eosio/chain/types.hpp>
+
+#include <boost/multi_index/hashed_index.hpp>
+#include <boost/multi_index/member.hpp>
+#include <boost/multi_index/ordered_index.hpp>
+#include <boost/multi_index/random_access_index.hpp>
+#include <boost/multi_index_container.hpp>
+
+#include <eosio/producer_plugin/snapshot_db_json.hpp>
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: yes, that is better :) done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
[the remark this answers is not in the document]
heifner: I would have made it the first include, but not a huge deal.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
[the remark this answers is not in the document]
heifner: `std::ifstream` and `std::ofstream` do not throw exceptions by default. Need to enable exceptions.
`file.exceptions(std::istream::failbit|std::istream::eofbit);`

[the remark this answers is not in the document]
766C6164: Thank you! Done.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+namespace eosio {
+
+namespace bmi = boost::multi_index;
+using chain::account_name;
+using chain::block_state_ptr;
+using chain::packed_transaction;
+using chain::transaction_id_type;
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: done

plugins/producer_plugin/producer_plugin.cpp
766C6164: done

plugins/producer_plugin/producer_plugin.cpp
766C6164: done

plugins/producer_plugin/producer_plugin.cpp
766C6164: on_start_block was the only hook that allowed execution of existing implementation of create_snapshot without massive chain of hard to trace exceptions. on_start_block hook executes create+snapshot perfectly with no issues.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: bug, that's what will happen, fixed

plugins/producer_plugin/producer_plugin.cpp
766C6164: removed

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
heifner: Would be good to add a test that verifies the correct block height snapshot is created.

heifner:
Would like see an integration test for this functionality.

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
```diff
@@ -158,6 +182,9 @@ FC_REFLECT(eosio::producer_plugin::greylist_params, (accounts));
 FC_REFLECT(eosio::producer_plugin::whitelist_blacklist, (actor_whitelist)(actor_blacklist)(contract_whitelist)(contract_blacklist)(action_blacklist)(key_blacklist) )
 FC_REFLECT(eosio::producer_plugin::integrity_hash_information, (head_block_id)(integrity_hash))
 FC_REFLECT(eosio::producer_plugin::snapshot_information, (head_block_id)(head_block_num)(head_block_time)(version)(snapshot_name))
+FC_REFLECT(eosio::producer_plugin::snapshot_request_information, (block_spacing)(start_block_num)(end_block_num)(snapshot_description))
+FC_REFLECT(eosio::producer_plugin::snapshot_request_id_information, (snapshot_request_id))
+FC_REFLECT(eosio::producer_plugin::get_snapshot_requests_result, (requests))
```
heifner, line 187: I thought you had to use FC_REFLECT_DERIVED for these derived types.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,90 @@
...
+            ssi.snapshot_description = req.second.get<std::string>("snapshot_description");
+            ssi.block_spacing = req.second.get<uint32_t>("block_spacing");
+            ssi.start_block_num = req.second.get<uint32_t>("start_block_num");
+            ssi.end_block_num = req.second.get<uint32_t>("end_block_num");
+            sr.push_back(ssi);
+         }
+      }
+      catch (std::ifstream::failure e) {
+          wlog( "unable to restore snapshots schedule from filesystem: ${details}",
+                  ("details",e.what()) );
```
heifner, line 50: Should be 3 space indent, an `elog`, and indicate the path in the log.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+      try {
+         std::ofstream file(get_json_path().string());
+         file.exceptions(std::istream::failbit|std::istream::eofbit);
+         boost::property_tree::write_json(file, root);
+         file.close();
+      }
+      catch (std::ofstream::failure e) {
+          wlog( "unable to store snapshots schedule to filesystem: ${details}",
+                  ("details",e.what()) );
```
heifner, line 80: Should be 3 space indent, an elog, and indicate the path in the log.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,159 @@
...
+   virtual producer_plugin::get_snapshot_requests_result get_snapshot_requests() {
+      producer_plugin::get_snapshot_requests_result result;
+      auto& asvector = _snapshot_requests.get<as_vector>();
+      result.requests.insert(result.requests.begin(), asvector.begin(), asvector.end());
+      return result;
+   }
+
+   // initialize with storage
+   void set_db_path(bfs::path db_path) {
+      _snapshot_db.set_path(db_path);
```
heifner, line 125: `std::move(db_path)`

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,159 @@
...
+         }
+         // remove expired request
+         if(req.end_block_num > 0 && height >= req.end_block_num) {
+            _snapshot_requests.erase(req.snapshot_request_id);
+         }
+      }
+   }
+   void on_block(uint32_t height) {}
+   void on_irreversible_block(uint32_t height) {}
+   void on_abort_block() {}
```
heifner, line 81: Why do we have these?

heifner (dismissed):

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,159 @@
...
+   void execute_snapshot() {
+      auto next = [](const std::variant<fc::exception_ptr, producer_plugin::snapshot_information>& result) {
+         if(std::holds_alternative<fc::exception_ptr>(result)) {
+            try {
+               std::get<fc::exception_ptr>(result)->dynamic_rethrow_exception();
+            } catch(const fc::exception& e) {
+               wlog( "snapshot creation error: ${details}", ("details",e.to_detail_string()) );
+               throw;
+            } catch(const std::exception& e) {
+               wlog( "snapshot creation error: ${details}", ("details",e.what()) );
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,159 @@
...
+   }
+
+
+   void execute_snapshot() {
+      auto next = [](const std::variant<fc::exception_ptr, producer_plugin::snapshot_information>& result) {
+         if(std::holds_alternative<fc::exception_ptr>(result)) {
+            try {
+               std::get<fc::exception_ptr>(result)->dynamic_rethrow_exception();
+            } catch(const fc::exception& e) {
+               wlog( "snapshot creation error: ${details}", ("details",e.to_detail_string()) );
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,159 @@
...
+      if(fc::exists(_snapshot_db.get_json_path())) {
+         std::vector<producer_plugin::snapshot_schedule_information> sr;
+         _snapshot_db >> sr;
+         _snapshot_requests.insert(sr.begin(), sr.end());
+      }
+   }
+
+   // snapshot executor
+   void set_create_snapshot_fn(std::function<void(producer_plugin::next_function<producer_plugin::snapshot_information>)> fn) {
+      _create_snapshot = fn;
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,159 @@
...
+
+      auto& vec = _snapshot_requests.get<as_vector>();
+      std::vector<producer_plugin::snapshot_schedule_information> sr(vec.begin(), vec.end());
+      _snapshot_db << sr;
+   }
+
+   virtual producer_plugin::get_snapshot_requests_result get_snapshot_requests() {
+      producer_plugin::get_snapshot_requests_result result;
+      auto& asvector = _snapshot_requests.get<as_vector>();
+      result.requests.insert(result.requests.begin(), asvector.begin(), asvector.end());
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: dropped

tests/nodeos_snapshot_programmable_test.py
```diff
@@ -0,0 +1,210 @@
+#!/usr/bin/env python3
+
```
heifner, line 2: This is almost identical to `nodeos_snapshot_diff_test.py`. Seems like they could easily be combined into one test that does both.

766C6164: Thank you! I updated nodeos_snapshot_diff_test to use scheduler and then compare that snapshot to irreversible.  I am working on re-purposing nodeos_snapshot_programmable_test for fork scenario

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,189 @@
...
+   using snapshot_requests = bmi::multi_index_container<
+         producer_plugin::snapshot_schedule_information,
+         indexed_by<
+               bmi::hashed_unique<tag<by_snapshot_id>, BOOST_MULTI_INDEX_MEMBER(producer_plugin::snapshot_request_id_information, uint32_t, snapshot_request_id)>,
+               bmi::random_access<tag<as_vector>>,
+               bmi::ordered_unique<tag<by_snapshot_value>,
+                                   composite_key<producer_plugin::snapshot_schedule_information,
+                                                 BOOST_MULTI_INDEX_MEMBER(producer_plugin::snapshot_request_information, uint32_t, block_spacing),
+                                                 BOOST_MULTI_INDEX_MEMBER(producer_plugin::snapshot_request_information, uint32_t, start_block_num),
+                                                 BOOST_MULTI_INDEX_MEMBER(producer_plugin::snapshot_request_information, uint32_t, end_block_num)>>>>;
```
greg7mdp, line 48: I feel that using a `multi-index-container`  is overkill here, and adds unnecessary complexity. I would suggest using an `unordered_set<producer_plugin::snapshot_schedule_information>` with the hash and compare done on the `snapshot_request_id`.

The serialization could as easily be done from the set as from the `<as_vector>` view.
The `<by_snapshot_value>` is used only to find out if a new request is a duplicate, and I think that doing a linear search is probably OK since we are unlikely to have many thousands of snapshot requests.

After all we already do a linear search at every `on_start_block`.

greg7mdp (changes requested):

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
```diff
@@ -51,6 +51,26 @@ class producer_plugin : public appbase::plugin<producer_plugin> {
...
+   struct snapshot_request_information {
+      uint32_t             block_spacing = 0;
+      uint32_t             start_block_num = 0;
+      uint32_t             end_block_num = 0;
+      std::string          snapshot_description = "";
+
+   };
+
+   struct snapshot_request_id_information {
+      uint32_t                            snapshot_request_id = 0;
```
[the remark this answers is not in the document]
766C6164: Thank you! Leftover after removing other long-typed struct members. Fixed.

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
```diff
@@ -0,0 +1,140 @@
...
+#include <eosio/producer_plugin/snapshot_db_json.hpp>
+#include <eosio/testing/tester.hpp>
+namespace {
+
+using namespace eosio;
+using namespace eosio::chain;
+
+BOOST_AUTO_TEST_SUITE(snapshot_scheduler_test)
+
+BOOST_AUTO_TEST_CASE(snapshot_scheduler_test) {
```
greg7mdp, line 15: Nice test!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
greg7mdp: @Vlad, I understand that you had a requirement to use `multi_index_container`, and not time to revisit this choice for now, so marking this as resolved.  Thanks!

tests/nodeos_snapshot_programmable_test.py
heifner: Are you planning on removing this test then?

tests/nodeos_snapshot_diff_test.py
```diff
@@ -152,10 +156,9 @@ def waitForBlock(node, blockNum, blockType=BlockType.head, timeout=None, reportI
     assert steadyStateAvg>=minRequiredTransactions, "Expected to at least receive %s transactions per block, but only getting %s" % (minRequiredTransactions, steadyStateAvg)
 
     Print("Create snapshot")
-    ret = nodeSnap.createSnapshot()
+    ret = nodeSnap.scheduleSnapshot()
```
heifner, line 159: I think this is probably fine. Would have been better to enhance this test to do both, but not a huge deal.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,90 @@
...
+            ssi.snapshot_description = req.second.get<std::string>("snapshot_description");
+            ssi.block_spacing = req.second.get<uint32_t>("block_spacing");
+            ssi.start_block_num = req.second.get<uint32_t>("start_block_num");
+            ssi.end_block_num = req.second.get<uint32_t>("end_block_num");
+            sr.push_back(ssi);
+         }
+      }
+      catch (std::ifstream::failure & e) {
+         elog( "unable to restore snapshots schedule from filesystem ${jsonpath}, details: ${details}",
+            ("jsonpath", get_json_path().string()) ("details",e.what()) );
```
heifner, line 50: I think we should consider an `app().quit()` here and below. Typically users will dedicate a node for creating snapshots. If this node is unable to do so, seems like it should indicate an error and exit.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,189 @@
...
+      auto next = [srid, this](const std::variant<fc::exception_ptr, producer_plugin::snapshot_information>& result) {
+         if(std::holds_alternative<fc::exception_ptr>(result)) {
+            try {
+               std::get<fc::exception_ptr>(result)->dynamic_rethrow_exception();
+            } catch(const fc::exception& e) {
+               elog( "snapshot creation error: ${details}", ("details",e.to_detail_string()) );
+               throw;
+            } catch(const std::exception& e) {
+               elog( "snapshot creation error: ${details}", ("details",e.what()) );
+               throw;
```
heifner, line 167: Does this result in a graceful shutdown? Can you manually verify that it does.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,189 @@
...
+         else if(!req.block_spacing || (!((height - req.start_block_num - 1) % req.block_spacing))) {
+            execute_snapshot(req.snapshot_request_id);
+         }       
+
+         // cleanup - remove expired (or invalid) request
+         if(!req.start_block_num || !req.block_spacing || (req.end_block_num > 0 && height >= req.end_block_num)) {
+            unschedule_snapshot(req.snapshot_request_id);
+         }
+      }
+   }
```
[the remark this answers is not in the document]
766C6164: Great suggestion, done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,189 @@
...
+      producer_plugin::get_snapshot_requests_result result;
+      auto& asvector = _snapshot_requests.get<as_vector>();
+      result.snapshot_requests.reserve(asvector.size());
+      result.snapshot_requests.insert(result.snapshot_requests.begin(), asvector.begin(), asvector.end());
+      return result;
+   }
+
+   // initialize with storage
+   void set_db_path(bfs::path db_path) {
+      _snapshot_db.set_path(std::move(db_path));
```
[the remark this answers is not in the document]
766C6164: yes, or probably clearing it up before insert is what should be expected behavior. done. Thank you!

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
```diff
@@ -0,0 +1,140 @@
...
+               // this means we are getting a snapshot for correct block # as well
+               BOOST_CHECK_EQUAL(9, pp->get_snapshot_requests().snapshot_requests.begin()->pending_snapshots->begin()->head_block_num);   
+            }
+         });
+       
+         producer_plugin::snapshot_request_information sri1 = {.block_spacing = 8, .start_block_num = 1, .end_block_num = 300000, .snapshot_description = "Example of recurring snapshot 2"};
+         producer_plugin::snapshot_request_information sri2 = {.block_spacing = 5000, .start_block_num = 100000, .end_block_num = 300000, .snapshot_description = "Example of recurring snapshot 2"};
+         producer_plugin::snapshot_request_information sri3 = {.block_spacing = 2, .start_block_num = 0, .end_block_num = 3, .snapshot_description = "Example of recurring snapshot 1"};
+
+         pp->schedule_snapshot(sri1);
```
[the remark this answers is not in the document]
766C6164: This is tricky, I am catching firing of sri1 snapshot request, which will immediately becomes pending and happening on start=1 + spacing=8. I'll see if I can do it in more clear/deterministic way. Thank you!

tests/nodeos_snapshot_diff_test.py
766C6164: Done, now 3 snapshots are being generated and compared to each other

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: No, it was abort, changed this to app().quit() and now shutdown graceful. Thank you!!!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
766C6164: good point, agreed, done. Thank you!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: added, reading this info from pending snapshot for simplicity

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,172 @@
...
+   virtual void on_block(uint32_t height) = 0;
+   virtual void on_irreversible_block(uint32_t height) = 0;
+   virtual void on_abort_block() = 0;
+};
+
+class snapshot_scheduler_handler {
+   virtual void schedule_snapshot(const snapshot_request_information& sri) = 0;
+   virtual void unschedule_snapshot(uint32_t sri) = 0;
+   virtual get_snapshot_requests_result get_snapshot_requests() = 0;
+};
```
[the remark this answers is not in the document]
766C6164: Dropped these, thank you!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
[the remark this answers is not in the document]
greg7mdp: Actually the `reserve()` is not needed because insert() already does it if needed.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,186 @@
...
+            if (req.block_spacing && height) {
+               auto& snapshot_by_id = _snapshot_requests.get<by_snapshot_id>();
+               auto it = snapshot_by_id.find(req.snapshot_request_id);
+               _snapshot_requests.modify(it, [&height](auto& p) { p.start_block_num = height; });
+               serialize_needed = true;              
+            }
+            execute_snapshot(req.snapshot_request_id);
+         }
+         // execute snapshot, -1 since its called from start block
+         else if(!req.block_spacing || (!((height - req.start_block_num - 1) % req.block_spacing))) {
```
greg7mdp, line 72: I believe this test is incorrect. Even if the snapshot specifies a block far in the future (for example `{.block_spacing = 0, .start_block_num = 1500, .end_block_num = 1500`, the snapshot will be taken the first time `on_start_block` is called after the snapshot is defined.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,186 @@
...
+            execute_snapshot(req.snapshot_request_id);
+         }       
+
+         // cleanup - remove expired (or invalid) request
+         if(!req.start_block_num || !req.block_spacing || (req.end_block_num > 0 && height >= req.end_block_num)) {
+            unschedule_snapshot(req.snapshot_request_id);
```
greg7mdp, line 78: suggestion: add `serialize_needed = true;` here and remove it from `unschedule_snapshot()`

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,186 @@
...
+         if(!req.start_block_num) {
+            // update start_block_num with current height only if this is recurring
+            // if non recurring, will be executed and unscheduled
+            if (req.block_spacing && height) {
+               auto& snapshot_by_id = _snapshot_requests.get<by_snapshot_id>();
+               auto it = snapshot_by_id.find(req.snapshot_request_id);
+               _snapshot_requests.modify(it, [&height](auto& p) { p.start_block_num = height; });
+               serialize_needed = true;              
+            }
+            execute_snapshot(req.snapshot_request_id);
```
greg7mdp, line 69: Whenever we call `execute_snapshot()`, we should `break;` out of the loop, as there is no reason to execute two snapshots at the same block height.

greg7mdp:
We really should have a debug log trace when we are taking snapshots I think.

greg7mdp:
If `block_spacing == 0`, should we require that `start_block == end_block`?
If they are different, are we taking the snapshot at `start_block` or `end_block`?

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,186 @@
...
+      }
+   }
+
+   // snapshot executor
+   void set_create_snapshot_fn(std::function<void(producer_plugin::next_function<producer_plugin::snapshot_information>)> fn) {
+      _create_snapshot = std::move(fn);
+   }
+
+   void execute_snapshot(uint32_t srid) {
+      _inflight_sid = srid;
```
greg7mdp, line 154: I think we should have a `dlog` call here indicating that we are taking a snapshot, and including the snapshot_request parameters (`start_block_num`, `end_block_num`, `block_spacing`), as well as the current block height.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: You totally right, i was tweaking this code and accidentally broke it. I tried to make it more readable now. Thank you!!!!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: unschedule_snapshot is also called directly from producer plugin that  expects serialization to happen as well

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: Agreed, done. Thank you!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
greg7mdp: OK, I missed that. So we could leave it as is, or add a parameter `bool serialize` to `unschedule_snapshot`. Up to you!

766C6164:
> If `block_spacing == 0`, should we require that `start_block == end_block`? If they are different, are we taking the snapshot at `start_block` or `end_block`?

by default start_block is used per our api specification and end_block ignored. In most cases user wont specify end_block at all (by default 0) and validation for end > start wont happen in this case. I am open to changing that if that sounds wrong

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: Thank you! Planned to do it, but forgot. Added dlog.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,205 @@
...
+         }        
+         else if(recurring_snapshot || onetime_snapshot) {
+            execute_snapshot_with_log(req);
+         }       
+
+         // cleanup - remove expired (or invalid) request
+         if(!req.start_block_num || !req.block_spacing || (req.end_block_num > 0 && height >= req.end_block_num)) {
+            unschedule_snapshot(req.snapshot_request_id);
+            snapshot_executed = true;
+         }
```
greg7mdp, line 95:
```suggestion
      bool snapshot_executed = false; 

      auto execute_snapshot_with_log = [this, &height, &snapshot_executed](const auto & req) {
         dlog("snapshot scheduler creating a snapshot from the request [start_block_num:${start_block_num}, end_block_num=${end_block_num}, block_spacing=${block_spacing}], height=${height}",
             ("start_block_num", req.start_block_num)
             ("end_block_num",   req.end_block_num)
             ("block_spacing",   req.block_spacing)
             ("height",          height));
             
         execute_snapshot(req.snapshot_request_id);
         snapshot_executed = true;
      };    

      for(const auto& req: _snapshot_requests.get<0>()) {
         // -1 since its called from start block
         bool recurring_snapshot =  req.block_spacing &&  (height >= req.start_block_num + 1) && (!((height - req.start_block_num - 1) % req.block_spacing));
         bool onetime_snapshot   = (!req.block_spacing) && (height == req.start_block_num + 1);
        
         // assume "asap" for snapshot with missed/zero start, it can have spacing
         if(!req.start_block_num) {
            // update start_block_num with current height only if this is recurring
            // if non recurring, will be executed and unscheduled
            if (req.block_spacing && height) {
               auto& snapshot_by_id = _snapshot_requests.get<by_snapshot_id>();
               auto it = snapshot_by_id.find(req.snapshot_request_id);
               _snapshot_requests.modify(it, [&height](auto& p) { p.start_block_num = height; });
               serialize_needed = true;              
            }
            execute_snapshot_with_log(req);
         }        
         else if(recurring_snapshot || onetime_snapshot) {
            execute_snapshot_with_log(req);
         }       

         // cleanup - remove expired (or invalid) request
         if(!req.start_block_num || !req.block_spacing || (req.end_block_num > 0 && height >= req.end_block_num)) {
            unschedule_snapshot(req.snapshot_request_id);
         }
```

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,205 @@
...
+            snapshot_executed = true;
+         }        
+         else if(recurring_snapshot || onetime_snapshot) {
+            execute_snapshot_with_log(req);
+         }       
+
+         // cleanup - remove expired (or invalid) request
+         if(!req.start_block_num || !req.block_spacing || (req.end_block_num > 0 && height >= req.end_block_num)) {
+            unschedule_snapshot(req.snapshot_request_id);
+            snapshot_executed = true;
```
greg7mdp, line 94: this `snapshot_executed = true;` should be after line 88. Even better capture `&snapshot_executed` in the `execute_snapshot_with_log` lambda and set it to true there.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
greg7mdp: Thanks for adding it!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: Ups... thank you!!! Fixed

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,205 @@
...
+            }
+            execute_snapshot_with_log(req);
+            snapshot_executed = true;
+         }        
+         else if(recurring_snapshot || onetime_snapshot) {
+            execute_snapshot_with_log(req);
+         }       
+
+         // cleanup - remove expired (or invalid) request
+         if(!req.start_block_num || !req.block_spacing || (req.end_block_num > 0 && height >= req.end_block_num)) {
```
greg7mdp, line 92: Hum, I believe this would remove all requests with `!req.block_spacing ` even before they are executed.

greg7mdp:
What about this simpler version:

```
   void on_start_block(uint32_t height) {
      bool serialize_needed  = false;

      for(const auto& req: _snapshot_requests.get<0>()) {
         if(!req.start_block_num && height) {
            // update start_block_num with current height only if not set
            auto& snapshot_by_id = _snapshot_requests.get<by_snapshot_id>();
            auto it = snapshot_by_id.find(req.snapshot_request_id);
            _snapshot_requests.modify(it, [&height](auto& p) { p.start_block_num = height - 1; });
            serialize_needed = true;              
         }
      }
      
      // store db to filesystem
      if (serialize_needed)
         x_serialize();

      for(const auto& req: _snapshot_requests.get<0>()) {   
         // -1 since its called from start block
         bool recurring_snapshot =  req.block_spacing &&  (height >= req.start_block_num + 1) && (!((height - req.start_block_num - 1) % req.block_spacing));
         bool onetime_snapshot   = (!req.block_spacing) && (height == req.start_block_num + 1);
        
         if(recurring_snapshot || onetime_snapshot) {
            dlog("snapshot scheduler creating a snapshot from the request [start_block_num:${start_block_num}, end_block_num=${end_block_num}, block_spacing=${block_spacing}], height=${height}",
                 ("start_block_num", req.start_block_num)
                 ("end_block_num",   req.end_block_num)
                 ("block_spacing",   req.block_spacing)
                 ("height",          height));
             
            execute_snapshot(req.snapshot_request_id);
            if (onetime_snapshot)
               unschedule_snapshot(req.snapshot_request_id);
            break; // stop iterating snapshot requests after snapshot execution
         }

         // cleanup - remove expired (or invalid) request
         if(req.block_spacing && (req.end_block_num > 0 && height >= req.end_block_num))
            unschedule_snapshot(req.snapshot_request_id);
      }
   }
```

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: Right! Thank you, appreciate it! I adjusted handling of this case and tested this scenario.

766C6164:
> What about this simpler version (untested):
> 
> ```
>    void on_start_block(uint32_t height) {
>       bool serialize_needed  = false;
> 
>       for(const auto& req: _snapshot_requests.get<0>()) {
>          if(!req.start_block_num && height) {
>             // update start_block_num with current height only if not set
>             auto& snapshot_by_id = _snapshot_requests.get<by_snapshot_id>();
>             auto it = snapshot_by_id.find(req.snapshot_request_id);
>             _snapshot_requests.modify(it, [&height](auto& p) { p.start_block_num = height - 1; });
>             serialize_needed = true;              
>          }
>       }
>       
>       // store db to filesystem
>       if (serialize_needed)
>          x_serialize();
> 
>       for(const auto& req: _snapshot_requests.get<0>()) {   
>          // -1 since its called from start block
>          bool recurring_snapshot =  req.block_spacing &&  (height >= req.start_block_num + 1) && (!((height - req.start_block_num - 1) % req.block_spacing));
>          bool onetime_snapshot   = (!req.block_spacing) && (height == req.start_block_num + 1);
>         
>          if(recurring_snapshot || onetime_snapshot) {
>             dlog("snapshot scheduler creating a snapshot from the request [start_block_num:${start_block_num}, end_block_num=${end_block_num}, block_spacing=${block_spacing}], height=${height}",
>                  ("start_block_num", req.start_block_num)
>                  ("end_block_num",   req.end_block_num)
>                  ("block_spacing",   req.block_spacing)
>                  ("height",          height));
>              
>             execute_snapshot(req.snapshot_request_id);
>             if (onetime_snapshot)
>                unschedule_snapshot(req.snapshot_request_id);
>             break; // stop iterating snapshot requests after snapshot execution
>          }
> 
>          // cleanup - remove expired (or invalid) request
>          if(req.block_spacing && req.end_block_num > 0 && height >= req.end_block_num)
>             unschedule_snapshot(req.snapshot_request_id);
>       }
>    }
> ```

That looks good, thank you! Before I saw this I adjusted my code to iterate through requests once, but completely since in one pass we are correcting missing start_block_num's and also executing snapshot if needed, but only once. Therefore I removed break, replaced with if in lambda

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,206 @@
...
+         }       
+
+         // cleanup - remove expired (or invalid) request
+         if(!req.start_block_num ||         
+            (!req.block_spacing && height >= (req.start_block_num + 1)) || 
+            (req.end_block_num > 0 && height >= (req.end_block_num + 1))) {
+            unschedule_snapshot(req.snapshot_request_id);
+         }
+      }
+
```
greg7mdp, line 101: There should be `if (snapshot_executed) break;` here.

766C6164: Yes, makes sense, added. Thank you!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,207 @@
...
+         else if(recurring_snapshot || onetime_snapshot) {
+            execute_snapshot_with_log(req);
+         }       
+
+         // cleanup - remove expired (or invalid) request
+         if(!req.start_block_num ||         
+            (!req.block_spacing && height >= (req.start_block_num + 1)) || 
+            (req.end_block_num > 0 && height >= (req.end_block_num + 1))) {
+            unschedule_snapshot(req.snapshot_request_id);
+            if (snapshot_executed) break;
```
greg7mdp, line 99: This `if` statement should be one line further (after the closing brace).

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,207 @@
...
+               serialize_needed = true;              
+            }
+            execute_snapshot_with_log(req);
+         }        
+         else if(recurring_snapshot || onetime_snapshot) {
+            execute_snapshot_with_log(req);
+         }       
+
+         // cleanup - remove expired (or invalid) request
+         if(!req.start_block_num ||
```
greg7mdp, line 95: This check `(!req.start_block_num || ` will remove all recurring snapshot requests which have `start_block_num == 0`, even if they are not expired. My version with two loops fixed that.

greg7mdp, line 95:
```suggestion
         if((!req.start_block_num && !req.block_spacing) ||         
```

766C6164: I am pretty sure at this line req.start_block_num==0 should not ever happen, being corrected few lines above, but I agree, your suggestion is safest one! Thank you!

766C6164: Done, thank you!!!

greg7mdp: > I am pretty sure at this line req.start_block_num==0 should not ever happen, being corrected few lines above, but I agree, your suggestion is safest one! Thank you!

I wasn't sure, because I see we update the request in boost multi-index, but `req` is a `const &`  to a view of that multi-index and I was not sure it was necessarily updated.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: This way it will break loop that may have another start_block_num's to correct... It needs to be completed fully and block for execution of snapshot is now in execute_snapshot_with_log lambda. What do you think?

greg7mdp: Yes, but we don't want to take two snapshots at the same block height.
In any case, if we already too a snapshot, it is OK to update the `start_block_num` the next time this function gets called (at height+1).

greg7mdp: Oh, yes, right, I missed that. So we probably can remove line 99, right?

766C6164: done! thanks again!

greg7mdp:
@766C6164 Thanks for all the changes! I think this is going to be a very popular feature :-).

Lin merged his large PR to main a few minutes ago. Maybe you should merge the latest changes from main (hopefully without too main conflicts), and then I'll do a last check and approve your PR, probably tomorrow morning.

766C6164:
> @766C6164 Thanks for all the changes! I think this is going to be a very popular feature :-).
> 
> Lin merged his large PR to main a few minutes ago. Maybe you should merge the latest changes from main (hopefully without too main conflicts), and then I'll do a last check and approve your PR, probably tomorrow morning.

Thank you so much!!! Really appreciate all your time and effort you put into reviewing this PR! Luckily picked up changes from main with only 3 minor conflicts :)

plugins/producer_api_plugin/producer_api_plugin.cpp
```diff
@@ -104,6 +109,18 @@ void producer_api_plugin::plugin_startup() {
             INVOKE_R_V(producer, get_greylist), 201),
        CALL_WITH_400(producer, producer, get_whitelist_blacklist,
             INVOKE_R_V(producer, get_whitelist_blacklist), 201),
+       CALL_WITH_400(producer, producer, set_whitelist_blacklist,
+            INVOKE_V_R(producer, set_whitelist_blacklist, producer_plugin::whitelist_blacklist), 201),
+       CALL_WITH_400(producer, producer, get_integrity_hash,
+            INVOKE_R_V(producer, get_integrity_hash), 201),
+       CALL_ASYNC(producer, producer, create_snapshot, producer_plugin::snapshot_information,
+            INVOKE_R_V_ASYNC(producer, create_snapshot), 201),
```
linh2931, line 117: Merge conflicts: `set_whitelist_blacklist`, `get_integrity_hash`, `get_integrity_hash` should be removed from here. They are not safe to read-only transactions and defined from line 147. Sorry for the trouble my PR causes.

linh2931 (changes requested):

plugins/producer_api_plugin/producer_api_plugin.cpp
linh2931: I checked out your branch and did a test. Confirmed the about problem caused CICD failures:

error 2023-03-16T13:52:42.400 nodeos    main.cpp:161                  main                 ] 3110006  plugin_config_exception: Incorrect plugin configuration
http url /v1/producer/create_snapshot is not unique
    {"u":"/v1/producer/create_snapshot"}
    nodeos  http_plugin.cpp:370 add_handler

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+            if (snapshot_req != snapshot_by_id.end()) {               
+               if (auto pending = snapshot_req->pending_snapshots; pending) {                 
+                  auto it = std::remove_if(pending->begin(), pending->end(), [&snapshot_info](const producer_plugin::snapshot_information & s){ return s.head_block_num <=  snapshot_info.head_block_num; });
+                  pending->erase(it, pending->end());
+                  _snapshot_requests.modify(snapshot_req, [&pending](auto& p) { 
+                     p.pending_snapshots = std::move(pending);
+                  });
+               }
+            }
```
linh2931, line 200: Do you want to report an error if `snapshot_req` was not found?

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,206 @@
...
+         }
+      };
+      _create_snapshot(next);
```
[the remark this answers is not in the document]
linh2931: https://github.com/AntelopeIO/leap/issues/830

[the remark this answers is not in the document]
heifner: Would be nice for that to be fixed before this is merged if possible, as that could cause a segfault I believe.

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
```diff
@@ -82,6 +82,26 @@ class producer_plugin : public appbase::plugin<producer_plugin> {
...
+      std::string          snapshot_description = "";
+
+   };
+
+   struct snapshot_request_id_information {
+      uint32_t snapshot_request_id = 0;
+   };
+
+   struct snapshot_schedule_information : public snapshot_request_id_information, public snapshot_request_information {
+      std::optional<std::vector<snapshot_information>> pending_snapshots;
```
greg7mdp, line 98: I don't see the benefit of making this a `std::optional`. Isn't just an empty vector the same?

Also, Just wondering, is this pending_snapshots vector, and its management in the functions  add_pending_snapshot_info and execute_snapshot used only in the tests?

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,92 @@
...
+   }
+
+   const snapshot_db_json& operator>>(std::vector<producer_plugin::snapshot_schedule_information>& sr) {
+      boost::property_tree::ptree root;
+
+      try {
+         std::ifstream file(get_json_path().string());
+         file.exceptions(std::istream::failbit|std::istream::eofbit);
+         boost::property_tree::read_json(file, root);
+         file.close();
```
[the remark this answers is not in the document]
766C6164: done, thank you!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,92 @@
...
+         node_srs.push_back(std::make_pair("", node));
+      }
+
+      root.push_back(std::make_pair("snapshot_requests", node_srs));
+
+      try {
+         std::ofstream file(get_json_path().string());
+         file.exceptions(std::istream::failbit|std::istream::eofbit);
+         boost::property_tree::write_json(file, root);
+         file.close();
```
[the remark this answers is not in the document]
766C6164: done!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_db_json.hpp
```diff
@@ -0,0 +1,92 @@
...
+      catch (std::ifstream::failure & e) {
+         elog( "unable to restore snapshots schedule from filesystem ${jsonpath}, details: ${details}",
+            ("jsonpath", get_json_path().string()) ("details",e.what()) );
+         appbase::app().quit();
+      }
+
+      return *this;
+   }
+
+   const snapshot_db_json& operator<<(std::vector<producer_plugin::snapshot_schedule_information>& sr) {
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,206 @@
...
+
+public:
+   snapshot_scheduler() = default;
+
+   // snapshot_scheduler_listener
+   void on_start_block(uint32_t height) {
+      bool serialize_needed  = false;
+      bool snapshot_executed = false; 
+
+      auto execute_snapshot_with_log = [this, &height, &snapshot_executed](const auto & req) {
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,206 @@
...
+            // if non recurring, will be executed and unscheduled
+            if (req.block_spacing && height) {
+               auto& snapshot_by_id = _snapshot_requests.get<by_snapshot_id>();
+               auto it = snapshot_by_id.find(req.snapshot_request_id);
+               _snapshot_requests.modify(it, [&height](auto& p) { p.start_block_num = height - 1; });
+               serialize_needed = true;              
+            }
+            execute_snapshot_with_log(req);
+         }        
+         else if(recurring_snapshot || onetime_snapshot) {
```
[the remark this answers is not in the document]
766C6164: done

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: done

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
```diff
@@ -0,0 +1,140 @@
...
+               num_empty = 10;
+            }
+         });
+         auto pp = appbase::app().find_plugin<producer_plugin>();
+         auto bs = chain_plug->chain().block_start.connect([&pp](uint32_t bn) {
+            // catching pending snapshot
+            if (pp->get_snapshot_requests().snapshot_requests.begin()->pending_snapshots && pp->get_snapshot_requests().snapshot_requests.begin()->pending_snapshots->size()==1) {
+               // lets check the head block num of it, it should be 8 + 1 = 9
+               // this means we are getting a snapshot for correct block # as well
+               BOOST_CHECK_EQUAL(9, pp->get_snapshot_requests().snapshot_requests.begin()->pending_snapshots->begin()->head_block_num);
```
greg7mdp, line 98: We never get here, the `pending_snapshots` vector is always empty.

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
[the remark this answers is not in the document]
766C6164: so @linh2931 do you think we should not change this? Thank you!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: not sure if we should gracefully exit here, probably not. un-finalized snapshot (should not happen) does not sound like terminal case.

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
greg7mdp: You are right, we get there. I'm not understanding why the check doesn't fail when we take a snapshot at block 3:

`snapshot scheduler creating a snapshot from the request [start_block_num:3, end_block_num=3, block_spacing=2], height=4`

766C6164: It fails for me in this case and fails hard, but not in this check if I modify sri1. Problem with start=end request that it will be removed immediately after execution so this code checking for "9" wont be reached

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
766C6164: no, this is exposed as api call, snapshot can take significant time to finalize so if there are any pending snapshot(s) triggered by a snapshot request, it will be returned in reply as json showing that it's in progress...

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -0,0 +1,208 @@
...
+               if (auto pending = snapshot_req->pending_snapshots; pending) {                 
+                  auto it = std::remove_if(pending->begin(), pending->end(), [&snapshot_info](const producer_plugin::snapshot_information & s){ return s.head_block_num <=  snapshot_info.head_block_num; });
+                  pending->erase(it, pending->end());
+                  _snapshot_requests.modify(snapshot_req, [&pending](auto& p) { 
+                     p.pending_snapshots = std::move(pending);
+                  });
+               }
+            } else {     
+               // should not happen, but if happens - not a terminal error          
+               elog( "unable to finalize pending snapshot request: ${details}", ("details", srid) );
```
greg7mdp, line 201: After your last change (addition of this line), I have the following error when running the test:

image

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
greg7mdp, line 98:
```suggestion
            auto& pending = pp->get_snapshot_requests().snapshot_requests.begin()->pending_snapshots;
            if (pending && pending->size()==1) {
               // lets check the head block num of it, it should be 8 + 1 = 9
               // this means we are getting a snapshot for correct block # as well
               BOOST_CHECK_EQUAL(9, pending->begin()->head_block_num);   
```

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
greg7mdp: I believe we hit this because the snapshot_request has already been removed from `_snapshot_requests` by the time `add_pending_snapshot_info` is called (even though we save its id in `_inflight_sid` - and I'm not sure why we do that - it is erased from the multi-index in `unschedule_snapshot`).

I suggest you just comment out this `elog` call for now - with a comment as why it is triggered.

greg7mdp (changes requested):

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: Thats totally correct, I was too fast adding this trace. It's normal behavior to have snapshot request removed (expired) and therefore any pendings should be gone as well.

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
766C6164: thats totally looks cleaner! thank you!

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
```diff
@@ -197,8 +197,8 @@ class snapshot_scheduler {
                   });
                }
             } else {     
-               // should not happen, but if happens - not a terminal error          
-               elog( "unable to finalize pending snapshot request: ${details}", ("details", srid) );
+               // this could happen when request is already "expired" so there is no value in reporting it          
+               // dlog( "unable to finalize pending snapshot request: ${details}", ("details", srid) );
```
linh2931, line 201: We normally do not leave commented out code behind. We remove them.

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
```diff
@@ -0,0 +1,141 @@
...
+         });
+         auto pp = appbase::app().find_plugin<producer_plugin>();
+         auto bs = chain_plug->chain().block_start.connect([&pp](uint32_t bn) {
+            // catching pending snapshot
+            auto& pending = pp->get_snapshot_requests().snapshot_requests.begin()->pending_snapshots;
+            if (pending && pending->size()==1) {
+               // lets check the head block num of it, it should be 8 + 1 = 9
+               // this means we are getting a snapshot for correct block # as well
+               BOOST_CHECK_EQUAL(9, pending->begin()->head_block_num);   
+            }
```
greg7mdp, line 100: This could happen if block_start is called before the snapshots are scheduled, of if we schedules only one non-recurring snapshot.

```suggestion
            if (!pp->get_snapshot_requests().snapshot_requests.empty()) {
               auto& pending = pp->get_snapshot_requests().snapshot_requests.begin()->pending_snapshots;
               if (pending && pending->size()==1) {
                  // lets check the head block num of it, it should be 8 + 1 = 9
                  // this means we are getting a snapshot for correct block # as well
                  BOOST_CHECK_EQUAL(9, pending->begin()->head_block_num);   
               }
            }
```

greg7mdp (changes requested):

plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
766C6164: lets remove it! thank you!

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
766C6164: agreed, done

greg7mdp (approved):

linh2931 (approved):

PR #770, opened 2023-03-02
Test: Add a test that verifies restart from genesis not allowed

heifner:
Add a test that verifies restart from genesis not allowed.
See #764

### DIFF
--- a/plugins/state_history_plugin/tests/session_test.cpp
+++ b/plugins/state_history_plugin/tests/session_test.cpp
@@ -425,11 +425,12 @@ BOOST_AUTO_TEST_CASE(store_with_existing) {
    // Do not allow starting from scratch for existing
    header.block_id     = block_id_for(1);
    BOOST_CHECK_EXCEPTION(
-         log.pack_and_write_entry(header, block_id_for(0),
-                                  [&](auto&& buf) { bio::write(buf, (const char*)data.data(), data.size() * sizeof(data[0])); }),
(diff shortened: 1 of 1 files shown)


plugins/state_history_plugin/tests/session_test.cpp
```diff
@@ -405,6 +405,33 @@ BOOST_AUTO_TEST_CASE(store_read_entry_prune_enabled) {
...
+   log.pack_and_write_entry(header, block_id_for(1),
+                            [&](auto&& buf) { bio::write(buf, (const char*)data.data(), data.size() * sizeof(data[0])); });
+
+   // Do not allow starting from scratch for existing
+   header.block_id     = block_id_for(1);
+   BOOST_CHECK_EXCEPTION(
+         log.pack_and_write_entry(header, block_id_for(0),
+                                  [&](auto&& buf) { bio::write(buf, (const char*)data.data(), data.size() * sizeof(data[0])); }),
+         eosio::chain::plugin_exception, [](const auto& e) {
+      return e.to_detail_string().find("Existing ship log") != std::string::npos;
```
greg7mdp, line 431: small detail. it feels like this line - the body of the lambda - should be indented more.

greg7mdp (approved):

plugins/state_history_plugin/tests/session_test.cpp
heifner: I agree, auto-format failed me here.

linh2931 (approved):

PR #774, opened 2023-03-03
Fix issue with reported fork head vs retrieval of block from fork db

### LINKED ISSUE #772: Test Failure: trx_finality_status_forked_test
https://github.com/AntelopeIO/leap/actions/runs/4318231823/jobs/7536367066

```
Traceback (most recent call last):
  File "/__w/leap/leap/build/tests/trx_finality_status_forked_test.py", line 189, in <module>
    assert prodC.waitForLibToAdvance(), \
AssertionError: ERROR: Network did not reach consensus after bridge node was restarted.
```

heifner:
When `IRREVERSIBLE` mode was added https://github.com/EOSIO/eos/pull/4424 access to fork db head was added to controller, during the refactor of fork database https://github.com/EOSIO/eos/pull/6624 access to fork db pending was added to deal with irreversible mode. 

This PR simplifies this interface by returning the expected values for fork db head according to `IRREVERSIBLE` mode or not. This conforms with the existing functionality of `fetch_block_state_by_number` which looks at `IRREVERSIBLE` mode to determine what branch to search on.

Not sure if this should be back-ported. We have not seen the test failure on previous versions. My guess is the performance improvements in 4.0 around not having to jump to the main thread to get blocks is causing this to be much more likely.

Resolves #772


libraries/chain/controller.cpp
```diff
@@ -3011,35 +3011,19 @@ block_state_ptr controller::head_block_state()const {
...
-   return my->fork_db.pending_head()->header.timestamp;
-}
-
-account_name  controller::fork_db_pending_head_block_producer()const {
-   return my->fork_db.pending_head()->header.producer;
+   if( my->read_mode == db_read_mode::IRREVERSIBLE ) {
+      return my->fork_db.pending_head()->id;
+   } else {
+      return my->fork_db.head()->id;
+   }
```
greg7mdp, line 3026: It looks like we could have an internal function:

`block_state_ptr get_fork_head() { return read_mode == db_read_mode::IRREVERSIBLE) ? fork_db.pending_head() : fork_db.head(); }`

which could be used to simplify the above functions, as well as `fetch_block_state_by_number` and other locations in controller.coo.

greg7mdp: it sounds counter-intuitive that, when asking for IRREVERSIBLE, we return the `pending` head?

greg7mdp (changes requested):

libraries/chain/controller.cpp
heifner: When running in IRREVERSIBLE, "head" is irreversible; but you still are receiving blocks and tracking head of the chain.

heifner: I thought about adding a `get_fork_head`, I didn't because I thought it was clearer to have it obvious in `fork_db_head_block_num` and `fork_db_head_block_id`. I can add it if you think it is cleaner.

greg7mdp: Hum. What about calling the function `get_irreversible_fork_head` to indicate it takes into account the mode?

heifner: It is only irreversible if you are running in IRREVERSIBLE mode.

greg7mdp: OK, my understanding :-). `get_fork_db_head` should be in a header.
```
block_state_ptr controller::get_fork_db_head() {
   // this function returns the last block, even when read_mode is IRREVERSIBLE
   // note: In normal mode, we return head(), which is the last validated block
   //       in IRREVERSIBLE mode, we return pending_head() which is the newest block we
   //       have, and which may or may not be validated.
   return read_mode == db_read_mode::IRREVERSIBLE ?
      fork_db.pending_head() :  // should we make sure we return only validated blocks?
      fork_db.head();
}

uint32_t controller::fork_db_head_block_num()const {
   return get_fork_db_head()->block_num;
}

block_id_type controller::fork_db_head_block_id()const {
   return get_fork_db_head()->id;
}
```

greg7mdp: I'm not quite understanding what `pending_head` is doing? it looks like it is either returning `my-head`, just like `head()`, or a block from indx if that block has a higher `dpos_irreversible_blocknum`?

```
   block_state_ptr fork_database::pending_head()const {
      std::shared_lock g( my->mtx );
      const auto& indx = my->index.get<by_lib_block_num>();

      auto itr = indx.lower_bound( false );
      if( itr != indx.end() && !(*itr)->is_valid() ) {
         if( first_preferred( **itr, *my->head ) )
            return *itr;
      }

      return my->head;
   }
```

heifner: The difference is `pending_head()` is returning the best/tip/preferred block that is not validated or head if head is best/tip/preferred.

heifner: Added `fork_db_head()`

greg7mdp (approved):

unittests/fork_test_utilities.cpp
```diff
@@ -9,10 +9,10 @@ public_key_type  get_public_key( name keyname, string role ){
 }
 
 void push_blocks( tester& from, tester& to, uint32_t block_num_limit ) {
-   while( to.control->fork_db_pending_head_block_num()
-            < std::min( from.control->fork_db_pending_head_block_num(), block_num_limit ) )
+   while( to.control->fork_db_head_block_num()
+            < std::min( from.control->fork_db_head_block_num(), block_num_limit ) )
    {
-      auto fb = from.control->fetch_block_by_number( to.control->fork_db_pending_head_block_num()+1 );
+      auto fb = from.control->fetch_block_by_number( to.control->fork_db_head_block_num()+1 );
```
greg7mdp, line 15: If this is used in non IRREVERSIBLE mode, it will not do the same thing. Is it on purpose?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2728,8 +2728,8 @@ namespace eosio {
       chain_lib_id = cc.last_irreversible_block_id();
       chain_head_blk_num = cc.head_block_num();
       chain_head_blk_id = cc.head_block_id();
-      chain_fork_head_blk_num = cc.fork_db_pending_head_block_num();
-      chain_fork_head_blk_id = cc.fork_db_pending_head_block_id();
+      chain_fork_head_blk_num = cc.fork_db_head_block_num();
+      chain_fork_head_blk_id = cc.fork_db_head_block_id();
```
greg7mdp, line 2732: in non-irreversible mode, we will store the same in `chain_head_blk_num` and `chain_fork_head_blk_num`. Same for id. Is this what we want?

unittests/fork_test_utilities.cpp
heifner: All tests pass ;). The point of this method is to copy blocks from one test chain to another. It uses `fetch_block_by_number` to do that, so it could only copy what that could find, so the actual functionality has not changed here.

unittests/state_history_tests.cpp
```diff
@@ -742,10 +742,10 @@ BOOST_AUTO_TEST_CASE(test_splitted_log) {
 }
 
 void push_blocks( tester& from, tester& to ) {
-   while( to.control->fork_db_pending_head_block_num()
-            < from.control->fork_db_pending_head_block_num() )
+   while( to.control->fork_db_head_block_num()
+            < from.control->fork_db_head_block_num() )
    {
-      auto fb = from.control->fetch_block_by_number( to.control->fork_db_pending_head_block_num()+1 );
+      auto fb = from.control->fetch_block_by_number( to.control->fork_db_head_block_num()+1 );
```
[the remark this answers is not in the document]
heifner: Same comment as above.

plugins/net_plugin/net_plugin.cpp
heifner: Thanks. `chain_head_*` is not used.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2726,21 +2724,19 @@ namespace eosio {
...
-   std::tuple<uint32_t, uint32_t, uint32_t, block_id_type, block_id_type, block_id_type>
+   //        lib_num, fork_head_blk_num, lib_id, fork_head_blk_id
+   std::tuple<uint32_t, uint32_t, block_id_type, block_id_type>
    net_plugin_impl::get_chain_info() const {
       std::lock_guard<std::mutex> g( chain_info_mtx );
       return std::make_tuple(
-            chain_lib_num, chain_head_blk_num, chain_fork_head_blk_num,
-            chain_lib_id, chain_head_blk_id, chain_fork_head_blk_id );
+            chain_lib_num, chain_fork_head_blk_num,
+            chain_lib_id, chain_fork_head_blk_id );
```
greg7mdp, line 2739: This would work as well: `return { chain_lib_num, chain_fork_head_blk_num, chain_lib_id, chain_fork_head_blk_id };`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1707,8 +1705,8 @@ namespace eosio {
 
       uint32_t lib_num = 0;
       uint32_t fork_head_block_num = 0;
-      std::tie( lib_num, std::ignore, fork_head_block_num,
-                std::ignore, std::ignore, std::ignore ) = my_impl->get_chain_info();
+      std::tie( lib_num, fork_head_block_num,
+                std::ignore, std::ignore ) = my_impl->get_chain_info();
```
greg7mdp, line 1710: I think it would be cleaner if `get_chain_info()` returned a `chain_info_t` struct with 4 members, instead of a tuple and the  extra variable declarations.

So these 4 lines would be replaced with:

`auto chain_info = my_impl->get_chain_info();`

greg7mdp (commented):
Nice, it is getting cleaner. I left a couple more suggestions.

plugins/net_plugin/net_plugin.cpp
heifner: Done.

plugins/net_plugin/net_plugin.cpp
heifner: Done.

greg7mdp: Nice, thank you @heifner

greg7mdp (approved):

linh2931 (approved):

Issue #760, opened 2023-03-01
Add configuration for 3-strike-rule time frame


huangminghuang:
Based on the conversion from https://github.com/AntelopeIO/leap/issues/315#issuecomment-1329968231

Add a new configuration to the existing 3-strike-rule
- Currently has configurable amount of strikes
  - `subjective-account-max-failures` - `Sets the maximum amount of failures that are allowed for a given account per block.` Defaults to 3.
  - Change this description to say `.. per subjective-account-max-failure-time-ms`
- Add new option
  - `subjective-account-max-failure-time-ms` - `Sets the time over which subjective-account-max-failures is enforced.` Defaults to 500ms.

heifner:
Target change for 3.1, 3,2, & main.

matthewdarwin:
So we would have the 3-strike-rule on a per block basis and a new rule for multiple blocks?  Can we have some names for these things so we can explain them?

heifner:
The idea is to change the hard-coded block time to be configurable.

matthewdarwin:
Ah, so not a different rule. So really it is adding a new configuration option, as was originally.  Title of this issue could probably be more clear.

heifner:
Would time in milliseconds or number of blocks be better?

DenisCarriere:
**Per block** seems like it would be a better fit

I'm assuming the milliseconds would need to match the block intervals (500ms), correct?

heifner:
> **Per block** seems like it would be a better fit
> 
> I'm assuming the milliseconds would need to match the block intervals (500ms), correct?

No, it could be reset at any time. If time based you would just check the time at each failure (or start of trx).

DenisCarriere:
Ahh.. well in that case i'll leave it to @matthewdarwin to provide a preference

I don't have any objections to either per block or per milliseconds

milliseconds would be the most flexible if it's not bound to block time intervals.

matthewdarwin:
Time is more flexible.

**But** there is discussion about having block intervals change (depending on the fullness of the block and release it early).  When that that features is implemented, then I think you need this feature to be by block and not by time.  Unless that feature is magically moving the time forward and the counter that would reset the 3-strikes-rule automatically moves forward as well.

heifner:
[...]

PR #777, opened 2023-03-03
[3.1] subjective-account-max-failures-window-size

huangminghuang:
This PR adds a new configuration option `--subjective-account-max-failures-window-size` to set the window size in number of blocks for `--subjective-account-max-failures`  option. 

Address IS #760

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -686,8 +686,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             auto first_auth = trx->packed_trx()->get_transaction().first_authorizer();
             if( _account_fails.failure_limit( first_auth ) ) {
                send_response( std::static_pointer_cast<fc::exception>( std::make_shared<tx_cpu_usage_exceeded>(
-                     FC_LOG_MESSAGE( error, "transaction ${id} exceeded failure limit for account ${a}",
-                                     ("id", trx->id())("a", first_auth) ) ) ) );
+                     FC_LOG_MESSAGE( error, "transaction ${id} exceeded failure limit for account ${a} until ${next_reset_time}",
+                                     ("id", trx->id())("a", first_auth)
+                                     ("next_reset_time", _account_fails.next_reset_timepoint(chain.head_block_num(),chain.head_block_time())) ) ) ) );
                return true;
             }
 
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -673,10 +685,14 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-                     FC_LOG_MESSAGE( error, "transaction ${id} exceeded failure limit for account ${a}",
-                                     ("id", trx->id())("a", first_auth) ) ) ) );
+            if (_account_fails.failure_limit(first_auth)) {
+               send_response(
+                     std::static_pointer_cast<fc::exception>(std::make_shared<tx_cpu_usage_exceeded>(FC_LOG_MESSAGE(
+                           error,
+                           "transaction ${id} exceeded failure limit for account ${a} until ${next_reset_timepoint}",
+                           ("id", trx->id())("a", first_auth)(
+                                 "next_reset_timepoint",
+                                 _account_fails.next_reset_timepoint(chain.head_block_num(), bt))))));
```
heifner, line 695: Please cleanup this formatting.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2023,10 +2047,13 @@ bool producer_plugin_impl::process_unapplied_trxs( const fc::time_point& deadlin
...
-                  itr->next( std::make_shared<tx_cpu_usage_exceeded>(
-                        FC_LOG_MESSAGE( error, "transaction ${id} exceeded failure limit for account ${a}",
-                                        ("id", trx->id())("a", first_auth) ) ) );
+               if (itr->next) {
+                  itr->next(std::make_shared<tx_cpu_usage_exceeded>(FC_LOG_MESSAGE(
+                        error,
+                        "transaction ${id} exceeded failure limit for account ${a} until ${next_reset_timepoint}",
+                        ("id", trx->id())("a", first_auth)(
+                              "next_reset_timepoint",
+                              _account_fails.next_reset_timepoint(chain.head_block_num(), chain.head_block_time())))));
```
heifner, line 2056: Please cleanup this formatting.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2439,8 +2466,7 @@ void producer_plugin_impl::produce_block() {
 
    block_state_ptr new_bs = chain.head_block_state();
 
-   _account_fails.report();
-   _account_fails.clear();
+   _account_fails.report(new_bs->block_num);
```
heifner, line 2469: I think you should make it `report_and_clear(uint32_t block_num)` and only call it from start_block. The `-1` in `report` vs `clear` seems a bit fragile.

heifner (changes requested):

heifner (approved):

heifner:
> Just curious, why do we do the `++` here?
> 
> https://github.com/AntelopeIO/leap/blob/71548a642577e0757f0e462bdaba20afa4f3841e/plugins/producer_plugin/producer_plugin.cpp#L223

Because we log the number of failures and want to include this failure in that number.

greg7mdp:
> Because we log the number of failures and want to include this failure in that number.

But there is an add() function above. It seems that we should have an add() function incrementing the number of failures, and a `failure_limit` to only check if the limit was exceeded?

heifner:
> > Because we log the number of failures and want to include this failure in that number.
> 
> But there is an add() function above. It seems that we should have an add() function incrementing the number of failures, and a `failure_limit` to only check if the limit was exceeded?

We don't want to add(), only increase the failure count.

greg7mdp (approved):

PR #778, opened 2023-03-03
[PH] Update Log Reader Tests For scrapeBlockTrxDataLog

ClaytonCalabrese:
A recent bug revealed a gap in the `log_reader_tests` for performance harness relating to `scrapeBlockTrxDataLog`, this PR addresses the location that this bug occurred so that if broken again in the future, it will be caught automatically.


tests/performance_tests/log_reader.py
```diff
@@ -237,11 +237,11 @@ def scrapeTrxGenLog(trxSent, path):
     with selectedopen(path, 'rt') as f:
         trxSent.update(dict([(x[0], x[1]) for x in (line.rstrip('\n').split(',') for line in f)]))
 
-def scrapeBlockTrxDataLog(trxDict, path):
+def scrapeBlockTrxDataLog(trxDict, path, nodeosVersion=None):
     #blockTrxData.txt
     selectedopen = selectedOpen(path)
     with selectedopen(path, 'rt') as f:
-        if Utils.getNodeosVersion().split('.')[0] == "v2":
+        if Utils.getNodeosVersion().split('.')[0] == "v2" or nodeosVersion == "v2":
```
oschwaldp-oci, line 244: Should just use `nodeosVersion` here and remove the `Utils.getNodeosVersion()` call.

Might be worth looking at our use of `Utils.getNodeosVersion()` throughout performance harness and see if we can limit this to a single place and then always get it out of `ClusterConfig`, for example.  Makes it easier to update in the future if we move away from `Utils.getNodeosVersion()` as the way we get nodeos version.

oschwaldp-oci (changes requested):

oschwaldp-oci (approved):

tests/performance_tests/log_reader.py
oschwaldp-oci: Thanks.

tests/performance_tests/log_reader_tests.py
```diff
@@ -192,6 +192,14 @@
 expectedTpsStats = log_reader.stats(41, 41, 41, 0.0, 0, 2)
 assert expectedTpsStats == stats , f"Error: Stats calculated: {stats} did not match expected stats: {expectedTpsStats}"
 
+#ensure that scraping of trxDataLog is compatible with 2.0
+trxDict = {}
+log_reader.scrapeBlockTrxDataLog(trxDict=trxDict, path=Path("tests")/"performance_tests"/"block_trx_data_log_2_0_14.txt", nodeosVers="v2")
```
oschwaldp-oci, line 197: block_trx_data_log_2_0_14.txt.gz

oschwaldp-oci (changes requested):

oschwaldp-oci (approved):

PR #785, opened 2023-03-07
[3.1 -> 3.2] subjective-account-max-failures-window-size

### LINKED ISSUE #760: Add configuration for 3-strike-rule time frame

huangminghuang:
This PR merges PR #777 into release/3.2 branch with the following changes

- adds a new configuration option --subjective-account-max-failures-window-size to set the window size in number of blocks for --subjective-account-max-failures option.
- extract the block tracking functionality from `account_failures` class into a `block_time_tracker` class.

Resolves #760

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -146,21 +146,6 @@ public:
       reset_window_size_in_num_blocks = size;
    }
 
-   void add_idle_time( const fc::microseconds& idle ) {
-      block_idle_time += idle;
-   }
-
-   void add_fail_time( const fc::microseconds& fail_time ) {
-      trx_fail_time += fail_time;
-      ++trx_fail_num;
-   }
-
-   void add_success_time( const fc::microseconds& time ) {
-      trx_success_time += time;
-      ++trx_success_num;
-   }
-
-
    void add( const account_name& n, int64_t exception_code ) {
       auto& fa = failed_accounts[n];
       ++fa.num_failures;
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -414,11 +436,10 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          auto& chain = chain_plug->chain();
 
          if( chain.is_building_block() ) {
-            _account_fails.report( _idle_trx_time );
+            _time_tracker.report_and_clear( _idle_trx_time );
```
heifner, line 439: It is ok to clear here, but not needed. So I think we can separate `report()` from `clear()` for `time_tracker`.

heifner (changes requested):
A similar fix as indicated in this review is needed in 3.1.

heifner (approved):

heifner:
> A similar fix as indicated in this review is needed in 3.1.

3.1 does not have reporting of block info like 3.2, no change is needed in 3.1.

PR #745, opened 2023-02-23
Remove HTTPS support from nodeos & keosd

greg7mdp:
Resolves #643.

- removed HTTPS support (and code) from http_plugin
- removed HTTPS support (and code) from http_client_plugin (mostly in libfc)
- removed `OpenSSL::SSL` from libraries/libfc/CMakeLists.txt

Issue #643, opened 2023-01-17
Remove HTTPS support from nodeos & keosd


spoonincode:
As documented in [Leap's deprecation notices](https://github.com/AntelopeIO/leap/wiki/Deprecations-and-Proposed-Deprecations-In-Leap-Software), HTTPS support in nodeos & keosd is to move removed.

This should be fairly simple, notable tasks are to
* Remove HTTPS support from http_plugin
* Remove HTTPS support from http_client_plugin

At the conclusion of this work there should no longer be the need for linking `OpenSSL::SSL` anywhere in Leap.

greg7mdp:
@spoonincode should I remove the https code from `libraries/libfc/src/network/http/` as well? This is what `http_client_plugin` uses.

PR #787, opened 2023-03-07
Cleanup some missed https support from test that was missed.

oschwaldp-oci:
Relates to https://github.com/AntelopeIO/leap/issues/643 and https://github.com/AntelopeIO/leap/pull/745


heifner (approved):

greg7mdp (approved):
Thanks for finding and fixing this!

PR #794, opened 2023-03-08
fix create_metrics_listener

huangminghuang:
The PR fixes a bug where it reuses a moved value

### DIFF
--- a/plugins/prometheus_plugin/prometheus_plugin.cpp
+++ b/plugins/prometheus_plugin/prometheus_plugin.cpp
@@ -127,9 +127,9 @@ namespace eosio {
       }
 
       metrics_listener create_metrics_listener(std::string plugin_name) {
-         return  [plugin_name{std::move(plugin_name)}, self=this] (vector<runtime_metric> metrics) mutable {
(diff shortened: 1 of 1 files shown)


plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -129,7 +129,7 @@ namespace eosio {
       metrics_listener create_metrics_listener(std::string plugin_name) {
          return  [plugin_name{std::move(plugin_name)}, self=this] (vector<runtime_metric> metrics) mutable {
```
heifner, line 130: You can remove `mutable` here.

heifner (changes requested):

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -129,7 +129,7 @@ namespace eosio {
...
             self->_prometheus_strand.post(
-                  [self, plugin_name{std::move(plugin_name)}, metrics{std::move(metrics)}]() mutable {
```
heifner, line 132: This `mutable` is needed. I was talking about the one on line 130.

heifner (approved):

PR #799, opened 2023-03-09
Search for LLVM versions 7-11, prefering latest version.

ScottBailey:
With multiple versions of LLVM installed, `find_package(LLVM REQUIRED CONFIG)` finds the latest version. Since I somehow seem to have versions 11, 13, 14, and 15 installed, this is just painful. This PR adds a few lines so that - when CMake supports it - `find_package` is called with version ranges. Because of API changes on LLVM minor modes and the way CMake evaluates versions, it's necessary to iterate over minor ranges as well as major.

I've tested this using: 
- Debian bookworm with llvm 11, 13, 14, & 15 installed
- Ubuntu 22 with llvm 11 and 14 installed.


CMakeLists.txt
```diff
@@ -77,9 +77,29 @@ if(ENABLE_OC AND CMAKE_SIZEOF_VOID_P EQUAL 8 AND NOT WIN32)
...
+               math(EXPR minor "${minor} - 1")
+            else()
+               math(EXPR major "${major} - 1")
+               set(minor ${max_minor})
+            endif()
+         endwhile()
+      endif()
       if(LLVM_VERSION_MAJOR VERSION_LESS 7 OR LLVM_VERSION_MAJOR VERSION_GREATER_EQUAL 12)
-	      message(FATAL_ERROR "Leap requires an LLVM version 7 through 11")
+	 message(FATAL_ERROR "Leap requires an LLVM version 7 through 11")
```
spoonincode, line 102: is there a tab here? otherwise LGTM. Consider changing the README to remove the `CMAKE_PREFIX_PATH`
https://github.com/AntelopeIO/leap/blob/8bf37aac8ab402a328035ca9f321f43e5097fc54/README.md?plain=1#L149

spoonincode: It looks like there was and still is a tab on this line. Pretty weird, wonder how it got there. Might as well remove it while we're here.

shot-2023-03-10_09-45-18

spoonincode (approved):

CMakeLists.txt
ScottBailey: So there is. I seem to have removed the automatic untabify from my emacs config. I've untabified and pushed the change.

heifner (approved):

PR #798, opened 2023-03-09
Remove deprecated `--private-key` option from nodeos

heifner:
Remove the the long deprecated `--private-key` option.
Use `--signature-provider` instead.

See https://github.com/AntelopeIO/leap/wiki/Deprecations-and-Proposed-Deprecations-In-Leap-Software

For example:
```
--private-key [\"EOS6hMjoWRF2L8x9YpeqtUEcsDKAyxSuM1APicxgRU1E3oyV5sDEg\",\"5JgbL2ZnoEAhTudReWH1RnMuQS6DBeLZt4ucV6t8aymVEuYg7sr\"]
```
Becomes:
```
--signature-provider EOS6hMjoWRF2L8x9YpeqtUEcsDKAyxSuM1APicxgRU1E3oyV5sDEg=KEY:5JgbL2ZnoEAhTudReWH1RnMuQS6DBeLZt4ucV6t8aymVEuYg7sr
```

Resolves #789

PR #802, opened 2023-03-10
remove eosio-blocklog

### LINKED ISSUE #790: Remove blocklog-util
We need to first verify that leap-util captures functionality parity with blocklog-util before removing. This does not include any functionality added in versions of eosio that were released after where mandel and subsequently antelope was forked from (there is know additional functionality in 2.2 and later).

Part of this work needs to verify whether only a rename is required to ingest this update, which we'd then cover in the release notes. If more support materials are required that should be communicated in the completion of this tasks.

huangminghuang:
The PR 
- removes the deprecated `eosio-blocklog` program. please use `leap-util block-log` instead.
- changes the output file name when using `leap-util block-log extract-blocklog`. Before the PR, the output filenames are `blocks.log` and `blocks.index` in the specified output directory. This PR makes the output filename suffixed with the start and end block numbers. For example, `blocks-100-200.log` and `blocks-100-200.index`.

See https://github.com/AntelopeIO/leap/wiki/Deprecations-and-Proposed-Deprecations-In-Leap-Software

Here are some examples of the equivalent command lines options between `eosio-blocklog` and `leap-util block-log`:

  - `eosio-blocklog --as-json-array --blocks-dir .` ==> `leap-util block-log print-log --as-json-array --blocks-dir .`
  - `eosio-blocklog --trim-blocklog -f 1 -l 100 --blocks-dir .` ==> `leap-util block-log trim-blocklog -f 1 -l 100 --blocks-dir .`
  - `eosio-blocklog --extract-blocks -f 1 -l 100 --blocks-dir . --output-dir newdir` ==> `leap-util block-log extract-blocklog -f 1 -l 100 --blocks-dir . --output-dir newdir`

Resolves #790


heifner (changes requested):
Update PR description with an example of using leap-util instead of eosio-blocklog.
Update PR description pointing to deprecation list, see for example: https://github.com/AntelopeIO/leap/pull/798
Update PR description with a Resloves.
Remove reference of eosio-blocklog from `docs/*`.
Fix reference to eosio-blocklog in `programs/leap-util/actions/blocklog.cpp`
Replace use in tests/TestHarness with leap-util.

docs/10_utilities/index.md
```diff
@@ -5,5 +5,4 @@ link_text: Antelope Utilities
 
 This section contains documentation for additional utilities that complement or extend `nodeos` and potentially other Antelope software:
```
766C6164, line 7: Not in the scope of this PR, but here probably where user would expect link to separate md file describing leap-util, I feel like with removal of eosio-blocklog.md users loose valuable guide, it just became a subcommand in leap-util now.

766C6164 (approved):

libraries/chain/block_log.cpp
```diff
@@ -1557,22 +1547,21 @@ namespace eosio { namespace chain {
    }
 
    // static
-   void block_log::extract_blocklog(const fc::path& log_filename, const fc::path& index_filename,
-                                    const fc::path& dest_dir, uint32_t start_block_num, uint32_t num_blocks) {
+   void block_log::extract_block_range(const fc::path& block_dir, const fc::path& dest_dir,
+                                       block_num_type start_block_num, block_num_type end_block_num) {
```
ScottBailey, line 1551: Generally, I'd prefer we use `end` as a synonym for 1 past `last` instead of `last`. This pattern is familiar to us and should be expected. Because we use end synonymous with last in other places in this file, I don't think it's necessary to change it in this one place unless we have bandwidth to change it throughout the file.

programs/leap-util/actions/blocklog.hpp
```diff
@@ -31,7 +31,7 @@ class blocklog_actions : public sub_command<blocklog_options> {
    void initialize();
    int trim_blocklog_end(bfs::path block_dir, uint32_t n);
    bool trim_blocklog_front(bfs::path block_dir, uint32_t n);
-   bool extract_block_range(bfs::path block_dir, bfs::path output_dir, uint32_t start, uint32_t end);
+   void extract_block_range(bfs::path block_dir, bfs::path output_dir, uint32_t start, uint32_t end);
```
ScottBailey, line 34: Without context, I would assume `end` was `last+1`. Let me suggest we update the variable to `last` here.
```suggestion
   void extract_block_range(bfs::path block_dir, bfs::path output_dir, uint32_t start, uint32_t last);
```

libraries/chain/include/eosio/chain/block_log.hpp
```diff
@@ -95,7 +95,7 @@ namespace eosio { namespace chain {
 
          static bool is_pruned_log(const fc::path& data_dir);
 
-         static bool extract_block_range(const fc::path& block_dir, const fc::path&output_dir, block_num_type& start, block_num_type& end, bool rename_input=false);
+         static void extract_block_range(const fc::path& block_dir, const fc::path&output_dir, block_num_type start, block_num_type end);
```
ScottBailey, line 98: `end` normal means `last+1`, I think switcheing `end` to `last` here would improve clarity.
```suggestion
         static void extract_block_range(const fc::path& block_dir, const fc::path&output_dir, block_num_type start, block_num_type last);
```

ScottBailey (approved):
I added a couple comments regarding `end` and `last`. It would be nice if these were updated; however, they aren't a priority.

ScottBailey (approved):
Thank you for the end/last changes!

heifner (approved):

PR #805, opened 2023-03-11
Improve logging during syncing when `producer_plugin` `debug` level is enabled

heifner:
If `producer_plugin` `debug` level is enabled during syncing up to head of chain there was too much logging that is not useful. Reduce the amount of logs during sync.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -410,8 +410,10 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          auto before = _unapplied_transactions.size();
          _unapplied_transactions.clear_applied( bsp );
          _subjective_billing.on_block( _log, bsp, fc::time_point::now() );
-         fc_dlog( _log, "Removed applied transactions before: ${before}, after: ${after}",
-                  ("before", before)("after", _unapplied_transactions.size()) );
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -465,9 +467,11 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          const auto& id = block_id ? *block_id : block->calculate_id();
          auto blk_num = block->block_num();
 
-         fc_dlog(_log, "received incoming block ${n} ${id}", ("n", blk_num)("id", id));
+         auto now = fc::time_point::now();
+         if (now - block->timestamp < fc::minutes(5) || (blk_num % 1000 == 0))
```
linh2931, line 471: Should `<` be `>` so you don't log again if the last logging is less than 5 minutes ago?

heifner: The `if` is to determine if this is a recently produced block or one from over 5 minutes ago.

linh2931: Oh I see. I first thought you wanted to log every 1000 blocks or every 5 minutes. Could you add a comment there?

linh2931 (approved):

PR #776, opened 2023-03-03
Parallelize Read-only Transaction Execution

linh2931:
PR https://github.com/AntelopeIO/leap/pull/558 introduced a new RPC endpoint `send_read_only_transaction` for pushing read-only transactions. Those read-only transactions were executed on the main thread. To take advantage of multi-core architecture, offloading them from the main thread and executing them multi-threaded will make more time for the main thread and improve throughput of read-only transactions.

The design is documented in https://github.com/eosnetworkfoundation/product/blob/main/transactions/read-only/parallel.md.

The main idea is as follows. The node toggles between `write` and `read` windows. In `write` window, the node operates normally, except read-only transactions are queued for later parallel execution in `read` window. In `read` window, the queued read-only transactions are executed in a dedicated thread pool, while operations safe to read-only transaction execution are executed on the main thread.

New configuration options are
```
  --read-only-threads arg (=0)          Number of worker threads in read-only
                                        transaction execution thread pool
  --read-only-write-window-time-us arg (=200000)
                                        time in microseconds the write window
                                        lasts
  --read-only-read-window-time-us arg (=60000)
                                        time in microseconds the read window
                                        lasts
```
If `--read-only-threads` is not configured or configured to 0, this feature is not enabled; read-only transactions are executed on the main thread.

`--read-only-threads` is not allowed  on producer node.

`--read-only-read-window-time-us` must be greater than `--max-transaction-time`.

New producer_plugin logging categories are added for transient transactions (dry-run (by `compute_transaction`) and read-only transactions). They are `transient_trx_success_tracing` and `transient_trx_failure_tracing` . Existing `transaction_success_tracing` and `transaction_failure_tracing` do not include the logging for transient transactions.

`max_action_return_value_size` is not enforced for read-only transactions.

PR #806, opened 2023-03-11
Execute read only safe tasks on read-only thread pool

heifner:

Builds on #776.


libraries/custom_appbase/tests/custom_appbase_tests.cpp
```diff
@@ -89,7 +89,7 @@ BOOST_AUTO_TEST_CASE( execute_from_read_queue ) {
    // stop application. Use lowest at the end to make sure this executes the last
    app->executor().post( priority::lowest, exec_queue::read_only_trx_safe, [&]() {
       // read_queue should have current function and write_queue should have all its functions
```
linh2931, line 91: This comment `read_queue should have current function` and the same ones below should be updated or removed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -421,6 +422,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       fc::microseconds                _ro_read_window_time_us{ 60000 };
       static constexpr fc::microseconds _ro_read_window_minimum_time_us{ 10000 };
       fc::microseconds                _ro_read_window_effective_time_us{ 0 }; // calculated during option initialization
+      fc::time_point                  _read_window_deadline; // only modified on app thread
```
linh2931, line 425: Can this be named as` _ro_read_window_deadline` to be consistent with other names?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -433,11 +435,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       void switch_to_read_window();
```
linh2931, line 435: Should `_ro_trx_queue` be renamed to something like `_ro_exhausted_read_only_trx_queue` as now it does not store regular read-only transactions?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -611,8 +611,13 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             // Parallel read-only trx execution enabled.
             // Store the transaction in read-only-trx queue so that it is
```
linh2931, line 612: This comment needs to be updated.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -611,8 +611,13 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
             // executed in read window
-            std::lock_guard<std::mutex> g( _ro_trx_queue.mtx );
-            _ro_trx_queue.queue.push_back({trx, std::move(next)});
+            app().executor().post(priority::low, exec_queue::read_only_trx_safe, [this, trx=trx, next{std::move(next)}]() mutable {
+               auto retry = process_read_only_transaction( trx, next );
```
linh2931, line 615: Now a read-only transaction can be executed in either read window or write window. If executed in write window, in `push_read_only_transaction` called by `process_read_only_transaction`, `_read_window_deadline` is unlimited; if executed in read window, `_read_window_deadline`  is `fc::time_point::now() + _ro_read_window_effective_time_us`. Is this what we want?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2813,28 +2820,29 @@ void producer_plugin_impl::switch_to_read_window() {
...
+         lck.unlock();
+
+         auto retry = process_read_only_transaction( trx.trx, trx.next );
+         if( retry ) {
+            lck.lock();
+            _ro_trx_queue.queue.push_front( trx );
+            // Do not schedule new execution
+            break;
+         }
+      } else {
```
linh2931, line 2842: `lck.unlock();` should be called.

plugins/producer_plugin/producer_plugin.cpp
heifner: Updated

plugins/producer_plugin/producer_plugin.cpp
heifner: The read-only-trx will be limited by `max-transaction-time` then. Since this is not allowed on a block producer there is no block deadline to enforce. However, I do think it should be limited to `_ro_write_window_time_us` for the write window. I have made that change.

plugins/producer_plugin/producer_plugin.cpp
heifner: Thanks! Fixed.

plugins/producer_plugin/producer_plugin.cpp
heifner: Renamed

plugins/producer_plugin/producer_plugin.cpp
heifner: Renamed

libraries/custom_appbase/tests/custom_appbase_tests.cpp
heifner: Updated

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -611,8 +611,13 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             // Parallel read-only trx execution enabled.
```
heifner: For consistency, I think we could post read-only-trx to the read_only_trx_safe queue always. Instead of sometimes being executed via `push_transaction`.

heifner: This is turning out to be a bit tricker than I anticipated. 
It requires the `process_read_only_transactions()` in start block. I still think it is cleaner than not piping read-only trxs through `push_transaction`, but I could see a desire to leave it where they flow through `push_transaction` if read-only thread pool is not enabled. Thoughts? @linh2931

linh2931: Oh no, I just realized we can not simply combine the two cases. For read-only trx with no threads, when you call `process_read_only_transactions`, you need to do set_db_read_only_mode and unset_db_read_only_mode, because those are done in window switching.

heifner: Thanks. Fixed.

linh2931: I think the change should work. I'm thinking into it further and will let you if anything comes up.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2912,6 +2915,16 @@ bool producer_plugin_impl::push_read_only_transaction(const transaction_metadata
          return true;
       }
 
+      // when executing on the main thread, need to switch db mode if not already in read only mode
```
linh2931, line 2922: To be precise, it is when executing on the main thread and in the write window

linh2931 (approved):

plugins/producer_plugin/producer_plugin.cpp
heifner: Updated

heifner:
#901 targets 4.0. If that is accepted then this PR will be closed.

PR #811, opened 2023-03-13
Pinned Build Script Changes

kj4ezj:
From [issue 735](https://github.com/AntelopeIO/leap/issues/735), this pull request fixes the pinned build by returning the `libcurl4-openssl-dev` dependency.

It also contains changes to the script output with the intention of making it easier to debug and easier to understand where in the process the script is. Mostly this means adding `echo` statements but, notably, I also switched from `wget` to `curl` because `wget` was throwing up in my terminal and `curl` gives us more control anyways. It is possible `wget` can be removed from the list of dependencies, but I did not remove it because I was not certain it was unused (for example, in the CMake).

## See Also
1. [Pull Request 833](https://github.com/AntelopeIO/leap/pull/833) - Fix Pinned Build on Ubuntu 22.04
    - Reverts [pull request 799](https://github.com/AntelopeIO/leap/pull/799)
1. [Pull Request 812](https://github.com/AntelopeIO/leap/pull/812) - Pinned Build Script Linter Changes
1. [Pull Request 835](https://github.com/AntelopeIO/leap/pull/835) - Change ASCII Art Banner in Build Script
1. ~~[Pull Request 832](https://github.com/AntelopeIO/leap/pull/832) - Roll `install_deps.sh` into `pinned_build.sh`~~
1. [Pull Request 847](https://github.com/AntelopeIO/leap/pull/847) - Restore `libcurl4-openssl-dev`
1. [Pull Request 849](https://github.com/AntelopeIO/leap/pull/849) - Fix `README.md` Formatting
1. [Pull Request 870](https://github.com/AntelopeIO/leap/pull/870) - Pinned Build Script Backport Omnibus
1. [Pull Request 857](https://github.com/AntelopeIO/leap/pull/857) - Simplify Pinned Build Scripts
1. [Pull Request 811](https://github.com/AntelopeIO/leap/pull/811) - Pinned Build Script Changes

Issue #735, opened 2023-02-22
Pinned Builds in CI


kj4ezj:
Take the existing pinned build scripts and run them in GitHub Actions. These builds will eventually be used for release assets, but attaching release assets should be a separate ticket. I have been asked not to run this build on every push, but the desired trigger conditions are not well-defined. Some suggestions include on pushes/merges to release branches, pushes/merges to `main`, or perhaps event to leave this as a manual run for now. Any of these conditions are acceptable to complete this ticket. The acceptance criteria are pinned build(s) which result in binaries for pinned builds, using the existing code where possible.

kj4ezj:
This issue was brought to me by @wanderingbort.

wanderingbort:
> using the existing code where possible

If this task cannot be completed **_without_** material changes to the pinned build scripts themselves then we should re-evaluate the scope. If this statement was more in the plumbing surrounding those scripts then :+1:

wanderingbort:
I would like to add a requirement that this uses our existing fleet of runners if possible.

bhazzard:
Zach is engaged on other priorities, and won't be available anytime soon. This is urgent enough that it is worth doing a quick solution. Make the script runnable in CI as an immediate fix. We'll reassign.

arhag:
Whoever takes this issue should talk to @kj4ezj to make sure hand-off occurs properly. Zach may have already done some work on this issue that shouldn't be duplicated.

bhazzard:
Blocked pending completion of https://github.com/AntelopeIO/leap/issues/1277

PR #812, opened 2023-03-13
Pinned Build Script Linter Changes

kj4ezj:
From [issue 735](https://github.com/AntelopeIO/leap/issues/735), I was having trouble running `scripts/pinned_build.sh` in a clean Ubuntu 22.04 container on the `HEAD` of `main`, which is expected to work, so I ran the [bashate](https://github.com/openstack/bashate) and [shellcheck](https://www.shellcheck.net) BASH linters on the script to look for errors.
```bash
bashate -i E006
shellcheck -x -f gcc
```
This pull request is **_only_** linting changes, I have introduced them in a pull request separate from my work to try to make the peer review less painful.

This will probably be easiest to review by ignoring whitespace...

GitHub peer review ignore whitespace

...and/or reviewing the commits individually, instead of as a group. The commits each have one linter rule in them they intend to address, which you can find online [here](https://github.com/openstack/bashate#currently-supported-checks) for `bashate` (the errors beginning with 'E'), or [here](https://www.shellcheck.net/wiki/Home) for ShellCheck.

## See Also

### DIFF
--- a/scripts/pinned_build.sh
+++ b/scripts/pinned_build.sh
@@ -57,7 +57,7 @@ popdir() {
 }
 
 try(){
-    eval "$@"
+    "$@"
     res=$?
     if [[ ${res} -ne 0 ]]; then
         exit 255
--- a/scripts/pinned_build.sh
+++ b/scripts/pinned_build.sh
@@ -7,7 +7,7 @@ if [[ "$(uname)" == "Linux" ]]; then
     if [[ -e /etc/os-release ]]; then
         # obtain NAME and other information
         . /etc/os-release
-        if [[ ${NAME} != "Ubuntu" ]]; then
+        if [[ "${NAME}" != "Ubuntu" ]]; then
             echo "Currently only supporting Ubuntu based builds. Proceed at your own risk."
         fi
     else
@@ -26,9 +26,9 @@ fi
 
 export CORE_SYM='EOS'
 # CMAKE_C_COMPILER requires absolute path
-DEP_DIR=`realpath $1`
-LEAP_DIR=$2
-JOBS=$3
+DEP_DIR=`realpath "$1"`
+LEAP_DIR="$2"
+JOBS="$3"
 CLANG_VER=11.0.1
 BOOST_VER=1.70.0
 LLVM_VER=7.1.0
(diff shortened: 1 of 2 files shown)


scripts/pinned_build.sh
```diff
@@ -1,140 +1,139 @@
-#!/usr/bin/env bash
+#!/bin/bash
+set -eo pipefail
```
spoonincode, line 2: I'm not sure I correctly understand the purpose of this change, at least from the description in the commit:
> Looking for BASH in the environment instead of invoking it directly enables attacks where malware could escalate privileges by pointing BASH to itself and tricking the user into entering a superuser password when a user invokes our script, a familiar request the user may be expecting.

This seems to be saying we want to guard our script from being compromised by a malicious application that is in the user's `PATH` that matches the name as an application the script uses. If that's really the case, then _all_ executables need to have an absolute path. `apt-get`, `cmake`, `gcc`, `clang`, `tar`, etc etc because any of those should be something malicious.

The practicality of trying to enforce this for the `pinned_build.sh` script seems tough. _Maybe_ it's more doable in `install_deps.sh`, and _maybe_ it's more pertinent in `install_deps.sh` since that one needs to be run as root and/or might be more expected to prompt a password.

kj4ezj: I agree with everything you wrote and I am certainly not advocating that we call other binaries in the script (such as aptitude) directly instead of relying on the system `PATH`. This shebang convention is something I researched many years ago and stuck to, so my explanation in the commit message as to why probably leaves a lot to be desired.

I looked it up again and found that, while the `PATH` exploit is a potential concern, the `env` binary itself is a bigger concern and it has been the subject of high-profile exploits in the past.
> Yes, there have been examples of malware that have exploited the use of the `env` command to execute malicious code.
>
> One such example is the "Shellshock" vulnerability, which was a series of security flaws discovered in the BASH shell in 2014. One of the exploits of the Shellshock vulnerability involved setting the `BASH_ENV` environment variable to a malicious script that was executed when a new BASH shell was launched. Attackers were able to use this exploit to execute arbitrary code on vulnerable systems.
>
> Another example is the "Dvmap" malware, which was discovered in 2017 and targeted Android devices. The malware exploited a vulnerability in the env command to install itself as a root application, giving attackers full control over the compromised device.
>
> In general, any command or utility that is used to execute code on a system can be vulnerable to exploits, including the `env` command. It's important to ensure that your system is up to date with the latest security patches and to follow best practices for securing your environment to minimize the risk of exploitation.

Another [disadvantage](https://unix.stackexchange.com/a/206366) to using `env` is that you cannot pass more than one argument to the binary you are looking for on some systems. That does not apply here.

People also [argue for its use](https://unix.stackexchange.com/a/206366).

I do not feel strongly enough to spend a ton of time on this, I will revert the change if you want. But, in general because of the security considerations and the fact that it has been exploited in the real world, I personally prefer to avoid using `env` when possible.

ScottBailey (approved):
Whitespace changes, adding quotes, and converting command substitution to the `$()` format.

If this works for you, then LGTM.

scripts/pinned_build.sh
spoonincode: imo it doesn't really do much for security. The rest of the applications the script runs (directly or indirectly) will suffer from the same problem. Plus, if you're to the point where you're worried about `PATH` being compromised what about `LD_PRELOAD` or others?

The script becomes less portable (bash on freebsd is in /usr/local/bin, for example) while not really gaining anything. But, portability is somewhat academic for this only-supported-on-ubuntu script anyways.

Issue #707, opened 2023-02-09
Add TestHarness python integration scripts to dev package


heifner:
It would be useful to have the python integration scripts `TestHarness/*` scripts in an installable package so integration tests could be created/run without having to build a complete Leap checkout.

Either include in the existing dev package or into a new `python3-leap-testharness` package.

stephenpdeos:
Agreed to keep in existing dev package.

heifner:
- Need to have the dev install use the dev installed: nodeos, cleos, eosio-launcher
- core_symbol.py - The def install will need a way to specify the CORE_SYMBOL instead of the default of SYS (the default leap is built with)
-

spoonincode:
When we discussed this at one point a week or two ago, one aspect that came up was what to do with the system contracts that the python stuff requires.

For libtester, the system contracts are embedded in to the library. The reason I did it this way was because it makes it easy to ensure the contracts are always available and not fuss around with what path to use (which may be different depending on whether the library has been installed or packaged or such). I wonder for Python what the best strategy would be?

PR #818, opened 2023-03-14
Add TestHarness python integration scripts to dev package

oschwaldp-oci:
### Provide the utilities of the `TestHarness` framework in the `dev-install` package of `leap`.

Now, when building `leap`, as part of the `make dev-install` installation, the `TestHarness`, as well as the contracts used in testing in `libraries/testing/contracts` and `unittests/contracts` will be made available for use outside of a `leap` source build tree.

### `TestHarness` Updates:

To make this possible, changes were needed in `TestHarness` to allow similar directory structures in both source build and install directory structures:

- `core_symbol.py` was moved into the `TestHarness` module and can now be altered by setting the `CORE_SYMBOL_NAME` environment variable. Otherwise it defaults to `SYS`.
- `logging.json` was left in place in the source tree for useful reference, however, in the build tree it has been placed in the `TestHarness` directory and renamed to `logging-template.json` to better express its use in the framework.
- `Cluster.py` now makes use of relative path references to find the `launcher.py` and contracts used for setting up test infrastructure.
- `testUtils.py` now makes use of the programs from the `bin/` directory instead of `programs/*` to make referencing in the install directory structure possible.  The `TestLogs` and `/etc/eosio/` logging directories are both now shifted to reference the current working directory for when making use of the utility outside of the build tree.

### Example Installing Use Case

To build, make sure you are in the root of the `leap` repo, then run the following commands:
`mkdir -p build`
`cd build`

#### Using `make dev-install`
Select where to install the `leap` `dev` package using `CMAKE_INSTALL_PREFIX`:
`cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=~/leap_install_dir ..`

Build and install `leap` `dev`:
`make dev-install`

#### Using `make package`
Select where to install the `leap` `dev` package using `CMAKE_INSTALL_PREFIX`:
`cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_INSTALL_PREFIX=~/leap_install_dir ..`

Build `leap` `dev` package:
`make package`

*Note* When doing the make package step, it is not a true error if you see:
`Error creating directory "/<leap_install_dir /lib/python3/dist-packages".`
`CMake Error: failed to create symbolic link '/<leap_install_dir >/lib/python3/dist-packages/TestHarness': no such file or directory`
`CMake Error: failed to create symbolic link '/<leap_install_dir >/share/leap_testing/bin': no such file or directory`

It is caused by a necessary step for `make install` to work on cmake versions < 3.21 but can flag as an error in the package step since those directories and symbolic links are created in the `postinit` script during the package install instead.  Couldn't find a way to differentiate between `make package` or `make install` in the cmake/makefile.

Then install the `dev` and `base` packages:

For testing, this was done in a container, but the steps are:

`apt-get update`
`apt-get install -y ./leap-dev_4.0.0-dev-ubuntu20.04_amd64.deb`
`apt-get install -y ./leap_4.0.0-dev-ubuntu20.04_amd64.deb`

### Validating install

This will install `leap` to the `~/leap_install_dir`.  Of note within the chosen install location for the `dev` package:

-  `~/leap_install_dir/share/leap_testing/` -> where the dev testing components are installed
  - `/leap_testing/libraries/testing/contracts/` -> contracts used during testing
  - `/leap_testing/unittests/contracts/` -> contracts used during testing
  - `/leap_testing/tests/TestHarness/` -> testing framework and utilities
  - `/leap_testing/bin` -> symbolic link pointing to `~/leap_install_dir/bin` to allow relative path referencing of programs by the  `TestHarness`
- `~/leap_install_dir/bin` -> installed versions of  `cleos`, `eosio-blocklog`, `keosd`, `leap-util`, `nodeos`, `trace_api_util`
- `~/leap_install_dir/lib/python3/dist-packages/TestHarness` -> symbolic link pointing to installed `TestHarness` used for python path for installed module discovery.

### Using Installed `dev` package

- Update the `PYTHONPATH` environment variable to allow python to find the installed `TestHarness` module. eg.:
  - `export PYTHONPATH=~/leap_install_dir/lib/python3/dist-packages`
- Optional: Update `CORE_SYMBOL_NAME`:
  - `export CORE_SYMBOL_NAME=EOS`
- Import necessary components from `TestHarness` into python script:
  - `from TestHarness import Cluster, TestHelper, Utils, WalletMgr, CORE_SYMBOL`
  - `from TestHarness.TestHelper import AppArgs`

Resolves: https://github.com/AntelopeIO/leap/issues/707


CMakeLists.txt
```diff
@@ -256,6 +256,26 @@ configure_file(programs/cleos/LICENSE.CLI11                                licen
 
 install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/licenses/leap" DESTINATION "${CMAKE_INSTALL_FULL_DATAROOTDIR}/licenses/" COMPONENT base)
 
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/libraries/testing/contracts DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/libraries/testing COMPONENT base
+        PATTERN "CMakeFiles" EXCLUDE
+        PATTERN "*.cmake" EXCLUDE
+        PATTERN "Makefile" EXCLUDE)
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/unittests/contracts DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/unittests COMPONENT base
```
spoonincode, line 263: should this have been `COMPONENT dev` here?

oschwaldp-oci: Yes.  Good catch. Thanks.

Addressed: [6f16347](https://github.com/AntelopeIO/leap/pull/818/commits/6f16347f6e71cb8a4fd924dcf35ed9f1eaf59a0f)

CMakeLists.txt
```diff
@@ -256,6 +256,26 @@ configure_file(programs/cleos/LICENSE.CLI11                                licen
...
+        FILES_MATCHING
+        PATTERN "*.py"
+        PATTERN "*.json"
+        PATTERN "__pycache__" EXCLUDE
+        PATTERN "CMakeFiles" EXCLUDE)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/tests/launcher.py DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/tests COMPONENT dev)
+
+install(CODE "execute_process( COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages)" COMPONENT dev)
+install(CODE "execute_process( COMMAND ${CMAKE_COMMAND} -E create_symlink ../../../share/leap_testing/tests/TestHarness ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages/TestHarness)" COMPONENT dev)
+install(CODE "execute_process( COMMAND ${CMAKE_COMMAND} -E create_symlink ../../bin ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/bin)" COMPONENT dev)
```
spoonincode, line 277: How do these work for the `leap-dev.deb` package? I downloaded the leap-dev.deb from the CI run, and when I
```
ar x leap-dev_4.0.0-dev-ubuntu20.04_amd64.deb
tar tlf data.tar.gz  | less
```
I don't see these symlinks anywhere. And I don't see any thing else magical from cmake in the package either. Admittedly I didn't try actually installing...

oschwaldp-oci: The symbolic links are now created in the install step making use of the `postinit` script that has been added.  Had to jump through some hoops since there are issues with `create_symlink` in cmake versions < 3.22, so there are 2 paths now based on cmake versions.  Added some notes in the PR description.

ScottBailey: Try it with the verbose flag, like this: 
```
ar x leap-dev_4.0.0-dev-ubuntu20.04_amd64.deb data.tar.gz
tar tlvf data.tar.gz | rg "\->"
```

If you haven't installed ripgrep yet, what are you waiting for? Its half the typing of grep! that alone is a win!

tests/TestHarness/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
 configure_file(__init__.py . COPYONLY)
+configure_file(core_symbol.py.in core_symbol.py)
```
ScottBailey, line 2: No need to duplicate the name. Also, in a sea of COPYONLY commands, it might be nice if we indicated this was not accidentally left out. Maybe there's a better comment than I've added?
```suggestion
configure_file(core_symbol.py.in .)  # Updates CORE_SYMBOL_NAME
```

CMakeLists.txt
```diff
@@ -257,6 +257,40 @@ configure_file(programs/cleos/LICENSE.CLI11                                licen
...
+        PATTERN "Makefile" EXCLUDE)
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/tests/TestHarness DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/tests COMPONENT dev
+        FILES_MATCHING
+        PATTERN "*.py"
+        PATTERN "*.json"
+        PATTERN "__pycache__" EXCLUDE
+        PATTERN "CMakeFiles" EXCLUDE)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/tests/launcher.py DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/tests COMPONENT dev)
+
+if(${CMAKE_VERSION} GREATER 3.22)
```
spoonincode, line 276: Probably should use `VERSION_GREATER` here (or maybe even `VERSION_GREATER_EQUAL` to make it a little clearer); also no need to expand the variable.

Can you explain a little more what the problem is for pre vs post 3.22 (3.23?)? For example, you only set the `CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA` in the pre-3.22/3 path; is it really unneeded in the post-3.22/3 path? Is it really not possible to just use the pre-3.22/3 path on all versions?

CMakeLists.txt
```diff
@@ -257,6 +257,40 @@ configure_file(programs/cleos/LICENSE.CLI11                                licen
...
+      COMMAND ${CMAKE_COMMAND} -E create_symlink ../../bin share/leap_testing/bin)
+
+   install(FILES ${CMAKE_CURRENT_BINARY_DIR}/lib/python3/dist-packages/TestHarness DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages COMPONENT dev)
+   install(FILES ${CMAKE_CURRENT_BINARY_DIR}/share/leap_testing/bin DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing COMPONENT dev)
+else()
+   install(CODE "execute_process( COMMAND ${CMAKE_COMMAND} -E make_directory ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages)" COMPONENT dev)
+   install(CODE "execute_process( COMMAND ${CMAKE_COMMAND} -E create_symlink ../../../share/leap_testing/tests/TestHarness ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages/TestHarness)" COMPONENT dev)
+   install(CODE "execute_process( COMMAND ${CMAKE_COMMAND} -E create_symlink ../../bin ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/bin)" COMPONENT dev)
+
+   set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA ${CMAKE_BINARY_DIR}/scripts/postinst COMPONENT dev)
```
spoonincode, line 290: I'm not sure this `set()` should have a `COMPONENT dev`.. that doesn't seem right. It seems like instead you'd want `set(CPACK_DEBIAN_DEV_PACKAGE_CONTROL_EXTRA ${CMAKE_BINARY_DIR}/scripts/postinst)`. I also would suggest considering placing this inside of `package.cmake` along with all the other package related stuff to keep all package related stuff together.

scripts/postinst
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+if [ ! -L ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages/TestHarness ]; then
+    mkdir -p ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages
+    ln -s ../../../share/leap_testing/tests/TestHarness ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages/TestHarness
```
spoonincode, line 5: Instead of `${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages`, would `${Python_SITELIB}` be more appropriate since the system python installation directory doesn't depend on the installation path of leap?

scripts/postinst
```diff
@@ -0,0 +1,11 @@
...
+fi
+
+if [ ! -L ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/bin ]; then
+    mkdir -p ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing
+    ln -s ../../bin ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/bin
```
spoonincode, line 10: I think we ought to consider a `postrm` (or whatever the proper script would be; not super familiar with debian packaging) that cleans up the symlinks we created so they aren't left dangling after uninstalling.

CMakeLists.txt
oschwaldp-oci: Just double checking, you're suggesting:

`if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.22)`

Yeah I can get on board with that, it is clearer.

oschwaldp-oci: > Can you explain a little more what the problem is for pre vs post 3.22 (3.23?)? For example, you only set the `CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA` in the pre-3.22/3 path; is it really unneeded in the post-3.22/3 path? Is it really not possible to just use the pre-3.22/3 path on all versions?

Because of the use of `install(CODE` and the false error reporting that comes out of that piece during a `make package`, was hoping to leave the >3.22 path in place with the hopes of remove the <3.22 path when we get to a point where we aren't supporting older versions of cmake where the `create_symlink` function has yet to be fixed.  And also alleviate that reporting for anyone building/installing on newer cmake toolchains.

CMakeLists.txt
```diff
@@ -257,6 +257,40 @@ configure_file(programs/cleos/LICENSE.CLI11                                licen
 
 install(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/licenses/leap" DESTINATION "${CMAKE_INSTALL_FULL_DATAROOTDIR}/licenses/" COMPONENT base)
 
+install(DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}/libraries/testing/contracts DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/libraries/testing COMPONENT dev
```
spoonincode, line 260: Unfortunately, for any place you have a `COMPONENT dev` it should also be `EXCLUDE_FROM_ALL`.

The problem is, we want `make install` to only install the files from the base component (or put another way, we want `make install` to ignore anything we classify as a dev install), but there is no way afaik to control the behavior of `make install` since it's a built in target that can't be overridden. So we have to resort to peeling away all these from the "install all" target.

The reason we go through this trouble is that historically, before the `COMPONENT dev EXCLUDE_FROM_ALL` behavior, a `make install` could install header files that were in a path with higher precedence than our own source tree. As you can imagine this led to users (and devs!) getting in to really wonky problems when the, for example, fc headers in /usr/local/include would be used instead of the fc headers in the current source tree. There is a little more about discussion on this in eosnetworkfoundation/mandel#277

CMakeLists.txt
spoonincode: Yes but `if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.22)` may be different then `if(${CMAKE_VERSION} GREATER 3.22)` because the original might not have been true for 3.22.0 since 3.22.0 might not have been greater than 3.22. So double check which version you really want here maybe it's 3.22 or 3.23

spoonincode:
We should probably add `python3-numpy` to the dev package DEPENDS list
https://github.com/AntelopeIO/leap/blob/d4ffa14da8e794219b387bdfa08a799c3305297d/package.cmake#L64

Is there any other external applications the python scripts depend on? I recall we got rid of curl in #442 that's the only one that comes to mind for me.

CMakeLists.txt
oschwaldp-oci: So it is actually due to cmake (fixed in 3.21) not allowing symbolic links to a directory to be installed via `install(FILES`

Cmake 3.21 fixed this (https://gitlab.kitware.com/cmake/cmake/-/commit/d71a7cc19d6e03f89581efdaa8d63db216184d40)
`Allow symbolic links to a directory to be installed via install(FILES ...)`

scripts/CMakeLists.txt
```diff
@@ -4,3 +4,4 @@ configure_file(eosio-tn_down.sh eosio-tn_down.sh COPYONLY)
 configure_file(eosio-tn_roll.sh eosio-tn_roll.sh COPYONLY)
 configure_file(eosio-tn_up.sh eosio-tn_up.sh COPYONLY)
 configure_file(abi_is_json.py abi_is_json.py COPYONLY)
+configure_file(postinst postinst)
```
ScottBailey, line 7: I wouldn't hate if `<output>` in [`configure_file()`](https://cmake.org/cmake/help/latest/command/configure_file.html) was `.` instead of a copy paste of `<input>`. But this is kind of an all or nothing thing for this file. No big deal if we don't do it.
```suggestion
configure_file(postinst .)
```

ScottBailey (changes requested):
Please address Matt's issues including adding the debian control script `prerm`.

scripts/postinst
spoonincode: (again, I'm not super familiar with debian packaging!), I found this kind of interesting:

https://www.debian.org/doc/manuals/maint-guide/dother.en.html#links
https://manpages.debian.org/testing/debhelper/dh_link.1.en.html#FILES

is the canonical way to create these sort of links through the `package.links` file, instead of a `postinst` script?

oschwaldp-oci: Not being super familiar either, but reading up on the Debian maintainer script flows, I think `scripts/prerm` might be the way to go here as well.  I'll work something up.

scripts/postinst
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
ScottBailey, line 1:
```suggestion
#!/bin/sh

set -e
```

See [Debian Policy Manual 6.1](https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#introduction-to-package-maintainer-scripts) regarding `set -e`.

I cannot find an entry in the manual regarding selection of shells for control scripts; however, the packages I surveyed on my machine only calls `sh`. I believe there is a guarantee that sh in installed and behaves correctly. It seems likely bash will always exist, but I don't think it's guaranteed.

scripts/postinst
ScottBailey: > is the canonical way to create these sort of links through the `package.links` file, instead of a `postinst` script?

Yes, but ONLY if we were using debhelper to package. So, in our case, using CPack, no.

scripts/CMakeLists.txt
oschwaldp-oci: Addressed: [be154da](https://github.com/AntelopeIO/leap/pull/818/commits/be154da6036dbe810631b53b32afee58491767fc)

oschwaldp-oci:
> We should probably add `python3-numpy` to the dev package DEPENDS list
> 
> https://github.com/AntelopeIO/leap/blob/d4ffa14da8e794219b387bdfa08a799c3305297d/package.cmake#L64
> 
> Is there any other external applications the python scripts depend on? I recall we got rid of curl in #442 that's the only one that comes to mind for me.

Did a quick scan of the dependencies and didn't find any outside `python3-numpy` like you already mentioned. 
Addressed: [be154da](https://github.com/AntelopeIO/leap/pull/818/commits/be154da6036dbe810631b53b32afee58491767fc)

CMakeLists.txt
oschwaldp-oci: Addressed: [be154da](https://github.com/AntelopeIO/leap/pull/818/commits/be154da6036dbe810631b53b32afee58491767fc)

CMakeLists.txt
oschwaldp-oci: Addressed: [be154da](https://github.com/AntelopeIO/leap/pull/818/commits/be154da6036dbe810631b53b32afee58491767fc)

scripts/postinst
oschwaldp-oci: Addressed: [be154da](https://github.com/AntelopeIO/leap/pull/818/commits/be154da6036dbe810631b53b32afee58491767fc)

CMakeLists.txt
oschwaldp-oci: Addressed: [be154da](https://github.com/AntelopeIO/leap/pull/818/commits/be154da6036dbe810631b53b32afee58491767fc)

oschwaldp-oci:
> Please address Matt's issues including adding the debian control script `prerm`.

Mostly addressed here: [be154da](https://github.com/AntelopeIO/leap/pull/818/commits/be154da6036dbe810631b53b32afee58491767fc)

CMakeLists.txt
```diff
@@ -257,6 +257,53 @@ configure_file(programs/cleos/LICENSE.CLI11                                licen
...
+        PATTERN "*.json"
+        PATTERN "__pycache__" EXCLUDE
+        PATTERN "CMakeFiles" EXCLUDE)
+install(FILES ${CMAKE_CURRENT_BINARY_DIR}/tests/launcher.py DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/tests COMPONENT dev EXCLUDE_FROM_ALL)
+
+if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.21)
+   # Cmake versions < 3.21 did not support installing symbolic links to a directory via install(FILES ...)
+   # Cmake 3.21 fixed this (https://gitlab.kitware.com/cmake/cmake/-/commit/d71a7cc19d6e03f89581efdaa8d63db216184d40)
+   # If/when the lowest cmake version supported becomes >= 3.21, the else block as well as the postinit script
+   # would be a good option to remove in favor of using the facilities in cmake directly.
```
ScottBailey, line 280: Minor issue, let's add prerm to the comment.
```suggestion
   # If/when the lowest cmake version supported becomes >= 3.21, the else block as well as the postinit and
   # prerm scripts would be a good option to remove in favor of using the facilities in cmake / cpack directly.
```

scripts/CMakeLists.txt
ScottBailey: Thanks!

scripts/postinst
oschwaldp-oci: Addressed: b6007e022d and [61b0efe](https://github.com/AntelopeIO/leap/pull/818/commits/61b0efe536f629077bcde91caf93b062c2749eb7)

scripts/prerm
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+if [ -L ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages/TestHarness ]; then
+    rmdir -p ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages/TestHarness
+fi
+
+if [ -L ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/bin ]; then
+    rmdir -p ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/bin
+fi
```
ScottBailey, line 9: I don't think `rmdir -p` is the correct command to remove a symbolic link. I think we can just `rm -f` these symlinks, existing or not. Plus, simpler is better, righ?

```suggestion
rm -f ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages/TestHarness
rm -f ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/bin
```

oschwaldp-oci: Addressed: [08b1eb5](https://github.com/AntelopeIO/leap/pull/818/commits/08b1eb525242be8746f68888b66cca9aae84be51)

ScottBailey (approved):
lgtm

package.cmake
```diff
@@ -70,6 +70,8 @@ if(DPKG_QUERY AND OS_RELEASE MATCHES "\n?ID=\"?ubuntu" AND LLVM_CMAKE_DIR)
       string(APPEND CPACK_DEBIAN_DEV_PACKAGE_DEPENDS ", ${LLVM_PKG_FIND_OUTPUT}")
    endif()
 endif()
+set(CPACK_DEBIAN_DEV_PACKAGE_CONTROL_EXTRA ${CMAKE_BINARY_DIR}/scripts/postinst)
+set(CPACK_DEBIAN_DEV_PACKAGE_CONTROL_EXTRA ${CMAKE_BINARY_DIR}/scripts/prerm)
```
spoonincode, line 74: This causes only `prerm` to be bundled. Probably should be a cmake list I guess

scripts/postinst
ScottBailey: I concur, finding `Python_SITELIB` on the host is probably not optimal.

But we could potentially use `/usr/lib/python3/dist-packages`. It would work for Debian Bookworm and Ubuntu 18 through 23.

package.cmake
oschwaldp-oci: Trying:
`set(CPACK_DEBIAN_DEV_PACKAGE_CONTROL_EXTRA ${CMAKE_BINARY_DIR}/scripts/postinst;${CMAKE_BINARY_DIR}/scripts/prerm)`

oschwaldp-oci: Tested and that worked:
Addressed: [8c8ad50](https://github.com/AntelopeIO/leap/pull/818/commits/8c8ad50862c76186a9a0a9428b4b20a9ae099963)

spoonincode (approved):
looks good, thanks

package.cmake
```diff
@@ -70,6 +70,7 @@ if(DPKG_QUERY AND OS_RELEASE MATCHES "\n?ID=\"?ubuntu" AND LLVM_CMAKE_DIR)
       string(APPEND CPACK_DEBIAN_DEV_PACKAGE_DEPENDS ", ${LLVM_PKG_FIND_OUTPUT}")
    endif()
 endif()
+set(CPACK_DEBIAN_DEV_PACKAGE_CONTROL_EXTRA ${CMAKE_BINARY_DIR}/scripts/postinst;${CMAKE_BINARY_DIR}/scripts/prerm)
```
ScottBailey, line 73: Probably better to add the quotes than not.

```suggestion
set(CPACK_DEBIAN_DEV_PACKAGE_CONTROL_EXTRA "${CMAKE_BINARY_DIR}/scripts/postinst;${CMAKE_BINARY_DIR}/scripts/prerm")
```

scripts/prerm
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+
+rm -f ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages/TestHarness
+rm -f ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/bin
```
ScottBailey, line 4: Sorry, I should have caught this before.
```suggestion
rm -rf ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages
rm -rf ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing
```

ScottBailey (changes requested):
Almost there! A couple quick changes to add. :-)

scripts/prerm
oschwaldp-oci: I'm nervous about this change.  What if it is installed into a common location where `python3/dist-packages/` could have non-leap owner resources we don't want to just destroy.  I almost think I need to check at each level and only remove if the dir is empty.

ScottBailey: Good point about dist packages. You could do a combination  'rmdir -p' for the directors and rm -r for the symlinks, and that would protect us, I think.

spoonincode: if you want to make it as simple as possible, I'd suggest removing the `mkdir`s since afaict they're guaranteed to be there by the time the `postinst` script runs. Then you're only responsible for removing the symlinks in `prerm` (unchanged from how it is now).

That allows the `postinst` & `prerm` to be 100% the opposite of each other without worrying about removing directories and whatnot.

oschwaldp-oci: Here is the solution I implemented and tested: [c46400c](https://github.com/AntelopeIO/leap/pull/818/commits/c46400c527ea956c06217a8474f0541de6734447)

```
Best effort attempt to remove the directories we may have created during the install.

Can ignore failed directory deletion as they may be installing to location that has more than leap in it, for instance the ~/lib/python3/dist-packages, however if it is empty it is a safe assumption that it is in the location the install created it and can be removed as the inverse uninstall operation.
```

scripts/prerm
```diff
@@ -0,0 +1,7 @@
...
+rmdir --ignore-fail-on-non-empty ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages
+rmdir --ignore-fail-on-non-empty ${CMAKE_INSTALL_FULL_LIBDIR}/python3
+rmdir --ignore-fail-on-non-empty ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing
```
spoonincode, line 7: when/how could this be empty so that it would succeed? Seems impossible: `leap_testing` is part of the package. Feels like we're over complicating this

oschwaldp-oci: You have a point, for `leap_testing` at least since that is part of the package.  Can probably remove the mkdir from the `postinst` script for `leap_testing` and the `rmdir` for it in `prerm`.

However, I think we need to keep the `python3/dist-packages` as it may be created from scratch in the `postinst` and isn't in the package.  Thus we should also attempt to clean it up.

spoonincode: Since the package depends on `python3-numpy` it's impossible for `/usr/lib/python3/dist-packages/` to not exist at the time of `postinst`

Now, of course, we're not linking to `/usr/lib/python3/dist-packages/` but rather `${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages`; that's kind of the disappointment with not being able to use `${Python_SITELIB}` as it would eliminate that academic difference.

oschwaldp-oci: Simplified per discussion and tested, still works: [e7e66a5](https://github.com/AntelopeIO/leap/pull/818/commits/e7e66a5d1991f60c1f27d4a97e518bfd3a30d9b1)

oschwaldp-oci: There is no strict guarantee that someone is installing to `/usr/lib/python3/dist-packages` so this solution was formulated to take that into account and still allow it to work.  It also takes into account that with the requirement of running with cmake 3.8, as we discussed, the use of `${Python_SITELIB}` wouldn't currently work as we were intending.

ScottBailey (approved):
lgtm

scripts/prerm
spoonincode: There is an important distinction though: while someone may be installing many different versions to various `CMAKE_INSTALL_PREFIX`s via `make dev-install`, they can, by rule, only install a single `leap-dev.deb`. So it seems plausible to consider the installed package's `TestHarness` as _the_ "system provided" `TestHarness` no matter where it is installed.

That actually makes me wonder if instead the `postinst` should be something like
```
ln -s ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/tests/TestHarness ${Python_SITELIB}/TestHarness
```
as this will make the installed package's TestHarness the system provided TestHarness no matter what `CMAKE_INSTALL_PREFIX` was used at build time (which should be `/usr` per-spec, but a user might not have configured it that way)

Anyways, we can debate this further post 4.0, where we actually have `${Python_SITELIB}`

oschwaldp-oci: I was thinking about this a lot, even before you posted this comment.  I tend to agree with you here.  It would be good to treat it that way.  As you said, let's revisit this.  I'll create an issue to track this.

scripts/postinst
oschwaldp-oci: Issue created to investigate this in the future: https://github.com/AntelopeIO/leap/issues/839

scripts/prerm
oschwaldp-oci: @spoonincode - Issue created to investigate this in the future: https://github.com/AntelopeIO/leap/issues/839

PR #820, opened 2023-03-15
get_raw_block api endpoint

### LINKED ISSUE #642: Creation of v1/chain/get_raw_block chain API endpoint
This is a transfer of eosnetworkfoundation/mandel#811 originally entered by @aaroncox, it is being moved to `leap` for easier visibility.

This is a feature request which will eliminate a bottleneck that our history solution (Roborovski) starts running into under high load. We imagine it'll be useful for other advanced applications as well, ones specifically that access raw block data.

We propose the addition of a new API endpoint:

`v2/chain/get_block` or `v1/chain/get_raw_block` (or some named variant of get_block)

This API endpoint will operate nearly identical to the `/v1/chain/get_block` API endpoint, except it will:

- Not decode any of the transactions using their ABIs, and instead just return the raw hex data.
- Run in a non-blocking thread, or be threaded as much as possible. (Note: we assume here the blocking nature of the current `get_block` API call is due to the abi decoding). 
- Maintain a similar API response structure to the `get_block` call.

These changes should offer a much more lightweight solution to retrieving a block and its data vs the current get_block implementation.

---

Side note:

As a more radical approach to solving this problem, we could also suggest making these modifications to the get_block call and deprecating the current behavior of decoding transactions. This makes sense since any historical block retrieved is going to be decoded with the current ABI, and not the ABI that was set historically at that point in time. The current get_block call results can often be invalid because of this. It's best to leave decoding historical blocks to some form of history solution which can reference the appropriate ABI.

However, since this would be a breaking change for many applications, the creation of a new API endpoint as suggested above is likely the safest path. Perhaps in the future the v1/chain/get_block call can be deprecated in favor of this new endpoint.

huangminghuang:
This PR 
- adds the following API endpoints 
   * `/v1/chain/get_raw_block`: returns the raw block content without ABI decoding
   * `/v1/chain/get_block_header`: returns the block header with optional block extensions
      - Input/output: 
   ```
   struct get_raw_block_header_params {
      string block_num_or_id;
      bool include_extensions = false; // include block extensions (requires reading entire block off disk)
   }; 
  
   struct get_raw_block_header_results {
      chain::block_id_type  id;
      fc::variant           signed_block_header;
      std::optional<chain::extensions_type> block_extensions;
   };
   ```
- add `cleos get block` options for the added API endpoints above
  * `--raw`: corresponding to `get_raw_block`
  * `--header`: corresponding to `get_block_header` with `include_extensions` set to false
  * `--header_with_extensions`: corresponding to `get_block_header` with `include_extensions` set to true

Resolves #642

### DIFF
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -185,15 +185,15 @@ namespace eosio { namespace chain {
       }
 
       template <typename Stream>
-      block_id_type read_block_id(Stream&& ds, uint32_t expect_block_num) {
-         block_header bh;
+      signed_block_header read_block_header(Stream&& ds, uint32_t expect_block_num) {
+         signed_block_header bh;
          fc::raw::unpack(ds, bh);
 
          EOS_ASSERT(bh.block_num() == expect_block_num, block_log_exception,
                     "Wrong block header was read from block log.",
                     ("returned", bh.block_num())("expected", expect_block_num));
 
-         return bh.calculate_id();
+         return bh;
       }
 
       /// Provide the read only view of the blocks.log file
@@ -474,8 +474,8 @@ namespace eosio { namespace chain {
          virtual void     reset(const chain_id_type& chain_id, uint32_t first_block_num)      = 0;
          virtual void     flush()                                                             = 0;
 
-         virtual signed_block_ptr read_block_by_num(uint32_t block_num)    = 0;
-         virtual block_id_type    read_block_id_by_num(uint32_t block_num) = 0;
+         virtual signed_block_ptr            read_block_by_num(uint32_t block_num)        = 0;
+         virtual std::optional<signed_block_header> read_block_header_by_num(uint32_t block_num) = 0;
 
          virtual uint32_t version() const = 0;
 
@@ -512,7 +512,7 @@ namespace eosio { namespace chain {
          void flush() final {}
 
          signed_block_ptr read_block_by_num(uint32_t block_num) final { return {}; };
-         block_id_type    read_block_id_by_num(uint32_t block_num) final { return {}; };
+         std::optional<signed_block_header> read_block_header_by_num(uint32_t block_num) final { return {}; };
 
          uint32_t         version() const final { return 0; }
          signed_block_ptr read_head() final { return {}; };
@@ -556,7 +556,7 @@ namespace eosio { namespace chain {
          virtual uint32_t         working_block_file_first_block_num() { return preamble.first_block_num; }
          virtual void             post_append(uint64_t pos) {}
          virtual signed_block_ptr retry_read_block_by_num(uint32_t block_num) { return {}; }
-         virtual block_id_type    retry_read_block_id_by_num(uint32_t block_num) { return {}; }
+         virtual std::optional<signed_block_header> retry_read_block_header_by_num(uint32_t block_num) { return {}; }
 
          void append(const signed_block_ptr& b, const block_id_type& id,
                      const std::vector<char>& packed_block) override {
@@ -609,14 +609,14 @@ namespace eosio { namespace chain {
             FC_LOG_AND_RETHROW()
          }
 
-         block_id_type read_block_id_by_num(uint32_t block_num) final {
+         std::optional<signed_block_header> read_block_header_by_num(uint32_t block_num) final {
             try {
                uint64_t pos = get_block_pos(block_num);
                if (pos != block_log::npos) {
                   block_file.seek(pos);
-                  return read_block_id(block_file, block_num);
+                  return read_block_header(block_file, block_num);
                }
-               return retry_read_block_id_by_num(block_num);
+               return retry_read_block_header_by_num(block_num);
             }
             FC_LOG_AND_RETHROW()
          }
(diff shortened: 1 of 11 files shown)


libraries/chain/block_log.cpp
```diff
@@ -474,8 +474,8 @@ namespace eosio { namespace chain {
          virtual void     reset(const chain_id_type& chain_id, uint32_t first_block_num)      = 0;
          virtual void     flush()                                                             = 0;
 
-         virtual signed_block_ptr read_block_by_num(uint32_t block_num)    = 0;
-         virtual block_id_type    read_block_id_by_num(uint32_t block_num) = 0;
+         virtual signed_block_ptr            read_block_by_num(uint32_t block_num)        = 0;
```
heifner, line 477: spacing looks odd here.

libraries/chain/block_log.cpp
```diff
@@ -1225,9 +1225,15 @@ namespace eosio { namespace chain {
...
 
-   block_id_type block_log::read_block_id_by_num(uint32_t block_num) const {
+   std::optional<signed_block_header> block_log::read_block_header_by_num(uint32_t block_num) const {
       std::lock_guard g(my->mtx);
-      return my->read_block_id_by_num(block_num);
+      return my->read_block_header_by_num(block_num);
+   }
+
+   block_id_type block_log::read_block_id_by_num(uint32_t block_num) const {
+      auto bh = read_block_header_by_num(block_num);
```
heifner, line 1234: Add comment: `// read_block_header_by_num acquires mutex`

libraries/chain/controller.cpp
```diff
@@ -3101,6 +3101,12 @@ signed_block_ptr controller::fetch_block_by_id( block_id_type id )const {
    return signed_block_ptr();
 }
 
+std::optional<signed_block_header> controller::fetch_block_header_by_id( block_id_type id )const {
```
heifner, line 3104: `const block_id_type& id`
Might as well fix `fetch_block_by_id` as well.

libraries/chain/controller.cpp
```diff
@@ -3101,6 +3101,12 @@ signed_block_ptr controller::fetch_block_by_id( block_id_type id )const {
...
+   auto state = my->fork_db.get_block(id);
+   if( state && state->block ) return state->header;
+   return my->blog.read_block_header_by_num( block_header::num_from_id(id) );
```
heifner, line 3107: Need to verify that the block id matches the one requested. See `fetch_block_by_id` implementation above.

tests/chain_plugin_tests.cpp
```diff
@@ -112,7 +112,7 @@ BOOST_FIXTURE_TEST_CASE( get_block_with_invalid_abi, TESTER ) try {
    BOOST_CHECK_THROW(resolver("asserter"_n), invalid_type_inside_abi);
 
    // get the same block as string, results in decode failed(invalid abi) but not exception
-   auto block2 = plugin.get_block(param, fc::time_point::maximum());
+   auto block2 = plugin.get_raw_block(param, fc::time_point::maximum());
```
heifner, line 115: Can you add a test for `get_block_header`.

heifner (changes requested):

heifner (approved):

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1930,6 +1930,47 @@ chain::signed_block_ptr read_only::get_block(const read_only::get_block_params&
    return block;
 }
 
+read_only::get_block_header_result read_only::get_block_header(const read_only::get_block_header_params& params, const fc::time_point& deadline) const{
+   std::optional<uint64_t> block_num;
+
+   EOS_ASSERT( !params.block_num_or_id.empty() && params.block_num_or_id.size() <= 64,
+               chain::block_id_type_exception,
+               "Invalid Block number or ID, must be greater than 0 and less than 64 characters"
```
linh2931, line 1938: You have an extra equal to 64 checked while greater than 0 is not checked. This is not blocking. You can fix in RC.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1930,6 +1930,47 @@ chain::signed_block_ptr read_only::get_block(const read_only::get_block_params&
...
+
+   try {
+      block_num = fc::to_uint64(params.block_num_or_id);
+   } catch( ... ) {}
+
+   if (!params.include_extensions) {
+      std::optional<signed_block_header> header;
+
+      if( block_num ) {
+         header = db.fetch_block_header_by_number( *block_num );
```
linh2931, line 1949: No try block for `db.fetch_block_header_by_number`?

linh2931 (approved):

plugins/chain_plugin/chain_plugin.cpp
huangminghuang: I believe the assertion condition is right but the message description was problematic, fixed in commit af1dcf23cf99a4236ef589f87e8c057167e0fece

plugins/chain_plugin/chain_plugin.cpp
huangminghuang: The try you saw in the else was thrown from the type conversion from `string` to `block_id_type`, not from the `fetch_block_header_by_id` itself. In this line, the block_num is already the right type; therefore, its not going to throw here. This is exactly the same logic with `get_raw_block()` in line 1920.

PR #829, opened 2023-03-16
[PH] fix bug and arg passing

oschwaldp-oci:
#### Fix bug causing possible division by 0.
    
If the `targetTps` is allowed to be 0, there can be a division by 0 error `TpsTrxGensConfig` because the `numGenerators` would be 0 and thus `initialTpsPerGenerator` would be division by 0.  Fix this error through starting search floor at min 1 as well as capping the min to the floor.

#### Fix command line argument passing.
    
When additional command line arguments were added to `performance_test_basic.py`, some were inadvertently missed being passed from `performance_test` down to the `performance_test_basic`.  Update the argument passing to match that in `performance_test_basic.py`.


heifner (approved):

ScottBailey (approved):
lgtm

PR #833, opened 2023-03-17
Fix Pinned Build on Ubuntu 22.04

kj4ezj:
From [issue 735](https://github.com/AntelopeIO/leap/issues/735), this reverts [pull request 799](https://github.com/AntelopeIO/leap/pull/799) (Search for LLVM versions 7-11, prefering latest version) to fix the pinned build on Ubuntu 22.04. There are no other changes in this PR besides the revert.

[Pull request 799](https://github.com/AntelopeIO/leap/pull/799) introduced code that attempts to automagically find the correct version of LLVM for the Leap build on systems that have multiple versions of LLVM installed and available. This code only runs when `cmake` is version 3.19 or newer. Ubuntu 22.04 ships with CMake version 3.22.1, following the new code path, whereas Ubuntu 20.04 ships with version 3.16.3 that does not follow the new code path. This explains why the pinned build only fails on Ubuntu 22.04.

I have verified the Ubuntu 22.04 pinned build works on this branch.

Credit to @spoonincode for identifying the cause of the bug.

## See Also

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,29 +77,9 @@ if(ENABLE_OC AND CMAKE_SIZEOF_VOID_P EQUAL 8 AND NOT WIN32)
       list(APPEND EOSIO_WASM_RUNTIMES eos-vm-oc)
       # EOS VM OC requires LLVM, but move the check up here to a central location so that the EosioTester.cmakes
       # can be created with the exact version found
-      # find_package VERSION range is only available in 3.19 and newer.
-      if(${CMAKE_VERSION} VERSION_LESS "3.19")
-         find_package(LLVM REQUIRED CONFIG)
-      else()
-         # Preferring the latest LLVM version, search through major 11..7 and minor 9..0 in reverse order.
-         # Minor versions may break API. This manifests in the LLVM config files such that a range of `11...<12`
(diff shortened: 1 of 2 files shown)


spoonincode (approved):
I'm fine with reverting this until a fix is decided on, so we can get the builds operational again asap

ScottBailey:
If #834 fixes the defect, this should be abandoned. #799 does not fail for my non-pinned builds.

spoonincode (approved):
#834 works but isn't production ready due to it not working with existing dep directories (ugh.. build scripts). #799 is basically causing a release blocker for a nice-to-have.. so I guess I'm back to my "revert it until someone comes up with a shipable fix" approval.

kj4ezj:
> Evaluate #834 and abandon this if its a solution.

The base branch is broken and the 4.0 release is blocked. The correct thing to do is to revert the breaking change.

If someone wants to try to re-implement the feature we were attempting to introduce, anyone is welcome to post up another PR. Reverting breaking changes is always the right thing to do because we want the base branch to be in a working state at any given point in time.

spoonincode:
fwiw I changed #834 to a draft after Scott made his comment -- after I realized it didn't work in all cases because of how the script works. It does otherwise fix the issue with a fresh depdir

ScottBailey (approved):

PR #836, opened 2023-03-17
Snapshot: fix issue with erasing element while iterating

### LINKED ISSUE #838: Test failure: nodeos_snapshot_forked_test
https://github.com/AntelopeIO/leap/actions/runs/4443562488/jobs/7800998721?pr=192

heifner:
Erasing an element from multi-index while iterating over it caused an abort (core dump) when running `nodeos_snapshot_forked_test.py`. Remove the element at the end instead.

Resolves #838

### DIFF
--- a/plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
+++ b/plugins/producer_plugin/include/eosio/producer_plugin/snapshot_scheduler.hpp
@@ -70,6 +70,7 @@ public:
          }
       };    
 
+      std::vector<uint32_t> unschedule_snapshot_request_ids;
       for(const auto& req: _snapshot_requests.get<0>()) {
          // -1 since its called from start block
          bool recurring_snapshot =  req.block_spacing &&  (height >= req.start_block_num + 1) && (!((height - req.start_block_num - 1) % req.block_spacing));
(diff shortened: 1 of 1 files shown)


greg7mdp (commented):
Instead of adding a vector, can we just loop a second time, as in:

```
      for(const auto& req: _snapshot_requests.get<0>()) {
         // -1 since its called from start block
         bool recurring_snapshot =  req.block_spacing &&  (height >= req.start_block_num + 1) && (!((height - req.start_block_num - 1) % req.block_spacing));
         bool onetime_snapshot   = (!req.block_spacing) && (height == req.start_block_num + 1);
        
         // assume "asap" for snapshot with missed/zero start, it can have spacing
         if(!req.start_block_num) {
            // update start_block_num with current height only if this is recurring
            // if non recurring, will be executed and unscheduled
            if (req.block_spacing && height) {
               auto& snapshot_by_id = _snapshot_requests.get<by_snapshot_id>();
               auto it = snapshot_by_id.find(req.snapshot_request_id);
               _snapshot_requests.modify(it, [&height](auto& p) { p.start_block_num = height - 1; });
               serialize_needed = true;              
            }
            execute_snapshot_with_log(req);
         } else if(recurring_snapshot || onetime_snapshot) {
            execute_snapshot_with_log(req);
         }       
      }

      for(const auto& req: _snapshot_requests.get<0>()) {
         // cleanup - remove expired (or invalid) request
            (!req.block_spacing && height >= (req.start_block_num + 1)) || 
            (req.end_block_num > 0 && height >= (req.end_block_num + 1))) {
            unschedule_snapshot(req.snapshot_request_id);
         }
      }
```

heifner:
> > Sure, but it is just a vector of ints, seems like better than looping again.
> 
> I'm fine either way. My preference is to not add extra data containers when what we have is perfectly fine. And the multi-index is only likely to have a couple entries in it. Up to you though.

This is holding up #192, so would like to get it in.

greg7mdp (approved):

linh2931 (approved):

PR #843, opened 2023-03-17
simplify read-only threads execution

linh2931:
This PR simplifies read-only thread lifetime management. 
- No task future queue is needed.
- No more push/pop tasks to the task queue in every window switching
- Switching to write window from only one place and simplifying reasoning.
- Fix a starvation problem

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -21,6 +21,7 @@
 #include <iostream>
 #include <algorithm>
 #include <mutex>
+#include <chrono>
 #include <boost/algorithm/string/predicate.hpp>
 #include <boost/range/adaptor/map.hpp>
 #include <boost/multi_index_container.hpp>
@@ -434,8 +435,12 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       fc::microseconds                _ro_max_trx_time_us{ 0 }; // calculated during option initialization
       ro_trx_queue_t                  _ro_trx_queue;
       std::atomic<uint32_t>            _ro_num_active_trx_exec_tasks{ 0 };
-      std::vector<std::future<bool>>  _ro_trx_exec_tasks_fut;
+      std::condition_variable         _switch_window_cond;
+      std::mutex                      _switch_window_mtx;
+      uint32_t                        _ro_idle_threads{ 0 };
+      std::condition_variable         _ro_more_work_cond;
 
+      void start_read_only_thread();
       void start_write_window();
       void switch_to_write_window();
       void switch_to_read_window();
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -434,8 +435,12 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       fc::microseconds                _ro_max_trx_time_us{ 0 }; // calculated during option initialization
       ro_trx_queue_t                  _ro_trx_queue;
       std::atomic<uint32_t>            _ro_num_active_trx_exec_tasks{ 0 };
-      std::vector<std::future<bool>>  _ro_trx_exec_tasks_fut;
+      std::condition_variable         _switch_window_cond;
+      std::mutex                      _switch_window_mtx;
+      uint32_t                        _ro_idle_threads{ 0 };
```
heifner, line 440: I think you want your condition variable to be the queue size. So you would use the `ro_trx_queue_t` mutex and not need `_ro_idle_threads`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2819,33 +2831,37 @@ void producer_plugin_impl::switch_to_read_window() {
...
-          } else if ( self ) {
-             self->_ro_trx_exec_tasks_fut.clear();
-          }
-       }));
+// Called from a read only trx thread. Run in parallel with main and other read only trx threads
+void producer_plugin_impl::start_read_only_thread() {
+   chain::controller& chain = chain_plug->chain();
+   chain.init_thread_local_data();
+
+   while ( 1 ) {
```
heifner, line 2846: The `named_thread_pool` does not have the opportunity to call `_ioc.run()`. That works now because you don't post anything to the `_ro_thread_pool` executor, but seems rather odd.

heifner (changes requested):
Stopping review since I believe to address current comments requires pretty much a complete rewrite of this PR.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -645,6 +650,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             // executed in read window
             std::lock_guard<std::mutex> g( _ro_trx_queue.mtx );
             _ro_trx_queue.queue.push_back({trx, std::move(next)});
+            if ( _ro_idle_threads > 0 ) {
+               _ro_more_work_cond.notify_all();
```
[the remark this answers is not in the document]
greg7mdp: Agree with Kevin comment on `notify_all`. 
I find it much easier to reason about, and to convince myself that the code is correct, when the mutex `lock()` and const `wait() / notify()` are internal to the trx queue class, as I did in my PR.

PR #844, opened 2023-03-17
Cleanup and prevent thread starvation while in the read window.

### LINKED ISSUE #822: Read-only transaction processing - prevent thread starvation while in the read window.
This is an issue related to the processing of read-only transactions in parallel during the read window.

When starting the read window, a task is posted for each thread in the ro thread pool (`_ro_thread_pool`), and this task has a loop doing approximately:

```
while (/* time left in read window */) {
      if (ro_trx_queue.empty())
          break;
      // pop and execute ro transaction
      ...
}
```

The issue is that, if for example we have only one ro trx in the queue when we enter the read window, the first thread will start executing this transaction, and all the other tasks will exit, leaving a single task to execute all further read transactions which may be posted while the read window is still active.

So essentially, for the remainder of the read window we would effectively execute readonly transactions on a single thread.

Instead of exiting the task, if there are other threads currently executing ro transactions, and there is time left in the read window, we should probably wait on a condition variable associated to the queue (`_ro_trx_queue`) instead of exiting the loop.

This will require changes to `_ro_trx_queue`.  In addition encapsulate queue lock/unlock by adding a push/pop members to `_ro_trx_queue`.

greg7mdp:
Resolves #822.

This PR encapsulate wait and lock operations in the ro transaction queue, and makes the tasks wait on a condition variable while the read window is active, and there are transactions either in the queue or being processed.

It addresses the issue of thread starvation described in issue #822.


heifner (changes requested):
Seems like this should be coordinated with #843 as it addresses some of the same issues.

greg7mdp:
> Seems like this should be coordinated with #843 as it addresses some of the same issues.

I agree. Areg asked me to work on this yesterday, and it seems both Lin and I worked on this in parallel.

linh2931:
Was focusing on fixing the main branch this afternoon. We should get together when Kevin is back to talk about the best way, keeping in mind to benefit Kevin's upcoming PR.

greg7mdp:
What test do we have that runds readonly transactions, with a read window active? I believe `test_read_only_trx` runds only without a read window.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -898,6 +978,8 @@ void producer_plugin::set_program_options(
           "time in microseconds the write window lasts")
          ("read-only-read-window-time-us", bpo::value<uint32_t>()->default_value(my->_ro_read_window_time_us.count()),
           "time in microseconds the read window lasts")
+         ("allow-ro-trx-on-producer-node", bpo::value<bool>()->default_value(false),
```
linh2931, line 981: We should not have this as an option. This is for testing only.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -417,11 +417,92 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+
+            if (queue.empty()) {
+               if (num_waiting + 1 == max_waiting) {
+                  notify_waiting(g, true);
+                  return false;
+               }
+               
+               ++num_waiting;
+               cond.wait(g);
+               --num_waiting;
```
heifner, line 460: This is a very odd use of a condition variable. I would recommend reviewing: https://embeddedartistry.com/blog/2022/01/10/remember-to-lock-around-all-stdcondition_variable-variables/

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -417,11 +417,92 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+         bool should_exit() {
+            return *received_block_ptr ||
+               fc::time_point::now() >= read_window_deadline;
+         }
+         
+         mutable std::mutex      mtx;
+         std::condition_variable cond;
+         deque<ro_trx_t>         queue;
+         uint32_t                num_waiting {0};
+         uint32_t                max_waiting {0};
```
heifner, line 500: These are accessed from multiple threads. As written this is not thread safe.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -417,11 +417,92 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+            return true;
+         }
+
+         // We exit the read window when either:
+         //    - all threads would be idle
+         //    - or the net_plugin received a block.
+         //    - or we have reached the read_window_deadline
+         void set_exit_criteria(uint32_t num_tasks, std::atomic<bool>* received_block, fc::time_point deadline) {
+            assert(num_tasks > 0 && num_waiting == 0);
+            assert(received_block && *received_block == false);
```
heifner, line 477: I thought we decided not to use `assert`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -417,11 +417,92 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+         bool pop_front(ro_trx_t &trx) {
+            std::unique_lock<std::mutex> g( mtx );
+            
+            if (should_exit()) {
+               notify_waiting(g, true);
+               return false;
+            }
+
+            if (queue.empty()) {
+               if (num_waiting + 1 == max_waiting) {
```
heifner, line 453: I think this needs to be refactored so you wait on a predicate on the conditions of the condition variable. Seems to me that condition is non-empty queue, received block, or timer fired. All threads can wait on that condition and be notified to check on a push read only trx (notify one) or timer (notify all).

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: Thanks Lin. I don't see another way to disable the checking for producer, so we can run the read window in the C++ test `test_read_only_trx.cpp`. If you can suggest a better solution, I'll be glad to implement it.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: They are accessed only inside the class `ro_trx_queue_t`, by member functions which take a lock (`std::unique_lock<std::mutex> g( mtx );`). Where is it not thread safe?

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: Can you elaborate on what is odd?
As expected, the wait() is called when the mutex lock is active, and the `notify_*` is done after unlocking.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: My memory of it is that we didn't reach a conclusion either way.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: In the following diagram, it shows that when we run out of trx to process in every thread, we switch back to the read window. That is what we did before this PR as well. 
image

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -417,11 +417,92 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+               --num_waiting;
+               
+               if (queue.empty() || should_exit()) // were we woken up by another thread?
+                  return false;
+            }
+            
+            trx = std::move(queue.front());
+            queue.pop_front();
+            return true;
+         }
```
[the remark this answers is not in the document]
greg7mdp: I don't understand what you mean by this.

plugins/producer_plugin/producer_plugin.cpp
heifner: `set_exit_criteria` does not take a lock.

plugins/producer_plugin/producer_plugin.cpp
heifner: You could create a header method `verify_no_producers(my->_producers.empty(), ""--read-only-threads not allowed on producer node"")` that had a `#ifdef`. You could add a static variable that you set before construction and check in `plugin_initialize`.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: `set_exit_criteria()` is called before we start the parallel tasks.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: I asked Areg on telegram. If he prefers not to add an option I'll go with the static variable.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: OK, got it, spurious notifications were not taken into account. Current version should be safe now.

plugins/producer_plugin/test/test_read_only_trx.cpp
```diff
@@ -150,8 +150,20 @@ BOOST_AUTO_TEST_CASE(no_read_only_threads) {
    test_trxs_common(specific_args);
 }
 
-// test read-only trxs on separate threads (with --read-only-threads)
-BOOST_AUTO_TEST_CASE(with_read_only_threads) {
+// test read-only trxs on 1 threads (with --read-only-threads)
+BOOST_AUTO_TEST_CASE(with_16_read_only_threads) {
```
linh2931, line 154: test name does not match the number f threads.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -417,11 +417,88 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+            assert(num_tasks > 0 && num_waiting == 0 && received_block != nullptr);
+            assert(received_block && *received_block == false);
+            max_waiting = num_tasks;
+            received_block_ptr = received_block;
+            read_window_deadline = deadline;
+         }
+
+      private:
+         bool should_exit() {
+            return *received_block_ptr || exiting_read_window || fc::time_point::now() >= read_window_deadline;
```
heifner, line 489: I would check `*received_block_ptr` last as it is more expensive to check than `exiting_read_window`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -474,33 +476,26 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-         void notify_waiting(bool all) {
-            if (num_waiting) {
-               all ? cond.notify_all() : cond.notify_one();
-            }
-         }
-            
          bool should_exit() {
-            return *received_block_ptr || exiting_read_window || 
-               fc::time_point::now() >= read_window_deadline;
+            return *received_block_ptr || exiting_read_window || fc::time_point::now() >= read_window_deadline;
```
linh2931, line 489: I think the condition in #862 is better and clearly spelled out (`num_waiting == max_waiting || fc::time_point::now() >= read_window_deadline || *received_block_ptr`). And the order reflects the likelihood of conditions too.

plugins/producer_plugin/test/test_read_only_trx.cpp
```diff
@@ -150,4 +150,29 @@ BOOST_AUTO_TEST_CASE(no_read_only_threads) {
...
+                                              "--read-only-threads=1",
+                                              "--max-transaction-time=10",
+                                              "--read-only-write-window-time-us=100000",
+                                              "--read-only-read-window-time-us=40000",
+                                              "--disable-subjective-billing=true" };
+   test_trxs_common(specific_args);
+}
+
+// test read-only trxs on 16 separate threads (with --read-only-threads)
+BOOST_AUTO_TEST_CASE(with_1_read_only_threads) {
```
heifner, line 166: Test name should be switched with the other test.

heifner (changes requested):

plugins/producer_plugin/test/test_read_only_trx.cpp
greg7mdp: done

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: done

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -417,11 +417,88 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+
+         bool empty() const {
+            std::lock_guard<std::mutex> g( mtx );
+            return queue.empty();
+         }
+
+         // may wait if the queue is empty, and not all other threads are already waiting.
+         // returns true if a transaction was dequeued and should be executed, or false 
+         // if conditions are met to stop processing transactions. 
+         bool pop_front(ro_trx_t &trx) {
```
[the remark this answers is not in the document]
greg7mdp: done

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: I don't want to test `num_waiting == max_waiting` here on purpose.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: > as it is more expensive

Because of the `atomic`? It is pronbably more expensive than `exiting_read_window`, but less than `fc::time_point::now()`

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: I switched to using a static variable as suggested by Kevin.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: Areg just said that he is OK with this use of `assert`.

PR #846, opened 2023-03-20
P2P improvement TR02/03

bxliu:
Add configuration items chain-alias, chain-id, chain-alias-mapping.

chain-alias-mapping is set in chain_alias_mapping.json from config dir.
JSON format:
{
    "mappings": {
      "eos": "aca376f206b8fc25a6ed44dbdc66547c36c6c33e3a119ffbeaef943642f0e906",
      "wax": "1064487b3cd1a897ce03ae5b6a865651747e2e152090f99c1d19d44e01aea5a4",
      "telos": "4667b205c6838ef70ff7988f6e8257e8be0e1284a2f59699054a018f743b1d11",
      "ux": "8fc6dce7942189f842170de953932b1f66693ad3788f766e777b6f9d22335c02",
      "test": "8a34ec7df1b8cd06ff4a8abbaa7cc50300823350cadc59ab296cb00d104d2b8f"
    },
    "version": 0
}

The rules are:
1. If there is a chain-alias, check the chain-alias-mapping, look for alias:id, if no matching id is found, stop the startup;
2. If there is a chain-id, no matter what step 1 does, it will overwrite the id;
3. Check whether the id is the same as the id calculated by genesis or snapshot or block, if not, stop the startup.


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -226,8 +228,8 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
    cfg.add_options()
          ("chain-alias", bpo::value<string>(), "Chain Alias, alias id mapping is needed. a bootnode server address must set in p2p-peer-address.")
          ("chain-id", bpo::value<string>(), "Chain ID, must match chain id in genesis or snapshot or block. a bootnode server address must set in p2p-peer-address.")
-         ("chain-alias-mapping", bpo::value< vector<string> >()->composing(),
-          "The mapping info in config.\n"
+         ("chain-alias-mapping", bpo::value< vector<string> >()->composing()->multitoken(),
```
systemzax, line 231: The alias to chain_id mappings should not be a configuration option, but rather should be part of the code distribution. For instance, it could be loaded as a .json file upon application start, that would exist in a specific folder.

The objective is to have an alias supported by default in the code base for the coalition chains, without requiring node operators to define these mappings themselves.

As new chains join the coalition, we can leverage the PR mechanism to have them added to the list of aliases.

Here are the mappings :

"eos" : "aca376f206b8fc25a6ed44dbdc66547c36c6c33e3a119ffbeaef943642f0e906"
"wax": "1064487b3cd1a897ce03ae5b6a865651747e2e152090f99c1d19d44e01aea5a4"
"telos" : "4667b205c6838ef70ff7988f6e8257e8be0e1284a2f59699054a018f743b1d11"
"ux" : "8fc6dce7942189f842170de953932b1f66693ad3788f766e777b6f9d22335c02"

systemzax (changes requested):
See comment regarding alias to chain_id mappings.

Otherwise, the rest looks good.

porkchop (approved):

libraries/chain/include/eosio/chain/config.hpp
```diff
@@ -7,6 +7,8 @@ namespace eosio { namespace chain { namespace config {
 
 typedef __uint128_t uint128_t;
 
+const static auto default_mapping_filename   = "mapping.json";
```
systemzax, line 10: Should be named something more specific, such as "chain_alias_mapping_filename"

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -898,21 +896,51 @@ void chain_plugin::plugin_initialize(const variables_map& options) {
...
-              for (const auto& entry : mapping) {
-                  auto colon_pos = entry.find(':');
-                  if (colon_pos != string::npos && entry.substr(0, colon_pos) == my->chain_alias) {
-                      auto id = entry.substr(colon_pos + 1);
-                      my->cfg_chain_id = id;
-                      break;
-                  }
+          bfs::path mapping_file;
+
+          if( options.count( "chain-alias-mapping-json" ) ) {
```
systemzax, line 901: Should be removed

systemzax (changes requested):
The mapping file (chain-alias-mapping-json) should not be a configurable option. The rest of the function looks good

systemzax (approved):

heifner (changes requested):
It is not clear to me how this provides:

TR02 The solution will allow the selection of a coalition network via alias or chain id.
TR03 The solution will allow the selection of a non-coalition network via chain id.

The PR as is simply sets `chain_alias` `cfg_chain_id` variables and then does nothing with them. Seems like this needs to be included with PR#918 and set up so that it knows a peer to connect to. 

I'm assuming that ENF will run a known peer discovery node. But maybe that will be a job of BPs. Regardless this needs some well known peer to connect to. Or I've complete misunderstood what is being accomplished here.

The PR description does not indicate the JSON format used for chain_alias_mapping.json.

bxliu:
> It is not clear to me how this provides:
> 
> TR02 The solution will allow the selection of a coalition network via alias or chain id. TR03 The solution will allow the selection of a non-coalition network via chain id.
> 
> The PR as is simply sets `chain_alias` `cfg_chain_id` variables and then does nothing with them. Seems like this needs to be included with #918 and set up so that it knows a peer to connect to.

This part of PR only check alias or chain id config, make sure that configuration matches the data.

> 
> I'm assuming that ENF will run a known peer discovery node. But maybe that will be a job of BPs. Regardless this needs some well known peer to connect to. Or I've completely misunderstood what is being accomplished here.

The prerequisite is indeed to set the corresponding P2P peer in the configuration file (in PR918), otherwise there is no way to know the bootnode (coalition/non-coalition).
From my understanding, the future goal of the project is to support multiple parachain projects. Therefore, it's also challenging to establish a universal P2P node configuration that covers all chain_ids. Node users will need to obtain P2P bootnode information from the project documentation.

> 
> The PR description does not indicate the JSON format used for chain_alias_mapping.json.

This modification was made to match the requirement as requested in the [comments](https://github.com/AntelopeIO/leap/pull/846#discussion_r1145281588) of this pull request.

ericpassmore:
closing as leap 6 IF changes have evolved significantly.

PR #835, opened 2023-03-17
Change ASCII Art Banner in Build Script

kj4ezj:
From [issue 735](https://github.com/AntelopeIO/leap/issues/735), the pinned build script prints a banner to the user's terminal spelling "LEAP" in ASCII art once the build is done. Our scripts have always done this as long as I have been involved in EOSIO/Antelope. This pull request upgrades our ASCII art! It now features an antelope jumping over the text "LEAP", keeping the 80-char width from before.

Don't worry, I did NOT spend a bunch of time on this. I used [DALL-E](https://openai.com/blog/dall-e) from OpenAI to generate an image based on this prompt...
> A pencil drawing of an antelope with horns jumping over the word "LEAP" at a 45 degree angle

...then I took the image it generated and fed it through [ASCIIart.club](https://asciiart.club) to display it using unicode characters. It took me no time at all! Doesn't it look cool? Maybe we should put this in the `README.md`, too.

## See Also
- [Pull Request 811](https://github.com/AntelopeIO/leap/pull/811) - Pinned Build Script Changes + Restore `libcurl4-openssl-dev`
- [Pull Request 812](https://github.com/AntelopeIO/leap/pull/812) - Pinned Build Script Linter Changes
- [Pull Request 832](https://github.com/AntelopeIO/leap/pull/832) - Roll `install_deps.sh` into `pinned_build.sh`
- [Pull Request 833](https://github.com/AntelopeIO/leap/pull/833) - Fix Pinned Build on Ubuntu 22.04
  - Reverts [pull request 799](https://github.com/AntelopeIO/leap/pull/799)
- [Pull Request 835](https://github.com/AntelopeIO/leap/pull/835) - Change ASCII Art Banner in Build Script

PR #847, opened 2023-03-20
Restore libcurl4-openssl-dev

kj4ezj:
From [issue 735](https://github.com/AntelopeIO/leap/issues/735), this pull request fixes the pinned build by returning the `libcurl4-openssl-dev` dependency.

## See Also

### DIFF
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -4,4 +4,23 @@ apt-get update --fix-missing
 DEBIAN_FRONTEND='noninteractive'
 TZ='Etc/UTC'
 apt-get install -y tzdata
-apt-get install -y zip unzip libncurses5 wget git build-essential cmake curl libgmp-dev libssl-dev libzstd-dev time zlib1g-dev libtinfo-dev bzip2 libbz2-dev python3 python3-numpy file
+apt-get install -y \
+    build-essential \
+    bzip2 \
+    cmake \
+    curl \
+    file \
+    git \
+    libbz2-dev \
+    libgmp-dev \
+    libncurses5 \
+    libssl-dev \
+    libtinfo-dev \
+    libzstd-dev \
+    python3 \
+    python3-numpy \
+    time \
+    unzip \
+    wget \
+    zip \
+    zlib1g-dev
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -12,6 +12,7 @@ apt-get install -y \
     file \
     git \
     libbz2-dev \
+    libcurl4-openssl-dev \
     libgmp-dev \
     libncurses5 \
     libssl-dev \
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -2,7 +2,7 @@
 apt-get update
 apt-get update --fix-missing
 export DEBIAN_FRONTEND='noninteractive'
-TZ='Etc/UTC'
+export TZ='Etc/UTC'
 apt-get install -y \
     build-essential \
     bzip2 \
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -3,7 +3,6 @@ apt-get update
 apt-get update --fix-missing
 DEBIAN_FRONTEND='noninteractive'
 TZ='Etc/UTC'
-apt-get install -y tzdata
 apt-get install -y \
     build-essential \
     bzip2 \
@@ -21,6 +20,7 @@ apt-get install -y \
     python3 \
     python3-numpy \
     time \
+    tzdata \
     unzip \
     wget \
     zip \
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
-
 apt-get update
 apt-get update --fix-missing
-DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get -y install tzdata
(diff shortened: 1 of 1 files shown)


kj4ezj:
> Before, these environment vatiable settings applied only to the `tzdata` install:
> 
> `DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get -y install tzdata`
> 
> I think that's why `tzdata` was installed separately. Was this change made for a specific purpose?

The `tzdata` install does not need to be separate. I am also not sure that line did what we think it is doing.

2023-03-20 18-03-29 - DEBIAN_FRONTEND

It is, at the very least, ambiguous behavior.

In any case, all installations are non-interactive, and all installations are intended to be in the context of a single timezone.

Is there something specific you see not working?

greg7mdp:
I have to leave in a couple minutes, but I'll be back later this evening, and I can approve if needed.

greg7mdp:
I'd recomment adding only what's missing (`libcurl4-openssl-dev` ?) and not make any other change.

kj4ezj:
> Since it appears to have been done on purpose, and as far as we know it is not broken, then why fix it?

I answered this question in my previous message - the behavior of the previous code is ambiguous. What happened to having a bias for change?

scripts/install_deps.sh
```diff
@@ -1,6 +1,27 @@
 #!/bin/bash
-
 apt-get update
 apt-get update --fix-missing
-DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get -y install tzdata
-apt-get -y install zip unzip libncurses5 wget git build-essential cmake curl libgmp-dev libssl-dev libzstd-dev time zlib1g-dev libtinfo-dev bzip2 libbz2-dev python3 python3-numpy file
+DEBIAN_FRONTEND='noninteractive'
+TZ='Etc/UTC'
+apt-get install -y tzdata
+apt-get install -y \
```
ScottBailey: Why is tzdata on it's own install line?

scripts/install_deps.sh
```diff
@@ -1,6 +1,27 @@
 #!/bin/bash
-
 apt-get update
 apt-get update --fix-missing
-DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get -y install tzdata
-apt-get -y install zip unzip libncurses5 wget git build-essential cmake curl libgmp-dev libssl-dev libzstd-dev time zlib1g-dev libtinfo-dev bzip2 libbz2-dev python3 python3-numpy file
+DEBIAN_FRONTEND='noninteractive'
+TZ='Etc/UTC'
```
ScottBailey: This won't work. Either set them before the call or `export` them.

ScottBailey (changes requested):
One package per line is easier to read and review. That part is a good change. Keep it. 

I'm not sure why `tzdata` would have been isolated to it's own line. My guess is that it was a copy paste from stackoverflow. Consider moving it into the larger list.

The settings `DEBIAN_FRONTEND` and `TZ` **must** either be on the same line as the call to apt **or** be export-ed. A simple set (i.e. `X=y`) won't put them into the environment for apt.

If we chose to call apt in interactive mode (why would we?!) we should comment on the difference between the two calls.

scripts/install_deps.sh
kj4ezj: Done.

scripts/install_deps.sh
kj4ezj: Done.

ScottBailey (approved):
lgtm

spoonincode (approved):

greg7mdp:
I still would have preferred the change to be only what was strictly necessary, since this is the day before the release, and there was enough back and forth to demonstrate that the additional changes are not quite trivial.

greg7mdp:
> What happened to having a bias for change?

You are quoting me wrong here. I said we shouldn't have a bias *against* change, which is really what happens. But we shouldn't have a bias for change either.

In special cases, like the days before a release, I think we should have a bias against change. In some companies they institute a "freeze" in the weeks before a release, where only strictly necessary and thoroughly vetted changes are allowed in.

spoonincode:
I was trying to not open up more discussion on the topic. I'm not really convinced either `DEBIAN_FRONTEND` nor `TZ` nor the `-y` should be here at all. Who are we to decide what time zone a user's box should be initialized to? Why should we not allow a user to interactively confirm what apt-get is about to do? This should just be left default and the CI can override it.

The change merged seems benign enough.

ScottBailey:
Those are good points, but no one is making the user run this script. And since it's self contained and easy to read, they are free to edit it or run the commands by hand.

spoonincode:
yeah that's the benefit of keeping it a separate script that the user themselves must sudo/run as root: they _know_ they're giving our script root privileges to do whatever because they must deliberately do so. Still, everything we do in a root context should be heavily scrutinized by ourselves, lest we end up with scripts from the 2.1 era that silently stuff a third party code signing key in to the user's apt database.

scripts/install_deps.sh
greg7mdp: > This won't work. Either set them before the call or `export` them

Actually, this would work. The variable values would change during the script, they just wouldn't be exported to child processes.

ScottBailey: Pedantically: the mode would remain interactive and the user would be queried for timezone.

PR #862, opened 2023-03-22
Simplify condition variable use

heifner:


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -419,18 +419,19 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
    
       // The queue storing read-only transactions to be executed by read-only threads
+      // Thread-safe
       class ro_trx_queue_t {
       public:
          void push_back(ro_trx_t&& trx) {
             std::unique_lock<std::mutex> g( mtx );
             queue.push_back(std::move(trx));
-            notify_waiting(false);
+            cond.notify_one();
```
greg7mdp, line 428:
```suggestion
            if (num_waiting)
               cond.notify_one();
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -473,34 +464,27 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          //    - all threads would be idle
          //    - or the net_plugin received a block.
          //    - or we have reached the read_window_deadline
-         void set_exit_criteria(uint32_t num_tasks, std::atomic<bool>* received_block, fc::time_point deadline) {
-            assert(num_tasks > 0 && num_waiting == 0);
+         void set_exit_criteria(uint32_t num_tasks, std::atomic<bool>* received_block, const fc::time_point& deadline) {
+            std::lock_guard<std::mutex> g( mtx );
```
greg7mdp, line 468: `std::lock_guard` unnecessary as this is called from a single thread

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -473,34 +464,27 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          //    - all threads would be idle
          //    - or the net_plugin received a block.
          //    - or we have reached the read_window_deadline
-         void set_exit_criteria(uint32_t num_tasks, std::atomic<bool>* received_block, fc::time_point deadline) {
-            assert(num_tasks > 0 && num_waiting == 0);
+         void set_exit_criteria(uint32_t num_tasks, std::atomic<bool>* received_block, const fc::time_point& deadline) {
```
greg7mdp, line 467: I think it is better to pass deadline by value https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -419,18 +419,19 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
             queue.push_back(std::move(trx));
-            notify_waiting(false);
+            cond.notify_one();
          }
          
          void push_front(ro_trx_t&& trx) {
             std::unique_lock<std::mutex> g( mtx );
             queue.push_front(std::move(trx));
-            notify_waiting(false);
+            cond.notify_one();
```
greg7mdp, line 434: same as above

heifner:
Comments are being included into #844

plugins/producer_plugin/producer_plugin.cpp
linh2931: Maybe we don't want to notify while in write window. `if (num_waiting)` can cover that.

PR #868, opened 2023-03-22
Fix import failure for platforms without unshare

huangminghuang:
This PR fixes the import failure for platform which doesn't support `unshare`.


spoonincode:
I think this should go to 4.0 first.

PR #873, opened 2023-03-22
[4.0 -> main] Increase robustness of snapshot diff python test

766C6164:
Bringing in https://github.com/AntelopeIO/leap/pull/856 to main

This PR mitigates rare and annoying issue of the test failing when there is a new block happening in between of scheduling and "now" create_snapshot calls. The fix is to execute create_snapshot first and schedule programmable snapshot at exact same block num. Same as with irreversible one.

### DIFF
--- a/tests/TestHarness/queries.py
+++ b/tests/TestHarness/queries.py
@@ -346,7 +346,7 @@ class NodeosQueries:
         return self.processCleosCmd(cmd, cmdDesc, silentErrors=False, exitOnError=exitOnError, exitMsg=msg, returnType=returnType)
 
     def getTable(self, contract, scope, table, exitOnError=False):
-        cmdDesc = "get table"
+        cmdDesc = "get table --time-limit 999"
         cmd="%s %s %s %s" % (cmdDesc, contract, scope, table)
         msg="contract=%s, scope=%s, table=%s" % (contract, scope, table);
         return self.processCleosCmd(cmd, cmdDesc, exitOnError=exitOnError, exitMsg=msg)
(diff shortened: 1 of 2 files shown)


tests/TestHarness/queries.py
```diff
@@ -346,7 +346,7 @@ def getEosAccount(self, name, exitOnError=False, returnType=ReturnType.json):
         return self.processCleosCmd(cmd, cmdDesc, silentErrors=False, exitOnError=exitOnError, exitMsg=msg, returnType=returnType)
 
     def getTable(self, contract, scope, table, exitOnError=False):
-        cmdDesc = "get table"
+        cmdDesc = "get table --time-limit 999"
```
ClaytonCalabrese, line 349: This will probably break 2.0 compatibility for performance test. This time limit is handled in `configure_version()` but was lost due to the splitting of Node.py into queries.

https://github.com/AntelopeIO/leap/pull/874/files#diff-4d1412150dda54fbc6280cc9053914a45d01e43085e35b0a5a68e9c8758ed42bR350 restores this in a way that won't break 2.0 compatibility.

766C6164: I see, that was brought from un-merged to main PR with Kevin's commit, should I just remove this change from this commit?  It has nothing to do with said python test

766C6164: removed it from this pr, so it properly has only one file with my changes

ClaytonCalabrese: Thanks

heifner (approved):

linh2931 (approved):

PR #872, opened 2023-03-22
[4.0] Test fix: Add max-transaction-time -1

heifner:
Use `--max-transaction-time -1` to avoid issue with `setcode` timing out.

PR #883, opened 2023-03-23
[4.0 -> main] Merge 4.0.0 rc2 version bump to main

linh2931:
This brings along tests/cli_test.py from #872.


linh2931:
Let #882 in first for clean history. Redo this in a new PR.

Issue #782, opened 2023-03-07
Add time summary logging for read-only trx processing


heifner:
https://github.com/AntelopeIO/leap/pull/776 Added multi-threaded read-only transaction execution.
https://github.com/AntelopeIO/leap/pull/145 Added logging of block trx summary for speculative and BP blocks.
When multi-threaded read-only transaction is enabled it would be nice to provide a log summary report of time spent on the read-only threads executing read-only transactions.

Something like:
```
debug 2022-09-12T18:47:37.241 test_trx_ producer_plugin.cpp:187       report               ] Read-only all threads idle: 131428us out of 150000us, success: 1128, 6622us, fail: 174, 2321us, other: 1054us
```

arhag:
Is this redundant with https://github.com/AntelopeIO/leap/issues/828?

heifner:
> Is this redundant with #828?

Oops. Yep. Forgot I had already created that.

PR #885, opened 2023-03-23
[4.0] Add time summary logging for transient trx processing and read-only threads

linh2931:
Resolve https://github.com/AntelopeIO/leap/issues/782

This PR provides a log summary report of time spent on transient transactions (dry-run and read-only), and the read-only threads executing read-only transactions.

- If `--read-only-threads` is NOT enabled, transient time is reported, as
`debug 2023-03-23T18:10:29.907 nodeos   producer_plugin.cpp:  278   report ] Block #199 trx idle: 573552us  out of 175us, success: 1, 660us,  fail: 0, 0us, transient: 1, 175us, other:  2345us`
- If `--read-only-threads` is enabled, time spent executing read-only transactions across the thread pool is reported, as
`debug 2023-03-23T22:25:01.077 nodeos  producer_plugin.cpp:2810  switch_to_write_wind ] Read-only threads 3, read window time 2831us, total all threads  6239us`


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -266,29 +266,37 @@ struct block_time_tracker {
...
+      transient_trx_time += time;
+      ++transient_trx_num;
+   }
+
    void report( const fc::time_point& idle_trx_time, uint32_t block_num ) {
       if( _log.is_enabled( fc::log_level::debug ) ) {
          auto now = fc::time_point::now();
          add_idle_time( now - idle_trx_time );
-         fc_dlog( _log, "Block #${n} trx idle: ${i}us out of ${t}us, success: ${sn}, ${s}us, fail: ${fn}, ${f}us, other: ${o}us",
+         fc_dlog( _log, "Block #${n} trx idle: ${i}us out of ${t}us, success: ${sn}, ${s}us, fail: ${fn}, ${f}us, transient trxs: ${tn}, ${t}us, other: ${o}us",
```
heifner, line 278: `trxs` is implied. I think: `... , transient: ${tn}, ...`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -688,7 +697,14 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
                      self->_idle_trx_time = fc::time_point::now();
-                     self->_time_tracker.add_fail_time(self->_idle_trx_time - start);
+                     auto dur = self->_idle_trx_time - start;
+                     if ( is_transient ) {
+                        // this function can only be on main thread.
+                        // transient time includes both success and fail time
+                        self->_time_tracker.add_transient_time(dur);
+                     } else {
+                        self->_time_tracker.add_fail_time(dur);
+                     }
```
heifner, line 707: I think this code would be simpler if instead you had `add_fail_time` and `add_success_time` take a `bool is_transient` so that this all becomes:
`self->_time_tracker.add_fail_time(self->_idle_trx_time - start, is_transient);`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2766,6 +2796,13 @@ void producer_plugin::log_failed_transaction(const transaction_id_type& trx_id,
 
 // Called from app thread
 void producer_plugin_impl::switch_to_write_window() {
+   if ( _log.is_enabled( fc::log_level::debug ) ) {
+      auto now = fc::time_point::now();
+      fc_dlog( _log, "Read-only threads #${n}, total time across all threads ${t}us",
```
heifner, line 2801: Recommend: `"Read-only threads ${n}, total all threads ${t}us"`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2766,6 +2796,13 @@ void producer_plugin::log_failed_transaction(const transaction_id_type& trx_id,
...
+               ("n", _ro_thread_pool_size)
+               ("t", now - _ro_read_window_start_time) );
```
heifner, line 2803: This is just the time spent in the read window. That would be good to report as well, but I was expecting this to be the total time spent on all threads executing read-only transactions. If there are 3 threads and 3 read-only transactions that each take 30ms (1 on each thread), then the total time would be 90ms worth of read-only transaction processing, not 30ms.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -688,7 +697,14 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                      self->log_trx_results( trx, nullptr, ex, 0, start, is_transient );
                      next( std::move(ex) );
                      self->_idle_trx_time = fc::time_point::now();
-                     self->_time_tracker.add_fail_time(self->_idle_trx_time - start);
+                     auto dur = self->_idle_trx_time - start;
+                     if ( is_transient ) {
+                        // this function can only be on main thread.
```
greg7mdp, line 702:
```suggestion
                        assert(is_on_main_thread());
```

linh2931: Thanks. The code is inside `app().executor().post` lambda and is on main thread. I will remove the redundant comment there.

greg7mdp: even better, add the assert so that it documents what you know is true.

linh2931: In this case, it seems an unnecessary assert. As we post to main, we would expect it is on main. All of our app().post() code assumes this.

plugins/producer_plugin/producer_plugin.cpp
linh2931: oh no, we cannot do this, as for transient transactions, we don't distinguish between fail and success.

plugins/producer_plugin/producer_plugin.cpp
linh2931: Thanks. I will count the total time of all the threads spending in `chain.push_transaction()`

plugins/producer_plugin/producer_plugin.cpp
heifner: Sure you can, something like:
```
   void add_fail_time( const fc::microseconds& fail_time, bool is_transient ) {
      if (is_transient) {
         transient_trx_time += time;
         ++transient_trx_num;
      } else {
         trx_fail_time += fail_time;
         ++trx_fail_num;
      }
   }

```
And similar change in `add_success_time`.

linh2931: Do you think this makes add_fail_time and add_success_time more complicated?

heifner: Yes, but it moves the complication into one place instead of duplicating the `if`

linh2931: But on the other hand, this hides how transient is handled. Will change.

heifner: It should be the time_trackers job to know what it is tracking. Then if we decide to also track fail/success separate it will be a trivial change.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -422,18 +436,29 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
       };
+      struct all_threads_exec_time_t {
+         std::mutex        mtx;
+         fc::microseconds  total_time{ 0 };
+
+         void add_time(const fc::microseconds& t) {
+            std::lock_guard<std::mutex>  g(mtx);
+            total_time += t;
+         }
+      };
```
heifner, line 447: Seems like `std::atomic<int64_t> total_time_us{0}` would be much simpler and more efficient.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2819,6 +2857,10 @@ void producer_plugin_impl::switch_to_read_window() {
    app().executor().set_to_read_window();
    chain_plug->chain().set_db_read_only_mode();
    _received_block = false;
+   if ( _log.is_enabled( fc::log_level::debug ) ) {
+      _ro_read_window_start_time = fc::time_point::now();
+      _ro_all_threads_exec_time.total_time = fc::microseconds(0);
+   }
```
heifner, line 2863: I don't think the `if` is worth it. Just always set them.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -256,39 +256,53 @@ struct block_time_tracker {
...
-      trx_success_time += time;
-      ++trx_success_num;
+   void add_success_time( const fc::microseconds& time, bool is_transient ) {
+      if( is_transient ) {
+         transient_trx_time += time;
+         ++transient_trx_num;
+      } else {
+         trx_success_time += time;
+         ++trx_success_num;
+      }
```
greg7mdp, line 277: These are awkward. I would do something like:

```
struct block_time_tracker {
   enum class trx { success, fail, transient, last };

   void add_trx_time( const fc::microseconds& fail_time, trx trx_type ) {
         processing_time[(int)trx_type] += fail_time;
         ++num_processed[(int)trx_type];
   }
   
   uint32_t num_processed[(int)trx::last];
   fc::microseconds processing_time[(int)trx::last];
};
```

and use like `_time_tracker.add_trx_time(end - start, trx::fail);`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2933,6 +2975,14 @@ bool producer_plugin_impl::push_read_only_transaction(
...
       // Ensure the trx to finish by the end of read-window.
       auto window_deadline = std::min( start + remaining_time_in_read_window_us, block_deadline );
+
+      auto track_exec_time = fc::make_scoped_exit([this,start](){
+         if ( _log.is_enabled( fc::log_level::debug ) ) {
+            // add_time is expensive as it is protected by a mutex.
+            // use it only when log is enabled
+            _ro_all_threads_exec_time.add_time( fc::time_point::now() - start );
+         }
+      });
```
heifner, line 2985: Seems like overkill. Would rather just see the add after the `chain.push_transaction` line.

plugins/producer_plugin/producer_plugin.cpp
heifner: I call that way more awkward. Looks like C code.

plugins/producer_plugin/producer_plugin.cpp
linh2931: We don't count if any exception happens?

heifner: Should not be possible for that to throw outside the bad_alloc cases which we will exit on anyway.

plugins/producer_plugin/producer_plugin.cpp
linh2931: My first version was `std::atomic<uint64_t>`, but I dropped it for concern of using correct microsecond type. But this is only for debug logging, I will go for simplicity.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: > I call that way more awkward. 

I guess we will have to agree to disagree. 

> Looks like C code.

What exactly? The casts?

heifner: The casts, the c-array, using the last enum value for the size of the c-array, and using an array for what in my opinion is better expressed as separate variables.

It is certainly more compact.

heifner (approved):

PR #888, opened 2023-03-23
Fix clang warning

heifner:
clang++11 warning:
```
In file included from /home/heifner/ext/leap/libraries/wasm-jit/Source/WASM/WASMSerialization.cpp:7:
/home/heifner/ext/leap/libraries/wasm-jit/Include/IR/Operators.h:682:55: warning: static_assert with no message is a C++17 extension [-Wc++17-extensions]
static_assert(sizeof(IR::OpcodeAndImm<IR::NoImm>) == 2);
```

### DIFF
--- a/libraries/wasm-jit/Include/IR/Operators.h
+++ b/libraries/wasm-jit/Include/IR/Operators.h
@@ -679,16 +679,16 @@ namespace IR
 }
 
 //paranoia for future platforms
-static_assert(sizeof(IR::OpcodeAndImm<IR::NoImm>) == 2);
-static_assert(sizeof(IR::OpcodeAndImm<IR::MemoryImm>) == 2);
-static_assert(sizeof(IR::OpcodeAndImm<IR::ControlStructureImm>) == 3);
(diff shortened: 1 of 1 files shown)


spoonincode:
What's going on here, this is completely valid c++17 right? and [this](https://en.cppreference.com/w/cpp/compiler_support/17) remarks that clang 2.5 added `static_assert` with no message. wat.

heifner:
> What's going on here, this is completely valid c++17 right? and [this](https://en.cppreference.com/w/cpp/compiler_support/17) remarks that clang 2.5 added `static_assert` with no message. wat.

In wasm-jit CMakeLists.txt
`set(CMAKE_CXX_STANDARD 11)`

spoonincode (approved):

spoonincode:
> ouch! hm... I wonder what the ramifications are for making that consistent. always hate to touch consensus code though..

Maybe when we bump the entire build to c++20, we can remove this configuration relic as we'll be touching all the consensus code at that time anyways

linh2931 (approved):

PR #893, opened 2023-03-24
[3.2] Use block_num for interrupt of start_block

### LINKED ISSUE #891: Interrupt start block when blocks ready to process
While testing https://github.com/AntelopeIO/leap/pull/867 it became apparent that https://github.com/AntelopeIO/leap/pull/543 did not prevent `start_block` from starting and delaying block production because the block was received before the start of the block. Even though this is more likely with #867, it is also possible pre-#867.

Instead of using a boolean for `received_block`, use an uint32_t block_num so we know if a block has been received that should be processed.

heifner:
Signal received block with the block number so that `start_block` can be interrupted even if the block is received before the start of `start_block`. This is particularly important for #867, but also useful as mitigation against race condition of receiving block right before start of the block in current implementation. See https://github.com/AntelopeIO/leap/pull/648.

Resolves #891

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -333,7 +333,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       std::optional<named_thread_pool>                          _thread_pool;
 
       std::atomic<int32_t>                                      _max_transaction_time_ms; // modified by app thread, read by net_plugin thread pool
-      std::atomic<bool>                                         _received_block{false}; // modified by net_plugin thread pool and app thread
+      std::atomic<uint32_t>                                     _received_block{false}; // modified by net_plugin thread pool
       fc::microseconds                                          _max_irreversible_block_age_us;
       int32_t                                                   _produce_time_offset_us = 0;
       int32_t                                                   _last_block_time_offset_us = 0;
@@ -693,7 +693,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          exhausted
       };
 
-      inline bool should_interrupt_start_block( const fc::time_point& deadline ) const;
+      inline bool should_interrupt_start_block( const fc::time_point& deadline, uint32_t pending_block_num ) const;
       start_block_result start_block();
 
       fc::time_point calculate_pending_block_time() const;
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1696,7 +1697,7 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
    if (_pending_block_mode == pending_block_mode::producing) {
       const auto start_block_time = block_time - fc::microseconds( config::block_interval_us );
       if( now < start_block_time ) {
-         fc_dlog(_log, "Not producing block waiting for production window ${n} ${bt}", ("n", hbs->block_num + 1)("bt", block_time) );
+         fc_dlog(_log, "Not producing block waiting for production window ${n} ${bt}", ("n", pending_block_num)("bt", block_time) );
```
greg7mdp, line 1700: Here we have assigned `pending_block_num = hbs->block_num + 1;`.

Is it the same as `chain.pending_block_num();`?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1798,7 +1799,7 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
          if( !remove_expired_blacklisted_trxs( preprocess_deadline ) )
             return start_block_result::exhausted;
          if( !_subjective_billing.remove_expired( _log, chain.pending_block_time(), fc::time_point::now(),
-                                                  [&](){ return should_interrupt_start_block( preprocess_deadline ); } ) ) {
+                                                  [&](){ return should_interrupt_start_block( preprocess_deadline, pending_block_num ); } ) ) {
```
greg7mdp, line 1802: Here we have assigned `pending_block_num = hbs->block_num + 1;`.

Is it the same as `chain.pending_block_num();`?

heifner: Yes. Head is what we are building on and pending is what we are building. Therefore, pending is always head+1.

greg7mdp: > Head is what we are building on and pending is what we are building

Hum, that would mean head = pending. What is head really? The last block we know about?

greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
heifner: Let me try that again. `[block#3][block#4 head][pending...` block#4 is a complete block. It is the head of the chain. pending is an incomplete block we are currently building.

greg7mdp: Oh, thanks, got you, I missed the `on` in `building on` in my first reading.

PR #895, opened 2023-03-24
set a maximum test time in CI via a ctest argument

### LINKED ISSUE #879: CI/CD: It happens that some tests fails to complete (no failure listed). When that happens, display the list of these in the log.
It would be nice to also display the command line that is used to launch the failed tests.

spoonincode:
An option to resolve #879

ENF runners have (external to the workflow and even VM environment) a maximum run time. This is intended to guard against a runner having a very long life (and thus a high :moneybag:  bill) due to a test getting wedged, or something buggy, or something malicious, etc.

Unfortunately it’s a little opaque when this protection kicks in. While it's better then it once was, it still just ends with a message `The operation was canceled` without an easy way to see the tests that were wedged.

This PR sets a global timeout on the tests via simply passing `--timeout` to ctest. The nice thing about this change is that it’s not very invasive. It doesn’t require changing the `CMakeLists.txt` all over. It only requires adding a new argument to existing ctest command (I don’t count the little more involved changes to parallel-ctest-containers because midterm that disappears). Adding zero lines is nice.

The not-so-nice thing is that it doesn’t really help someone running ctest locally unless they add the argument too. Trade off feels okay to me though.

Picked times are... 7 minutes for P & NP, 30 minutes for LR.

The documentation states
> sets a timeout on all tests that do not already have a timeout set on them via the `TIMEOUT` property

so I’ve elected not to remove `TIMEOUT` properties that are _smaller_ than the times selected above. Maybe those more strict timeouts are an important aspect of the test (idk). I did remove a couple longer `TIMEOUT`s on LR tests.


greg7mdp:
Were the ENF runner timeouts removed? And if not are these timeouts less than the ENF runner timeouts?

greg7mdp:
>  I did remove a couple longer TIMEOUTs on LR tests.

Hopefully the tests still pass with the shorter timeout.

spoonincode:
> Were the ENF runner timeouts removed? And if not are these timeouts less than the ENF runner timeouts?

They still exist. `enf-x86-beefy`, `enf-x86-hightier`, `enf-x86-midtier` (which are used for build, P tests, and NP tests currently) are all 10 minutes. While `enf-x86-lowtier`, used for LR tests currently, is 60 minutes.

Keep in mind that time is for the entire life of the VM, after it starts. So time spent on steps like cloning the repo or "initializing containers" are included.

> Hopefully the tests still pass with the shorter timeout.

the latest run on main was ~5.5 minutes for NP tests, ~20 minutes for LR. That's where the 7 & 30 came from. Clearly it's easy enough to change as needed in the future.

greg7mdp:
Sounds good, thanks!

greg7mdp (approved):

heifner (approved):

greg7mdp:
@spoonincode, maybe we should try adding a wait in a test, just to see how the change works in practice. Or I guess we can just check this in and watch for it, we'll see eventually.

spoonincode:
Here is a run with the NP limit turned down to 4m
https://github.com/AntelopeIO/leap/actions/runs/4519776232/jobs/7960484293#step:4:23736
It's a lot of output because of `--output-on-failure` but at least the log will clearly indicate something timed out

greg7mdp:
Excellent! Love it.

PR #901, opened 2023-03-25
[4.0] Execute read only safe tasks on read-only thread pool

### LINKED ISSUE #810: Move all read-only operations to parallel execution of all read operations
Depends on first identifying all read-only operations that are safe to executing in parallel during the read window and documenting this (see https://github.com/AntelopeIO/leap/issues/813).

Then this issue is to actually implement parallel execution of those identified transactions.

heifner:

Builds on #776. 
Resolves #810


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -445,24 +468,26 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
       std::atomic<int64_t>            _ro_all_threads_exec_time_us; // total time spent by all threads executing transactions. use atomic for simplicity and performance
       fc::time_point                  _ro_read_window_start_time;
+      fc::time_point                  _ro_window_deadline; // only modified on app thread
       boost::asio::deadline_timer     _ro_write_window_timer;
       boost::asio::deadline_timer     _ro_read_window_timer;
       fc::microseconds                _ro_max_trx_time_us{ 0 }; // calculated during option initialization
-      ro_trx_queue_t                  _ro_trx_queue;
-      std::atomic<uint32_t>           _ro_num_active_trx_exec_tasks{ 0 };
-      std::vector<std::future<bool>>  _ro_trx_exec_tasks_fut;
+      ro_trx_queue_t                  _ro_exhausted_trx_queue;
```
greg7mdp, line 475: Why the name change? I don't see why `exhausted`?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2829,98 +2856,126 @@ void producer_plugin_impl::start_write_window() {
...
+   // 4. A transaction execution is exhausted (means end of read window)
+   bool exhausted_trx_queue_empty = false;
+   while ( fc::time_point::now() < _ro_window_deadline && !_received_block && !_ro_exiting_read_window ) {
+      if (!exhausted_trx_queue_empty) {
+         ro_trx_t t;
+         exhausted_trx_queue_empty = !_ro_exhausted_trx_queue.pop_front(t);
+         if (!exhausted_trx_queue_empty) {
+            auto retry = process_read_only_transaction( t.trx, t.next );
+            if( retry ) {
+               _ro_exiting_read_window = true;
```
greg7mdp, line 2928: Don't we need to push the transaction back on the queue?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2829,98 +2856,126 @@ void producer_plugin_impl::start_write_window() {
...
-         // Do not schedule new execution
-         break;
+   // 2. net_plugin receives a block
+   // 3. No more transactions in the exhausted read-only queue and no read-only tasks to execute
+   // 4. A transaction execution is exhausted (means end of read window)
+   bool exhausted_trx_queue_empty = false;
+   while ( fc::time_point::now() < _ro_window_deadline && !_received_block && !_ro_exiting_read_window ) {
+      if (!exhausted_trx_queue_empty) {
+         ro_trx_t t;
+         exhausted_trx_queue_empty = !_ro_exhausted_trx_queue.pop_front(t);
```
greg7mdp, line 2924: It seems that we aren't pushing ro transactions to the `_ro_exhausted_trx_queue` queue anymore (so we won't process them in parallel).

plugins/producer_plugin/producer_plugin.cpp
heifner: Because now it only contains previously exhausted read-only transactions.

plugins/producer_plugin/producer_plugin.cpp
heifner: They are pushed into the read_only task queue and run in parallel out of there along with other read_only tasks.

greg7mdp: so why do we even need `_ro_exhausted_trx_queue`?

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: What's an `exhausted read-only transactions`? One where `start_recover_keys` took so long it didn't finish within the read window?
If the read window is very small, couldn't that keep happening?

plugins/producer_plugin/producer_plugin.cpp
heifner: I considered just removing and re-posting to the read-only queue. 

However, we currently use the unapplied transaction queue for non-read-only trxs instead of re-posting. That keeps us from processing them more than once per block. My understanding is the original design it was decided to keep them out of the unapplied trx queue and have their own queue. So to mirror that behavior I kept the exhausted read-only queue.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2801,26 +2825,29 @@ void producer_plugin_impl::switch_to_write_window() {
...
+// Called from only one read_only thread & called from app thread, but not concurrently
 void producer_plugin_impl::start_write_window() {
    chain::controller& chain = chain_plug->chain();
 
    app().executor().set_to_write_window();
    chain.unset_db_read_only_mode();
+   _ro_in_read_only_mode = false;
    _idle_trx_time = fc::time_point::now();
 
+   _ro_window_deadline = _idle_trx_time + _ro_write_window_time_us; // not allowed on block producers, so no need to limit to block deadline
```
linh2931, line 2845: Will multiple threads write on `_ro_window_deadline` now that this function can be called from main and a read-only thread.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2953,23 +3009,24 @@ bool producer_plugin_impl::push_read_only_transaction(
...
-      auto window_deadline = std::min( start + remaining_time_in_read_window_us, block_deadline );
+      // when executing on the main thread while in the write window, need to switch db mode to read only
+      // _ro_in_read_only_mode can only be false if running on main thread as it is only modified from the main thread
+      const bool switch_read_mode = !_ro_in_read_only_mode;
+      auto db_read_only_mode_guard = fc::make_scoped_exit([&]{
+         if( switch_read_mode )
+            chain.unset_db_read_only_mode();
+      });
+      if( switch_read_mode )
+         chain.set_db_read_only_mode();
```
linh2931, line 3026: Now those 3 statements will be always executed, even when read-only threads are enabled.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1126,25 +1154,26 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
          }
       }
 #endif
+   }
```
greg7mdp, line 1157: this `}` should not have been moved up I think.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1090,11 +1114,15 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
       }
    }
 
-   my->_ro_thread_pool_size = options.at( "read-only-threads" ).as<uint16_t>();
+   if ( options.count( "read-only-threads" ) ) {
+      my->_ro_thread_pool_size = options.at( "read-only-threads" ).as<uint16_t>();
```
linh2931, line 1118: Can someone type `--read-only-threads 0` on producer node, but we don't allow?

plugins/producer_plugin/producer_plugin.cpp
heifner: exhausted read-only transaction is one that didn't execute in the window. We require the read-only window to be greater than the max-transaction time, so if executed first in the read-window it should either finish or hit max-transaction-time which does not re-queue.

plugins/producer_plugin/producer_plugin.cpp
heifner: Yeah, it should be once per read-window. So the queue keeps us from doing that. Without the queue (or some mechanism) we would potentially try it again in the same read-window.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2829,98 +2856,126 @@ void producer_plugin_impl::start_write_window() {
...
+            }
+         }
+      }
+   }
+}
+
 // Called from a read only trx thread. Run in parallel with app and other read only trx threads
 // Return whether the trx needs to be retried in next read window
-bool producer_plugin_impl::process_read_only_transaction(const packed_transaction_ptr& trx, const next_function<transaction_trace_ptr>& next, const fc::time_point& read_window_start_time) {
+bool producer_plugin_impl::process_read_only_transaction(const packed_transaction_ptr& trx, const next_function<transaction_trace_ptr>& next) {
```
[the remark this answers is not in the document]
heifner: `next` was changed to a `const&` because it is normally not moved/copied, only used.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: Not really. Once in the `_ro_exhausted_trx_queue`, they could be tried multiple times within a read window, right?

plugins/producer_plugin/producer_plugin.cpp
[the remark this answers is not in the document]
heifner: It could take `ro_trx` by value. Would require creating a `ro_trx` in `on_incoming_transaction_async` but we are creating affectively the same thing. I would lean toward not doing that because I feel `ro_trx_t` is more of an implementation type and hence would prefer to keep it out of function signatures. But I don't feel strongly either way. I'll switch to by value if you want?

plugins/producer_plugin/producer_plugin.cpp
heifner: That was on purpose. I needed the setting of _ro_* outside of `if ( my->_ro_thread_pool_size > 0 )` because they are used even when there is no active read-only thread pool.

README.md
```diff
@@ -72,7 +72,7 @@ git clone --recursive https://github.com/AntelopeIO/leap.git
 git clone --recursive git@github.com:AntelopeIO/leap.git
 ```
 
-> ℹ️ **HTTPS vs. SSH Clone** ℹ️
+> ℹ️ **HTTPS vs. SSH Clone** ℹ️
```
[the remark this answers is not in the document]
heifner: It pulled in Zach's PR. I'm not sure why, but thought they looked like harmless changes.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2011,6 +2040,8 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
             process_scheduled_and_incoming_trxs( scheduled_trx_deadline, incoming_itr );
          }
 
+         process_read_only_transactions( preprocess_deadline );
```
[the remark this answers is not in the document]
heifner: I don't understand. Line #s have shifted since you wrote this comment, but I do not see us canceling processing ro transactions when building a block.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1090,11 +1114,15 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
       }
    }
 
-   my->_ro_thread_pool_size = options.at( "read-only-threads" ).as<uint16_t>();
+   if ( options.count( "read-only-threads" ) ) {
+      my->_ro_thread_pool_size = options.at( "read-only-threads" ).as<uint16_t>();
+      EOS_ASSERT( my->_producers.empty(), plugin_config_exception, "--read-only-threads not allowed on producer node" );
+   } else if ( my->_producers.empty() ) {
+      // default to 3 threads for non producer nodes if not specified
+      my->_ro_thread_pool_size = 3;
```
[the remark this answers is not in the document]
heifner: No, better to find out right away in my opinion by turning it on by default. We can always tell people to disable it as a work-around if they hit some kind of issue.

plugins/producer_plugin/producer_plugin.cpp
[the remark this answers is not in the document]
greg7mdp: Oh, my bad, I read the check incorrectly:

image

I thought we were returning when `chain.is_building_block()` (missed the `!`).

But I'm a bit puzzled. I thought ro transactions did not go into blocks?

plugins/producer_plugin/producer_plugin.cpp
heifner: No, once we hit exhausted on a read-only we exit the read-window.

plugins/producer_plugin/producer_plugin.cpp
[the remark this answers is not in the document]
heifner: ro trx do not go into blocks. However, we require a pending block for the execution of read-only transactions. This provides host function capability such as `get_block_num()`

plugins/producer_plugin/producer_plugin.cpp
heifner: Fixed.

plugins/producer_plugin/producer_plugin.cpp
heifner: It is only ever written to at start of write window or start of read window which we already depend on only be executed on one thread at a time.

plugins/producer_plugin/producer_plugin.cpp
heifner: Yes. Do you think that is a problem?

tests/read_only_trx_test.py
```diff
@@ -21,6 +21,7 @@
```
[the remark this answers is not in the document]
heifner: Added more tests.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1164,7 +1109,20 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
       }
    }
 
-   my->_ro_thread_pool_size = options.at( "read-only-threads" ).as<uint16_t>();
+   if ( options.count( "read-only-threads" ) ) {
+      my->_ro_thread_pool_size = options.at( "read-only-threads" ).as<uint16_t>();
+      EOS_ASSERT( test_mode_ || my->_ro_thread_pool_size == 0 || my->_producers.empty(), plugin_config_exception, "--read-only-threads not allowed on producer node" );
```
greg7mdp, line 1114: Move this line after 1125, and 
```
      if (!test_mode_)
         EOS_ASSERT( my->_producers.empty(), plugin_config_exception, "--read-only-threads not allowed on producer node" );
```
can be removed

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2087,6 +2046,8 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
             process_scheduled_and_incoming_trxs( scheduled_trx_deadline, incoming_itr );
          }
 
+         process_read_only_transactions( preprocess_deadline );
```
greg7mdp, line 2049: Should this be renamed something like `repost_exhausted_transactions`?

heifner: Yes, and should honor the deadline.

libraries/custom_appbase/include/eosio/chain/application.hpp
```diff
@@ -1,87 +1,98 @@
...
+            read_write_queue_.execute_highest();
+         } else if( !read_only_queue_.empty() ) {
+            read_only_queue_.execute_highest();
          }
-         return !read_only_trx_safe_queue_.empty() || !general_queue_.empty();
+         return !read_only_queue_.empty() || !read_write_queue_.empty();
       } else {
-         return read_only_trx_safe_queue_.execute_highest();
+         // When in read window, multiple threads including main app thread are accessing two_queue_executor, locking required
+         return read_only_queue_.execute_highest_locked(false);
```
greg7mdp, line 66: parameter should be `true` I think.

heifner: This is called by the main thread. The main thread is in `app` `exec()` which does it own blocking.

libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
```diff
@@ -0,0 +1,233 @@
...
+      assert(num_threads > 0 && num_waiting_ == 0);
+      lock_enabled_ = true;
+      max_waiting_ = num_threads;
+      should_exit_ = std::move(should_exit);
+      exiting_blocking_ = false;
+   }
+
+   void disable_locking() {
+      lock_enabled_ = false;
+      should_exit_ = [](){ assert(false); return true; }; // should not be called when locking is disabled
```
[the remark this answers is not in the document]
greg7mdp: Sure, that makes sense. This should never happen, but I guess it is better to have the abort disabled in release mode.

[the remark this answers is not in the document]
heifner: A slightly more efficient approach is we just remove the line. `should_exit_` is not called and if it was it would most likely return true because of deadline being exceeded.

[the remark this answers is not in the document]
greg7mdp: Actually I prefer it as is, because if the `assert()` is triggered it will alert us that something is not working as expected.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2909,147 +2863,143 @@ void producer_plugin_impl::start_write_window() {
...
+   _ro_in_read_only_mode = true;
    _ro_read_window_start_time = fc::time_point::now();
    _ro_all_threads_exec_time_us = 0;
 
    // start a read-only transaction execution task in each thread in the thread pool
-   _ro_num_active_trx_exec_tasks = _ro_thread_pool_size;
-   auto start_time = fc::time_point::now();
-   _ro_trx_queue.set_exit_criteria(_ro_thread_pool_size, &_received_block, pending_block_num, start_time + _ro_read_window_effective_time_us);
+   _ro_num_active_exec_tasks = _ro_thread_pool_size;
    for (auto i = 0; i < _ro_thread_pool_size; ++i ) {
```
greg7mdp, line 2894: Maybe add `_ro_exec_tasks_fut.resize(0)` before this line, and we won't need the two `self->_ro_exec_tasks_fut.clear();` in the lambda of the async_wait.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2909,147 +2863,143 @@ void producer_plugin_impl::start_write_window() {
...
+         // will be executed from the main app thread because all read-only threads are idle now
          self->switch_to_write_window();
       } );
+      // last thread post any exhausted back into read_only queue with slightly higher priority (low+1) so they are executed first
+      ro_trx_t t;
+      while( _ro_exhausted_trx_queue.pop_front(t) ) {
+         app().executor().post(priority::low+1, exec_queue::read_only, [this, trx{std::move(t.trx)}, next{std::move(t.next)}]() mutable {
+            push_read_only_transaction( std::move(trx), std::move(next) );
+         } );
+      }
```
greg7mdp, line 2949:
```suggestion
      repost_exhausted_transactions();
```

note: remove unused `deadline` parameter in `repost_exhausted_transactions`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2909,147 +2863,143 @@ void producer_plugin_impl::start_write_window() {
...
 
    try {
-      const auto block_deadline = calculate_block_deadline( chain.pending_block_time() );
       auto start = fc::time_point::now();
-      auto time_used_in_read_window_us = ( start - read_window_start_time );
-      if ( time_used_in_read_window_us >= _ro_read_window_time_us ) {
-         // already passed read-window deadline, try next round
+      chain::controller& chain = chain_plug->chain();
+      if ( !chain.is_building_block() ) {
+         _ro_exhausted_trx_queue.push_front( {std::move(trx), std::move(next)} );
```
greg7mdp, line 2978: Issue here I think! We could have a loop where `repost_exhausted_transactions` pops from `_ro_exhausted_trx_queue` and we push back here.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2909,147 +2863,143 @@ void producer_plugin_impl::start_write_window() {
...
-
-      auto retry = process_read_only_transaction( trx.trx, trx.next, start );
-      if ( retry ) {
-         _ro_trx_queue.push_front(std::move(trx));
-         // Do not schedule new execution
+   // 2. net_plugin receives a block
+   // 3. no read-only tasks to execute
+   while ( fc::time_point::now() < _ro_window_deadline && _received_block < pending_block_num ) {
+      bool more = app().executor().execute_highest_read_only(); // blocks until all read only threads are idle
+      if ( !more ) {
```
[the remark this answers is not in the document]
heifner: Having it inside the loop makes it much easier to set a breakpoint without having to step over `now()`. Plus I think it is easier to read.

plugins/producer_plugin/producer_plugin.cpp
heifner: I prefer the current `clear()` so the futures are destroyed earlier rather than later.

plugins/producer_plugin/producer_plugin.cpp
heifner: It should be used and it is now.

plugins/producer_plugin/producer_plugin.cpp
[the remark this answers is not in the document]
greg7mdp: Fair enough! Debugging convenience is important :-).

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2909,147 +2862,143 @@ void producer_plugin_impl::start_write_window() {
...
-         _ro_all_threads_exec_time_us += dur.count();
+      // when executing on the main thread while in the write window, need to switch db mode to read only
+      // _ro_in_read_only_mode can only be false if running on main thread as it is only modified from the main thread
+      auto db_read_only_mode_guard = fc::make_scoped_exit([&]{
+         if( !_ro_in_read_only_mode )
+            chain.unset_db_read_only_mode();
+      });
+      if( !_ro_in_read_only_mode ) {
+         chain.set_db_read_only_mode();
+         _ro_window_deadline = calculate_block_deadline( chain.pending_block_time() );
```
greg7mdp, line 2989: Aren't we overwriting a variable set also in start_write_window. Why don't we use a local variable instead of `_ro_window_deadline`?

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: Why are we checking the deadline in `repost_exhausted_transactions`? This moving from a queue to another one is pretty much instant if I understand correctly. It is not like we are processing anything.

heifner: It is just a safe guard in case somehow the queue gets extra large.  All `while` loops in start block call `should_interrupt_start_block()`. I've updated this to do the same.

plugins/producer_plugin/producer_plugin.cpp
heifner: Hmm. This is rather messy. `_ro_in_read_only_mode` is always `false` if `_ro_thread_pool_size` is `0`. That is what we really want to know. If not in a write window we need to use `block_deadline` instead.

greg7mdp: What about the case where `_ro_thread_pool_size` is not 0.

In `start_write_window`, we set `_ro_window_deadline`:

```c++
void producer_plugin_impl::start_write_window() {

   app().executor().set_to_write_window();
   chain.unset_db_read_only_mode();
   _ro_in_read_only_mode = false;
   _idle_trx_time = fc::time_point::now();

   _ro_window_deadline = _idle_trx_time + _ro_write_window_time_us; // not allowed on block producers, so no need to limit to block deadline
```

It seems strange to be that we overwrite it in `push_read_only_transaction`? I'm not understanding the logic here.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2909,147 +2862,143 @@ void producer_plugin_impl::start_write_window() {
...
-         auto dur = fc::time_point::now() - start;
-         _ro_all_threads_exec_time_us += dur.count();
+      // when executing on the main thread while in the write window, need to switch db mode to read only
+      // _ro_in_read_only_mode can only be false if running on main thread as it is only modified from the main thread
+      auto db_read_only_mode_guard = fc::make_scoped_exit([&]{
+         if( !_ro_in_read_only_mode )
+            chain.unset_db_read_only_mode();
+      });
+      if( !_ro_in_read_only_mode ) {
+         chain.set_db_read_only_mode();
```
greg7mdp, line 2990: Why do we need to set and unset the `db_read_only_mode` when we are not in read only mode (so executing on a single thread).

plugins/producer_plugin/producer_plugin.cpp
heifner: The current code does not overwrite for the `_ro_thread_pool_size == 0` case.

plugins/producer_plugin/producer_plugin.cpp
heifner: Because they are interleaved with non-read-only transactions. We set the db read only mode so we enforce read-only behavior for read-only transactions during the execution of the read-only transaction.

greg7mdp: I talked with Kevin and I understand now why we want to fail the transaction even in single thread mode (where it could execute successfully), because a user should have non readonly transactions submitted as readonly fail consistently.

plugins/producer_plugin/producer_plugin.cpp
heifner: `remove_expired_trxs` uses `pending_block_time` not `preprocess_deadline` for determination of which trxs to expire.
The `preprocess_deadline` is exclusively used for determining if too much time has passed in `start_block`.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: I don't really see a benefit to have the futures destroyed a little bit earlier, and it requires to have to check the logic to convince yourself that the `_ro_exec_tasks_fut` vector is empty before we start emplacing futures into it.
Also the resize(0) would have the additional benefit that the vector capacity is preserved, so that we don't reallocate memory every time. 

If you prefer to keep it as is, would you consider adding:

```
   assert(_ro_exec_tasks_fut.empty());
   _ro_exec_tasks_fut.reserve(_ro_thread_pool_size);
```

greg7mdp: Oh, my bad, I see that `clear()` leaves the vector capacity unchanged. For some reason I thought otherwise. So we're good as is (though I think the `assert()` would be nice :-)!

heifner: `clear()` is not required to change capacity and, in my experience, most (maybe all) do not change capacity. `std::vector::resize(0)` c++ spec does not say anything about capacity. Although cppreference does say it maintains capacity, it makes that determination because iterators are maintained. But maintaining iterators does not matter if resizing to 0. So I think a compliant implementation could reduce capacity for `resize(0)`. It would have to special case for the value of `0`, but there might be benefit. Regardless, in practice I think `clear()` and `resize(0)` do the same thing for most/all implementations.

We could add the `_ro_exec_tasks_fut.resize(0)` while maintaining the existing `_ro_exec_tasks_fut.clear()` calls. Ok with that?

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: Oh, yes, got it, my bad, thanks!

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: > We could add the `_ro_exec_tasks_fut.resize(0)` while maintaining the existing `_ro_exec_tasks_fut.clear()` calls. Ok with that?

Sure. or the `assert(_ro_exec_tasks_fut.empty());`. I'm fine with either.

greg7mdp (approved):
Awesome work!

plugins/producer_plugin/producer_plugin.cpp
heifner: Done.

PR #903, opened 2023-03-27
Change from 500 to 400 http error code for exceptions during http processing

### LINKED ISSUE #591: Fix get_account fail case, RPC endpoint for is_valid_account
Use case:

Any app creating a user account must check if the account exists. 

Current solution:
Calling get_account - get_account is relatively expensive when it succeeds but assumes that the account passed in is valid.

get_account does not handle non-existing account well: It crashes with a 500 internal server error, which might be a bug and exploitable by malicious actors. 

Fixing get_account is likely also a good idea, so it fails gracefully on a non-existing account

This is a fail case trace:

```
❯ cleosm --print-request --print-response get account nnnnn       
REQUEST:
---------------------
POST /v1/chain/get_account HTTP/1.1
Host: node.hypha.earth
content-length: 29
Accept: */*
Connection: close

{
  "account_name": "nnnnn"
}
---------------------
RESPONSE:
---------------------
{
  "code": 500,
  "message": "Internal Service Error",
  "error": {
    "code": 0,
    "name": "exception",
    "what": "unspecified",
    "details": [{
        "message": "unknown key (boost::tuples::tuple<bool, eosio::chain::name, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type>): (0 nnnnn)",
        "file": "http_plugin.cpp",
        "line_number": 936,
        "method": "handle_exception"
      }
    ]
  }
}
---------------------
error 2022-12-22T03:39:42.285 thread-0  main.cpp:4371                 operator()           ] Failed with error: unspecified (0)
unknown key (boost::tuples::tuple<bool, eosio::chain::name, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type, boost::tuples::null_type>): (0 nnnnn)

```

heifner:
FC and STD Exceptions during http processing is almost always because a resource can't be found. Modify reporting of exceptions from current `500` to `400`. An exception doesn't normally mean there is something wrong with `nodeos`.

See https://moz.com/learn/seo/http-status-codes
4xxs – Client errors: Page not found. The site or page couldn’t be reached. (The request was made, but the page isn’t valid — this is an error on the website’s side of the conversation and often appears when a page doesn’t exist on the site.)

HTTP Status Code 500 - Internal Server Error
Instead of the problem being with pages missing or not found, this status code indicates a problem with the server. A 500 is a classic server error and will affect access to your site. Human visitors and bots alike will be lost, and your link equity will go nowhere fast. Search engines prefer sites that are well maintained, so you’ll want to investigate these status codes and get these fixed as soon as you encounter them.

Resolves #591

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -2374,6 +2374,7 @@ read_only::get_raw_abi_results read_only::get_raw_abi( const get_raw_abi_params&
 }
 
 read_only::get_account_results read_only::get_account( const get_account_params& params, const fc::time_point& deadline )const {
+   try {
    get_account_results result;
    result.account_name = params.account_name;
 
@@ -2541,6 +2542,7 @@ read_only::get_account_results read_only::get_account( const get_account_params&
       }
    }
    return result;
+   } EOS_RETHROW_EXCEPTIONS(chain::account_query_exception, "unable to retrieve account info")
 }
 
 read_only::get_required_keys_result read_only::get_required_keys( const get_required_keys_params& params, const fc::time_point& deadline )const {
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -403,13 +403,13 @@ namespace eosio {
             fc_elog( logger(), "Unable to parse arguments to ${api}.${call}", ("api", api_name)( "call", call_name ) );
             fc_dlog( logger(), "Bad arguments: ${args}", ("args", body) );
          } catch (fc::exception& e) {
-            error_results results{500, "Internal Service Error", error_results::error_info(e, verbose_http_errors)};
-            cb( 500, fc::time_point::maximum(), fc::variant( results ));
+            error_results results{400, "Exception", error_results::error_info(e, verbose_http_errors)};
+            cb( 400, fc::time_point::maximum(), fc::variant( results ));
             fc_dlog( logger(), "Exception while processing ${api}.${call}: ${e}",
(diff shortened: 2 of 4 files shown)


linh2931 (approved):

spoonincode (changes requested):
This is changing behavior that has existed since 1.0. It's something we've always discussed changing, but why now in the middle of a RC?

spoonincode:
This seems to be a really broad change to the API, if I'm understanding right. For `read_only::get_account()` you added
https://github.com/AntelopeIO/leap/blob/32daed0840801e6a72b55159b732e03f42dc83d8/plugins/chain_plugin/chain_plugin.cpp#L2545
but then changed _any_ `fc::exception` or `std::exception` (besides `unknown_block_exception`, `invalid_http_request`, `unsatisfied_authorization`, `tx_duplicate`, or `eof_exception`) to be 400 instead of 500
https://github.com/AntelopeIO/leap/blob/32daed0840801e6a72b55159b732e03f42dc83d8/plugins/http_plugin/http_plugin.cpp#L405-L415
How come `account_query_exception` wasn't specialized as a 400 or 404 above instead? Wouldn't that isolate the 500->400 change to just account lookup, making the fallout from this change a little more predictable?

heifner:
Reworked via https://github.com/AntelopeIO/leap/pull/912

PR #913, opened 2023-03-28
Replace needless std::map with std::vector, and std::move in capture

greg7mdp:
small optimizations.

### DIFF
--- a/plugins/http_plugin/include/eosio/http_plugin/macros.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/macros.hpp
@@ -15,7 +15,7 @@ struct async_result_visitor : public fc::visitor<fc::variant> {
          auto params = parse_params<api_namespace::call_name ## _params, params_type>(body);\
          FC_CHECK_DEADLINE(deadline);\
          api_handle.call_name( std::move(params), \
-            [cb, body](const std::variant<fc::exception_ptr, call_result>& result){\
+               [cb=std::move(cb), body=std::move(body)](const std::variant<fc::exception_ptr, call_result>& result){ \
                if (std::holds_alternative<fc::exception_ptr>(result)) {\
                   try {\
                      std::get<fc::exception_ptr>(result)->dynamic_rethrow_exception();\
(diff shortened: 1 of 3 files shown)


plugins/http_plugin/include/eosio/http_plugin/macros.hpp
```diff
@@ -15,7 +15,7 @@ struct async_result_visitor : public fc::visitor<fc::variant> {
          auto params = parse_params<api_namespace::call_name ## _params, params_type>(body);\
          FC_CHECK_DEADLINE(deadline);\
          api_handle.call_name( std::move(params), \
-            [cb, body](const std::variant<fc::exception_ptr, call_result>& result){\
+               [cb=std::move(cb), body=std::move(body)](const std::variant<fc::exception_ptr, call_result>& result){ \
```
heifner, line 18: `cb` and `body` are used in the catch.

plugins/producer_api_plugin/producer_api_plugin.cpp
```diff
@@ -43,7 +43,7 @@ struct async_result_visitor : public fc::visitor<fc::variant> {
 {std::string("/v1/" #api_name "/" #call_name), \
    [&api_handle](string&&, string&& body, url_response_callback&& cb) mutable { \
       if (body.empty()) body = "{}"; \
-      auto next = [cb, body](const std::variant<fc::exception_ptr, call_result>& result){\
+      auto next = [cb=std::move(cb), body=std::move(body)](const std::variant<fc::exception_ptr, call_result>& result){ \
```
heifner, line 46: `body` is used by `INVOKE`.

heifner (changes requested):

linh2931 (approved):

plugins/http_plugin/include/eosio/http_plugin/macros.hpp
greg7mdp: Hum, my understanding is that the lambda we construct (into which I mode `cb` and `body`) is passed to the API as `next`, and if the API fails for any reason it invokes the `next()` function and doesn't rethrow. That is why we check for exceptions inside the lambda.

So either we construct the lambda, and we pass it to the API, or there was an exception which happened *before* the API was called in which case the bottom catc clause is used.

plugins/producer_api_plugin/producer_api_plugin.cpp
greg7mdp: same here.

heifner (approved):

PR #914, opened 2023-03-28
Added leap-util subcommand to check for a clean shutdown

766C6164:
This PR adds command line option to leap-util to indicate if the last shutdown was clean or not. This would allow someone to better control logic about perhaps automatically starting from snapshot if the state database was crashed.

### DIFF
--- a/programs/leap-util/actions/chain.cpp
+++ b/programs/leap-util/actions/chain.cpp
@@ -47,6 +47,16 @@ void chain_actions::setup(CLI::App& app) {
       // properly return err code in main
       if(rc) throw(CLI::RuntimeError(rc));
    });
+
+  auto* sstate =  sub->add_subcommand("last-shutdown-state", "indicate whether last shutdown was clean or not");
+  sstate->add_option("--state-dir,-o", opt->sstate_state_dir, "The location of the state directory (absolute path or relative to the current directory)")->capture_default_str();
+
+  sstate->callback([&]() {
+      int rc = run_subcommand_sstate();
+      // properly return err code in main
+      if(rc) throw(CLI::RuntimeError(rc));
+   });
+
 }
 
 int chain_actions::run_subcommand_build() {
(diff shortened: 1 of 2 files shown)


programs/leap-util/actions/chain.cpp
```diff
@@ -62,5 +72,47 @@ int chain_actions::run_subcommand_build() {
...
+   if(hs.fail()) {
+      std::cerr << "Unable to read database status: file invalid or corrupt" << shared_mem_path <<  std::endl;
+      return -1;
+   }
+
+   chainbase::db_header* dbheader = reinterpret_cast<chainbase::db_header*>(header);
+   if(dbheader->dirty) {
+      std::cout << "Database dirty flag is set, shutdown was not clean" << std::endl;
+      return -1;
+   }
```
spoonincode, line 114: Should check that dbheader's `header_id` is expected value, as it's possible the format will change in the future

heifner:
Can you add to the PR description details about the new option and an example of the output.

programs/leap-util/actions/chain.cpp
766C6164: thank you, done

programs/leap-util/actions/chain.cpp
```diff
@@ -47,6 +47,16 @@ void chain_actions::setup(CLI::App& app) {
       // properly return err code in main
       if(rc) throw(CLI::RuntimeError(rc));
    });
+
+  auto* sstate =  sub->add_subcommand("last-shutdown-state", "indicate whether last shutdown was clean or not");
+  sstate->add_option("--state-dir,-o", opt->sstate_state_dir, "The location of the state directory (absolute path or relative to the current directory)")->capture_default_str();
```
arhag, line 52: Drop the `-o` short option name and move this option to the parent command `chain-state`.

766C6164: Thank you, done!

arhag (approved):

Issue #900, opened 2023-03-25
4.0.0-rc2: gelf logging is busted when using dns pointing to both IPv4 and IPv6 addresses


matthewdarwin:
Upgrading a nodeos 3.2.2 node to 4.0.0-rc2 and enabling the prometheus plugin, the gelf logging shows now output with 
```
$ tcpdump udp port 12201
```

Probably prometheus has no impact, but that was the only config change made.

logging.json was unchanged, which in part has this:

```
    {
      "name": "net",
      "type": "gelf",
      "args": {
        "endpoint": "prod-server:12201",
        "host": "jungle4-bp203",
        "_operator": "eosnationftw",
        "_network": "jungle4"
      },
      "enabled": true
    },
    {
      "name": "test",
      "type": "gelf",
      "args": {
        "endpoint": "test-server:12201",
        "host": "jungle4-bp203",
        "_operator": "eosnationftw",
        "_network": "jungle4"
      },
      "enabled": true
    }

...

    {
      "name": "transaction_failure_tracing",
      "level": "debug",
      "enabled": true,
      "additivity": false,
      "appenders": [
        "net",
        "test"
      ]
    },
    {
      "name": "transaction_success_tracing",
      "level": "debug",
      "enabled": true,
      "additivity": false,
      "appenders": [
        "net",
        "test"
      ]
    }
```

matthewdarwin:
Works with IP address, but not DNS which resolves to both IPv4 and IPv6 address.

heifner:
Need to remove the use of libfc udp_socket which is hard coded to use IPv4 only. `gelf_appender` should use boost directly.

spoonincode:
yes that clean up should have been done, but worst case to resolve this issue quickly in 4.0 can simply change the resolver to v4 only.

PR #917, opened 2023-03-28
[4.0] fix gelf DNS resolves to ipv6 address

huangminghuang:
The PR

- fixes the issue when the DNS resolve to both ipv4 and ipv6 address,
- removes fc udp_socket and resolver,
- eliminate unnecessary buffer copy for sending gelf message.

resolve https://github.com/AntelopeIO/leap/issues/900


spoonincode:
I believe there are a couple subtle changes to how the UDP socket is handled: it's now a blocking socket, and it's now a "connected" socket. Have these two changes been studied carefully to ensure no adverse behavior?

My worry with the former is that it may be possible to concoct scenarios where the thread never completes thus preventing nodeos from shutting down. Consider a network interface that is "stuck" (such as being commanded to halt xmit via flow control): the socket buffer may never clear. But possibly a UDP socket buffer never blocks?

My worry with the latter is that errors on the socket may be more "sticky". For example, would an ICMP Port Unreachable from the destination somehow cause the socket to enter a failure state it would have previously ignored when it was not a "connected" UDP socket? It has been a long time since I've studied the socket rules here.

huangminghuang:
First of all, the GELF code has always been using blocking socket. I merely change the sending thread from main thread to anew thread in my previous PR. Secondly, as far as I understand, a UDP send() can only block when the OS trying to send ARP request in order to get the hardware address of the next hop. By using connected socket, it would allow the OS to save a route lookup result. Therefore, it is recommended to use connected socket for outgoing traffic by https://blog.cloudflare.com/everything-you-ever-wanted-to-know-about-udp-sockets-but-were-afraid-to-ask-part-1/. I am not an expert on this topic, I just follow the recommendation. If you want me to change it back to unconnected socket, I can do that.

spoonincode:
> GELF code has always been using blocking socket

The prior code was
https://github.com/AntelopeIO/leap/blob/fca96e56ebc9f79a7d94e3fcc1cf33692fb443fc/libraries/libfc/src/log/gelf_appender.cpp#L133-L137
where `open()`,
https://github.com/AntelopeIO/leap/blob/fca96e56ebc9f79a7d94e3fcc1cf33692fb443fc/libraries/libfc/src/network/udp_socket.cpp#L85-L89
def looks like non blocking to me. And,
https://github.com/AntelopeIO/leap/blob/fca96e56ebc9f79a7d94e3fcc1cf33692fb443fc/libraries/libfc/src/network/udp_socket.cpp#L38-L47
expressly handled the case where the socket buffer was full and returned an error due to it being in non blocking mode.

Since we're refactoring code in the middle of RC stage the most conservative approach is to match previous behavior unless we have full confidence we're not introducing some other potential defect or undesirable behavior.

Another change I noticed is that previously any error on the send would be silently swallowed, where as now every single error results in an `fprintf(stderr)`. I'm not sure if flooding stderr would be a good behavior or not here; it's somewhat dependent on how likely the socket is to return an error on the send I guess.

huangminghuang:
My bad. I didn't even looked into the implementation of `udp_socket` class and assume it worked like regular socket object.

spoonincode (approved):

heifner (approved):

PR #918, opened 2023-03-29
Peer discovery

bxliu:
Here is the [link to RFP #11](https://github.com/eosnetworkfoundation/Coalition-RFPs/blob/main/2022%2011%2007%20RFP%20-%20P2P%20Peer%20Discovery.pdf) for this PR.

The mechanism is as follows, some need discuss:

Added new connection types: host:port:type, where the type can be trx/blk/peer/all. If the type is not specified, it defaults to 'trx' and 'blk' (compatible with historical configurations). If set to 'all' or 'peer', the current connection will participate in peer exchange, otherwise it will not. 
TR06/07 have been resolved by controlling the type setting (perhaps an additional switch is needed):
TR06: The solution sets all new nodes as publicly discoverable peers by default.
TR07: The solution allows nodes to opt out of being discoverable.

Added an address_manager class to manage addresses in a thread-safe manner, replacing the previous address management mechanism. Peer addresses from the configuration file will be marked as 'manual', but there are currently no corresponding processing requirements for such manual addresses. Perhaps an additional switch could be added to retrieve only manual addresses (the collection of addresses through last_active would be more effective):
TR08: The solution allows nodes to manually configure a list of peer nodes in addition to or to override automated peer discovery.

Added a message communication mechanism for synchronization messages, including address_request_message for sending address requests, with request types including pull, push, latest_active, and manual. The latest_active request type is not yet supported and the address transfer limit needs to be added.
Requests are executed every 100 handshakes, starting from the first handshake. Once p2p-peer-address is set (in compliance with the connection type requirements of all or peer), a list of peer nodes will be automatically retrieved, resolving TR04: The solution will automatically retrieve a list of peer nodes once the network has been selected.

Added a min_peers configuration, which is checked by the connection_monitor. If the number of connections does not meet the min_peers requirement, new connections will be added from the address manager. However, the mechanism for starting up when the number of connected nodes is insufficient but has been unable to meet the required number needs to be improved. The active connection nodes need to be obtained through last_active. 

TR05(The solution will automatically request a new list of peers if none from a previous list are available.) needs further discussion as the address nodes may come from different boot_nodes and invalid connections may actually be a state (which may become valid after a period of time). Therefore, when the minimum connection number cannot be met, the solution is to try to create more new connections.

TR09( The solution will maintain a configurable minimum number of connected peer nodes.
) has been partially resolved, and the mechanism for starting up when the number of updated nodes is insufficient but has been unable to meet the required number needs to be improved.


plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
```diff
@@ -135,15 +135,141 @@ namespace eosio {
...
+            if (type_str == "trx") {
+                address_type = trx;
+            } else if (type_str == "blk") {
+                address_type = blk;
+            } else if (type_str == "peer") {
+                address_type = peer;
+            } else if (type_str.empty()) {
+                address_type = both;
+            } else
+                address_type = all;
```
smlu, line 219: I suggest making separate inline utility function to convert string to address type. It will make the code leaner and more readable, plus the function can be reused elsewhere if needed.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
```diff
@@ -135,15 +135,141 @@ namespace eosio {
       uint32_t end_block{0};
    };
 
-   using net_message = std::variant<handshake_message,
+    enum address_type_enum {
+        blk      = 1 << 0, // 1
+        trx      = 1 << 1, // 2
+        peer     = 1 << 2, // 4
+        both     = 0x3,
+        all      = 0x7
```
smlu, line 143: I suggest using enumerators instead of fixed values. It makes it more readable.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
bxliu: // determine address type
    inline address_type_enum str_to_address_type(const std::string& address_type_str) {
        static const std::unordered_map<std::string, address_type_enum> address_type_map = {
                {"blk", blk},
                {"trx", trx},
                {"peer", peer},
                {"", both},
                {"all", all},
        };
        const auto it = address_type_map.find(address_type_str);
        if (it != address_type_map.end()) {
            return it->second;
        }
        // if address type not match, return all by default
        return all;
    }

plugins/net_plugin/net_plugin.cpp
```diff
@@ -112,6 +112,145 @@ namespace eosio {
...
+    }
+
+    void address_manager::add_address_str(const std::string& address, bool is_manual ) {
+        peer_address addr = peer_address::from_str(address, is_manual);
+        // same address with different configurations is ignored
+        add_address(addr);
+    }
+
+    void address_manager::add_addresses(const std::vector<std::string>& new_addresses, bool is_manual ) {
+        std::unique_lock<std::mutex> lock(addresses_mutex);
```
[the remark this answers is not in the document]
bxliu: In my opinion, because we will add timeout or connectivity checks to the address pool in the future, frequent modifications to the addresses vector by multiple threads may cause issues if another thread deletes an element from the vector while an iterator is being incremented. this modification may result in undefined behavior and locking the entire loop may be necessary.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
```diff
@@ -135,15 +135,141 @@ namespace eosio {
...
+            return address;
+        }
+
+    };
+
+    enum request_type_enum {
+        push,   //push and pull
+        pull,
+        latest_active,
+        manual, //only request manual addresses
```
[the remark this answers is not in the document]
bxliu: enum request_type_enum {
        push = 0,           // push and pull
        pull = 1,
        latest_active = 2,
        manual = 3,         // only request manual addresses
    };

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
smlu: LGTM

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
smlu: If you put mutex lock just before if statement: `if (std::find(addresses.begin(), addresses.end(), addr) == addresses.end())` should be fine since the `addresses` vector will be locked from further modification.

However, as previously mentioned, you'd have to simulate different situations and measure whether there is any real benefit to it. Since leap is designed to be highly performant, it's reasonable to assume that every unlocked mutex would result in improved throughput. Than again maybe it's not the case here.

systemzax:
Since you are adding new messages types to the protocol (address_request_message, address_sync_message), it would be a good idea to also increment and describe the protocol version number.

See here : 

https://github.com/bxliu/leap/blob/peer-discovery/plugins/net_plugin/net_plugin.cpp#L585

Then, when appropriate, you would verify the protocol version of a peer, and only send the new messages to peers with a protocol version equal or higher.

This way, the protocol remains backward compatible.

For instance, you could have a check here, and only send address request messages to peers that support the protocol version : https://github.com/bxliu/leap/blob/peer-discovery/plugins/net_plugin/net_plugin.cpp#L1363

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
bxliu: https://github.com/AntelopeIO/leap/commit/9b213992192e0c09c7f579ddff42a504f756af24

In the given scenario, with 1000 threads concurrently adding 1000 new addresses to the existing 30000 addresses, the first function (i.e., add_addresses) performs better. 

Further testing is needed for more scenarios and both add and remove operations are present.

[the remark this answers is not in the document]
bxliu: https://github.com/AntelopeIO/leap/pull/918/commits/3567cebddae04441f1dbf85d3e92258c918084bc


Since we are discussing performance, why use vector...

Code has been updated to use unordered_maps, which better meets the requirements for address management.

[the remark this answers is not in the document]
bxliu: vector add:
30000, 1000 threads add 1000 address:
image

unordered_maps: 
30000, 1000 threads add 10 address, 300 threads remove 10 address, first is better.
image

unordered_maps:
30000, 1000 threads add 1000 address, 300 threads remove 1000 address, result is complex.
image

image

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3464,15 +3823,31 @@ namespace eosio {
          }
          ++it;
       }
+      if(it == connections.end())
```
porkchop, line 3826: `it == connections.end()` seems guaranteed at this point no? I see no breaks that would short circuit the main while condition

porkchop (approved):
Some minor suggestions for you to consider. The PR looks great. Very easy to follow and the ideas makes sense

systemzax (approved):

greg7mdp:
Can you please add a link to the RFP in the PR description?

bxliu:
> Can you please add a link to the RFP in the PR description?

https://github.com/eosnetworkfoundation/Coalition-RFPs/blob/main/2022%2007%20RFP%20-%20P2P%20Improvements.pdf

heifner:
> > Can you please add a link to the RFP in the PR description?
> 
> https://github.com/eosnetworkfoundation/Coalition-RFPs/blob/main/2022%2007%20RFP%20-%20P2P%20Improvements.pdf

I don't think that is the correct document. That includes swarm download and rewrite of net_plugin with an explicit state machine among other things.  

Is it not? https://github.com/eosnetworkfoundation/Coalition-RFPs/blob/main/2022%2011%2007%20RFP%20-%20P2P%20Peer%20Discovery.pdf

bxliu:
> > > Can you please add a link to the RFP in the PR description?
> > 
> > 
> > https://github.com/eosnetworkfoundation/Coalition-RFPs/blob/main/2022%2007%20RFP%20-%20P2P%20Improvements.pdf
> 
> I don't think that is the correct document. That includes swarm download and rewrite of net_plugin with an explicit state machine among other things.
> 
> Is it not? https://github.com/eosnetworkfoundation/Coalition-RFPs/blob/main/2022%2011%2007%20RFP%20-%20P2P%20Peer%20Discovery.pdf

I'm sorry, that is the correct documentation link.

tests/plugin_http_api_test.py
```diff
@@ -791,7 +791,7 @@ def test_NetApi(self) :
         ret_json = self.nodeos.processUrllibRequest(resource, command, self.empty_content_dict, endpoint=endpoint)
         self.assertEqual(ret_json["code"], 400)
         self.assertEqual(ret_json["error"]["code"], 3200006)
-        payload = "localhost"
+        payload = "localhost:1234"
```
heifner, line 794: Instead of modifying the test, I think you need to modify the code to continue to handle the case of no port given.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -143,6 +144,170 @@ namespace eosio {
...
+        std::lock_guard<std::mutex> lock(addresses_mutex);
+        auto it = addresses.find(updated_address.to_key());
+        if (it != addresses.end()) {
+            it->second = updated_address;
+        }
+    }
+
+    std::unordered_set<std::string> address_manager::get_addresses() const {
+        std::lock_guard<std::mutex> lock(addresses_mutex);
+        std::unordered_set<std::string> result;
```
heifner, line 256: `result.reserve(addresses.size());`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1492,10 +1735,52 @@ namespace eosio {
...
+   void connection::enqueue_address_request(request_type_enum request_type) {
+       // only send request to peers that support the protocol
+       if(protocol_version >= proto_address_sync) {
+           peer_dlog(this, "Sending address request type ${type}", ("type", request_type_str(request_type)));
+           address_request_message request_msg;
+
+           request_msg.request_type = request_type;
+           // request push type will send addresses to other node
+           if (request_type == push) {
+               std::unordered_set<std::string> addresses = my_impl->address_master->get_addresses();
```
heifner, line 1768: if `addresses.size()` <= `my_impl->max_addresses_per_request` you can just `std::move(addresses)` into `request_msg.addresses`. Otherwise you can erase a distance to reduce the size of addresses and then move. Or better yet pass `my_impl->max_addresses_per_request` to `get_addresses()` and only return a max of that many.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2128,7 +2412,8 @@ namespace eosio {
    // called from c's connection strand
    void sync_manager::recv_handshake( const connection_ptr& c, const handshake_message& msg, uint32_t nblk_combined_latency ) {
 
-      if (!c->is_blocks_connection())
+      // since connection type changed, choose connection accept blocks
```
heifner, line 2415: I don't understand this comment.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -790,8 +983,22 @@ namespace eosio {
...
+      bool is_all_connection() const { return connection_type == all; }
+      bool is_both_connection() const { return connection_type == both; }
+
+      bool is_transactions_connection() const { return connection_type == transactions_only || is_both_connection() || is_all_connection(); }
+      bool is_blocks_connection() const { return connection_type == blocks_only || is_both_connection() || is_all_connection(); }
+      bool is_peers_connection() const { return connection_type == peers_only || is_all_connection(); }
+
+      bool is_no_transactions_connection() const { return !is_transactions_connection(); }
+      bool is_no_blocks_connection() const { return !is_blocks_connection(); }
+      bool is_no_peers_connection() const { return !is_peers_connection(); }
```
heifner, line 999: Not used, remove.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -790,8 +983,22 @@ namespace eosio {
...
+
+      bool is_all_connection() const { return connection_type == all; }
+      bool is_both_connection() const { return connection_type == both; }
+
+      bool is_transactions_connection() const { return connection_type == transactions_only || is_both_connection() || is_all_connection(); }
+      bool is_blocks_connection() const { return connection_type == blocks_only || is_both_connection() || is_all_connection(); }
+      bool is_peers_connection() const { return connection_type == peers_only || is_all_connection(); }
+
+      bool is_no_transactions_connection() const { return !is_transactions_connection(); }
+      bool is_no_blocks_connection() const { return !is_blocks_connection(); }
```
heifner, line 998: I would prefer using `!is_blocks_connection()` instead of adding `is_no_blocks_connection()`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -790,8 +983,22 @@ namespace eosio {
...
+      bool is_peers_only_connection() const { return connection_type == peers_only; }
+
+      bool is_all_connection() const { return connection_type == all; }
+      bool is_both_connection() const { return connection_type == both; }
+
+      bool is_transactions_connection() const { return connection_type == transactions_only || is_both_connection() || is_all_connection(); }
+      bool is_blocks_connection() const { return connection_type == blocks_only || is_both_connection() || is_all_connection(); }
+      bool is_peers_connection() const { return connection_type == peers_only || is_all_connection(); }
+
+      bool is_no_transactions_connection() const { return !is_transactions_connection(); }
```
heifner, line 997: I would prefer using `!is_transactions_connection()` instead of adding `is_no_transactions_connection()`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -790,8 +983,22 @@ namespace eosio {
       void set_connection_type( const string& peer_addr );
       bool is_transactions_only_connection()const { return connection_type == transactions_only; } // thread safe, atomic
       bool is_blocks_only_connection()const { return connection_type == blocks_only; }
-      bool is_transactions_connection() const { return connection_type != blocks_only; } // thread safe, atomic
-      bool is_blocks_connection() const { return connection_type != transactions_only; } // thread safe, atomic
+
+      //add more connection type
+      bool is_peers_only_connection() const { return connection_type == peers_only; }
```
heifner, line 988: I don't understand a peers only connection. You run a node that does nothing more than share out peers it is connected to? It does not forward blocks or transactions?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -321,12 +487,16 @@ namespace eosio {
...
    constexpr auto     def_max_clients = 25; // 0 for unlimited clients
+   constexpr auto     def_min_peers = 0;
    constexpr auto     def_max_nodes_per_host = 1;
    constexpr auto     def_conn_retry_wait = 30;
    constexpr auto     def_txn_expire_wait = std::chrono::seconds(3);
    constexpr auto     def_resp_expected_wait = std::chrono::seconds(5);
    constexpr auto     def_sync_fetch_span = 1000;
    constexpr auto     def_keepalive_interval = 10000;
+   constexpr auto     def_max_address_per_request = 10000;
+   constexpr auto     def_address_request_last_active_secs = std::chrono::seconds(600);
```
heifner, line 498: 10 minutes seems like a really long time for this. Seems like this should be more like 5 seconds.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3899,6 +4259,8 @@ namespace eosio {
       if( is_transactions_only_connection() ) hello.p2p_address += ":trx";
       // if we are not accepting transactions tell peer we are blocks only
       if( is_blocks_only_connection() || !my_impl->p2p_accept_transactions ) hello.p2p_address += ":blk";
+      if( is_peers_only_connection() ) hello.p2p_address += ":peer";
+      if( is_all_connection() ) hello.p2p_address += ":all";
```
heifner, line 4263: This logic does not look correct. If `!my_impl->p2p_accept_transactions` then `hello.p2p_address` already contains `:blk`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4045,8 +4418,10 @@ namespace eosio {
 
          std::vector<std::string> peers;
          if( options.count( "p2p-peer-address" )) {
-            peers = options.at( "p2p-peer-address" ).as<vector<string>>();
```
heifner, line 4048: This breaks `p2p-auto-bp-peer` below.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3539,6 +3843,62 @@ namespace eosio {
...
+               if(address != this->peer_address()) {
+                   addresses_msg.addresses.insert(address);
+                   count++;
+                   if(count >= my_impl->max_addresses_per_request) {
+                       break;
+                   }
+               }
+           }
+
+           enqueue(addresses_msg);
```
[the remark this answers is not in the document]
bxliu: https://github.com/AntelopeIO/leap/blob/bf6290ac4d9ae0ca93b952d30737bec09dc44181/plugins/net_plugin/net_plugin.cpp#L1764

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4437,11 +4827,15 @@ namespace eosio {
 
    // called from any thread
    void connections_manager::connection_monitor(const std::weak_ptr<connection>& from_connection) {
+      bool from_begin = false;
```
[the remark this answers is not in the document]
bxliu: The purpose of from_begin is to ensure that the connection_monitor has completed all checks without exceeding the timeout.

plugins/net_plugin/include/eosio/net_plugin/address_manager.hpp
```diff
@@ -0,0 +1,194 @@
...
+                address.host = host_str;
+                address.port = port_str;
+                address.address_type = str_to_address_type(type_str);
+                address.receive = fc::time_point::now();
+                address.last_active = fc::time_point::min();
+                address.manual = is_manual;
+                return address;
+            }
+            catch (const std::exception& e) {
+                throw std::invalid_argument(std::string("Invalid peer address string: ") + e.what());
```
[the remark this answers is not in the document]
bxliu: e.what() include input_address_str

plugins/net_plugin/include/eosio/net_plugin/address_manager.hpp
```diff
@@ -0,0 +1,194 @@
+#pragma once
```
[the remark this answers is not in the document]
bxliu: TR05(The solution will automatically request a new list of peers if none from a previous list are available.) needs further discussion as the address nodes may come from different boot_nodes and invalid connections may actually be a state (which may become valid after a period of time). Therefore, when the minimum connection number cannot be met, the solution is to try to create more new connections.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3931,14 +4293,17 @@ namespace eosio {
...
+           "  'all' means trx and blk and peer, no type means trx and blk "
            "  Examples:\n"
            "    p2p.eos.io:9876\n"
            "    p2p.trx.eos.io:9876:trx\n"
            "    p2p.blk.eos.io:9876:blk\n")
-         ( "p2p-max-nodes-per-host", bpo::value<int>()->default_value(def_max_nodes_per_host), "Maximum number of client nodes from any single IP address")
+         ( "p2p-max-nodes-per-host", bpo::value<uint32_t>()->default_value(def_max_nodes_per_host), "Maximum number of client nodes from any single IP address")
+         ( "p2p-max-addresses-per-request", bpo::value<uint32_t>()->default_value(def_max_address_per_request), "Maximum number of addresses in address request or sync message")
          ( "p2p-accept-transactions", bpo::value<bool>()->default_value(true), "Allow transactions received over p2p network to be evaluated and relayed if valid.")
+         ( "p2p-only-send-manual-addresses", bpo::value<bool>()->default_value(false), "Only send addresses from configuration instead from automated peer discovery")
```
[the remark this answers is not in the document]
bxliu: Although P2P forwarding can be restricted through the use of the 'type' and 'allowed_connections' parameters, in order to ensure clear configuration, the 'p2p-exchange-peers' setting is added with a default value of 'true'. if set to false, peers will remove 'peer' from connection_type, trx and blk still works.

plugins/net_plugin/net_plugin.cpp
bxliu: changed the connection type enum logic.

plugins/net_plugin/net_plugin.cpp
bxliu: We can set up a service node dedicated to P2P exchanges, especially within a large-scale network.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1492,10 +1735,52 @@ namespace eosio {
                        ("lib", last_handshake.last_irreversible_block_num)
                        ("head", last_handshake.head_num)("id", last_handshake.head_id.str().substr(8,16)) );
             c->enqueue( last_handshake );
+            // only send request to peers allowed connection
+            // every 100 handshakes will send a request
+            if(c->is_peers_connection() && c->sent_handshake_count % my_impl->address_request_frequency == 1) {
```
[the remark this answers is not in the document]
bxliu: handshake will triggered by connect/handle_message/check_heartbeat, i think that's enough for start handshaking.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -321,12 +487,16 @@ namespace eosio {
...
    constexpr auto     def_max_consecutive_immediate_connection_close = 9; // back off if client keeps closing
    constexpr auto     def_max_clients = 25; // 0 for unlimited clients
+   constexpr auto     def_min_peers = 0;
    constexpr auto     def_max_nodes_per_host = 1;
    constexpr auto     def_conn_retry_wait = 30;
    constexpr auto     def_txn_expire_wait = std::chrono::seconds(3);
    constexpr auto     def_resp_expected_wait = std::chrono::seconds(5);
    constexpr auto     def_sync_fetch_span = 1000;
    constexpr auto     def_keepalive_interval = 10000;
+   constexpr auto     def_max_address_per_request = 10000;
```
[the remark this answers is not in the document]
bxliu: The value of 10,000 is empirical. For instance, in the P2P implementation of Bitcoin, I would incorporate additional test checks to determine an acceptable quantity.

ericpassmore:
closing as leap 6 IF changes have evolved significantly.

PR #919, opened 2023-03-29
Add substitution plugin

guilledk:
## Description
This adds the substitution plugin, which allows to replace a wasm contract on chain at runtime, this is mostly used for performance / debugging reasons.

Closes https://github.com/telosnetwork/telos.evm/issues/103

Most of the line diff is from adding the `abieos` and `debug_eos_vm` libraries back to the repo.


guilledk:
Wooopsss, meant to open this on `telosnetwork/leap` fork, sorry folks.

PR #920, opened 2023-03-29
bump minimum cmake requirement to 3.16

spoonincode:
With the removal of Ubuntu 18.04 support we're clear to bump the minimum cmake requirement to 3.16 (the version included in Ubuntu 20.04). Interest in C++20 is primary immediate driver.

The only other notable platform loss with this change is Debian 10 which, up until this change, could still build Leap completely with system provided packages (Debian 10 included boost 1.67 even). With Debian 12 shortly releasing this isn't much of a concern really.

PR includes a few obvious small changes to remove now unneeded `CMAKE_VERSION` checks; I certainly don't claim to have refactored everything that could have been refactored for newer cmake here in this PR.


ScottBailey (approved):
LGTM, thanks!

spoonincode:
oh hm, I just realized Debian 11 ships with 3.13. Does it make any sense to just go to 3.13 instead? is there anything useful added between 3.14 and 3.16? I looked through the release notes and nothing huge stood out...

ScottBailey:
> oh hm, I just realized Debian 11 ships with 3.13. Does it make any sense to just go to 3.13 instead? is there anything useful added between 3.14 and 3.16? I looked through the release notes and nothing huge stood out...

Given Bookworm (12) releases soon and I feel Bullseye (11) user should expect they will need to either upgrade to 12 or add bullseye-backports to get CMake 3.16, I think it's acceptable to leave this as is. Additionally, it's fairly easy to build a deb package from CMake source.

ScottBailey (approved):

linh2931:
While we always try to support as many as platforms as possible by default,  as we officially support only Ubuntu 20.04 and 22.04, it is OK for us to go to the lowest supported OS's cmake 3.16. You can call this out in the 4.0 release note along with the removal of ubuntu 18.04 announcement.

linh2931 (approved):

spoonincode:
Yeah, combined
* Debian isn't a primary support target
* 3.16 is available in bullseye-backports, and
* By the the time 5.0 is released Debian 12 will have been out for months

makes it seem like not a major hindrance.

Issue #911, opened 2023-03-27
Test failure: resource_monitor_plugin_test


heifner:
https://github.com/AntelopeIO/leap/actions/runs/4535476529/jobs/7991066032

May be same issue as #906

heifner:
Issue with launcher.py --kill not working correctly. Handed off to Jonathan to fix.

PR #925, opened 2023-03-29
[4.0 -> main] Implement support for -k/--kill in launcher.

jgiszczak:
Launcher now honors `-k/--kill <signal_number>` argument. Cluster was updated for the Python launcher but support in the launcher itself was missed. Resolves https://github.com/AntelopeIO/leap/issues/911.

Merges #921 into `main`.


heifner:
Create a branch that does this, don't do it directly.

PR #928, opened 2023-03-30
[3.2] SHiP flush logs on write

### LINKED ISSUE #596: 3.2: when ship node exits with error it usually doesn't start up again properly with snapshot
Normally when a ship node crashes, you can start it again by using a snapshot, and then it will start with log messages like:

```
info  2022-12-23T20:32:22.959 nodeos    controller.cpp:494            replay               ] existing block log, attempting to replay from 218085489 to 218258039 blocks
info  2022-12-23T20:32:24.373 nodeos    log.hpp:479                   truncate             ] fork or replay: removed 172884 blocks from trace_history.log
info  2022-12-23T20:32:24.629 nodeos    log.hpp:479                   truncate             ] fork or replay: removed 172884 blocks from chain_state_history.log
```

Then it runs replay and all is good.

With 3.2 it seems very often the trace history index is somehow corrupted, and it re-builds the entire state history from scratch.. which takes an inordinate amount of time on a large blockchain (eg wax), so just better to restore from backup than to let this continue.

image

It is always the case that the trace history index index can be corrupt on an unclean shutdown, but there is something in 3.2 that makes it corrupt more than in previous versions.  It seems like it is never valid?

heifner:
Flush the `state_history_plugin` logs to disk at the end of each write to make it more likely that a `kill -9` or crash leaves valid logs.

Resolves #596

### DIFF
--- a/libraries/state_history/include/eosio/state_history/log.hpp
+++ b/libraries/state_history/include/eosio/state_history/log.hpp
@@ -263,6 +263,9 @@ class state_history_log {
          const uint32_t num_blocks_in_log = _end_block - _begin_block;
          fc::raw::pack(log, num_blocks_in_log);
       }
+
+      log.flush();
+      index.flush();
    }
 
    // returns cfile positioned at payload


libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -263,6 +263,9 @@ class state_history_log {
          const uint32_t num_blocks_in_log = _end_block - _begin_block;
          fc::raw::pack(log, num_blocks_in_log);
       }
+
+      log.flush();
+      index.flush();
```
linh2931, line 268: Will this have any performance impact? But correctness is more important and this is only on ship node. It should be OK.

Do we have tests which repeatedly kill and shutdown ship?

heifner: Yes, there is a performance impact. Likely fairly small.
We do not test `kill -9` with SHiP as there is no guarantee it will work. If the `kill -9` happens during or between the flushes then this will not help. However, during manual testing I was able to `kill -9` and recover from snapshot. Since most of the time is outside the write/flush this does help.

spoonincode (approved):

Issue #837, opened 2023-03-17
Read-only transaction processing - Interrupt executing transactions at the end of the read window


greg7mdp:
Currently, when the timer marking the end of the read window is triggered, we wait for the readonly tasks to finish, so we don't strictly respect the requested window duration.

heifner:
Also would like to interrupt read-only and normal trxs when a block arrives at the node.

linh2931:
We don't have this issue. In `push_read_only_transaction`, we call `chain.push_transaction( trx, window_deadline, _ro_max_trx_time_us, 0, false, 0 );`, where `window_deadline` is  `std::min( start + remaining_time_in_read_window_us, block_deadline );`. A transaction will be interrupted if the window_deadline is reached, and the thread won't schedule any any transactions when this happens.

So when read window timer expires, all threads have finish execution. For safety, before switching to write window, we make sure all threads indeed finish executing. This is not the read-window duration is not respected.

arhag:
> Also would like to interrupt read-only and normal trxs when a block arrives at the node.

@heifner: 
We discussed already privately, but I am summarizing here as well for others to make it clear we didn't miss the above quoted point when closing this issue either.


The summary is that we do not think it is worth the added complexity of interrupting any actively executing read-only transactions (or for that matter any other read-only tasks) when a new block is received. Already when a new block is received, it will end the read window after all read threads finish the task they are currently working on. 

In the case of read-only transactions that should not exceed max-transaction-time (or the time until the end of the read window whichever is sooner). In the case of other read operations, they are already expected to have their own sensible limits to protect against abuse (remember that prior to https://github.com/AntelopeIO/leap/pull/901 these read operations were running serially on the main application thread and so they were delaying other high priority operations that had just been queued and were waiting to run on the main application thread like processing a newly received block). So the transition from the read window to the write window may be delayed the maximum of one of these sensible limits.

Perhaps that delay would be unacceptable on a block producer node (or any of their canary nodes they depend on to receive blocks) because getting and processing blocks quickly is important to minimize microforks. But those types of nodes should not be running nodeos in a mode that has a read window anyway (the read window is really there for API nodes). In fact, nodeos prevents it from even being configured with producers and positive read threads at the same time.

For other types of nodes that have a reason to enable read threads, the slight delay this can add to processing the latest block (which again should be significantly less than 500 ms) is not that big of a deal and so we can avoid the extra complexity of trying to interrupt the read tasks upon receiving a new block.

PR #929, opened 2023-03-30
[4.0 -> main] Execute read only safe tasks on read-only thread pool

### LINKED ISSUE #810: Move all read-only operations to parallel execution of all read operations

heifner:

Note read and write windows are interrupted when new blocks are received.

`--read-only-threads` now defaults to 3 on non-producer configured `nodeos` that enable the `eosio::chain_api_plugin`.
Block relay nodes or canary nodes should consider `--read-only-threads=0`. Since `eosio::chain_api_plugin` may be enabled so `/v1/chain/get_info` is available. If other http request and push_* operations of `eosio::chain_api_plugin` are not used there is no benefit to enabling `read-only-threads`. If `read-only-threads` are enabled, but not utilized, there is a very small overhead of switching back and forth between the read/write windows.
See: https://github.com/AntelopeIO/leap/issues/837#issuecomment-1489542655

Merges #901 into `main`

Builds on #776. 
Resolves #810


greg7mdp (approved):

linh2931 (approved):
We need to revisit if read-only transactions should run on main thread. That can be another PR if decision is made to no allow on main thread.

PR #931, opened 2023-03-30
[4.0] Improved test_snapshot_scheduler stability

766C6164:
This PR improves test stability, in particular tolerance to ordering of snapshot requests in multiindex container.

### DIFF
--- a/plugins/producer_plugin/test/test_snapshot_scheduler.cpp
+++ b/plugins/producer_plugin/test/test_snapshot_scheduler.cpp
@@ -95,11 +95,11 @@ BOOST_AUTO_TEST_CASE(snapshot_scheduler_test) {
             if (!pp->get_snapshot_requests().snapshot_requests.empty()) {
                const auto& snapshot_requests = pp->get_snapshot_requests().snapshot_requests;
                auto it = find_if(snapshot_requests.begin(), snapshot_requests.end(), [](const producer_plugin::snapshot_schedule_information& obj) {return obj.snapshot_request_id == 0;});
-               if (it!=snapshot_requests.end()) {
-                  auto& pending = it->pending_snapshots;
-                  if (pending && pending->size()==1) {
-                     BOOST_CHECK_EQUAL(9, pending->begin()->head_block_num);   
-                  }
+               // we should have a pending snapshot for request id = 0
+               BOOST_REQUIRE(it != snapshot_requests.end());
+               auto& pending = it->pending_snapshots;
+               if (pending && pending->size()==1) {
+                  BOOST_CHECK_EQUAL(9, pending->begin()->head_block_num);   
                }
             }
          });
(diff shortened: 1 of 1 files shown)


spoonincode (approved):

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
```diff
@@ -118,8 +120,13 @@ BOOST_AUTO_TEST_CASE(snapshot_scheduler_test) {
          // one of the snapshots is done here and request, corresponding to it should be deleted
          BOOST_CHECK_EQUAL(2, pp->get_snapshot_requests().snapshot_requests.size());
 
-         // check whether no pending snapshots present
-         BOOST_CHECK_EQUAL(0, pp->get_snapshot_requests().snapshot_requests.begin()->pending_snapshots->size());
+         // check whether no pending snapshots present for a snapshot with id 0
+         const auto& snapshot_requests = pp->get_snapshot_requests().snapshot_requests;
+         auto it = find_if(snapshot_requests.begin(), snapshot_requests.end(),[](const producer_plugin::snapshot_schedule_information& obj) {return obj.snapshot_request_id == 0;});
+         if (it!=snapshot_requests.end()) {
```
arhag, line 126: At this point, we should know that the snapshot with request ID 0 exists and there is no reason for it to expire.

So instead of an if statement it seems more appropriate to require that it is found before moving on to checking its pending snapshots state.

```cpp
BOOST_REQUIRE(it != snapshot_requests.end());
```

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
```diff
@@ -118,8 +120,13 @@ BOOST_AUTO_TEST_CASE(snapshot_scheduler_test) {
...
+            auto& pending = it->pending_snapshots;
+            BOOST_CHECK_EQUAL(0, pending ? pending->size() : 0);
```
arhag, line 128: Is there any situation in which `pending_snapshots` can exist and its value is 0? 

If so, what is the meaning behind such a case and how does it differ from `pending_snapshots` existing but being empty.

If not, this check should be replaced with the stricter check `BOOST_CHECK(!it->pending_snapshots);`.

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
766C6164: Agreed, done.

plugins/producer_plugin/test/test_snapshot_scheduler.cpp
766C6164: Thank you, updated. pending_snapshots is lazy initialized when first pending snapshot is created and will be returned empty when there is no pending snapshots at the moment. Probably as improvement this can be changed in future to either always return pending_snapshots in request reply or just reset() an optional when no pending snapshot present.

arhag (approved):

Issue #800, opened 2023-03-09
Avoid compiling the same contract on each thread in parallel read-only transaction execution in eos-vm and eos-vm-jit


linh2931:
In current implementation of parallel read-only transaction execution, each thread maintains a thread local `wasmif`, which can result in compile the same contract on each thread. The performance can be improved if the compiled contract is shared among the threads.

linh2931:
It is decided to revisit all memory usage issues related to read-only transactions after 4.0 RC.

arhag:
Replaced by https://github.com/AntelopeIO/leap/issues/1158.

BenjaminGormanPMP:
@linh2931 can you provide a brief description (1 or 2 sentences) for discard of issue #800.  This will help for historical reference, lessons learned, and retrospective? Thanks.

PR #932, opened 2023-03-30
[4.0] wasm interface cleanups for parallel read-only transaction execution

linh2931:
This PR
- Reuses compiled wasm module for main and read-only threads
- Fixes an issue wasm_interface methods not applied to wasmif held by read-only threads
- Replaces thread_local for wasmif with unordered_map

Resolved https://github.com/AntelopeIO/leap/issues/800


libraries/chain/controller.cpp
```diff
@@ -2688,11 +2690,30 @@ struct controller_impl {
       else
 #endif
          // Non-EOSVMOC needs a wasmif per thread
-         wasmif_thread_local = std::make_unique<wasm_interface>( conf.wasm_runtime, conf.eosvmoc_tierup, db, conf.state_dir, conf.eosvmoc_config, !conf.profile_accounts.empty());
+         threaded_wasmifs[std::this_thread::get_id()] = std::make_unique<wasm_interface>( conf.wasm_runtime, conf.eosvmoc_tierup, db, conf.state_dir, conf.eosvmoc_config, !conf.profile_accounts.empty());
```
heifner, line 2693: This is called from multiple threads at startup and it is not thread safe. Since it is only modified during start of the threads you can likely just use a mtx only here. 

Also to be extra cautious you should probably wait in `producer_plugin::startup()` until all `init_thread_local_data()` have been called. You could use some futures in `producer_plugin::startup()` to do that. That seems extremely unlikely, but technically possible I think.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -173,21 +173,10 @@ namespace eosio { namespace chain {
...
-               WASM::serialize(stream, module);
-               module.userSections.clear();
-            } catch (const Serialization::FatalSerializationException& e) {
-               EOS_ASSERT(false, wasm_serialization_error, e.message.c_str());
-            } catch (const IR::ValidationException& e) {
-               EOS_ASSERT(false, wasm_serialization_error, e.message.c_str());
-            }
+            IR::Module module = trx_context.control.get_wasm_module_cache().get_module(code_hash, vm_type, vm_version, bytes);
 
             wasm_instantiation_cache.modify(it, [&](auto& c) {
```
heifner, line 181: Is there a reason not to have a shared `wasm_instantiation_cache` ?

heifner: Seems like they could use the same boost multi index that stores both the `IR::Module` and the `wasm_instantiated_module_interface`.

libraries/chain/include/eosio/chain/wasm_module_cache.hpp
```diff
@@ -0,0 +1,47 @@
...
+   struct wasm_module_cache {
+      const IR::Module& get_module( const digest_type& code_hash, uint8_t vm_type, uint8_t vm_version, std::vector<U8>& bytes );
+      void code_block_num_last_used(const digest_type& code_hash, uint8_t vm_type, uint8_t vm_version, uint32_t block_num);
+      void current_lib(uint32_t lib);
+
+      struct module_entry {
+         digest_type   code_hash;
+         uint8_t       vm_type = 0;
+         uint8_t       vm_version = 0;
+         uint32_t      first_block_num_used;
```
heifner, line 19: Does not appear to be used.

libraries/chain/wasm_module_cache.cpp
```diff
@@ -0,0 +1,49 @@
...
+            EOS_ASSERT(false, wasm_serialization_error, e.message.c_str());
+         } catch (const IR::ValidationException& e) {
+            EOS_ASSERT(false, wasm_serialization_error, e.message.c_str());
+         }
+
+         it = module_cache.emplace( module_entry {
+                                      .code_hash = code_hash,
+                                      .vm_type = vm_type,
+                                      .vm_version = vm_version,
+                                      .module = module
```
heifner, line 28: Should init `last_block_num_used`

libraries/chain/include/eosio/chain/wasm_module_cache.hpp
```diff
@@ -0,0 +1,47 @@
...
+         uint32_t      last_block_num_used;
```
heifner, line 20: I think both of these should be initialized.

libraries/chain/include/eosio/chain/wasm_module_cache.hpp
heifner: Also does not appear to be used in wasm_interface_cache. Removing that unused index would be a performance boost.

heifner (changes requested):

libraries/chain/controller.cpp
linh2931: Thanks! I forgot about the mutex here.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
linh2931: I initially tried shared `wasm_instantiation_cache`. It did not work, likely due to some other deep complications. For RCs, I did not want to make too tricky changes.

libraries/chain/include/eosio/chain/wasm_module_cache.hpp
linh2931: I don't see it used either.  As it is defined in wasm_interface_private.hpp, an implementation class, it is safe to remove. Will remove it from both files.

libraries/chain/CMakeLists.txt
```diff
@@ -129,7 +130,7 @@ add_library( eosio_chain
              )
 
 target_link_libraries( eosio_chain PUBLIC bn256 fc chainbase eosio_rapidjson Logging IR WAST WASM Runtime
-                       softfloat builtins ${CHAIN_EOSVM_LIBRARIES} ${LLVM_LIBS} ${CHAIN_RT_LINKAGE}
+                       softfloat builtins custom_appbase ${CHAIN_EOSVM_LIBRARIES} ${LLVM_LIBS} ${CHAIN_RT_LINKAGE}
```
heifner, line 133: We have tried to keep chain from depending on appbase. It looks like this is only used for EOS_ASSERTS. I would recommend reverting this and not make chain depend on appbase.

libraries/chain/controller.cpp
```diff
@@ -337,7 +337,12 @@ struct controller_impl {
       set_activation_handler<builtin_protocol_feature_t::crypto_primitives>();
 
       self.irreversible_block.connect([this](const block_state_ptr& bsp) {
-         get_wasm_interface().current_lib(bsp->block_num);
+         EOS_ASSERT( appbase::app().executor().is_write_window(), misc_exception, "write window expected for irreversible_block signal"); // read-only threads are idle, threaded_wasmifs are not used
```
heifner, line 340: If signals are called from read-window all kinds of assumptions do not hold. You could move this assert into `producer_plugin`'s handling of `irreversible_block signal`.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
heifner: What problems did you run into with a shared `wasm_instantiation_cache`?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1266,6 +1265,17 @@ void producer_plugin::plugin_startup()
             chain.init_thread_local_data();
          });
 
+      // Do a test run to make sure all threads have started
+      std::vector<std::future<bool>> fut;
+      for (auto i = 0; i < my->_ro_thread_pool_size; ++i ) {
+         fut.emplace_back( post_async_task( my->_ro_thread_pool.get_executor(), [] () {
```
heifner, line 1271: These do not all necessarily execute on diff threads. I would say it is unlikely that they would. Instead, I would use an `std::atomic<uint32_t>` that was incremented after `chain.init_thread_local_data();` above. Then either wait on condition variable that is signaled after the uint32_t increment. Or just a simple while loop waiting on the uint32_t to equal `my->_ro_thread_pool_size`. The condition variable needs a mutex which seems like a lot for this. Too bad we don't have c++20 yet as this could be done trivially with a `std::latch` or `std::atomic<>::wait`. For now I would say the simple loop with a 1ms sleep is likely the best with a comment that this can switch to a `std::latch` once C++20 is available.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
linh2931: Instantiated modules rely on runtime interface and cannot be simply shared. nodeos can run a few multi-threaded read-only trx for a few seconds and crashes with memory corruption. I plan to look into this further after RC. For now focus on cleaning up other loose ends.

libraries/chain/CMakeLists.txt
linh2931: As now read-only trxs can run in both read and write window, we need some way to know which window we are in in OC code, where we will only evict cache entries in write window in which only one thread is executing transactions. Currently we us is_main_thread. I am going to make that changes in a separate PR.

Any other way we can get window infor from producer plugin? Passing window in push transaction? Do you think that incurs some performance penality?

libraries/chain/controller.cpp
linh2931: Will move this to producer plugin.

plugins/producer_plugin/producer_plugin.cpp
linh2931: Thanks for your detailed suggestion. Will use a simple loop since this only executes once at startup.

libraries/chain/CMakeLists.txt
heifner: Can you use `chainbase._read_only_mode` or tie it to `chain.set_db_read_only_mode()` maybe renaming to `chain.set_read_only_mode()`?

linh2931: Thanks. I will look into that direction. 

I will revert appbase change. For the assert in `code_block_num_last_used`, we don't need it here as in `eosio_contract.cpp'`s  `apply_eosio_setcode` where `code_block_num_last_used` is called, we already have an assert it cannot be called by any read-only transactions.

libraries/chain/wasm_module_cache.cpp
```diff
@@ -0,0 +1,49 @@
...
+                                   } ).first;
+      }
+
+      return it->module;
+   }
+
+   void wasm_module_cache::code_block_num_last_used(const digest_type& code_hash, uint8_t vm_type, uint8_t vm_version, uint32_t block_num) {
```
heifner, line 35: Add lock of mutex, or comment why one is not needed.

libraries/chain/wasm_module_cache.cpp
```diff
@@ -0,0 +1,49 @@
...
+      auto it = module_cache.find(boost::make_tuple(code_hash,vm_type, vm_version));
+      if(it != module_cache.end())
+         module_cache.modify(it, [block_num](module_entry& e) {
+            e.last_block_num_used = block_num;
+         });
+      }
+
+   void wasm_module_cache::current_lib(uint32_t lib) {
```
heifner, line 43: Add lock of mutex or comment why one is not needed.

heifner (approved):

spoonincode:
I'm not understanding how this resolves #800. It could be we just have a different definition of what 'compile' means, or maybe I just have not studied this enough yet. But since there is still a wasm_interface per thread, wouldn't that mean if there are 64 RO threads, EOS VM JIT still has to compile the contract (up to) 64 times, which means each contract consumes 64x the resources as it previously did?

Seeing `IR::Module` as the cached type doesn't feel right, but I need to look at this more to articulate a better definitive reason. My feel comes about be cause EOS VM doesn't use `IR::Module` -- so why would we keep one of those around? `IR::Module` is used to run some checks on a `setcode` prior to `CONFIGURABLE_WASM_LIMITS2`, but that's just in `setcode` and should never need to leak out of there I believe.

linh2931:
Thanks Matt. Would caching the memory image from `parse_initial_memory` be a step in the right direction, if instantiated_module is not made shared during this RC time?

spoonincode:
> caching the memory image from `parse_initial_memory`

I'm not sure that's even used any more? Best I can tell both `eosvmoc_instantiated_module` and `eos_vm_instantiated_module` don't require it.

arhag:
We agreed to not target this for `release/4.0`.

Issue #824, opened 2023-03-15
Parallel read-only transaction execution further improvement


linh2931:
During code walkthrough led by Kevin, the following improvements were suggested. They can be done during RC phases or later.

- Use `deque` instead of `std::deque`. PR #946
- Pick a value that seems somewhat reasonable for read and window time. Or put a comment indicating the default values are just numbers. No need to do in code. Should be in a document.
- ~~Would be nice to encapsulate mutex in `_ro_trx_queue` by adding a `push/pop` to `_ro_trx_queue`.~~ (moved to gh-822)
- Add comment that window timers only accessed from the main thread. PR #946
- Consider using only one timer since only one window should be active at any given time. PR #905
- Pass in `read_window_deadline` so that it is calculated before these are spawning. #PR 901
- `start_recover_keys` may not needed. PR #901
- `EOS_ASSERT` `chain.is_building_block()`. No longer applicable as we allow producer in test mode.
- `calculate_block_deadline` is not needed. No longer applicable as the code is refactored to handle read-only trxs in both main thread and read-only threads.
- Update cap on read-only threads when runninng with OC to reflect behavior described in document. PR #946

PR #933, opened 2023-03-31
[4.0] Small improvements for parallelizing read-only transactions and tasks

linh2931:
Resolved the remaining items in https://github.com/AntelopeIO/leap/issues/824

- update eos-vm-oc threads limit calculation based on the design
- use faster boost deque instead of std:deque


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -931,9 +931,9 @@ void producer_plugin::set_program_options(
          ("read-only-threads", bpo::value<uint16_t>(),
           "Number of worker threads in read-only execution thread pool")
          ("read-only-write-window-time-us", bpo::value<uint32_t>()->default_value(my->_ro_write_window_time_us.count()),
-          "time in microseconds the write window lasts")
+          "Time in microseconds the write window lasts. The default value is not a representative one. Choose a proper one based on your use cases.")
```
heifner, line 934: This seems odd to say in a help message. Maybe we should have a guide on use of `read-only-threads` in https://docs.eosnetwork.com/ which could discuss this in detail. I would suggest just leaving this as `Time in microseconds the write window lasts.`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -931,9 +931,9 @@ void producer_plugin::set_program_options(
...
          ("read-only-read-window-time-us", bpo::value<uint32_t>()->default_value(my->_ro_read_window_time_us.count()),
-          "time in microseconds the read window lasts")
+          "Time in microseconds the read window lasts. The default value is not a representative one. Choose a proper one based on your use cases.")
```
heifner, line 936: Same comment as above.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2975,10 +2971,7 @@ bool producer_plugin_impl::push_read_only_transaction(transaction_metadata_ptr t
    try {
       auto start = fc::time_point::now();
       chain::controller& chain = chain_plug->chain();
-      if ( !chain.is_building_block() ) {
-         _ro_exhausted_trx_queue.push_front( {std::move(trx), std::move(next)} );
-         return true;
-      }
+      EOS_ASSERT( chain.is_building_block(), producer_exception, "Expecting chain is building block" );
```
heifner, line 2974: I don't think this is correct. We allow read-only trxs on a producer node. Those trxs could run when we are not building a block. I think this change should be reverted.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
linh2931: Oh now single thread and multi-threaded resad-only trxs use the same method. For `--read-only-threads` enabled nodes, will the assert be applicable?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1146,14 +1146,10 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
...
          // reserve 1 for the app thread, 1 for anything else which might use VM
-         int num_threads_supported = (vm_total_kb - vm_used_kb) / 4200000000 - 2;
-         ilog("vm total in kb: ${total}, vm used in kb: ${used}, number of EOS VM OC threads supported ((vm total - vm used)/4.2 TB - 2): ${supp}", ("total", vm_total_kb) ("used", vm_used_kb) ("supp", num_threads_supported));
-         EOS_ASSERT( num_threads_supported >= my->_ro_thread_pool_size, plugin_config_exception, "--read-only-threads (${th}) greater than number of threads supported for EOS VM OC (${supp}) by the system virtual memory size", ("th", my->_ro_thread_pool_size) ("supp", num_threads_supported) );
-
-         if ( my->_ro_thread_pool_size > my->_ro_max_eos_vm_oc_threads_allowed ) {
-            wlog("--read-only-threads (${th}) greater than maximum number of threads allowed (${allowed}) for EOS Vm OC. Set it to ${allowed}", ("th", my->_ro_thread_pool_size) ("allowed", my->_ro_max_eos_vm_oc_threads_allowed));
-            my->_ro_thread_pool_size = my->_ro_max_eos_vm_oc_threads_allowed;
-         }
+         uint16_t num_threads_supported = (vm_total_kb - vm_used_kb) / 4200000000 - 2;
```
greg7mdp, line 1149: Why change from `int` to `uint16_t`? Don't we have a potential overflow here if `(vm_total_kb - vm_used_kb) / 4200000000` < 2?

plugins/producer_plugin/producer_plugin.cpp
heifner: Not for our test case were we allow that with producer configured. I think best to just have it gracefully handle it by pushing it back into the exhausted queue. i.e. revert this change.

plugins/producer_plugin/producer_plugin.cpp
linh2931: The std::min below requires the same type. The intermediate computational results on the right hand will not be stored on the left hand during calculation. But to be safe in future machines which can have support 65535 threads, I will change to uint32_t. Thanks.

linh2931: thanks.

greg7mdp: @linh2931 your change does not fix the issue https://godbolt.org/z/aE7cE15xs

greg7mdp: link works for me. Thy this, you get 16 instead of the correct answer 1 (because `num_threads_supported` == 65535)

```
#include <iostream>

int main() {
   size_t vm_total_kb = 5200000000;
   size_t vm_used_kb = 10;
   uint32_t _ro_max_eos_vm_oc_threads_allowed = 16;
   
   uint32_t num_threads_supported = (vm_total_kb - vm_used_kb) / 4200000000;
   num_threads_supported -= 2;
   auto actual_threads_allowed = std::min(_ro_max_eos_vm_oc_threads_allowed, num_threads_supported);
   std::cout << actual_threads_allowed << '\n';
}
```

linh2931: This assert `EOS_ASSERT( num_threads_supported > 2` will prevent the execution from going further.

greg7mdp: Yes, that's fine, but it should be `EOS_ASSERT( num_threads_supported >= 2`.

greg7mdp: Message is still not right: `Available system virtual memory cannot support mininum of 3 minimum threads`.

Maybe something like:

`"With the EOS VM OC configured, there is not enough system virtual memory to support the required minimum of 3 threads`

greg7mdp (approved):

Issue #899, opened 2023-03-25
4.0.0-rc2: put the prometheus exporter on a different port


matthewdarwin:
When you put `plugin = eosio::prometheus_plugin` it adds a new /v1/prometheus/metrics URL.

You want this URL to be available to the monitoring system.
But this same port is used for the producer plugin which needs to be highly restricted (and definately not available to the monitoring system).

For security reasons, the prometheus export must run on a dedicated port.  Otherwise node operators need to have a proxy server in between to handle the security.  This is nonsensical requirement to use a monitoring plugin.

Look at any other blockchain software that has prometheus feature.  Always runs on a separate port.

spoonincode:
I'm moving this out of actionable to discussion because maybe we want to shoehorn in something special for prometheus in 4.0, but maybe we need to reflect on this a little more as the "run such-and-such-plugin on a separate port" is such a common request (#88, eosnetworkfoundation/mandel#343, eosnetworkfoundation/mandel#360, etc etc). I'd like some consideration given before rushing in to anything that may be short sighted.

bhazzard:
This is a missed requirement on my part. @heifner can you propose a solution here and have @huangminghuang immediately start work on it for inclusion in rc3?

We need a solution that is trivial to implement (ideally that can be designed, developed, tested & reviewed in about 1 week).

If there is such a trivial solution that addresses Matt's point about the need for a more general solution, please propose it.

Otherwise, please propose a focussed solution to the immediate problem at hand. Consider how we might move towards a more generalized solution as part of our focussed solution.

PR #939, opened 2023-03-31
[4.0] prometheus on different port

huangminghuang:
This PR put prometheus metrics exporter on a different port.

Added option 
- prometheus-exporter-address : The local IP and port to listen for incoming prometheus metrics http request.

The API resource name remains to be `/v1/prometheus/metrics`

Resolves IS #899


plugins/prometheus_plugin/include/eosio/prometheus_plugin/simple_rest_server.hpp
```diff
@@ -0,0 +1,223 @@
...
+      T* self() { return static_cast<T*>(this); }
+
+      void fail(beast::error_code ec, char const* what) { self()->log_error(what, ec.message()); }
+      // Return a response for the given request.
+      http::response<http::string_body> handle_request(http::request<http::string_body>&& req) {
+         // Returns a bad request response
+         auto const bad_request = [&req](beast::string_view why) {
+            http::response<http::string_body> res{ http::status::bad_request, req.version() };
+            res.set(http::field::server, BOOST_BEAST_VERSION_STRING);
+            res.set(http::field::content_type, "text/html");
```
heifner, line 28: Seems like these should be "text/plain" as they are not `html`.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -214,24 +220,35 @@ namespace eosio {
...
+      string port = lipstr.substr(host.size() + 1, lipstr.size());
+
+      boost::system::error_code ec;
+      using tcp = boost::asio::ip::tcp;
+      tcp::resolver resolver(app().get_io_service());
 
-      prometheus_api handle(*my, max_response_time);
-      app().get_plugin<http_plugin>().add_async_api({
-        CALL_ASYNC_WITH_400(prometheus, handle, eosio, metrics, std::string, 200, http_params_types::no_params)}, http_content_type::plaintext);
+      my->_endpoint = *resolver.resolve(tcp::v4(), host, port, ec);
```
heifner, line 241: Is there some reason we can't support IPv6?

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -179,32 +206,11 @@ namespace eosio {
...
       }
 
-      void metrics_async(chain::plugin_interface::next_function<std::string> results) {
-         _prometheus_strand.post([self=this, results=std::move(results)]() {
-            results(self->metrics());
-         });
+      void start() {
+         run(_prometheus_thread_pool.get_executor(), _endpoint);
+         _prometheus_thread_pool.start(
+               1, [](const fc::exception& e) { elog("Prometheus excpetion ${e}:${l}", ("e", e)); });
```
heifner, line 212: `exception`. What is `${l}` suppose to be associated with?

heifner (changes requested):

plugins/prometheus_plugin/prometheus_plugin.cpp
huangminghuang: Our existing HTTP plugin use ipv4 only and I couldn't find any example from goole which listens an interface returned from tcp::resolver with both ipv4 and ipv6 enabled. I was just not confident that I won't introduce unforeseeable bugs if I dit that. If we want to support ipv6, I believe we should wait until 5.0 and do it with the existing http plugin.

spoonincode: Yeah, I think holding off on IPv6 is the right choice until there is more reflection on how to handle the various options discussed in #622

spoonincode (changes requested):

plugins/prometheus_plugin/include/eosio/prometheus_plugin/simple_rest_server.hpp
```diff
@@ -0,0 +1,223 @@
...
+            res.keep_alive(req.keep_alive());
+            res.body() = "The resource '" + std::string(target) + "' was not found.";
+            res.prepare_payload();
+            return res;
+         };
+
+         // Returns a server error response
+         auto const server_error = [&req](beast::string_view what) {
+            http::response<http::string_body> res{ http::status::internal_server_error, req.version() };
+            res.set(http::field::server, BOOST_BEAST_VERSION_STRING);
```
spoonincode, line 49: I'm not sure if there is an easy way to plumb it through, but consider changing this to be identical to
https://github.com/AntelopeIO/leap/blob/4099678077cded77a7219e54b00f55863b7fe793/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp#L121-L122
which is
https://github.com/AntelopeIO/leap/blob/4099678077cded77a7219e54b00f55863b7fe793/programs/nodeos/main.cpp#L125

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -110,12 +113,41 @@ namespace eosio {
...
+         return method == http::verb::get;
+      }
+
+      std::optional<http::response<http::string_body>>
+       on_request(http::request<http::string_body>&& req) {
+         if(req.target() != prometheus_api_name)
+            return {};
+         http::response<http::string_body> res{ http::status::ok, req.version() };
+         // Respond to GET request
+         res.set(http::field::server, BOOST_BEAST_VERSION_STRING);
```
spoonincode, line 137: This server string is now different then what is returned on the error responses

plugins/prometheus_plugin/include/eosio/prometheus_plugin/simple_rest_server.hpp
```diff
@@ -0,0 +1,234 @@
...
+
+#include <boost/beast/core.hpp>
+#include <boost/beast/http.hpp>
+#include <boost/beast/version.hpp>
+
+namespace eosio { namespace rest {
+
+   // The majority of the code here are derived from boost source
+   // libs/beast/example/http/server/async/http_server_async.cpp
+   // with minium modification and yet reusable.
```
spoonincode, line 11: `s/minium/minimum/`

spoonincode (approved):
Seems okay since we're trying to make this remain close to the beast example, even though the example is overkill in some ways for our usage here.

plugins/prometheus_plugin/include/eosio/prometheus_plugin/simple_rest_server.hpp
```diff
@@ -0,0 +1,234 @@
...
+   template <typename T>
+   class simple_server {
+      T* self() { return static_cast<T*>(this); }
+
+      void fail(beast::error_code ec, char const* what) { self()->log_error(what, ec.message()); }
+      // Return a response for the given request.
+      http::response<http::string_body> handle_request(http::request<http::string_body>&& req) {
+         auto server_header = self()->server_header();
+         // Returns a bad request response
+         auto const bad_request = [&req, &server_header](beast::string_view why) {
```
heifner, line 26: Can you use `std::string_view` instead of `beast::string_view`?

plugins/prometheus_plugin/include/eosio/prometheus_plugin/simple_rest_server.hpp
```diff
@@ -0,0 +1,234 @@
...
+namespace eosio { namespace rest {
+
+   // The majority of the code here are derived from boost source
+   // libs/beast/example/http/server/async/http_server_async.cpp
+   // with minium modification and yet reusable.
+
+   namespace beast = boost::beast;         // from <boost/beast.hpp>
+   namespace http  = beast::http;          // from <boost/beast/http.hpp>
+   namespace net   = boost::asio;          // from <boost/asio.hpp>
+   using tcp       = boost::asio::ip::tcp; // from <boost/asio/ip/tcp.hpp>
```
heifner, line 16: Not a fan of these `using` and `namespace` aliases in header files; although these seem rather harmless.

plugins/prometheus_plugin/prometheus_plugin.cpp
```diff
@@ -179,32 +211,11 @@ namespace eosio {
...
       }
 
-      void metrics_async(chain::plugin_interface::next_function<std::string> results) {
-         _prometheus_strand.post([self=this, results=std::move(results)]() {
-            results(self->metrics());
-         });
+      void start() {
+         run(_prometheus_thread_pool.get_executor(), _endpoint);
+         _prometheus_thread_pool.start(
+               1, [](const fc::exception& e) { elog("Prometheus excpetion ${e}", ("e", e)); });
```
heifner, line 217: s/excpetion/exception/

heifner (changes requested):

heifner (approved):

PR #945, opened 2023-03-31
[PH] Reduce log Usage and Other Performance Harness Improvements

ClaytonCalabrese:
This PR tackles a number of issues that have lingered in performance test.

1. There was a dependency on the nodeosLogs in order to populate `blockData` in the `chainData::blockLog`, which left a vulnerability: If a fork occured, and there were 2 of the same block in the the log, that information would be double counted in `updateTotal`. This lead to cases where more transactions than were generated were found, causing test failure, and junk data. 

2. There were 2 classes with similar names and similar data, `blkData` and `blockData`. This was confusing and unnecessary.  The two have been combined into `blockData`.

3.  Block latency was an unwanted data point being tracked, it and references to it have been removed.

4. Some junk remained from when nodeos logs were an acceptable entrypoint to the performance harness, those have been cleaned up 

5.  In an effort to have less bulk and accurate information, `scrapeBlockTrxDataLog` and `scrapeBlockDataLog` are removed. These were processing files which is written out during `queryBlockTrxData` and were populating dictionaries with block and transaction information.  This meant we were processing the same data twice rather than just saving it the first time. As such, `queryBlockTrxData` now creates both `blockDict` and `trxDict` as information is being processed. `isImportantTransaction` is used to determine whether a transaction being processed is one that has been generated, or is not.

6. Similarly, there was repeated information being passed around and stored. `blockDict` and `chainData::blockLog` were holding essentially the same information, but in different places. To remedy this, they now both point to the same `blockData` objects. As python is pass by reference, this means it is now lightweight to have both `blockDict` and `blockLog`. Additionally, `blockDict` and `trxDict` are now both part of `chainData`. 

7. Forked blocks and dropped blocks were only being counted for the first nonproducing node. `ArtifactPaths::nodeosLogDir`  and `chainData::numNodes` were added and is used to process each node in the test run for forked and dropped blocks, rather than just one. This also required a little changing to how the report is formatted to display each node's forked and dropped blocks. 

8. Some parts of `performance_test.py` were still using `stem[0]` meaning rather than `performance_test` and so the folder that contained it's output was actually just `p`. This has been fixed. 

9. `chainData::blockLog` has been renamed to `chainData::blockList` to better represent what the member actually is.


oschwaldp-oci (approved):

ScottBailey (approved):
lgtm

tests/performance_tests/performance_test_basic.py
```diff
@@ -236,16 +232,34 @@ def fileOpenMode(self, filePath) -> str:
...
-        for blockNum in range(startBlockNum, endBlockNum):
+        for blockNum in range(startBlockNum, endBlockNum + 1):
+            blockCpuTotal, blockNetTotal, blockTransactionTotal = 0, 0, 0
             block = node.fetchBlock(blockNum)
             btdf_append_write = self.fileOpenMode(blockTrxDataPath)
             with open(blockTrxDataPath, btdf_append_write) as trxDataFile:
-                self.clusterConfig.writeTrx(trxDataFile, block, blockNum)
-
+                for transaction in block['payload']['transactions']:
+                    if self.isImportantTransaction(transaction):
```
heifner, line 250: This should be called `isOnBlockTransaction` and be changed to `if not self.isOnBlockTransaction(transaction)`

tests/performance_tests/performance_test_basic.py
```diff
@@ -236,16 +232,34 @@ def fileOpenMode(self, filePath) -> str:
             append_write = 'w'
         return append_write
 
+    def isImportantTransaction(self, transaction):
+        if self.clusterConfig.nodeosVers == "v2":
```
heifner, line 236: Add comment that v2 history does not include `onblock`.

heifner (changes requested):

tests/performance_tests/performance_test_basic.py
oschwaldp-oci: Addressed: [3127329](https://github.com/AntelopeIO/leap/pull/945/commits/31273296fee664f68d6978efa136db27a0df5ac4)

tests/performance_tests/performance_test_basic.py
oschwaldp-oci: Addressed: [3127329](https://github.com/AntelopeIO/leap/pull/945/commits/31273296fee664f68d6978efa136db27a0df5ac4)

oschwaldp-oci: And here: [cd87f42](https://github.com/AntelopeIO/leap/pull/945/commits/cd87f42a9b5f672a98afc978403c960ddad6f332)

heifner (approved):

jgiszczak (approved):

PR #947, opened 2023-04-02
[4.0] Correct time summary calculation for transient trxs due to change of read-only trxs windows

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/950

https://github.com/AntelopeIO/leap/pull/776 assumed read-only transactions to only run on read-only threads. https://github.com/AntelopeIO/leap/pull/901 changed that and read-only transactions can run on main thread too.

- Existing calculation of transient trxs time was done on the main thread. This is not correct any more as read-only trxs on the main thread in the read window will be double counted. The solution is to count transient time only in the write window (dry-run trxs only run in write window). This change makes the intention more clearer and safer, as in write window all processing is single threaded and no read-only threads are running.
- Refactored `push_read_only_transaction` for read-only tasks missed idle time accounting. Add it now.

A sample report looks like
`debug 2023-04-02T20:36:41.903 nodeos   producer_plugin.cpp:284  report  ] Block   #259 trx idle: 555758us out of 565688us, success: 0, 0us, fail: 0, 0us, transient: 14, 2281us,   other: 7649us`

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -281,11 +281,11 @@ struct block_time_tracker {
       if( _log.is_enabled( fc::log_level::debug ) ) {
          auto now = fc::time_point::now();
          add_idle_time( now - idle_trx_time );
-         fc_dlog( _log, "Block #${n} trx idle: ${i}us out of ${t}us, success: ${sn}, ${s}us, fail: ${fn}, ${f}us, transient: ${tn}, ${t}us, other: ${o}us",
+         fc_dlog( _log, "Block #${n} trx idle: ${i}us out of ${t}us, success: ${sn}, ${s}us, fail: ${fn}, ${f}us, transient: ${trn}, ${tr}us, other: ${o}us",
                   ("n", block_num)
                   ("i", block_idle_time)("t", now - clear_time)("sn", trx_success_num)("s", trx_success_time)
                   ("fn", trx_fail_num)("f", trx_fail_time)
-                  ("tn", transient_trx_num)("t", transient_trx_time)
+                  ("trn", transient_trx_num)("tr", transient_trx_time)
                   ("o", (now - clear_time) - block_idle_time - trx_success_time - trx_fail_time - transient_trx_time) );
       }
    }
@@ -2882,6 +2882,7 @@ void producer_plugin_impl::switch_to_read_window() {
       return;
    }
 
+
    auto& chain = chain_plug->chain();
    uint32_t pending_block_num = chain.head_block_num() + 1;
    _ro_read_window_start_time = fc::time_point::now();
@@ -2996,6 +2997,12 @@ bool producer_plugin_impl::push_read_only_transaction(transaction_metadata_ptr t
       if( !_ro_in_read_only_mode ) {
          chain.set_db_read_only_mode();
       }
+
+      if ( app().executor().is_write_window() ) {
+         auto idle_time = fc::time_point::now() - _idle_trx_time;
+         _time_tracker.add_idle_time( idle_time );
+      }
+
       // use read-window/write-window deadline if there are read/write windows, otherwise use block_deadline if only the app thead
       auto window_deadline = (_ro_thread_pool_size != 0) ? _ro_window_deadline : calculate_block_deadline( chain.pending_block_time() );
 
@@ -3008,6 +3015,10 @@ bool producer_plugin_impl::push_read_only_transaction(transaction_metadata_ptr t
       retry = pr.trx_exhausted;
       if( retry ) {
          _ro_exhausted_trx_queue.push_front( {std::move(trx), std::move(next)} );
+      }
+
+      if ( app().executor().is_write_window() ) {
+         _idle_trx_time = fc::time_point::now();
       }
    } catch ( const guard_exception& e ) {
       chain_plugin::handle_guard_exception(e);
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -281,11 +281,11 @@ struct block_time_tracker {
       if( _log.is_enabled( fc::log_level::debug ) ) {
          auto now = fc::time_point::now();
          add_idle_time( now - idle_trx_time );
-         fc_dlog( _log, "Block #${n} trx idle: ${i}us out of ${t}us, success: ${sn}, ${s}us, fail: ${fn}, ${f}us, transient: ${tn}, ${t}us, other: ${o}us",
+         fc_dlog( _log, "Block #${n} trx idle: ${i}us out of ${t}us, success: ${sn}, ${s}us, fail: ${fn}, ${f}us, transient: ${trn}, ${tr}us, other: ${o}us",
                   ("n", block_num)
                   ("i", block_idle_time)("t", now - clear_time)("sn", trx_success_num)("s", trx_success_time)
                   ("fn", trx_fail_num)("f", trx_fail_time)
-                  ("tn", transient_trx_num)("t", transient_trx_time)
+                  ("trn", transient_trx_num)("tr", transient_trx_time)
```
heifner, line 288: Very minor thing, but I actually prefer `tn`, `t` over `trn`,`tr`. If you are going to rename these, maybe `trx_num`, `trx_time`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2996,6 +2997,12 @@ bool producer_plugin_impl::push_read_only_transaction(transaction_metadata_ptr t
       if( !_ro_in_read_only_mode ) {
          chain.set_db_read_only_mode();
       }
+
+      if ( app().executor().is_write_window() ) {
```
heifner, line 3001: Seems like this should be `!_ro_in_read_only_mode` to be consistent. Same for below.

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2988,6 +2997,12 @@ bool producer_plugin_impl::push_read_only_transaction(transaction_metadata_ptr t
...
+         auto idle_time = fc::time_point::now() - _idle_trx_time;
+         _time_tracker.add_idle_time( idle_time );
```
greg7mdp, line 3003: It looks like idle time is not tracked for a specific transaction type (read_only or transient), but for both. So it seems to me that this code will count the read_window time as idle. Is is what we want?

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: really it looks like `_ro_in_read_only_mode` and `app().executor().is_read_window()` are tracking the same state. Why the duplication?

linh2931: We cannot use `!_ro_in_read_only_mode` to tell if a read-only trx is running exclusively on the main thread (in write window) or running on the main thread in parallel with read-only threads (in read-window).
For any read-only trxs in either read window or write window,
`_ro_in_read_only_mode` is always true. So `!_ro_in_read_only_mode` is always false.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2306,7 +2306,11 @@ producer_plugin_impl::handle_push_result( const transaction_metadata_ptr& trx,
...
    push_result pr;
    if( trace->except ) {
-      if ( chain.is_on_main_thread() ) {
+      // Transient trxs are dry-run or read-only.
+      // Dry-run trxs only run in write window. Read-only trxs can run in
+      // both write and read windows; time spent in read window is counted
+      // by read window summary.
+      if ( app().executor().is_write_window() ) {
          auto dur = end - start;
          _time_tracker.add_fail_time(dur, trx->is_transient());
```
greg7mdp, line 2315:
```suggestion
         assert(!trx->is_read_only());
         _time_tracker.add_fail_time(dur, trx->is_transient());
```

plugins/producer_plugin/producer_plugin.cpp
heifner: Maybe I'm having a slow start this Monday, but it appears to me that `!_ro_in_read_only_mode` <==> `app().executor().is_write_window()`. As such, I think we should remove `_ro_in_read_only_mode` and only use `app().executor().is_write_window()`

`_ro_in_read_only_mode` is set at the same time/place as `app().executor().set_to_*_window()`.

plugins/producer_plugin/producer_plugin.cpp
heifner: This idle time is idle time of the main thread that is producing a block (or speculatively executing trxs into a ephemeral block), so yes.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: > That assert is not valid. These can be read-only trxs.

in the write window?

plugins/producer_plugin/producer_plugin.cpp
linh2931: I was overlooking and thought `_ro_in_read_only_mode` as chain's `db_read_only_mode`. Will replace `_ro_in_read_only_mode` with `app().executor().is_write_window()`.

plugins/producer_plugin/producer_plugin.cpp
heifner: yes

PR #958, opened 2023-04-03
Test: Add SHiP forking and have_positions unittest

heifner:
Add a unittest for `state_history_plugin` handling of forks and `have_positions` in `get_blocks_request`.

Adds reflection of `get_blocks_result_v0` type used for debugging.

Related to #955

### DIFF
--- a/plugins/state_history_plugin/tests/session_test.cpp
+++ b/plugins/state_history_plugin/tests/session_test.cpp
@@ -80,7 +80,7 @@ fc::datastream<ST>& operator>>(fc::datastream<ST>& ds, eosio::state_history::get
 
 //------------------------------------------------------------------------------
 
-std::map<uint32_t, eosio::chain::block_id_type> block_ids;
+std::unordered_map<uint32_t, eosio::chain::block_id_type> block_ids;
 fc::sha256 block_id_for(const uint32_t bnum, const std::string& nonce = {}) {
    if (block_ids.count(bnum))
       return block_ids[bnum];
--- a/plugins/state_history_plugin/tests/session_test.cpp
+++ b/plugins/state_history_plugin/tests/session_test.cpp
@@ -80,9 +80,13 @@ fc::datastream<ST>& operator>>(fc::datastream<ST>& ds, eosio::state_history::get
 
 //------------------------------------------------------------------------------
 
-fc::sha256 block_id_for(const uint32_t bnum) {
-   fc::sha256 m = fc::sha256::hash(fc::sha256::hash(std::to_string(bnum)));
+std::map<uint32_t, eosio::chain::block_id_type> block_ids;
+fc::sha256 block_id_for(const uint32_t bnum, const std::string& nonce = {}) {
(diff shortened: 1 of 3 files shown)


huangminghuang (approved):

plugins/state_history_plugin/tests/session_test.cpp
```diff
@@ -80,9 +80,13 @@ fc::datastream<ST>& operator>>(fc::datastream<ST>& ds, eosio::state_history::get
 
 //------------------------------------------------------------------------------
 
-fc::sha256 block_id_for(const uint32_t bnum) {
-   fc::sha256 m = fc::sha256::hash(fc::sha256::hash(std::to_string(bnum)));
+std::map<uint32_t, eosio::chain::block_id_type> block_ids;
```
linh2931, line 83: Do you want to use unordered_map?

plugins/state_history_plugin/tests/session_test.cpp
```diff
@@ -80,9 +80,13 @@ fc::datastream<ST>& operator>>(fc::datastream<ST>& ds, eosio::state_history::get
...
+fc::sha256 block_id_for(const uint32_t bnum, const std::string& nonce = {}) {
+   if (block_ids.count(bnum))
```
linh2931, line 85: Can change to contains when c++-20 is enabled.

linh2931 (approved):

plugins/state_history_plugin/tests/session_test.cpp
```diff
@@ -80,9 +80,13 @@ fc::datastream<ST>& operator>>(fc::datastream<ST>& ds, eosio::state_history::get
 
 //------------------------------------------------------------------------------
 
-fc::sha256 block_id_for(const uint32_t bnum) {
-   fc::sha256 m = fc::sha256::hash(fc::sha256::hash(std::to_string(bnum)));
+std::unordered_map<uint32_t, eosio::chain::block_id_type> block_ids;
+fc::sha256 block_id_for(const uint32_t bnum, const std::string& nonce = {}) {
+   if (block_ids.count(bnum))
+      return block_ids[bnum];
```
greg7mdp, line 86: Small inefficiency. This (or using .contains()) will do 2 lookups of the ley in the unordered_map. It would be more optimal to do:
```
if (auto it = block_ids.find(bnum); it != block_ids.end())
    return it->second;
```

greg7mdp (approved):

plugins/state_history_plugin/tests/session_test.cpp
heifner: Sure, it is test code and the second lookup should be fast as it just found it.

greg7mdp: Indeed. It seems I can't help myself be a stickler for details :-).
The second lookup will be faster only because the memory is in the processor cache, otherwise, if there are collisions, it may still iterate over a linked list and do comparisons.

Issue #951, opened 2023-04-03
Update EOS-VM-OC read-only trxs processing  due to the changes of execution window behavior


linh2931:
https://github.com/AntelopeIO/leap/pull/776 assumed read-only transactions to only run on read-only threads. https://github.com/AntelopeIO/leap/pull/901 changed that and read-only transactions can run on main thread too. 
EOS-VM-OC cache entries eviction assumed read-only trxs do not run on main thread. This needs to be changed.

Close  https://github.com/AntelopeIO/leap/issues/930 for better problem tracking.

PR #960, opened 2023-04-03
[4.0] Use is_write_window in eos-vm-oc main to determine if it is safe to update code cache

linh2931:
Resolve https://github.com/AntelopeIO/leap/issues/951.

We used to use `is_main_thread` to determine if it was safe to update eos-vm-cc code cache, as a transaction on the main thread did not run in parallel with read-only threads. Now a read-only transaction on the main thread can run in parallel with read-only threads. It is not safe for a read-only transaction on the main thread to modify code cache data structures. The solution is to use app's execution window. When in write window, all transactions are executed sequentially and no read-only threads are executing transactions.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -257,7 +257,7 @@ struct controller_impl {
    // which overwrites the same static wasmif, is used for eosvmoc too.
    wasm_interface  wasmif;  // used by main thread and all threads for EOSVMOC
    thread_local static std::unique_ptr<wasm_interface> wasmif_thread_local; // a copy for each read-only thread, used by eos-vm and eos-vm-jit
-   app_window_type app_window;
+   app_window_type app_window = app_window_type::write;
 
    typedef pair<scope_name,action_name>                   handler_key;
    map< account_name, map<handler_key, apply_handler> >   apply_handlers;
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -210,6 +210,12 @@ struct pending_state {
 };
 
 struct controller_impl {
+   enum class app_window_type {
+      write, // Only main thread is running; read-only threads are not running.
+             // All read-write and read-only tasks are sequentially executed.
+      read   // Main thread and read-only threads are running read-ony tasks in parallel.
+             // Read-write tasks are not being executed.
+   };
 
    // LLVM sets the new handler, we need to reset this to throw a bad_alloc exception so we can possibly exit cleanly
    // and not just abort.
@@ -251,6 +257,7 @@ struct controller_impl {
    // which overwrites the same static wasmif, is used for eosvmoc too.
    wasm_interface  wasmif;  // used by main thread and all threads for EOSVMOC
    thread_local static std::unique_ptr<wasm_interface> wasmif_thread_local; // a copy for each read-only thread, used by eos-vm and eos-vm-jit
+   app_window_type app_window;
 
    typedef pair<scope_name,action_name>                   handler_key;
    map< account_name, map<handler_key, apply_handler> >   apply_handlers;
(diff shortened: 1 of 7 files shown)


libraries/chain/controller.cpp
```diff
@@ -2694,6 +2701,16 @@ struct controller_impl {
...
 
+   void set_to_write_window() {
+      app_window = app_window_type::write;
+   }
+   void set_to_read_window() {
+      app_window = app_window_type::read;
+   }
+   bool is_write_window() const {
+      return app_window == app_window_type::write;
+   }
```
heifner, line 2712: I think you should only have one of `set_db_read_only_mode` and `set_to_read_window`. Can you just use the db read only state? `context.control.db().is_read_only_mode()` Would need to add the `is_read_only_mode()` accessor to chainbase.

heifner (changes requested):

libraries/chain/controller.cpp
```diff
@@ -251,6 +257,7 @@ struct controller_impl {
    // which overwrites the same static wasmif, is used for eosvmoc too.
    wasm_interface  wasmif;  // used by main thread and all threads for EOSVMOC
    thread_local static std::unique_ptr<wasm_interface> wasmif_thread_local; // a copy for each read-only thread, used by eos-vm and eos-vm-jit
+   app_window_type app_window;
```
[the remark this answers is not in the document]
linh2931: Thank you so much for checking my tests. I was away from my desk for a while. Will fix it now.

libraries/chain/controller.cpp
linh2931: We need to keep the both. `db_read_only_mode` is for protecting chainbase not to be updated while a read-only transaction is executed, either in read window or write window. It cannot tell whether the app() is in write window (tasks are running sequentially and safe to update global data like eos-vm-oc code cache) or in read window (tasks are running in parallel to each other).

heifner: That makes sense. I'm not thinking of anyway to make this cleaner at the moment. I do think this is better than a registered function that calls `app().executor().is_write_window()`. 

You can change all the `producer_plugin` calls of `app().executor().is_write_window()` with `chain.is_write_window()`.

spoonincode:
I think this should be pointed to `release/4.0` branch?

linh2931:
Thanks for noticing the bad base branch! Sorry about it. Too many PRs today.

spoonincode (approved):

libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
```diff
@@ -136,11 +137,11 @@ const code_descriptor* const code_cache_async::get_descriptor_for_code(const dig
...
    code_cache_index::index<by_hash>::type::iterator it = _cache_index.get<by_hash>().find(boost::make_tuple(code_id, vm_version));
    if(it != _cache_index.get<by_hash>().end()) {
-      if (is_main_thread())
+      if (is_write_window)
          _cache_index.relocate(_cache_index.begin(), _cache_index.project<0>(it));
       return &*it;
    }
-   if(!is_main_thread()) // on read-only thread
+   if(!is_write_window)
       return nullptr;
```
spoonincode, line 145: one initial concern I had was what would happen for a contract _only_ used for ROtrxs -- would it never get tiered up? @heifner pointed out to me that we do run ROtrxs during the write window as a low priority task, so it seems like a contract being used only for ROtrxs would indeed have an opportunity to be tiered up during that period

heifner (approved):

libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
linh2931: Thanks for digging this out. I had another concern last night when re-reading the code and was going to ask today. After a ROtrx's contract is evicted, how is it got back into the code cache again if it happens to be scheduled to run in the read window? Where is the code to make it wait until write window? Thanks again.

linh2931: I merge the PR now. Will do further work if necessary.

Issue #863, opened 2023-03-22
(non fatal) cpack error when creating leap-dev package on cmake <3.21


spoonincode:
When performing a `make package` on cmake <3.21 (such as on ubuntu 20.04):
```
CMake Error: failed to create symbolic link '/usr/local/share/leap_testing/bin': no such file or directory
```
This is definitely some fallout from #818 and now that I go back and read it it even points this error out. I'd like to see this resolved as it's a bad look.

oschwaldp-oci:
This has to do with the following code segment found here (https://github.com/AntelopeIO/leap/blob/5bfb498b45483155cb08ec061d60f9f23043ddbb/CMakeLists.txt#L256-L285), and in particular with the `else` block:

```if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.21)
   # Cmake versions < 3.21 did not support installing symbolic links to a directory via install(FILES ...)
   # Cmake 3.21 fixed this (https://gitlab.kitware.com/cmake/cmake/-/commit/d71a7cc19d6e03f89581efdaa8d63db216184d40)
   # If/when the lowest cmake version supported becomes >= 3.21, the else block as well as the postinit and prerm scripts
   # would be a good option to remove in favor of using the facilities in cmake / cpack directly.

   add_custom_target(link_target ALL
      COMMAND ${CMAKE_COMMAND} -E make_directory lib/python3/dist-packages
      COMMAND ${CMAKE_COMMAND} -E create_symlink ../../../share/leap_testing/tests/TestHarness lib/python3/dist-packages/TestHarness
      COMMAND ${CMAKE_COMMAND} -E make_directory share/leap_testing
      COMMAND ${CMAKE_COMMAND} -E create_symlink ../../bin share/leap_testing/bin)

   install(FILES ${CMAKE_CURRENT_BINARY_DIR}/lib/python3/dist-packages/TestHarness DESTINATION ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages COMPONENT dev EXCLUDE_FROM_ALL)
   install(FILES ${CMAKE_CURRENT_BINARY_DIR}/share/leap_testing/bin DESTINATION ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing COMPONENT dev EXCLUDE_FROM_ALL)
else()
   # The following install(CODE ...) steps are necessary for `make dev-install` to work on cmake versions < 3.21.
[...]

PR #972, opened 2023-04-05
[4.0] fix cpack error message

huangminghuang:
The PR fixes IS #863 to avoid cpack error messages with cmake version < 3.21


oschwaldp-oci (approved):

heifner:
I think this should be targeted to 4.0

scripts/install_testharness_symlinks.cmake
ScottBailey, line 1: Pedantically, I wonder if `${CMAKE_COMMAND}` should be escaped rather than populated by configure_file(), but I don't think it makes any practical difference.

If you were to do this you'd want to change `${CMAKE_INSTALL_FULL_LIBDIR}` to `@CMAKE_INSTALL_FULL_LIBDIR@` and similar for `${CMAKE_INSTALL_FULL_DATAROOTDIR}`. Additionally, the configure file command would need to include the `@ONLY` command.

ScottBailey (approved):
Consider the comment I made in `install_testharness_symlinks.cmake`.

Issue #969, opened 2023-04-05
remove/refactor global `WASM::check_limits` which is accessed from multiple threads


spoonincode:
There is a global here,
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/wasm-jit/Source/WASM/WASMSerialization.cpp#L178-L183
which is accessed throughout WAVM's parsing code.

This global `bool` is toggled via `scoped_skip_checks`
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/wasm-jit/Include/WASM/WASM.h#L12-L18

`scoped_skip_checks` is notably used within `wasm_interface_impl`, part of  `wasm_interface`
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/chain/include/eosio/chain/wasm_interface_private.hpp#L183

There are multiple `wasm_interface`s, one per a thread. This may result in unsafe access to this variable causing some sort of undesirable behavior.

I suspect we can remove the entire parsing inside of `wasm_interface_impl::get_instantiated_module()` because I don't believe there is a need to `parse_initial_memory()` any more.

The other use of  `scoped_skip_checks`,
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/chain/webassembly/runtimes/eos-vm-oc/compile_trampoline.cpp#L27
I'm not as sure about. I think we may need it there. But, _worst case_, we can leave the `scoped_skip_checks` in `compile_trampoline` because it's _process local_. It'd be nice to remove `check_limits` entirely though.

linh2931:
`instantiate_module` in eos-vm and eos-vm-oc do not use the `initial_memory` parameter. I gave it a try removing it plus removing `WASM::scoped_skip_checks`. Tests passed.

linh2931:
Because of the global `WASM::scoped_skip_checks`, it seems this should be fixed before rc3.

Issue #673, opened 2023-01-30
refactor EOS VM OC to use EOS VM's WASM parser


spoonincode:
EOS VM OC needs to be refactored to use EOS VM's WASM parser. This will allow commonality between the two implementations and is probably tech debt number one for EOS VM OC.

This is a carryover of eosnetworkfoundation/mandel#640 however I have dropped the discussion about wholesale removal of WAVM's wasm parser. It's not clear that is the correct approach: prior to `CONFIGURABLE_WASM_LIMITS2` I believe WAVM's parser is always clutched in in front of EOS VM's parser. We may want to maintain that behavior indefinitely just to ensure we don't goof up something. Leaving WAVM's parser in the code indefinitely doesn't seem too high overhead.

spoonincode:
Completely remove `WASM::check_limits` as part of this, see #969

PR #975, opened 2023-04-05
[4.0] Remove global WASM::check_limits and initial memory from instantiate_module

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/969

https://github.com/AntelopeIO/leap/issues/969 identified global scoped_skip_checks can be accessible by multiple read-only threads. This variable is not needed in `wasm_interface_impl::get_instantiated_module`. 

Another finding was initial memory argument in `instantiate_module` was not needed in all the  `instantiate_module` methods. I double checked that it was un-named in all the implementations of `instantiate_module`. Safe to remove.

This PR removed both of them.

### DIFF
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -156,11 +156,8 @@ namespace eosio { namespace chain {
             });
             trx_context.pause_billing_timer();
             IR::Module module;
-            std::vector<U8> bytes = {
-                (const U8*)codeobject->code.data(),
-                (const U8*)codeobject->code.data() + codeobject->code.size()};
             wasm_instantiation_cache.modify(it, [&](auto& c) {
-               c.module = runtime_interface->instantiate_module((const char*)bytes.data(), bytes.size(), code_hash, vm_type, vm_version);
+               c.module = runtime_interface->instantiate_module(codeobject->code.data(), codeobject->code.size(), code_hash, vm_type, vm_version);
             });
          }
          return it->module;
--- a/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
+++ b/libraries/chain/include/eosio/chain/wasm_interface_private.hpp
@@ -109,24 +109,6 @@ namespace eosio { namespace chain {
          return it != wasm_instantiation_cache.end();
       }
 
-      std::vector<uint8_t> parse_initial_memory(const Module& module) {
-         std::vector<uint8_t> mem_image;
-
-         for(const DataSegment& data_segment : module.dataSegments) {
-            EOS_ASSERT(data_segment.baseOffset.type == InitializerExpression::Type::i32_const, wasm_exception, "");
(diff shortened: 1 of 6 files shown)


libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -180,7 +162,6 @@ namespace eosio { namespace chain {
             try {
                Serialization::MemoryInputStream stream((const U8*)bytes.data(),
```
spoonincode, line 163: need to remove all of this section

spoonincode (changes requested):

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
linh2931: oha, thanks! that's the reason of the test failure.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -177,20 +159,8 @@ namespace eosio { namespace chain {
             std::vector<U8> bytes = {
                 (const U8*)codeobject->code.data(),
                 (const U8*)codeobject->code.data() + codeobject->code.size()};
```
spoonincode, line 161: I think we can get rid of `bytes` too? could be up to 1MB, on the big side for a needless copy

linh2931: You are right. `bytes` is not needed any more. Removing it saves an unnecessary copy.

spoonincode:
@heifner my hunch is this solves https://github.com/AntelopeIO/leap/pull/964#issuecomment-1497791395 because it ought to make WAVM only ever touched on a single thread (...in a process). If it's easy to retest would be interesting to confirm.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -174,23 +156,8 @@ namespace eosio { namespace chain {
             });
             trx_context.pause_billing_timer();
             IR::Module module;
```
spoonincode, line 158: one more line here that can get the chop

heifner:
> @heifner my hunch is this solves [#964 (comment)](https://github.com/AntelopeIO/leap/pull/964#issuecomment-1497791395) because it ought to make WAVM only ever touched on a single thread (...in a process). If it's easy to retest would be interesting to confirm.

Yes, it was rather easy to reproduce, I'll re-test with this PR.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
linh2931: Sorry I should be more thorough. Don't know why my build did not show unused variables warning.

spoonincode (approved):
Looks good. I've made note in #673 to remove `WASM::check_limits` entirely at that time

heifner (approved):

Issue #970, opened 2023-04-05
parallel ROtrx dispatched to baseline runtime while OC enabled may not be thread safe


spoonincode:
In the case of EOS VM OC being enabled, only a single `wasm_interface wasmif` is created in controller
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/chain/controller.cpp#L2690-L2700
and remember that this is the only `wasm_interface` that is accessible
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/chain/controller.cpp#L2714-L2723
Recall how execution for a wasm is kicked off
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/chain/wasm_interface.cpp#L89-L113
Assuming `my->eosvmoc` is populated (that is, OC tier up enable), which it will be when `--eos-vm-oc-enable`, then nominally it tries to get the `code_descriptor`
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/chain/wasm_interface.cpp#L96
And then executes that entry
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/chain/wasm_interface.cpp#L107
Keep in mind that `exec` is `thread_local`
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/chain/include/eosio/chain/wasm_interface_private.hpp#L66
and `cc` is expected to be safe to access from multiple threads as long as `is_write_window() == false`. So this is correctly thread safe nominally.

However, if obtaining the `code_descriptor` fails -- such as when OC is still compiling the contract and/or has blacklisted a compiled contract due to subjective failure -- then the code will fall back to the baseline runtime
https://github.com/AntelopeIO/leap/blob/d7eb6d7f9fa150a7e2f0e648325ce58ec35b79f9/libraries/chain/wasm_interface.cpp#L112
This appears to then enter code that is _not_ thread safe.

linh2931:
[...]

PR #984, opened 2023-04-06
[4.0] Fix ROtrx dispatched to baseline runtime when compiled code not ready

linh2931:
Resolve https://github.com/AntelopeIO/leap/issues/970

https://github.com/AntelopeIO/leap/issues/970 identified a problem when `get_descriptor_for_code` failed, the execution will go on the non-thread-safe path.

This PR implements a workaround. `get_descriptor_for_code` returns failure reasons when it fails. Read-only trxs will retry in the future or fail permanently.


linh2931:
@spoonincode and @heifner, please double check if my assignment of failure reasons in `get_descriptor_for_code` is accurate or not. Thanks.

libraries/chain/wasm_interface.cpp
```diff
@@ -93,7 +93,15 @@ namespace eosio { namespace chain {
...
+            cd = my->eosvmoc->cc.get_descriptor_for_code(code_hash, vm_version, context.control.is_write_window(), failure);
+            if (!cd && context.trx_context.is_read_only()) {
+               if (failure == eosio::chain::eosvmoc::get_cd_failure::temporary) {
+                  EOS_ASSERT(false, ro_trx_temporary, "get_descriptor_for_code failed with temporary failure");
+               } else {
+                  EOS_ASSERT(false, ro_trx_permanent, "get_descriptor_for_code failed with permanent failure");
+               }
+            }
          }
          catch(...) {
```
heifner, line 106: This is not quite right as the `catch(...)` catches these `EOS_ASSERT`. This should be more like:
```
index 4cbdfdbadf..5cd75786f0 100644
--- a/libraries/chain/wasm_interface.cpp
+++ b/libraries/chain/wasm_interface.cpp
@@ -92,16 +92,9 @@ namespace eosio { namespace chain {
 #ifdef EOSIO_EOS_VM_OC_RUNTIME_ENABLED
       if(my->eosvmoc) {
          const chain::eosvmoc::code_descriptor* cd = nullptr;
+         eosio::chain::eosvmoc::get_cd_failure failure = eosio::chain::eosvmoc::get_cd_failure::temporary;
          try {
-            eosio::chain::eosvmoc::get_cd_failure failure = eosio::chain::eosvmoc::get_cd_failure::temporary;
             cd = my->eosvmoc->cc.get_descriptor_for_code(code_hash, vm_version, context.control.is_write_window(), failure);
-            if (!cd && context.trx_context.is_read_only()) {
-               if (failure == eosio::chain::eosvmoc::get_cd_failure::temporary) {
-                  EOS_ASSERT(false, ro_trx_temporary, "get_descriptor_for_code failed with temporary failure");
-               } else {
-                  EOS_ASSERT(false, ro_trx_permanent, "get_descriptor_for_code failed with permanent failure");
-               }
-            }
          }
          catch(...) {
             //swallow errors here, if EOS VM OC has gone in to the weeds we shouldn't bail: continue to try and run baseline
@@ -114,6 +107,12 @@ namespace eosio { namespace chain {
          if(cd) {
             my->eosvmoc->exec->execute(*cd, my->eosvmoc->mem, context);
             return;
+         } else if (context.trx_context.is_read_only()) {
+            if (failure == eosio::chain::eosvmoc::get_cd_failure::temporary) {
+               EOS_ASSERT(false, ro_trx_temporary, "get_descriptor_for_code failed with temporary failure");
+            } else {
+               EOS_ASSERT(false, ro_trx_permanent, "get_descriptor_for_code failed with permanent failure");
+            }
          }
       }
 #endif
```

libraries/chain/include/eosio/chain/exceptions.hpp
```diff
@@ -276,6 +276,10 @@ namespace eosio { namespace chain {
                                     3040017, "Transaction includes disallowed extensions (invalid block)" )
       FC_DECLARE_DERIVED_EXCEPTION( tx_resource_exhaustion, transaction_exception,
                                     3040018, "Transaction exceeded transient resource limit" )
+      FC_DECLARE_DERIVED_EXCEPTION( ro_trx_temporary, transaction_exception,
+                                    3040019, "Read-only transaction temporary failure" )
+      FC_DECLARE_DERIVED_EXCEPTION( ro_trx_permanent, transaction_exception,
+                                    3040020, "Read-only transaction permanent failure" )
```
heifner, line 282: Seems like these should go under `resource_exhausted_exception` instead.
Also I think they could have a better description, maybe something like:
`Read-only transaction eos-vm-oc compile [temporary|permanent] failure.`
Also maybe call them `_ro_trx_vm_oc_compile_temporary` or something like that.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/code_cache.hpp
```diff
@@ -37,6 +37,12 @@ using allocator_t = bip::rbtree_best_fit<bip::null_mutex_family, bip::offset_ptr
 
 struct config;
 
+// get_descriptor_for_code failure reasons
+enum class get_cd_failure {
+   temporary, // oc compile not done yet, users like read-only trxs can retry
+   permanent  // oc will not start, users should not retry
+};
```
heifner, line 44: This can be moved into `code_cache_base` class I think and then it could be `code_cache_base::cd_failure` which seems more descriptive and easier to read/understand.

libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
```diff
@@ -141,33 +141,44 @@ const code_descriptor* const code_cache_async::get_descriptor_for_code(const dig
...
+   if(!codeobject) { //should be impossible right?
+      failure = get_cd_failure::permanent; // Compile will not start
       return nullptr;
+   }
 
    _outstanding_compiles_and_poison.emplace(ct, false);
    std::vector<wrapped_fd> fds_to_pass;
    fds_to_pass.emplace_back(memfd_for_bytearray(codeobject->code));
    write_message_with_fds(_compile_monitor_write_socket, compile_wasm_message{ ct }, fds_to_pass);
+   failure = get_cd_failure::temporary; // Compile might be be done yet
```
heifner, line 181: I think all these comments should be `Compile might not be done yet` instead of `Compile might be be done yet`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2337,6 +2338,9 @@ producer_plugin_impl::handle_push_result( const transaction_metadata_ptr& trx,
          if ( !trx->is_read_only() )
             pr.block_exhausted = block_is_exhausted(); // smaller trx might fit
          pr.trx_exhausted = true;
+      } else if (trace->except->code() == ro_trx_temporary::code_value) {
+         fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Read-only trx execution temporary failure, tx: ${txid} RETRYING", ("txid", trx->id()));
+         pr.retry = true;
```
heifner, line 2343: Instead of this `else if`, I think you should add `ro_trx_temporary::code_value` to `exception_is_exhausted` list and consider this just another exhausted case.

Then the above `fc_dlog(_trx_failed_trace_log` in the `exception_is_exhausted` `if` block should be `fc_dlog( is_transient ? _transient_trx_failed_trace_log : _trx_failed_trace_log`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -333,6 +333,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          bool block_exhausted = false;
          bool trx_exhausted = false;
          bool failed = false;
+         bool retry = false;
```
heifner, line 336: Don't think this is needed, see other comment.

heifner (changes requested):

libraries/chain/wasm_interface.cpp
linh2931: Thanks for raising the issue and detailed solution.

libraries/chain/include/eosio/chain/exceptions.hpp
linh2931: I was thinking of generic exceptions for read-only transactions. But it is better to be more specific.

libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
linh2931: "be be" should have "not be".

plugins/producer_plugin/producer_plugin.cpp
linh2931: Thanks. This simplifies coding.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2329,7 +2330,7 @@ producer_plugin_impl::handle_push_result( const transaction_metadata_ptr& trx,
       }
       if( exception_is_exhausted( *trace->except ) ) {
          if( _pending_block_mode == pending_block_mode::producing ) {
-            fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Block ${block_num} for producer ${prod} COULD NOT FIT, tx: ${txid} RETRYING ",
+            fc_dlog(trx->is_transient() ? _transient_trx_failed_trace_log : _trx_failed_trace_log, "[TRX_TRACE] Block ${block_num} for producer ${prod} COULD NOT FIT, tx: ${txid} RETRYING ",
                     ("block_num", chain.head_block_num() + 1)("prod", get_pending_block_producer())("txid", trx->id()));
          } else {
             fc_dlog(_trx_failed_trace_log, "[TRX_TRACE] Speculative execution COULD NOT FIT tx: ${txid} RETRYING", ("txid", trx->id()));
```
heifner, line 2336: `fc_dlog(trx->is_transient() ? _transient_trx_failed_trace_log : _trx_failed_trace_log` needed here as well.

heifner (approved):

spoonincode (approved):

Issue #613, opened 2023-01-04
consider making cleos easier/clearer to set wildcard linkauth


spoonincode:
cleos' `set action permission` (which performs `linkauth`) states
```
Set paramaters dealing with account permissions
Usage: cleos set action permission [OPTIONS] account code type requirement

Positionals:
  account TEXT REQUIRED  The account to set/delete a permission authority for
  code TEXT REQUIRED     The account that owns the code for the action
  type TEXT REQUIRED     The type of the action
  requirement TEXT REQUIRED
                         [delete] NULL, [set/update] The permission name require for executing the given action
```
And [docs](https://github.com/eosnetworkfoundation/eos-system-contracts/blob/7cec470b17bd53b8c78465d4cbd889dbaf1baffb/contracts/eosio.system/include/eosio.system/native.hpp#L206) also state that "Link authorization action assigns _a specific action_ from a contract to a permission you have created" (emphasis mine).

Antelope also allows a wildcard linkauth as well though. But, despite `cleos get account` showing these wildcard linkauths clearly, there doesn't seem an easy or documented way to convey that to `set action permission`? The only way to convey this wildcard is through an empty `type` string. that is
```
cleos set action permission account contract '' eosio.any
```
then you can see in `cleos get account`
```
permission links:
     eosio.any:
          contract::*
```
(which is actually quite nicely displayed!)

Maybe `type` can accept '*'? But that has shell expansion grief. Maybe just update cleos' help text to state an empty string is a wildcard? Is there precedent somewhere else we can pull from? Passing an empty string like this on the command line is a little wonky.

heifner:
We use `NULL` for the requirement text. Seems like we could use uppercase `ALL` as it has to be a valid `eosio::name` and so `ALL` could be used.

spoonincode:
Yeah, that's not a bad simple change that is similar to existing behavior. To be clear though
https://github.com/AntelopeIO/leap/blob/61b2f98a6fc8e34214847d12cf05c36aa622bb56/programs/cleos/main.cpp#L1066
so "null" is allowed which means you can't linkauth to a permission named "null" today. Likewise for an action named "all" if making this change; so technically breaking, but probably fine.

heifner:
I think we only allow `ALL` and not `all`. Null should have only allowed `NULL` and not `null` in my opinion because of that.

PR #991, opened 2023-04-07
cleos wildcard linkauth

huangminghuang:
The PR address IS #613 to
- make cleos wildcard linkauth easier by specify 'ALL' for the type parameter of `cleos set action permission`,
- disallow 'null' (allows 'NULL' only) for the requirement parameter of `cleos set action permission`.

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -1057,7 +1057,7 @@ struct set_action_permission_subcommand {
       auto permissions = actionRoot->add_subcommand("permission", localized("Set parameters dealing with account permissions"));
       permissions->add_option("account", accountStr, localized("The account to set/delete a permission authority for"))->required();
       permissions->add_option("code", codeStr, localized("The account that owns the code for the action"))->required();
-      permissions->add_option("type", typeStr, localized("The type of the action"))->required();
+      permissions->add_option("type", typeStr, localized("The type of the action [use ALL for all types]"))->required();
       permissions->add_option("requirement", requirementStr, localized("[delete] NULL, [set/update] The permission name require for executing the given action"))->required();
 
       add_standard_transaction_options_plus_signing(permissions, "account@active");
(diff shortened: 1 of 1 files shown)


heifner (approved):

programs/cleos/main.cpp
```diff
@@ -1057,16 +1057,16 @@ struct set_action_permission_subcommand {
       auto permissions = actionRoot->add_subcommand("permission", localized("Set parameters dealing with account permissions"));
       permissions->add_option("account", accountStr, localized("The account to set/delete a permission authority for"))->required();
       permissions->add_option("code", codeStr, localized("The account that owns the code for the action"))->required();
-      permissions->add_option("type", typeStr, localized("The type of the action"))->required();
+      permissions->add_option("type", typeStr, localized("The type of the action [use ALL for all types]"))->required();
       permissions->add_option("requirement", requirementStr, localized("[delete] NULL, [set/update] The permission name require for executing the given action"))->required();
```
spoonincode, line 1061: I think it may be worth considering updating the text of this item to be similar to the formatting and syntax above. Something like
> The permission name required for executing the given action [use NULL to delete]

or something

programs/cleos/main.cpp
```diff
@@ -1057,16 +1057,16 @@ struct set_action_permission_subcommand {
       auto permissions = actionRoot->add_subcommand("permission", localized("Set parameters dealing with account permissions"));
       permissions->add_option("account", accountStr, localized("The account to set/delete a permission authority for"))->required();
       permissions->add_option("code", codeStr, localized("The account that owns the code for the action"))->required();
-      permissions->add_option("type", typeStr, localized("The type of the action"))->required();
+      permissions->add_option("type", typeStr, localized("The type of the action [use ALL for all types]"))->required();
```
spoonincode, line 1060: do we think this should change `type` to `action` here? `type` is pretty weird and atypical for our modern terminology I think

heifner: I found `type` very odd as well, @huangminghuang actually asked me the same question. I agree `action` would be much better name. Didn't know if we want to update that.

spoonincode: maybe it makes sense to keep the argument name as "type", since that's the name in the json that is sent? :man_shrugging: I'm fine either way with the new description

spoonincode (approved):

PR #994, opened 2023-04-10
Small but measurable improvement to executor queue

heifner:
For short running tasks, e.g. token transfer transaction, it is faster to remain in the executor before returning to the `appbase` `exec()` loop. There is a trade-off here on how long to potentially delay a high priority task from executing. This PR uses 3ms. 5ms or even 10ms would be better, but this PR goes with the more conservative delay. Opinions on the delay amount welcome.

### DIFF
--- a/libraries/custom_appbase/include/eosio/chain/application.hpp
+++ b/libraries/custom_appbase/include/eosio/chain/application.hpp
@@ -2,6 +2,7 @@
 
 #include <appbase/application_base.hpp>
 #include <eosio/chain/exec_pri_queue.hpp>
+#include <chrono>
 #include <mutex>
 
 /*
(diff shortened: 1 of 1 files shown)


spoonincode:
Do you have metrics on how much faster?

heifner:
> Do you have metrics on how much faster?

For Perf Benchmark of TPS (token transfer) ~3% for the 3ms. More for larger delays. I ran a 10ms months ago and saw something like 9%, but I would need to re-run as a lot has changed since then. The 3% was for `release/4.0` with `3ms`.

libraries/custom_appbase/include/eosio/chain/application.hpp
```diff
@@ -52,23 +57,41 @@ class two_queue_executor {
...
-      return read_only_queue_.execute_highest_locked(true);
+      // execute for at least minimum runtime
+      const auto end = std::chrono::high_resolution_clock::now() + std::chrono::milliseconds(minimum_runtime_ms);
+
+      bool more = false;
+      while (true) {
+         more = read_only_queue_.execute_highest_locked( true );
+         if (!more || std::chrono::high_resolution_clock::now() > end)
+            break;
+      }
```
greg7mdp, line 93: Matter of preference of course!
```suggestion
       do {
         more = read_only_queue_.execute_highest_locked( true );
       } while (more && std::chrono::high_resolution_clock::now() < end);
```

greg7mdp (approved):

Issue #980, opened 2023-04-05
Restore deprecated read-mode=speculative


heifner:
PR #414 removed `read-mode` `speculative` & `disable-api-persisted-trx`.
Potential use-cases for `read-mode=speculative` have been brought forth by the community.
Restore the removed `read-mode=speculative` but keep `read-mode=head` as the default.
`disable-api-persisted-trx` will not be restored.

PR #995, opened 2023-04-10
 [4.0 -> main] make eosio_exit impl more direct 

### LINKED ISSUE #956: investigate "this assumes only one [wasm] can possibly run at a time" comment in context of ROtrx
A `wasm_runtime_interface` exposes
https://github.com/AntelopeIO/leap/blob/d382d71c7e1b9ec9e8e14fae5d0b80e1f2dfee81/libraries/chain/include/eosio/chain/webassembly/runtime_interface.hpp#L26-L27
This is pretty interesting as it suggests this function only has defined behavior if a single wasm is executing, which of course is not the case now with ROtrx.

This is called via `eosio_exit` host function.
https://github.com/AntelopeIO/leap/blob/d382d71c7e1b9ec9e8e14fae5d0b80e1f2dfee81/libraries/chain/webassembly/cf_system.cpp#L48-L50
via
https://github.com/AntelopeIO/leap/blob/d382d71c7e1b9ec9e8e14fae5d0b80e1f2dfee81/libraries/chain/wasm_interface.cpp#L115-L117

So it can absolutely be called via multiple ROtrx at a time.

But... Most likely this is a relic of the way WAVM and/or wabt runtimes were integrated. OC stubs this out as a do-nothing, and EOS VM looks like it throws an exception that likely has proper semantics. I think this "singleton" `immediately_exit_currently_running_module()` should be completely removed.

spoonincode:
main merge of #993 

> This is simply a clean up change to try and make it more explicit how `eosio_exit` works by eliminating some needless abstraction that contains a deceptive comment.
> 
> When the host function `eosio_exit` is called the contract execution is considered successfully complete immediately.
> 
> I don't remember the exact historical reasons the code was written the way it was, so I'll go with a plausible theory.
> 
>A reasonable way to accomplish the required behavior is to throw an exception in the `eosio_exit` impl that is caught as a do-nothing in the appropriate location (ideally immediately after the line that 'entered' wasm execution). This catch continues to live on
https://github.com/AntelopeIO/leap/blob/dfe829e7dc0d7420e73905cba7d520a3f8dc8769/libraries/chain/apply_context.cpp#L96-L98
> 
> However, at some point the performance of this exception was quite poor. Early in EOS' chain history `eosio_exit` is used _a lot_. And old EOSIO/Leap implementations had extremely poor exception handling performance when WAVM was being used (the process could spend >50% of its time just handling the `eosio_exit` host function exceptions!). At some point I did a refactor skipping the exception in WAVM's case to improve performance. Since there isn't an object that represents the "currently running wasm instance", instead `wasm_interface` gained a new function which purpose was to exit the currently running wasm instance, and then each runtime (like WAVM, etc) could implement this as appropriate.
> 
> That's a bad abstraction in the new world of multiple wasm instances executing at the same time.
> 
> So this PR restores what was likely similar to the original implementation: just simply throw the exception in `eosio_exit` host function impl. However, OC continues to use an exception-free optimization for `eosio_exit` by handling that host function internal to its implementation. So note that on the host function's implementation.

Resolves #956

### DIFF
--- a/docs/01_nodeos/03_plugins/chain_plugin/index.md
+++ b/docs/01_nodeos/03_plugins/chain_plugin/index.md
@@ -120,7 +120,7 @@ Config Options for eosio::chain_plugin:
                                         applied to them (may specify multiple
                                         times)
   --read-mode arg (=head)               Database read mode ("head",
-                                        "irreversible").
+                                        "irreversible", "speculative").
                                         In "head" mode: database contains state
                                         changes up to the head block;
                                         transactions received by the node are
@@ -131,7 +131,14 @@ Config Options for eosio::chain_plugin:
                                         received via the P2P network are not
                                         relayed and transactions cannot be
                                         pushed via the chain API.
-
+                                        In "speculative" mode: (DEPRECATED:
+                                        head mode recommended) database
+                                        contains state changes by transactions
+                                        in the blockchain up to the head block
+                                        as well as some transactions not yet
+                                        included in the blockchain;
+                                        transactions received by the node are
+                                        relayed if valid.                                        
   --api-accept-transactions arg (=1)    Allow API transactions to be evaluated
                                         and relayed if valid.
   --validation-mode arg (=full)         Chain validation mode ("full" or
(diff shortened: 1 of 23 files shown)


heifner (approved):

spoonincode:
> Any reason not to head of `main` but to `GH-980-restore-read-mode-speculative-main`?

just stacking up the 4.0->main merges in a clean order

linh2931:
Oh, I didn't know this trick. Thanks.

linh2931 (approved):

Issue #859, opened 2023-03-21
3.1 vs 3.2 HTTP server behavior change when `Host` header value not allowed


spoonincode:
On 3.1 when an HTTP client's `Host` header is not allowed by `http_plugin`, an empty 400 response is returned to the client. On 3.2+, the connection is simply closed without any headers sent to the client.

This is easy to reproduce by simply
```
nodeos --plugin eosio::chain_api_plugin
```
and then
```
curl -v http://localhost:8888/v1/chain/get_info
```
(`http_plugin` listens on 127.0.0.1 by default, and thus will not accept a `Host: localhost:8888` header)

IMO an error response should be sent (that is, the behavior change in 3.2 is undesirable). I would go further and suggest returning a small body describing the error would be very helpful.

bhazzard:
Patched into 3.2, bring forward to 4.0 and beyond.

Caused confusion internally as well. Lack of information on why connection closes.

ericpassmore:
Sorry mouse slip closing issue, and I shouldn’t have opened it. I ran into this issue last night. 

I feel http://local host should work. That’s a separate issue.

PR #996, opened 2023-04-10
[3.2] response to disallowed host

huangminghuang:
This PR fixes the problem where HTTP plugin doesn't response to requests with disallowed HTTP HOST header.

Resolve IS #859

### DIFF
--- a/tests/http_plugin_test.py
+++ b/tests/http_plugin_test.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
 
 from TestHarness import Cluster, Node, ReturnType, TestHelper, Utils
+import urllib.request
+import sys
 
 ###############################################################
 # http_plugin_tests.py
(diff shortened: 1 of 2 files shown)


heifner (changes requested):
Can you add a test case for this.

heifner (approved):

tests/http_plugin_test.py
```diff
@@ -35,21 +37,34 @@
     TestHelper.printSystemInfo("BEGIN")
 
     Print("Stand up cluster")
-    # standup cluster with HTTPS enabled, but not configured
-    # HTTP should still work
-    extraArgs={ 0 : "--https-server-address 127.0.0.1:5555" }
-    # specificExtraNodeosArgs=extraArgs
 
-    if cluster.launch(dontBootstrap=True, loadSystemContract=False) is False:
+    if cluster.launch(dontBootstrap=True, loadSystemContract=False, specificExtraNodeosArgs = {0: "--http-validate-host true"}) is False:
```
spoonincode, line 41: Do you really need `--http-validate-host true` here? It's the default.

huangminghuang: Yes, I have to. Somewhere in the TestHarness changed it to false.

spoonincode: ah okay, it's interesting because you can't `--http-validate-host true --http-validate-host true` nor
```
http-validate-host = true 
http-validate-host = true 
```

But you can `http-validate-host = true` & `--http-validate-host true`

PR #1011, opened 2023-04-11
Replace boost/fc filesystem

### LINKED ISSUE #620: replace `boost::filesystem` with `std::filesystem`
This is a transfer of eosnetworkfoundation/mandel#321& eosnetworkfoundation/mandel-fc#47

In general when the C++ standard library and Boost provide the same functionality we want to use the standard library. History has shown us it's just less grief to do it that way. Migration to std::filesystem has been blocked in the past due to macOS 10.14 not supporting std::filesystem. Obviously that's not an issue any longer, and afaik even Ubuntu 18.04 has a working std::filesystem.

Effort here would be to:
* Remove fc's wrapper over boost::filesystem and just use std::filesystem directly instead of through fc.
* Provide temporary directory support that uses std::filesystem (a refactor for fc::temp_directory etc)
* Deal with any other minor differences (possibly `remove_all()` etc)

I don't expect this to be a simple 1:1 regex conversion, as it's not clear the fc wrapper was 1:1 in all cases.

There are, however, some additional complications. Older stdlibs (which we do support and do compile Leap just fine today) sometimes require extra linker flags for std::filesystem. libstdc++7 (which will be used on Ubuntu 18.04 with newer clang from Ubuntu's packages, and does compile Leap fine today this way) actually puts std::filesystem in an experimental namespace. A cmake `INTERFACE` target can wrangle these differences but then we have submodules -- something like appbase & chainbase should also be moved to std::filesystem. That means we may end up needing a new submodule that provides a cmake `INTERFACE` target that can be used by all submodules & Leap.

huangminghuang:
This PR
- replaces all usage of `boost` and `fc` filesystem with `std::filesystem`,
- creates a new class `temp_cfile` to replace the usages of `boost::filesystem::unique_path` which doesn't exists in `std::filesystem` due to safety concern,

Includes
https://github.com/AntelopeIO/appbase/pull/15
https://github.com/AntelopeIO/appbase/pull/16
https://github.com/AntelopeIO/appbase/pull/17
https://github.com/AntelopeIO/appbase/pull/18
https://github.com/AntelopeIO/appbase/pull/19
https://github.com/AntelopeIO/appbase/pull/20
https://github.com/AntelopeIO/appbase/pull/21
https://github.com/AntelopeIO/chainbase/pull/5
https://github.com/AntelopeIO/chainbase/pull/6
https://github.com/AntelopeIO/chainbase/pull/7
https://github.com/AntelopeIO/chainbase/pull/8
https://github.com/AntelopeIO/chainbase/pull/9

Resolves #620


spoonincode:
This change will require gcc 9.1+ or clang 9+ (technically libstdc++ & libc++). Please increase the minimum version in the root cmake,
https://github.com/AntelopeIO/leap/blob/69c4e006137d1c58f0aeff25ab84a7e0f09226fe/CMakeLists.txt#L38-L47
(I think it's okay to remove `AppleClang` completely from this check; what's there isn't correct anyways)

spoonincode:
Can you test if it's possible to remove this workaround now
https://github.com/AntelopeIO/leap/blob/69c4e006137d1c58f0aeff25ab84a7e0f09226fe/.github/workflows/build.yaml#L134-L136

spoonincode:
There is at least one remaining `boost::filesystem` here
https://github.com/AntelopeIO/leap/blob/6726ca6ed388b322ba3bf77b52e470681e17606a/plugins/trace_api_plugin/include/eosio/trace_api/store_provider.hpp#L8
https://github.com/AntelopeIO/leap/blob/6726ca6ed388b322ba3bf77b52e470681e17606a/plugins/trace_api_plugin/include/eosio/trace_api/store_provider.hpp#L16

This causes a compile error in gcc12.

unittests/deep_mind_tests.cpp
```diff
@@ -100,14 +104,15 @@ BOOST_FIXTURE_TEST_CASE(deep_mind, deep_mind_tester)
       return std::find(argv, argv + argc, std::string("--save-dmlog")) != (argv + argc);
    }();
 
+   auto log_output_path = tmp.file().get_file_path();
+
    if(save_log)
    {
-      // Cannot use fc::copy as it does not copy to an existing destination file 
-      fc::rename(log_output.path().preferred_string(), DEEP_MIND_LOGFILE);
+      std::filesystem::copy(log_output_path, DEEP_MIND_LOGFILE, std::filesystem::copy_options::update_existing);
```
spoonincode, line 111: This `fc::rename()` became `fs::copy(.. update_existing)`. The semantics are clearly different between the two. Any reason for the change?

plugins/trace_api_plugin/configuration_utils.cpp
```diff
@@ -7,15 +7,15 @@
...
    {
       fc::variant abi_variant;
-      auto abi_path = fc::path(file_name);
+      auto abi_path = std::filesystem::path(file_name);
       if (abi_path.is_relative()) {
          abi_path = data_dir / abi_path;
       }
 
-      EOS_ASSERT(fc::exists(abi_path) && !fc::is_directory(abi_path), chain::plugin_config_exception, "${path} does not exist or is not a file", ("path", abi_path.generic_string()));
+      EOS_ASSERT(std::filesystem::exists(abi_path) && !std::filesystem::is_directory(abi_path), chain::plugin_config_exception, "${path} does not exist or is not a file", ("path", abi_path.generic_string()));
```
spoonincode, line 18: Thoughts on adding a `to_variant(fs::path)` so we don't need to put `generic_string()` in all these log/asserts?

heifner: +1. Although a new logging framework is likely going to make us add `fmt::formatter<>` for all our types.

libraries/libfc/include/fc/io/cfile.hpp
```diff
@@ -221,11 +235,43 @@ class cfile {
...
    detail::unique_file   _file;
 };
 
+class temp_cfile {
+   cfile _impl;
+public:
+   temp_cfile(const char* mode = "wb"){
+      std::filesystem::path template_path{ std::filesystem::temp_directory_path() / "fc-XXXXXX" };
+      char                  tmp_buf[4096];
+      strncpy(tmp_buf, template_path.c_str(), 4096);
```
spoonincode, line 249: This can result in `tmp_buf` lacking a null terminator.

plugins/trace_api_plugin/configuration_utils.cpp
heifner: Looks like `to_variant` is already provided.

programs/cleos/program_location.hpp
```diff
@@ -0,0 +1,65 @@
+#pragma once
+#include <boost/predef/os.h>
+#include <filesystem>
+
+// adapted from boost source boost/dll/detail/posix/program_location_impl.hpp
```
spoonincode, line 5: Can we continue to use boost DLL with `BOOST_DLL_USE_STD_FS`, so we don't need this local reimplementation?

unittests/deep_mind_tests.cpp
huangminghuang: `log_output_path` is derived from `tmp` of type `temp_cfile object which` would remove the file when `tmp` is out of scope. Simply rename the file would cause the desctructor tryiing to remove an non-existent file.

spoonincode:
Includes
AntelopeIO/appbase#15
AntelopeIO/appbase#16
AntelopeIO/appbase#17
AntelopeIO/appbase#18
AntelopeIO/appbase#19
AntelopeIO/appbase#20
AntelopeIO/appbase#21
AntelopeIO/chainbase#5
AntelopeIO/chainbase#6
AntelopeIO/chainbase#7
AntelopeIO/chainbase#8
AntelopeIO/chainbase#9

spoonincode (approved):

heifner (changes requested):

libraries/libfc/include/fc/io/cfile.hpp
```diff
@@ -221,11 +235,42 @@ class cfile {
...
 
+class temp_cfile {
+   cfile _impl;
+public:
+   temp_cfile(const char* mode = "wb"){
+      std::filesystem::path template_path{ std::filesystem::temp_directory_path() / "fc-XXXXXX" };
+      std::string tmp = template_path.string();
+      int fd = mkstemp(tmp.data());
+      if (fd == -1)
+         throw std::system_error(errno, std::generic_category(), __PRETTY_FUNCTION__);
```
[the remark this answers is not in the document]
huangminghuang: All the existing `ios_base::failure` get the error from the `ferror` API which apply only to C file pointer. Here, `mkstemp`  is called before the file pointer exists and the error is derived from `errno`.

heifner (approved):

PR #1020, opened 2023-04-12
P2P:  Inform peer if not accepting transactions

### LINKED ISSUE #767: nodeos is sending extra p2p traffic when not required - need new configuration option(s)
In our environment, our API nodes don't need to receive p2p transactions, but they need to send it out.
More specifically, they are configured with: `p2p-accept-transactions = false`, but they accept transactions via API.

Each API node is connected to multiple "sentry" nodes (aka transaction relay nodes). They are bi-directionally connected to each other via `p2p-peer-address = n.n.n.n` so either side can initiate the connection (one gets dropped as duplicate).

- We need transactions to flow from API -> Sentry. (any transaction pushed to the API endpoint flows to the network)
- We don't want transactions to flow from Sentry -> API (they get dropped on receipt because `p2p-accept-transactions = false`)

The sentry nodes are wasting a lot of time and network bandwidth sending transactions to nodes that don't need them.  On WAX, that is currently ~3MiB/sec network traffic between 1 sentry node and the multiple API nodes we have running.

Alternately we could change our infrastructure to have the API nodes not accept transactions via HTTP and use a load balancer to route the transaction related HTTP traffic to the sentry nodes directly.  Then we could use ":blk" on the p2p connection to avoid sending any transactions.

heifner:
Optimize network traffic by informing peers when node is configured for `p2p-accept-transactions=false`. Since the transactions will be immediately dropped, no reason for the peer to send them. 
If configured to not accept transactions, `p2p-accept-transactions=false`, then inform peer the node is a blocks only connection.

Resolves #767

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3598,7 +3598,8 @@ namespace eosio {
          hello.token = sha256();
       hello.p2p_address = my_impl->p2p_address;
       if( is_transactions_only_connection() ) hello.p2p_address += ":trx";
-      if( is_blocks_only_connection() ) hello.p2p_address += ":blk";
+      // if we are not accepting transactions tell peer we are blocks only
+      if( is_blocks_only_connection() || !my_impl->p2p_accept_transactions ) hello.p2p_address += ":blk";
       hello.p2p_address += " - " + hello.node_id.str().substr(0,7);
 #if defined( __APPLE__ )
       hello.os = "osx";


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3598,7 +3598,8 @@ namespace eosio {
          hello.token = sha256();
       hello.p2p_address = my_impl->p2p_address;
       if( is_transactions_only_connection() ) hello.p2p_address += ":trx";
-      if( is_blocks_only_connection() ) hello.p2p_address += ":blk";
+      // if we are not accepting transactions tell peer we are blocks only
+      if( is_blocks_only_connection() || !my_impl->p2p_accept_transactions ) hello.p2p_address += ":blk";
```
linh2931, line 3602: Can `!my_impl->p2p_accept_transactions` be folded inside `is_blocks_only_connection()`?

heifner: No, as other places `is_blocks_only_connection()` is called we don't want it to be true. For example `is_blocks_only_connection()` is called in the loop to send out trxs. If it was true we would not send out a trx to our peer. We want to send trxs out, just not allow transactions to flow in.

linh2931: Thanks. Maybe you can add a comment here.

linh2931 (approved):

plugins/net_plugin/net_plugin.cpp
greg7mdp, line 3602: I think this might be better because it also updates `connection_type` and logs the change.

```c++
       if(hello.sig == chain::signature_type())
          hello.token = sha256();
       hello.p2p_address = my_impl->p2p_address;
+      if (!my_impl->p2p_accept_transactions && !is_blocks_only_connection()) {
+         fc_dlog( logger, "Setting connection ${c} type for: ${peer} to transactions only", ("c", connection_id)("peer", peer_addr) );
+         connection_type = blocks_only;
+      }
       if( is_transactions_only_connection() ) hello.p2p_address += ":trx";
       if( is_blocks_only_connection() ) hello.p2p_address += ":blk";
       hello.p2p_address += " - " + hello.node_id.str().substr(0,7);
```

heifner: We do not want to set `connection_type = blocks_only` see my comment to Lin above.
Adding a log statement we are doing it is a good idea.

greg7mdp: Right now, we are keeping state describing what the connection permits in two places, `connection_type ` and `p2p_accept_transaction`.

I am suggesting keeping it only in one place: `connection_type `.

heifner: That is not how I see it. `connection_type` is the type of connection I have with my peer. That does not imply it is the same connection they have with me. `p2p_accept_transaction` is an indication of what I allow. A peer might not honor my request for what they send me, so I need to enforce it myself. In my mind these are two separate concepts.

greg7mdp: Well, I understand this is how it is setup today. However, if you just consider what you allow on your connection to your peer:

- send_transactions
- accept_transactions
- send_blocks
- accept_blocks

I don't see why these capabilities should not be stored together in one field. It is certainly less confusing than the current way imho.

greg7mdp: However I'm fine approving as-is.

greg7mdp (approved):

PR #1023, opened 2023-04-12
use python_SITELIB

### LINKED ISSUE #839: Address `TestHarness` installation location
Follow on to `Add TestHarness python integration scripts to dev package` (https://github.com/AntelopeIO/leap/pull/818)

Revisit how and where the python dev modules are installed. In particular investigate using `${Python_SITELIB}` once supported versions of cmake all support `Python_SITELIB`.

Discussions pertaining to this topic for context:
- https://github.com/AntelopeIO/leap/pull/818#discussion_r1139665438
- https://github.com/AntelopeIO/leap/pull/818#discussion_r1139255215

huangminghuang:
This PR uses ${Python_SITELIB} as the location for installing python modules.

Resolves #839


oschwaldp-oci (approved):
In light of comment: https://github.com/AntelopeIO/leap/pull/818#discussion_r1139255215, the use of `FindPackage(Python)` at build time verses the discovery and setting of `Python_SITELIB` at installation time - The reason for the decision to opt for the installation time choice was to more accurately reflect and support the use case where build machine and install machine could be different and the location on the installation machine is what is important here.

scripts/postinst
```diff
@@ -1,13 +1,11 @@
 #!/bin/sh
 
 set -e
+Python_SITELIB=$(python3 -c "from distutils import sysconfig;print(sysconfig.get_python_lib(plat_specific=False,standard_lib=False))")
```
ScottBailey, line 4: As of python 3.10 (in Ubuntu 22), `distutils` is deprecated. It will be removed in 3.12. Additionall pain: distutils must be installed before use!

I'm not 100% certain this suggestion works in non Debian derived environments, but it works in my Debain Python 3.11 environment as well as a mostly vanilla Ubuntu 18 environment. [Here is the documentation on Debian dist-packages.](https://wiki.debian.org/Python)
```suggestion
Python_SITELIB=$(python3 -c "import sysconfig; print(sysconfig.get_path('stdlib'))")/dist-packages
```

If this is not the solution, we should find a better solution than `distutils`.

ScottBailey (changes requested):
See comments: `distutils` requires install and is deprecated.

scripts/postinst
spoonincode: I couldn't get that to work in Ubuntu nor Debian
```
# ls $(python3 -c "import sysconfig; print(sysconfig.get_path('stdlib'))")/dist-packages
ls: cannot access '/usr/lib/python3.8/dist-packages': No such file or directory
```

spoonincode: I think you want `platlib`

```
# python3 -c "import sysconfig;  print(sysconfig.get_path('platlib'))"
/usr/lib/python3.9/site-packages
```

huangminghuang: Currently, it is implemented with `python3 -c "import sysconfig; print(sysconfig.get_path('purelib'))"` as described in https://cmake.org/cmake/help/latest/module/FindPython3.html

ScottBailey: According to that link, dist-packages seem to be the correct path. I wonder if we need to `mkdir -p` to ensure it exists. @spoonincode This worked for me in Debian 11/12 and Ubuntu 18.

ScottBailey (approved):

scripts/postinst
ScottBailey: `purelib` seems to be the correct answer. :-) Approved.

spoonincode (changes requested):
holding

spoonincode:
When I try to install the package, I get
```
ln: failed to create symbolic link '/usr/lib/python3.8/site-packages/TestHarness': No such file or directory
```
I'm not sure if we're expected to create `site-packages` or what...

scripts/postinst
oschwaldp-oci: Are we sure we should be installing into `site-packages`?

```dist-packages instead of site-packages. Third party Python software installed from Debian packages goes into dist-packages, not site-packages. This is to reduce conflict between the system Python, and any from-source Python build you might install manually.```
https://wiki.debian.org/Python

File: scripts/prerm [hunk 2]
```diff
@@ -1,10 +1,6 @@
...
 
 # Cleanup symbolic links created during postinst
-rm -f ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages/TestHarness
-rm -f ${CMAKE_INSTALL_FULL_DATAROOTDIR}/leap_testing/bin
-
-# Attempt to clean up directories that may have been created during postinst
-# Also may have already existed, so it is acceptable to leave them in place if the rmdir fails
-rmdir --ignore-fail-on-non-empty ${CMAKE_INSTALL_FULL_LIBDIR}/python3/dist-packages
-rmdir --ignore-fail-on-non-empty ${CMAKE_INSTALL_FULL_LIBDIR}/python3
+Python_DISTLIB=$(python3 -c "import sysconfig; print(sysconfig.get_path('purelib'))" | sed 's/site-packages/dist-packages/')
```
ScottBailey, line 4: I really don't think this is the right answer.

ScottBailey (changes requested):
I think we need to look at how official Ubuntu and Debian packages set distlib.

scripts/prerm
spoonincode: I agree. Something feels wrong about the trouble we're having here and that's usually a reasonable indicator we're on the wrong path.

.github/workflows/build.yaml
```diff
@@ -103,12 +103,21 @@ jobs:
           zstdcat build.tar.zst | tar x
           cd build
           cpack
+      - name: Install dev package
+        run: |
+          apt update
+          export DEBIAN_FRONTEND=noninteractive
+          export TZ="America/New_York"
+          apt install -y ./build/leap_*.deb
+          apt install -y ./build/leap-dev*.deb
```
spoonincode, line 112: I want to keep the workflow as concise and clutter free as possible, so two changes I'd like here:

1. move the `DEBIAN_FRONTEND` and `TZ` to the Dockerfile
2. combine the two installs, either as `./build/leap_*.deb ./build/leap-dev*.deb` or `./build/leap*.deb` or whatever

.github/workflows/build.yaml
```diff
@@ -103,12 +103,21 @@ jobs:
           zstdcat build.tar.zst | tar x
           cd build
           cpack
+      - name: Install dev package
+        run: |
+          apt update
```
spoonincode, line 108: Should be `update && upgrade` here? I'm not sure how problematic it will be on Ubuntu to have "mixed age" packages installed: the base image is cached and could have many month old packages installed alongside new packages installed here.

scripts/prerm
huangminghuang: This implementation is based on the description of https://wiki.debian.org/Python, first point of the section `Deviations from upstream`. In debian/Ubuntu, using the deprecated `distutils.sysconfig.get_python_lib(plat_specific=False,standard_lib=False)` does return the right answer `/usr/lib/python3/dist-packages` while using sysconfig does not. Based on the [StackOverflow post](https://stackoverflow.com/questions/122327/how-do-i-find-the-location-of-my-python-site-packages-directory) 

> The root cause of the discrepancy is because Debian [patch the distutils install layout](https://salsa.debian.org/cpython-team/python3/-/blob/master/debian/patches/distutils-install-layout.diff), to correctly reflect their changes to the site, but they fail to patch the sysconfig module.

I am not sure how to move forward from here. Here are the list of solutions I can think of (I am welcome any other solutions)
- just use `/usr/lib/python3/dist-packages`,
- use distutils (at least it's still available in our supported distributions),
- keep the current implementation.

spoonincode: I think my vote is to just use distutils and sort out the problem with python 3.12 + ubuntu when that actually arrives.

ScottBailey: As much as I dislike using distutils, it seems like the least painful solution.

spoonincode (approved):

ScottBailey (approved):

PR #1027, opened 2023-04-13
Fix compilation warnings with clang-16 

greg7mdp:
mostly:
- warning: unqualified call to 'std::forward'
- warning: unqualified call to 'std::move'

### DIFF
--- a/libraries/libfc/include/fc/crypto/common.hpp
+++ b/libraries/libfc/include/fc/crypto/common.hpp
@@ -178,7 +178,7 @@ namespace fc { namespace crypto {
       {}
 
       shim(data_type&& data)
-      :_data(forward<data_type>(data))
+         :_data(std::move(data))
       {}
 
       shim(const data_type& data)
--- a/libraries/libfc/include/fc/crypto/private_key.hpp
+++ b/libraries/libfc/include/fc/crypto/private_key.hpp
@@ -53,7 +53,7 @@ namespace fc { namespace crypto {
          storage_type _storage;
 
          private_key( storage_type&& other_storage )
-         :_storage(forward<storage_type>(other_storage))
+            :_storage(std::move(other_storage))
          {}
 
          friend bool operator == ( const private_key& p1, const private_key& p2);
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -69,7 +69,7 @@ class maybe_session {
       maybe_session() = default;
 
       maybe_session( maybe_session&& other)
-      :_session(move(other._session))
+         :_session(std::move(other._session))
       {
       }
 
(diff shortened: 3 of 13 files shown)


libraries/libfc/include/fc/crypto/private_key.hpp
```diff
@@ -53,7 +53,7 @@ namespace fc { namespace crypto {
          storage_type _storage;
 
          private_key( storage_type&& other_storage )
-         :_storage(forward<storage_type>(other_storage))
+            :_storage(std::move(other_storage))
```
heifner, line 56: `other_storage` is trivially copiable, I think just remove `std::move`.

libraries/libfc/include/fc/crypto/common.hpp
```diff
@@ -178,7 +178,7 @@ namespace fc { namespace crypto {
       {}
 
       shim(data_type&& data)
-      :_data(forward<data_type>(data))
+         :_data(std::move(data))
```
heifner, line 181: This one should be `std::forward<Data>` since it based on the class template type `Data`.

heifner (changes requested):

libraries/libfc/include/fc/crypto/common.hpp
greg7mdp: Hum, no, forward just works for function templates, not for regular member functions of a templated class.

libraries/libfc/include/fc/crypto/private_key.hpp
greg7mdp: Yes, thanks!

libraries/libfc/include/fc/crypto/common.hpp
greg7mdp: the `fact_t_forward<int&> ff2(x);` call is calling the default copy constructor. Add `fact_t_forward(val_t &) = delete;`  to see.

greg7mdp: https://rextester.com/FUUAH38143

greg7mdp: https://rextester.com/YBWP43749

heifner: Ah right, thanks. `std::forward` is for universal references and a class template type is set at time of object creation making it not universal. The member of the templated class is just a normal member at the point of use.

heifner (approved):

libraries/libfc/include/fc/crypto/common.hpp
greg7mdp: Yes, exactly. I created a better example demonstrating this https://rextester.com/LBXE59514

linh2931 (approved):

Issue #452, opened 2022-11-09
Propose block buffer design for P2P syncing processes


stephenpdeos:
Currently if blocks are received out of order, future blocks are discarded until preceding blocks are applied. An optimization would be to cache future blocks up to a configurable buffer size (or block count) until they can be applied.

Considerations
- Size of the cache of future blocks received. Should this be configurable?

Reqs
- Part of net plugin architecture
- All state managed in memory
- Cache is transient, does not persist across restarts

PR #1040, opened 2023-04-15
P2P: Misc improvements

heifner:
In preparation of work for #452 some initial cleanup.

Testing with performance harness when sending more transactions than nodeos can process it fills up the write queue. This is designed to then close the connection. nodeos was generating thousands of write queue full messages (example below) because the posted `close()` call would have to wait until all the currently queued up tasks on the strand finished. This delayed the close for seconds. Added a `closing` flag to stop existing tasks quickly, resulting in only one message instead of thousands and an immediate close.
```
error 2023-04-15T16:46:16.320 net-0     net_plugin.cpp:2494           start_read_message   ] ["localhost:9877 - d7c63c8" - 1 127.0.0.1:37658] write queue full 71936405 bytes, giving up on connection, closing
```

Added better encapsulation of the `sync_mtx`.
Added `alignas(hardware_destructive_interference_size)` to avoid cache synchronization.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -122,7 +122,7 @@ namespace eosio {
       static constexpr int64_t block_interval_ns =
             std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::milliseconds(config::block_interval_ms)).count();
 
-      mutable std::mutex sync_mtx;
+      std::mutex     sync_mtx;
       uint32_t       sync_known_lib_num{0};
       uint32_t       sync_last_requested_num{0};
       uint32_t       sync_next_expected_num{0};
@@ -150,10 +150,8 @@ namespace eosio {
       void sync_update_expected( const connection_ptr& c, const block_id_type& blk_id, uint32_t blk_num, bool blk_applied );
       void recv_handshake( const connection_ptr& c, const handshake_message& msg );
       void sync_recv_notice( const connection_ptr& c, const notice_message& msg );
-      inline std::unique_lock<std::mutex> locked_sync_mutex() {
-         return std::unique_lock<std::mutex>(sync_mtx);
-      }
-      inline void reset_last_requested_num(const std::unique_lock<std::mutex>& lock) {
+      inline void reset_last_requested_num() {
+         std::lock_guard g(sync_mtx);
          sync_last_requested_num = 0;
       }
    };
@@ -215,7 +213,6 @@ namespace eosio {
    class net_plugin_impl : public std::enable_shared_from_this<net_plugin_impl>,
                            public auto_bp_peering::bp_connection_manager<net_plugin_impl, connection> {
     public:
-      using connection_t = connection;
       unique_ptr<tcp::acceptor>        acceptor;
       std::atomic<uint32_t>            current_connection_id{0};
 
@@ -1590,7 +1587,7 @@ namespace eosio {
 
          // if closing the connection we are currently syncing from, then reset our last requested and next expected.
          if( c == sync_source ) {
-            reset_last_requested_num(g);
+            sync_last_requested_num = 0;
             // if starting to sync need to always start from lib as we might be on our own fork
             uint32_t lib_num = my_impl->get_chain_lib_num();
             sync_next_expected_num = lib_num + 1;
@@ -1676,7 +1673,7 @@ namespace eosio {
          fc_elog( logger, "Unable to continue syncing at this time");
          if( !new_sync_source ) sync_source.reset();
          sync_known_lib_num = chain_info.lib_num;
-         reset_last_requested_num(g_sync);
+         sync_last_requested_num = 0;
          set_state( in_sync ); // probably not, but we can't do anything else
          return;
       }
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3252,18 +3285,6 @@ namespace eosio {
...
-      if( !c->socket_is_open() ) {
-         if( bsp ) {
-            // valid bsp means add_peer_block already called, need to remove it since we are not going to process the block
-            // call on dispatch strand to serialize with the add_peer_block calls
-            my_impl->dispatcher->strand.post( [blk_id]() {
-               my_impl->dispatcher->rm_block( blk_id );
-            } );
-         }
-         return;
-      }
```
heifner, line 3265: This very much looks like an anti-pattern to me. As the comment describes care has to be taken to call `rm_block` otherwise we may end up skipping a block. There is no reason we need to stop processing a block just because the socket has been closed since we received the block. It is cleaner to just to process the block as normal.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1059,6 +1061,7 @@ namespace eosio {
       self->flush_queues();
       self->connecting = false;
       self->syncing = false;
+      self->closing = false;
```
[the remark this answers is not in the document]
heifner: I do think that is cleaner, yes. Moved.

linh2931 (approved):

greg7mdp:
Just curious. What led you to adding `alignas(hardware_destructive_interference_size) to avoid cache synchronization.`?
Could you measure whether it made a difference in performance?

heifner:
> Just curious. What led you to adding `alignas(hardware_destructive_interference_size) to avoid cache synchronization.`? Could you measure whether it made a difference in performance?

Just something I've been meaning to add. No, I have not measured the performance difference.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -629,7 +655,7 @@ namespace eosio {
       std::shared_ptr<tcp::socket>              socket; // only accessed through strand after construction
 
       fc::message_buffer<1024*1024>    pending_message_buffer;
-      std::atomic<std::size_t>         outstanding_read_bytes{0}; // accessed only from strand threads
+      std::size_t                      outstanding_read_bytes{0}; // accessed only from strand threads
```
greg7mdp, line 658: I'm sure there is a good reason, but I'm not getting why `std::atomic` is not needed?

heifner: It is only accessed from each individual connection strand.

greg7mdp:
> Just something I've been meaning to add. No, I have not measured the performance difference.

Very cool change!

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1590,7 +1625,7 @@ namespace eosio {
 
          // if closing the connection we are currently syncing from, then reset our last requested and next expected.
          if( c == sync_source ) {
-            reset_last_requested_num(g);
```
greg7mdp, line 1593: I think the purpose for this passing a reference to a unique_lock to `reset_last_requested_num()` was to ensure that a lock was taken before `sync_last_requested_num` was updated.
While it is still true that the lock is taken before this variable is reset, maybe in the future a code copy/past will be done where this is not true anymore.
Did you think that this "insurance" was not worth it?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2373,8 +2378,7 @@ namespace eosio {
                   c->send_handshake();
                }
             } else {
-               fc_elog( logger, "connection failed to ${host}:${port} ${error}",
-                        ("host", endpoint.address().to_string())("port", endpoint.port())( "error", err.message()));
+               fc_elog( logger, "connection failed to ${a}, ${error}", ("a", c->peer_address())( "error", err.message()));
```
greg7mdp, line 2381: wouldn't `"connection to ${a} failed, ..."` be better english?

heifner: Maybe. I would say "connection failed to" is easier to grep.

plugins/net_plugin/net_plugin.cpp
greg7mdp: Oh, yes, thanks!

plugins/net_plugin/net_plugin.cpp
heifner: No, the socket can be closed at any moment, so the code is already aware of that condition.

plugins/net_plugin/net_plugin.cpp
greg7mdp: odd justification, but OK :-). If we want to optimize for grep, maybe we should add random typos as well.

heifner: lol, I have found random typos in the past to be helpful actually. I don't have a strong opinion about this at all. Happy to change it if you like.

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
greg7mdp: No if you like it better like that, it is fine, you certainly look  at logs much more than me.

PR #1045, opened 2023-04-17
P2P: Save and retry unlinkable blocks

### LINKED ISSUE #452: Propose block buffer design for P2P syncing processes

heifner:
Keep a small cache of unlinkable blocks to retry. This is mainly to address the issue of blocks coming in on multiple connections and getting processed out of order. On an unlinkable block exception, store the block and retry it if its `previous` block is applied.

Resolves #452

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -132,12 +132,12 @@ namespace eosio {
    };
 
    struct by_timestamp;
-   struct by_blk;
+   struct by_block_num_id;
    struct by_prev;
    typedef multi_index_container<
          eosio::unlinkable_block_state,
          indexed_by<
-               ordered_unique< tag<by_blk>,
+               ordered_unique< tag<by_block_num_id>,
                      composite_key< unlinkable_block_state,
                            const_mem_fun<unlinkable_block_state, uint32_t , &eosio::unlinkable_block_state::block_num>,
                            member<unlinkable_block_state, block_id_type, &eosio::unlinkable_block_state::id>
@@ -241,7 +241,7 @@ namespace eosio {
       void rm_block(const block_id_type& blkid);
 
       void add_unlinkable_block( signed_block_ptr b, const block_id_type& id );
-      unlinkable_block_state get_possible_linkable_block(const block_id_type& blkid);
+      unlinkable_block_state pop_possible_linkable_block(const block_id_type& blkid);
 
       bool add_peer_txn( const transaction_id_type& id, const time_point_sec& trx_expires, uint32_t connection_id,
                          const time_point_sec& now = time_point::now() );
@@ -2168,7 +2168,7 @@ namespace eosio {
       }
    }
 
-   unlinkable_block_state dispatch_manager::get_possible_linkable_block(const block_id_type& blkid) {
+   unlinkable_block_state dispatch_manager::pop_possible_linkable_block(const block_id_type& blkid) {
       std::lock_guard g(unlinkable_blk_state_mtx);
       auto& index = unlinkable_blk_state.get<by_prev>();
       auto blk_itr = index.find( blkid );
@@ -2225,7 +2225,7 @@ namespace eosio {
       }
       {
          std::lock_guard<std::mutex> g( unlinkable_blk_state_mtx );
-         auto& stale_blk = unlinkable_blk_state.get<by_blk>();
+         auto& stale_blk = unlinkable_blk_state.get<by_block_num_id>();
          stale_blk.erase( stale_blk.lower_bound( 1 ), stale_blk.upper_bound( lib_num ) );
       }
    }
@@ -3408,8 +3408,8 @@ namespace eosio {
             fc_dlog( logger, "accepted signed_block : #${n} ${id}...", ("n", blk_num)("id", blk_id.str().substr(8,16)) );
             dispatcher->add_peer_block( blk_id, c->connection_id );
 
-            while (true) { // schedule any possible previously unlinkable blocks
-               unlinkable_block_state prev_unlinkable = my_impl->dispatcher->get_possible_linkable_block(blk_id);
+            while (true) { // attempt previously unlinkable blocks where prev_unlinkable->block->previous == blk_id
+               unlinkable_block_state prev_unlinkable = my_impl->dispatcher->pop_possible_linkable_block(blk_id);
                if (!prev_unlinkable.block)
                   break;
                fc_dlog( logger, "retrying previous unlinkable block #${n} ${id}...",
@@ -3428,7 +3428,7 @@ namespace eosio {
          c->strand.post( [sync_master = my_impl->sync_master.get(), dispatcher = my_impl->dispatcher.get(), c,
                           block{std::move(block)}, blk_id, blk_num, reason]() mutable {
             if( reason == unlinkable || reason == no_reason ) {
-               my_impl->dispatcher->add_unlinkable_block( std::move(block), blk_id );
+               dispatcher->add_unlinkable_block( std::move(block), blk_id );
             }
             // reason==no_reason means accept_block() return false because we are producing, don't call rejected_block which sends handshake
             if( reason != no_reason ) {
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -163,15 +163,39 @@ namespace eosio {
       // almost immediately (blocks came in from multiple peers out of order). 30 allows for one block per
       // producer round until lib. When queue larger than max, remove by block timestamp farthest in the past.
       static constexpr size_t max_unlinkable_cache_size = 30;
-      dispatch_manager& dispatch;
 
    public:
-      explicit unlinkable_block_state_cache(dispatch_manager& dispatch)
-      : dispatch(dispatch) {};
+      // returns block id of any block removed because of a full cache
+      std::optional<block_id_type> add_unlinkable_block( signed_block_ptr b, const block_id_type& id ) {
+         std::lock_guard g(unlinkable_blk_state_mtx);
+         unlinkable_blk_state.insert( {id, std::move(b)} ); // does not insert if already there
+         if (unlinkable_blk_state.size() > max_unlinkable_cache_size) {
+            auto& index = unlinkable_blk_state.get<by_timestamp>();
+            auto begin = index.begin();
+            block_id_type rm_block_id = begin->id;
+            index.erase( begin );
+            return rm_block_id;
+         }
+         return {};
+      }
 
-      void add_unlinkable_block( signed_block_ptr b, const block_id_type& id );
-      unlinkable_block_state pop_possible_linkable_block( const block_id_type& blkid );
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -121,6 +122,36 @@ namespace eosio {
...
+      block_id_type    id;
+      signed_block_ptr block;
+
+      uint32_t block_num() const { return block_header::num_from_id(id); }
+      const block_id_type& prev() const { return block->previous; }
+      const block_timestamp_type& timestamp() const { return block->timestamp; }
+   };
+
+   struct by_timestamp;
+   struct by_blk;
```
linh2931, line 135: It might be more clearer called `by_block_num`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -201,6 +240,9 @@ namespace eosio {
       bool have_block(const block_id_type& blkid) const;
       void rm_block(const block_id_type& blkid);
 
+      void add_unlinkable_block( signed_block_ptr b, const block_id_type& id );
+      unlinkable_block_state get_possible_linkable_block(const block_id_type& blkid);
```
linh2931, line 244: Should `get_possible_linkable_block` be named something like `get_and_remove_possible_linkable_block` as it removes the block from the cache?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3330,22 +3404,31 @@ namespace eosio {
...
       if( accepted ) {
-         boost::asio::post( my_impl->thread_pool.get_executor(), [dispatcher = my_impl->dispatcher.get(), cid=c->connection_id, blk_id, msg]() {
-            fc_dlog( logger, "accepted signed_block : #${n} ${id}...", ("n", msg->block_num())("id", blk_id.str().substr(8,16)) );
-            dispatcher->add_peer_block( blk_id, cid );
+         boost::asio::post( my_impl->thread_pool.get_executor(), [dispatcher = my_impl->dispatcher.get(), c, blk_id, blk_num]() {
+            fc_dlog( logger, "accepted signed_block : #${n} ${id}...", ("n", blk_num)("id", blk_id.str().substr(8,16)) );
+            dispatcher->add_peer_block( blk_id, c->connection_id );
+
+            while (true) { // schedule any possible previously unlinkable blocks
+               unlinkable_block_state prev_unlinkable = my_impl->dispatcher->get_possible_linkable_block(blk_id);
```
linh2931, line 3412: In this while loop, `blk_id` is never updated.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3330,22 +3404,31 @@ namespace eosio {
...
-         c->strand.post( [sync_master = my_impl->sync_master.get(), dispatcher = my_impl->dispatcher.get(), c, blk_id, blk_num, reason]() {
+         c->strand.post( [sync_master = my_impl->sync_master.get(), dispatcher = my_impl->dispatcher.get(), c,
+                          block{std::move(block)}, blk_id, blk_num, reason]() mutable {
             if( reason == unlinkable || reason == no_reason ) {
-               // unlinkable may be linkable in the future, so indicate we have not received it
-               // call on dispatch strand to serialize with the add_peer_block calls
-               my_impl->dispatcher->strand.post( [blk_id]() {
-                  my_impl->dispatcher->rm_block( blk_id );
-               } );
+               my_impl->dispatcher->add_unlinkable_block( std::move(block), blk_id );
```
linh2931, line 3431: Can `my_impl->dispatcher` jus be `dispatcher`?

linh2931:
Will it be easy to add a unit or integration test for this?

plugins/net_plugin/net_plugin.cpp
heifner: Right. We want to try any blocks with a `previous` equal to the just applied `block_id`. I updated the comment. Hopefully clearer now.

plugins/net_plugin/net_plugin.cpp
heifner: Renamed to `pop_possible_linkable_block`.

heifner:
> Will it be easy to add a unit or integration test for this?

There already are integration tests that verify this works. My first commit actually failed: https://github.com/AntelopeIO/leap/actions/runs/4724092570 because I was using the wrong `id`.

plugins/net_plugin/net_plugin.cpp
heifner: Yes

plugins/net_plugin/net_plugin.cpp
```diff
@@ -121,6 +122,36 @@ namespace eosio {
...
+
+      uint32_t block_num() const { return block_header::num_from_id(id); }
+      const block_id_type& prev() const { return block->previous; }
+      const block_timestamp_type& timestamp() const { return block->timestamp; }
+   };
+
+   struct by_timestamp;
+   struct by_block_num_id;
+   struct by_prev;
+   typedef multi_index_container<
```
greg7mdp, line 137: I think `using` is nicer than `typedef` these days :-).

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3330,22 +3404,31 @@ namespace eosio {
...
       if( accepted ) {
-         boost::asio::post( my_impl->thread_pool.get_executor(), [dispatcher = my_impl->dispatcher.get(), cid=c->connection_id, blk_id, msg]() {
-            fc_dlog( logger, "accepted signed_block : #${n} ${id}...", ("n", msg->block_num())("id", blk_id.str().substr(8,16)) );
-            dispatcher->add_peer_block( blk_id, cid );
+         boost::asio::post( my_impl->thread_pool.get_executor(), [dispatcher = my_impl->dispatcher.get(), c, blk_id, blk_num]() {
+            fc_dlog( logger, "accepted signed_block : #${n} ${id}...", ("n", blk_num)("id", blk_id.str().substr(8,16)) );
+            dispatcher->add_peer_block( blk_id, c->connection_id );
+
+            while (true) { // attempt previously unlinkable blocks where prev_unlinkable->block->previous == blk_id
+               unlinkable_block_state prev_unlinkable = my_impl->dispatcher->pop_possible_linkable_block(blk_id);
```
greg7mdp, line 3412:
```suggestion
               unlinkable_block_state prev_unlinkable = dispatcher->pop_possible_linkable_block(blk_id);
```

plugins/net_plugin/net_plugin.cpp
```diff
@@ -121,6 +122,36 @@ namespace eosio {
...
+   struct by_prev;
+   typedef multi_index_container<
+         eosio::unlinkable_block_state,
+         indexed_by<
+               ordered_unique< tag<by_block_num_id>,
+                     composite_key< unlinkable_block_state,
+                           const_mem_fun<unlinkable_block_state, uint32_t , &eosio::unlinkable_block_state::block_num>,
+                           member<unlinkable_block_state, block_id_type, &eosio::unlinkable_block_state::id>
+                     >,
+                     composite_key_compare< std::less<>, sha256_less >
```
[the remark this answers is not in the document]
heifner: `std::less<fc::sha256>` uses memcmp.
`sha256_less` uses `<` on each `uint64_t`.
See https://lemire.me/blog/2018/08/22/avoid-lexicographical-comparisons-when-testing-for-string-equality/

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3330,22 +3404,31 @@ namespace eosio {
...
 
       if( accepted ) {
-         boost::asio::post( my_impl->thread_pool.get_executor(), [dispatcher = my_impl->dispatcher.get(), cid=c->connection_id, blk_id, msg]() {
-            fc_dlog( logger, "accepted signed_block : #${n} ${id}...", ("n", msg->block_num())("id", blk_id.str().substr(8,16)) );
-            dispatcher->add_peer_block( blk_id, cid );
+         boost::asio::post( my_impl->thread_pool.get_executor(), [dispatcher = my_impl->dispatcher.get(), c, blk_id, blk_num]() {
+            fc_dlog( logger, "accepted signed_block : #${n} ${id}...", ("n", blk_num)("id", blk_id.str().substr(8,16)) );
+            dispatcher->add_peer_block( blk_id, c->connection_id );
+
+            while (true) { // attempt previously unlinkable blocks where prev_unlinkable->block->previous == blk_id
```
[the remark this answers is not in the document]
heifner: Doesn't matter if the connection is open or not. We should still attempt these blocks if ready.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2075,8 +2137,43 @@ namespace eosio {
...
+      std::unique_lock g(unlinkable_blk_state_mtx);
+      unlinkable_blk_state.insert( {id, std::move(b)} ); // does not insert if already there
+      if (unlinkable_blk_state.size() > max_unlinkable_cache_size) {
+         auto& index = unlinkable_blk_state.get<by_timestamp>();
+         auto begin = index.begin();
+         block_id_type rm_block_id = begin->id;
+         index.erase( begin );
+         g.unlock();
+         // rm_block since we are no longer tracking this not applied block, allowing it to flow back in if needed
+         dispatch.rm_block(rm_block_id);
```
greg7mdp, line 2152: Thanks a lot for making the change, Kevin. This line is the only location where `dispatch` is used. Instead of creating a dependency, why not have this function returning `std::optional<block_id_type>` as in my original suggestion? That way `unlinkable_block_state_cache` doesn't know about `dispatch_manager`. And you won't need the `g.unlock()`.

Also, why not have the functions inline in the class? It is all in the `.cpp` file anyways.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -121,6 +122,81 @@ namespace eosio {
...
+   struct unlinkable_block_state {
+      block_id_type    id;
+      signed_block_ptr block;
+
+      uint32_t block_num() const { return block_header::num_from_id(id); }
+      const block_id_type& prev() const { return block->previous; }
+      const block_timestamp_type& timestamp() const { return block->timestamp; }
+   };
+
+   class dispatch_manager;
```
greg7mdp, line 134: I don't believe you nee this forward declaration anymore.

greg7mdp (approved):
Thanks for the changes, Kevin, looks great!

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
greg7mdp: So it it only for performance? I think the difference may be only when checking for equality, not for `operator<`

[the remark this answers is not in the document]
heifner: Could be, but yes the idea here is performance.

PR #1049, opened 2023-04-18
FC cleanup

heifner:
- Remove unused files.
- Remove fc::string alias (this is most of the changes).


greg7mdp (approved):
Cool, very nice to see these cleanups!

libraries/libfc/include/fc/string.hpp
```diff
@@ -2,151 +2,29 @@
...
-  fc::string to_string( uint64_t );
-  fc::string to_string( int64_t );
-  fc::string to_string( uint16_t );
+  int64_t  to_int64( const std::string& );
+  uint64_t to_uint64( const std::string& );
+  double   to_double( const std::string& );
+  std::string to_string( double );
+  std::string to_string( uint64_t );
+  std::string to_string( int64_t );
+  std::string to_string( uint16_t );
```
linh2931, line 16: Why do we still need this string.hpp file? Can we use STD's `to_string`?

libraries/libfc/include/fc/string.hpp
```diff
@@ -2,151 +2,29 @@
...
   std::string to_pretty_string( int64_t );
-  inline fc::string to_string( int32_t v ) { return to_string( int64_t(v) ); }
-  inline fc::string to_string( uint32_t v ){ return to_string( uint64_t(v) ); }
+  inline std::string to_string( int32_t v ) { return to_string( int64_t(v) ); }
```
linh2931, line 18: Why cast int32_t to int64_t? Can those be templtaed?

libraries/libfc/include/fc/string.hpp
heifner: `to_string( double )` has a "unique" impl that we do rely on it looks like.
The other `to_string` could likely be `std::_to_string` but I would need to verify none of them are used in consensus code before we moved from `lexical_cast` to `std::to_string`.

libraries/libfc/include/fc/string.hpp
heifner: It think they can just be removed. See above comment. I'll create a follow on issue for this.
https://github.com/AntelopeIO/leap/issues/1057

libraries/libfc/include/fc/string.hpp
heifner: It does look like there are a few more unused functions in there I can kill.

linh2931 (approved):

greg7mdp (approved):

PR #1052, opened 2023-04-19
[PH] Add http client async

oschwaldp-oci:
Create boost beast async http client for use in transaction generator when creating transactions pointed at API Node.

Added unit test around initial http async client with simple http server.


tests/trx_generator/trx_generator_tests.cpp
```diff
@@ -435,4 +488,46 @@ BOOST_AUTO_TEST_CASE(account_name_generator_tests)
...
+   auto const test_body2 = "{\"return_failure_trace\":true,\"retry_trx\":false,\"transaction\":{\"signatures\":[\"SIG_K1_JyzLqbvpdybyujtiN1YdY2FWcBBi8dWWiFgZ515qyyqgKJJ6892i4rXTHdw5KGYut6EBuXPR3ExRwPSioSZ2bZ1RjNUXVj\"],\"compression\":\"none\",\"packed_context_free_data\":\"\",\"packed_trx\":\"848a34641800f994a24e00000000030000000000ea305500409e9a2264b89a0160ae423ad15b974a00000000a8ed32326660ae423ad15b974a1042088a4dd35057010000000100038d26b3d5ce8c7d76ef00d3d586a3d7bbc76c42f0b0719cc6f7b0cce1790622c301000000010000000100028dc3921705c71d30b0b26674536fff934f8e43890c980aa1d2c168f00f406539010000000000000000ea3055000000004873bd3e0160ae423ad15b974a00000000a8ed32322060ae423ad15b974a1042088a4dd35057009435770000000004535953000000000000000000ea305500003f2a1ba6a24a0160ae423ad15b974a00000000a8ed32323160ae423ad15b974a1042088a4dd3505740420f0000000000045359530000000040420f000000000004535953000000000000\"}}";
+
+   // The io_context is required for all I/O
+   net::io_context ioc;
+
+   // Launch the asynchronous operation
+   std::future<http::response<http::string_body>> future_response = std::make_shared<session>(ioc)->run(host, port_str, target, version, content_type, test_body);
```
huangminghuang, line 517: It's very clear to me what scenarios the http client code will be used; however, I prefer to use callback model than the std::future model. Furthermore, you can wrap this statement into a more readable function.

tests/trx_generator/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ target_include_directories(trx_generator PUBLIC ${CMAKE_CURRENT_BINARY_DIR} ${CM
 target_link_libraries( trx_generator
                        PRIVATE eosio_chain fc chain_plugin ${CMAKE_DL_LIBS} ${PLATFORM_SPECIFIC_LIBS} )
 
-add_executable(trx_generator_tests trx_generator_tests.cpp trx_provider.cpp)
+add_executable(trx_generator_tests trx_generator_tests.cpp trx_provider.cpp )
```
huangminghuang, line 8: you need `trx_generator.cpp` here

oschwaldp-oci: Addressed: b9f2d280a1

huangminghuang (approved):

tests/trx_generator/http_client_async.hpp
```diff
@@ -0,0 +1,166 @@
...
+#include <cstdlib>
+#include <functional>
+#include <iostream>
+#include <memory>
+#include <string>
+
+namespace beast = boost::beast;         // from <boost/beast.hpp>
+namespace http  = beast::http;          // from <boost/beast/http.hpp>
+namespace net   = boost::asio;          // from <boost/asio.hpp>
+using tcp       = boost::asio::ip::tcp; // from <boost/asio/ip/tcp.hpp>
```
heifner, line 35: I think we should avoid namespace alias and using in header files, especially in new code.

tests/trx_generator/http_client_async.hpp
```diff
@@ -0,0 +1,166 @@
...
+
+   void on_read(beast::error_code ec, std::size_t bytes_transferred) {
+      boost::ignore_unused(bytes_transferred);
+
+      if (ec) {
+         response_callback_(ec, {});
+         return fail(ec, "read");
+      }
+
+      // Write the message to promise result
```
heifner, line 136: Comment is wrong, no longer using a promise.

tests/trx_generator/http_client_async.hpp
```diff
@@ -0,0 +1,166 @@
...
+namespace beast = boost::beast;         // from <boost/beast.hpp>
+namespace http  = beast::http;          // from <boost/beast/http.hpp>
+namespace net   = boost::asio;          // from <boost/asio.hpp>
+using tcp       = boost::asio::ip::tcp; // from <boost/asio/ip/tcp.hpp>
+
+//------------------------------------------------------------------------------
+
+namespace eosio {
+
+using response_callback_t = std::function<void(beast::error_code ec, http::response<http::string_body>)>;
```
heifner, line 41: Seems like another namespace of `http_client_async` would be good for this and the function at the end.

tests/trx_generator/http_client_async.hpp
```diff
@@ -0,0 +1,166 @@
...
+struct http_request_params {
+   net::io_context& ioc;
+   char const*      host;
+   char const*      port;
+   char const*      target;
+   int              version;
+   char const*      content_type;
+};
+
+inline void initiate_async_http_request(http_request_params& req_params, char const* request_body,
```
heifner, line 160: I think I would call this `async_http_request`

tests/trx_generator/simple_rest_server.hpp
```diff
@@ -0,0 +1,234 @@
...
+namespace eosio { namespace rest {
+
+   // The majority of the code here are derived from boost source
+   // libs/beast/example/http/server/async/http_server_async.cpp
+   // with minimum modification and yet reusable.
+
+   namespace beast = boost::beast;         // from <boost/beast.hpp>
+   namespace http  = beast::http;          // from <boost/beast/http.hpp>
+   namespace net   = boost::asio;          // from <boost/asio.hpp>
+   using tcp       = boost::asio::ip::tcp; // from <boost/asio/ip/tcp.hpp>
```
heifner, line 16: Would like to avoid these in header files.

tests/trx_generator/trx_generator_tests.cpp
```diff
@@ -1,10 +1,58 @@
...
+      // Respond to POST request
+      res.set(http::field::server, server_header());
+      res.set(http::field::content_type, "text/plain");
+      res.keep_alive(req.keep_alive());
+      res.body() = req.body();
+      res.prepare_payload();
+      return res;
+   }
+
+   eosio::chain::named_thread_pool<struct prom> _trx_gen_server_thread_pool;
```
heifner, line 38: Use a diff struct name here.

tests/trx_generator/trx_generator_tests.cpp
```diff
@@ -1,10 +1,58 @@
...
+
+   std::optional<http::response<http::string_body>> on_request(http::request<http::string_body>&& req) {
+      if (req.target() != api_name)
+         return {};
+      http::response<http::string_body> res{http::status::ok, req.version()};
+      // Respond to POST request
+      res.set(http::field::server, server_header());
+      res.set(http::field::content_type, "text/plain");
+      res.keep_alive(req.keep_alive());
+      res.body() = req.body();
```
heifner, line 33: Add comment that this echos back the request. I would name it `echo_server`.

heifner (changes requested):

tests/trx_generator/trx_generator_tests.cpp
```diff
@@ -435,4 +483,71 @@ BOOST_AUTO_TEST_CASE(account_name_generator_tests)
...
+   server.start(endpoint);
+
+   // Start Client
+
+   // The io_context is required for all I/O
+   net::io_context ioc;
+   auto const      target        = "/v1/chain/test";
+   int             version       = 11;
+   auto const      content_type  = "text/plain";
+   auto const      content_type2 = "application/json";
```
greg7mdp, line 506: Do you specify `const` to mean the pointer cannot be changed? `auto x = "string";` is already a `const char *`.

And why `int` for version instead of auto? `auto version = 11;` is an int by default.

I think these could all be `constexpr auto`. And I would make the strings be `std::string` by adding the `s` suffix, so that there isn't a conversion to `std::string` for every comparison, which is what I think occurs.

tests/trx_generator/http_client_async.hpp
greg7mdp: This is true in general, but in this case, because it is a header file used only in a .cpp in the same directory, I think I is actually preferable to having the code be less readable. Personal opinion.

greg7mdp: I think the solution is to move those 4 lines (32 to 35) after `namespace http_client_async_details {`, so these aliases don't leak out.

tests/trx_generator/http_client_async.hpp
oschwaldp-oci: Addressed: [9ff417b](https://github.com/AntelopeIO/leap/pull/1052/commits/9ff417bc69e9b87b352ebb9d606c34821547023b)

tests/trx_generator/trx_generator_tests.cpp
oschwaldp-oci: `constexpr` for `string_literals` isn't available until C++20. So opted for a different direction here.

Addressed: [f02287e](https://github.com/AntelopeIO/leap/pull/1052/commits/f02287e44535e51b7ae21d1c90db73c40f51d644)

tests/trx_generator/trx_generator_tests.cpp
oschwaldp-oci: Addressed: [f02287e](https://github.com/AntelopeIO/leap/pull/1052/commits/f02287e44535e51b7ae21d1c90db73c40f51d644)

tests/trx_generator/trx_generator_tests.cpp
oschwaldp-oci: Addressed: [f02287e](https://github.com/AntelopeIO/leap/pull/1052/commits/f02287e44535e51b7ae21d1c90db73c40f51d644)

tests/trx_generator/http_client_async.hpp
oschwaldp-oci: Addressed: [f02287e](https://github.com/AntelopeIO/leap/pull/1052/commits/f02287e44535e51b7ae21d1c90db73c40f51d644)

tests/trx_generator/http_client_async.hpp
oschwaldp-oci: Addressed: [f02287e](https://github.com/AntelopeIO/leap/pull/1052/commits/f02287e44535e51b7ae21d1c90db73c40f51d644)

tests/trx_generator/http_client_async.hpp
oschwaldp-oci: Addressed: [f02287e](https://github.com/AntelopeIO/leap/pull/1052/commits/f02287e44535e51b7ae21d1c90db73c40f51d644)

heifner (approved):

PR #1058, opened 2023-04-19
Added missing namespace to logger in case of FC_USE_PTHREAD_NAME_NP not defined

766C6164:
When compiled not on Linux (BSD here)  FC_USE_PTHREAD_NAME_NP not defined and code for this scenario was missing std:: for string

### DIFF
--- a/libraries/libfc/src/log/logger_config.cpp
+++ b/libraries/libfc/src/log/logger_config.cpp
@@ -151,7 +151,7 @@ namespace fc {
          }
 #else
          static int thread_count = 0;
-         thread_name = string( "thread-" ) + fc::to_string( thread_count++ );
+         thread_name = std::string( "thread-" ) + fc::to_string( thread_count++ );
 #endif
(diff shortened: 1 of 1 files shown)


spoonincode (approved):
Clearly good and fine to go in as is.

But!

I don't really like this code as-is and it could use a clean up imo. It's not clear why there are two separate `set_os_thread_name()` & `set_thread_name()` vs just a single `set_thread_name()` that also sets the OS thread name. And if that's the case `get_thread_name()` can just return the `thread_name` and not query the OS. And if that's the case we don't need `pthread_getname_np()` and all. And if that's the case we probably don't even need the cmake compile time check and `FC_USE_PTHREAD_NAME_NP` because FreeBSD, linux+glibc, and linux+musl all have `pthread_setname_np()` (macos also has `pthread_setname_np()` but with a slightly different interface: that can be `#ifdef __APPLE__` easily enough though)

Issue #815, opened 2023-03-14
Nodeos instability issue during re-start under FreeBSD


766C6164:
With a full understanding that FreeBSD is not officially supported but feeling that fixing this issue may improve general stability of nodeos on supported platforms as well. This issue is not reproducible on Ubuntu/Debian.

System:
FreeBSD bsd 13.1-RELEASE FreeBSD 13.1-RELEASE releng/13.1-n250148-fc952ac2212 GENERIC amd64

Steps to reproduce:
1) clean-up any leftovers of ~/.local/share/eosio
2) start producing node:
./bin/nodeos -e -p eosio --plugin eosio::producer_plugin --plugin eosio::producer_api_plugin
3) wait for several blocks
4) terminate it with ctrl+c
5) re-start node with the same command
./bin/nodeos -e -p eosio --plugin eosio::producer_plugin --plugin eosio::producer_api_plugin

node will start producing and after several blocks will start to error out:

`
./bin/nodeos -e -p eosio --plugin eosio::producer_plugin --plugin eosio::producer_api_plugin
info  2023-03-14T17:38:15.205 thread-0  chain_plugin.cpp:470          plugin_initialize    ] initializing chain plugin
info  2023-03-14T17:38:15.207 thread-0  protocol_feature_manag:835    operator()           ] Support for builtin protocol feature 'PREACTIVATE_FEATURE' (with digest of '0ec7e080177b2c02b278d5088611686b49d739925a92d9bfcacd7fc6b74053bd') is enabled without activation restrictions
info  2023-03-14T17:38:15.207 thread-0  protocol_feature_manag:822    operator()           ] Support for builtin protocol feature 'ONLY_LINK_TO_EXISTING_PERMISSION' (with digest of '1a99a59d87e06e09ec5b028a9cbb7749b4a5ad8819004365d02dc4379a8b7241') is enabled with preactivation required
info  2023-03-14T17:38:15.207 thread-0  protocol_feature_manag:822    operator()           ] Support for builtin protocol feature 'FORWARD_SETCODE' (with digest of '2652f5f96006294109b3dd0bbde63693f55324af452b799ee137a81a905eed25') is enabled with preactivation required
[...]

PR #1059, opened 2023-04-19
Stability issues and improvements on BSD 

766C6164:
This PR addresses multiple platform specific for BSD issues, in particular:
- using kqueue based timer on BSD, same way as on OSX, posix one is not behaving well
- avoid making any pthread calls such as pthread_mutex_lock of boost/interprocess until it is properly fixed in Boost (PR in chainbase - https://github.com/AntelopeIO/chainbase/pull/13)

Original issue https://github.com/AntelopeIO/leap/issues/815 ended up being fixed with https://github.com/AntelopeIO/chainbase/pull/13. This PR brings additional robustness by using a better timer for FreeBSD and bumps chainbase dependency


libraries/chain/platform_timer_bsd.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+#include <mutex>
+#include <thread>
+
+#include <sys/types.h>
+#include <sys/event.h>
+#include <sys/time.h>
+
+#ifndef KEVENT_FLAG_IMMEDIATE
+#define KEVENT_FLAG_IMMEDIATE 0x001
```
spoonincode, line 17: Isn't used. Looks like a holdover from trying to make macos & bsd impls shared? I think it's worth trying to find a way to make them shared since they do seem really similar -- something a couple `ifdefs` might be able to resolve?

766C6164: Yes, absolutely agreed, I am already working on single kevent based file for both osx/bsd, since they are very similar. It was just easier to play with separate file while debugging. Thank you!

linh2931 (approved):

spoonincode (approved):
nice, the differences ended up being small!

Issue #579, opened 2022-12-16
action return incorrect int64_t value


quocle108:
I expected to receive an action value is -909090909090909**09**. But the action return the value is -909090909090909**10**

```
  [[eosio::action]] int64_t getvalue2() {
        int64_t value = -90909090909090909;
        eosio::print("getvalue2: ", value);
        return value;
  }
```

You can reproduce this issue through this repo https://github.com/quocle108/return-error

heifner:
Thanks for the excellent bug report.

spoonincode:
There appears to be an oversight in fc's json conversion that doesn't stringify small values that can't be represented in a javascript number: -90909090909090909 is smaller than -(2<sup>53</sup> - 1)

https://github.com/AntelopeIO/leap/blob/167bcf781cc07f0111d29a8489b87244f73ff6ce/libraries/libfc/src/io/json.cpp#L620-L627

This would only be a problem when consuming the JSON returned from a trace via javascript though (i.e. cleos is fine); but that's likely what is happening here since using reporter is using qtest

quocle108:
> This would only be a problem when consuming the JSON returned from a trace via javascript though (i.e. cleos is fine); but that's likely what is happening here since using reporter is using qtest

it might be. Qtest gets the return value from eosjs. And javascript use Number (64-bit Floating Point) to present the number value, so rounding the number might be the issue?

BTW, it should be returned string in both cases even the value i > 0xffffffff or <= 0xffffffff.

spoonincode:
Right, that number cannot be represented in a javascript primitive type. Just running
```js
let x = -90909090909090909;
console.log(x);
```
in node,
```
$ node num.js
-90909090909090910
```

It would need to be represented as a bigint (or string).

PR #1060, opened 2023-04-19
Properly stringify both large and small int64's when converting to json

766C6164:
This PR contains fix (and ut's) for the following issue (https://github.com/AntelopeIO/leap/issues/579):

I expected to receive an action value is -90909090909090909. But the action return the value is -90909090909090910

  [[eosio::action]] int64_t getvalue2() {
        int64_t value = -90909090909090909;
        eosio::print("getvalue2: ", value);
        return value;
  }
You can reproduce this issue through this repo https://github.com/quocle108/return-error


libraries/libfc/src/io/json.cpp
```diff
@@ -596,7 +596,7 @@ namespace fc
          {
               int64_t i = v.as_int64();
               if( format == json::output_formatting::stringify_large_ints_and_doubles &&
-                  i > 0xffffffff )
+                  (i * ((i>0)-(i<0)) > 0xffffffff))
```
heifner, line 599: I find this construct rather odd and I am a bit surprised that compilers would not give a warning about bool used in asthmatic statement.

Seems like: `(i > 0xffffffff) || (i < -0xffffffff)` would be clearer.

mikelik: how about `std::abs(i) > 0xffffffff`
probably `cmath` should be included

smlu: Casting to `uint64_t` should also do the job, i.e. `static_cast<uint64_t>(i) > 0xffffffff`

766C6164: Thank you, all great suggestions! Changed.

libraries/libfc/src/io/json.cpp
```diff
@@ -595,12 +595,13 @@ namespace fc
          case variant::int64_type:
          {
               int64_t i = v.as_int64();
+              int64_t max_value(0xffffffff);
               if( format == json::output_formatting::stringify_large_ints_and_doubles &&
```
greg7mdp, line 599: I wonder if it shouldn't be `||`?

libraries/libfc/src/io/json.cpp
```diff
@@ -595,12 +595,13 @@ namespace fc
          case variant::int64_type:
          {
               int64_t i = v.as_int64();
+              int64_t max_value(0xffffffff);
```
mikelik, line 598:
```suggestion
              constexpr int64_t max_value(0xffffffff);
```

libraries/libfc/src/io/json.cpp
```diff
@@ -595,12 +595,13 @@ namespace fc
...
               if( format == json::output_formatting::stringify_large_ints_and_doubles &&
-                  i > 0xffffffff )
+                  (i > max_value || i < -max_value))
                  os << '"'<<v.as_string()<<'"';
               else
                  os << i;
-
+
```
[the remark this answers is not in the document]
766C6164: removed

libraries/libfc/src/io/json.cpp
766C6164: added, thank you

mikelik (approved):

heifner (approved):

PR #1070, opened 2023-04-21
Changes for building with c++20.

greg7mdp:
With these changes, plus the ones in the PRs opened in the following submodules:

- bn256
- eos-vm
- CLI11
- abieos

Leap builds with clang-16 in c++20 mode.


libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
```diff
@@ -37,11 +37,14 @@ struct unapplied_transaction {
...
    fc::time_point_sec expiration()const { return trx_meta->packed_trx()->expiration(); }
-
+#if __cplusplus >= 202002L
+   [[no_unique_address]] noncopyable _;
+#else
    unapplied_transaction(const unapplied_transaction&) = delete;
    unapplied_transaction() = delete;
    unapplied_transaction& operator=(const unapplied_transaction&) = delete;
    unapplied_transaction(unapplied_transaction&&) = default;
+#endif
```
heifner, line 47: Instead, I think it would be better to just add the needed constructor:
```
   unapplied_transaction(transaction_metadata_ptr t, trx_enum_type type, bool return_failure_trace = false, next_func_t next = {})
   : trx_meta(std::move(t)), trx_type(type), return_failure_trace(return_failure_trace), next(std::move(next)) {}
```

plugins/chain_plugin/trx_retry_db.cpp
```diff
@@ -52,11 +52,14 @@ struct tracked_transaction {
...
    size_t memory_size()const { return ptrx->get_estimated_size() + trx_trace_v.estimated_size() + sizeof(*this); }
-
+#if __cplusplus >= 202002L
+   [[no_unique_address]] noncopyable _;
+#else
    tracked_transaction(const tracked_transaction&) = delete;
    tracked_transaction() = delete;
    tracked_transaction& operator=(const tracked_transaction&) = delete;
    tracked_transaction(tracked_transaction&&) = default;
+#endif
```
heifner, line 62: I think we should instead just add the needed constructor that will work in pre-c++20 as well.

spoonincode:
It seems gcc12 is failing with a problem in LLVM's headers

greg7mdp:
> It seems gcc12 is failing with a problem in LLVM's headers

Yes, that is correct, that is the one issue I wasn't able to fix yet. That is why I wrote that it builds with clang-16. I did fix some other issues with gcc-12, but I was stumped on that one. 

I actually was thinking of asking you (once this was the last issue) to see if you had an idea, since you understand what is done there much better than I do, but otherwise I'll keep looking at it.

plugins/chain_plugin/trx_retry_db.cpp
greg7mdp: Does it work though? We are using the aggregate initialization when inserting into the boost multi-index, and adding any declared constructor makes the struct not an aggregate anymore in c++20.

greg7mdp:
> It seems gcc12 is failing with a problem in LLVM's headers

This is a known issue with the LLVM headers, see [this](https://stackoverflow.com/questions/67698523/error-in-llvm-stlextras-expected-unqualified-id-before-const-with-gcc11).

I changed /usr/include/llvm-11/llvm/ADT/STLExtras.h to remove the unneeded `<R>` and it does build fine.

libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
greg7mdp: I would have to change the `insert` calls, as in:
```
         auto insert_itr = queue.insert(
               { trx, api_trx ? trx_enum_type::incoming_api : trx_enum_type::incoming_p2p, return_failure_trace, std::move( next ) } );
```
because wiith the constructor you suggest does not accept initializer lists.

Also the intention appeared to have been to disable copied (by deleting copy constructor and assignment operator).

heifner: In my opinion modifying the insert calls or adding additional constructors or removing the disable of copy anything is better than the #ifdef.

plugins/chain_plugin/trx_retry_db.cpp
heifner: Anything is better than than #ifdef. Find a solution that works for both. Easiest is probably to just remove the `=delete`. This is all in a cpp file, we don't need those.

greg7mdp: I'll remove the `#ifdef` once we switch the build to `-std=c++20`. This is just temporary.

heifner: I think this is rather ugly, would not want to see the addition of `struct noncopyable` and `#ifdef` even temporarily.

greg7mdp: OK, I removed the whole `#ifdef`.

libraries/chain/include/eosio/chain/unapplied_transaction_queue.hpp
greg7mdp: OK, I removed the whole `#ifdef`.

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -87,6 +87,14 @@ namespace eosio::chain {
 #endif
    struct void_t{};
 
+   struct noncopyable {
+      noncopyable() = default;
+      noncopyable(noncopyable&&) = default;
+      noncopyable& operator=(noncopyable&&) = default;
+      noncopyable(const noncopyable&) = delete;
+      noncopyable& operator=(const noncopyable&) = delete;
+   };
```
heifner, line 96: No longer used.

greg7mdp: removed.

heifner (approved):

greg7mdp:
@spoonincode @linh2931 @arhag @larryk85 can I get a second approval in this? Thanks!

linh2931 (approved):

Issue #621, opened 2023-01-05
possible to register a host function and neglect to map it to an OC call index w/o compile error


spoonincode:
Transfer of eosnetworkfoundation/mandel#197

In the 2.0 implementation of OC's host function plumbing, a compiler error would result if a developer added a host function but neglected to assign it an OC call index.

There appears to be a regression in 3.1+ due to new host function code being backported from 2.1 that effectively removed this check. That is, one can add a
```c++
REGISTER_LEGACY_CF_HOST_FUNCTION(banana_stand)
```
but then not modify `libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/intrinsic_mapping.hpp` and complete the compile successfully.

This isn't a user facing bug, so I've just marked it as tech debt. imo (from my memory of doing this work years ago in 2.0) it's much easier to implement this check (and fix the regression) once we go c++20 due to increased constexpr across the stdlib

bhazzard:
Deferring for now, until we've switched to c++20 (https://github.com/AntelopeIO/leap/issues/922).

PR #1071, opened 2023-04-22
BLS12-381 Crypto Primitives

mschoenebeck:
Added library [mschoenebeck/bls12-381](https://github.com/mschoenebeck/bls12-381) as a submodule. For `x86_64` there are high-speed asm implementations of all low-level arithmetic functions available, which optionally make use of `adc` and `bmi2` CPU extensions (if available on host machine, live dispatcher via `init` call).

Added the following types to `leap`:
```
bls_scalar_type  = std::array<uint64_t, 4>;
bls_g1_type      = bls12_381::g1;
bls_g2_type      = bls12_381::g2;
bls_gt_type      = bls12_381::fp12;
bls_fp_type      = bls12_381::fp;
bls_fp2_type     = bls12_381::fp2;
```

Added the following host functions to `leap` (according to https://eips.ethereum.org/EIPS/eip-2537):
```
bls_g1_add(bls_g1_type op1, bls_g1_type op2, bls_g1_type result);
bls_g2_add(bls_g2_type op1, bls_g2_type op2, bls_g2_type result);
bls_g1_mul(bls_g1_type point, bls_scalar_type scalar, bls_g1_type result);
bls_g2_mul(bls_g2_type point, bls_scalar_type scalar, bls_g2_type result);
bls_g1_exp(bls_g1_type[] points, bls_scalar_type[] scalars, bls_g1_type result);
bls_g2_exp(bls_g2_type[] points, bls_scalar_type[] scalars, bls_g2_type result);
bls_pairing(bls_g1_type[] g1_points, bls_g2_type[] g2_points, bls_gt_type result);
bls_g1_map(bls_fp_type e, bls_g1_type result);
bls_g2_map(bls_fp2_type e, bls_g2_type result);
bls_fp_mod(std::array<uint8_t, 64> s, bls_fp_type result);
```

Check out [mschoenebeck/aggsigtest](https://github.com/mschoenebeck/aggsigtest) for a sample contract using those new host functions.


libraries/chain/controller.cpp
```diff
@@ -338,6 +338,8 @@ struct controller_impl {
       set_activation_handler<builtin_protocol_feature_t::get_code_hash>();
       set_activation_handler<builtin_protocol_feature_t::get_block_num>();
       set_activation_handler<builtin_protocol_feature_t::crypto_primitives>();
+      set_activation_handler<builtin_protocol_feature_t::bls_primitives>();
+      bls12_381::init();
```
mschoenebeck, line 342: Is this the correct place for the library's `init` function? The call to `init` only needs to happen once when `nodeos` starts. It is a live dispatcher that checks if `adc` and `bmi2` cpu features are available and if so sets the faster asm routines.

libraries/chain/protocol_feature_manager.cpp
```diff
@@ -258,6 +258,17 @@ Adds new cryptographic host functions
 Builtin protocol feature: GET_BLOCK_NUM
 
 Enables new `get_block_num` intrinsic which returns the current block number.
+*/
+            {}
+         } )
+         (  builtin_protocol_feature_t::bls_primitives, builtin_protocol_feature_spec{
+            "BLS_PRIMITIVES",
+            fc::variant("2c302889ce2db124878f7c6092cf27519d450559ed2fdfbad14f532d90fc5139").as<digest_type>(),
```
mschoenebeck, line 266: This SHA256 hash value of the comment message below is probably incorrect. I haven't been able to reproduce any of the other hashes, so I assume the one I generated is probably wrong.

heifner: The SHA256 hash includes the trailing newline.
For example, GET_BLOCK_NUM, hash the string between the /**/, quoted below, including the trailing newline but not the newline before "Builtin".
```
"Builtin protocol feature: GET_BLOCK_NUM

Enables new `get_block_num` intrinsic which returns the current block number.
"
```

heifner: So yours would be `01969c44de35999b924095ae7f50081a7f274409fdbccb9fc54fa7836c76089c`

mschoenebeck: Hello Kevin. Thank you very much for clarifying that. I tried a few different things but not that. I am now able to reproduce all the hashes.

Corrected the hash value.

libraries/chain/webassembly/crypto.cpp
```diff
@@ -250,4 +251,105 @@ namespace eosio { namespace chain { namespace webassembly {
...
+
+   void interface::bls_g1_exp(span<const char> points, span<const char> scalars, const uint32_t n, span<char> result) const
+   {
+       std::vector<bls12_381::g1> pv;
+       std::vector<std::array<uint64_t, 4>> sv;
+       pv.reserve(n);
+       sv.reserve(n);
+       for(uint32_t i = 0; i < n; i++)
+       {
+           bls12_381::g1 p = bls12_381::g1::fromJacobianBytesLE({reinterpret_cast<const uint8_t*>(points.data() + i*144), 144}, false, true);
```
spoonincode, line 294: Before using the `span<>`s passed as arguments, need to validate that the size of the span is what is required.

Consider, for example, a contract currently with 64KB of memory calling `bls_g1_exp(64KB-8, 4, ..., n=1, ...)`. The host function validation wrappers will allow this call because the `points` span is 4 bytes long and within the 0 to 64KB-1 valid range of memory. However, `bls12_381::g1::fromJacobianBytesLE({reinterpret_cast<const uint8_t*>(points.data() + i*144), 144}` then goes on to use 144 bytes starting from 64KB-8. This will access invalid WASM memory leading to undefined behavior (what will likely happen _in this case_ and _at this point of time_ is a recovered SEGV which leaks the memory allocated by the `std::vector<>::reserve()` above).

Host functions must ensure they only access memory within the passed `span<>` ranges. So instead, need to validate that `points.size() == n*144`. Consider, for example, this host function
https://github.com/AntelopeIO/leap/blob/ba919c794165a9702df7c2ed6b09ac85ee64bd74/libraries/chain/webassembly/crypto.cpp#L119-L126
notice how if `op1` is not 64 bytes an error is returned. Since `bls_g1_exp()` does not return anything, the only reasonable option for `bls_g1_exp` may be to assert that `points.size() == n*144` (i.e. fail the transaction).

This same guidance applies to `result` as well. Notice that the above `alt_bn128_add` example checks that `result` is _at least_ 64 bytes. That may not be ideal and may have been an oversight in the original implementation -- I would suggest enforcing that the `result` here is exactly 144 bytes.

(this feedback applies to all the added host functions)

libraries/chain/webassembly/crypto.cpp
```diff
@@ -250,4 +251,141 @@ namespace eosio { namespace chain { namespace webassembly {
...
+      v.reserve(n);
+      for(uint32_t i = 0; i < n; i++)
+      {
+         bls12_381::g1 p_g1 = bls12_381::g1::fromJacobianBytesLE({reinterpret_cast<const uint8_t*>(g1_points.data() + i*144), 144}, false, true);
+         bls12_381::g2 p_g2 = bls12_381::g2::fromJacobianBytesLE({reinterpret_cast<const uint8_t*>(g2_points.data() + i*288), 288}, false, true);
+         bls12_381::pairing::add_pair(v, p_g1, p_g2);
+         if(i%4 == 0)
+            context.trx_context.checktime();
+      }
+      bls12_381::fp12 r = bls12_381::pairing::calculate(v);
```
spoonincode, line 356: In the case of `v.size() == 80000` how long does `bls12_381::pairing::calculate()` take?

libraries/chain/webassembly/crypto.cpp
mschoenebeck: Thanks for the detailed explanation and example. Much appreciated! I was expecting the execution to fail in those cases but didn't consider potential memory leaks. Of course, these checks are very important.

Updates in my most recent commit:

- added checks for all `span<>` ranges
- make use of `return_code::success`/`return_code::failure` as in BN host functions
- also see related changes in `cdt` (proper error propagation back to smart contract and then handling with `eosio::check`, again, exactly as in BN host functions)
- fixed indentation of tabs (3 ws instead of 4)

libraries/chain/webassembly/crypto.cpp
```diff
@@ -250,4 +251,105 @@ namespace eosio { namespace chain { namespace webassembly {
...
+   {
+       std::vector<std::tuple<bls12_381::g1, bls12_381::g2>> v;
+       v.reserve(n);
+       for(uint32_t i = 0; i < n; i++)
+       {
+           bls12_381::g1 p_g1 = bls12_381::g1::fromJacobianBytesLE({reinterpret_cast<const uint8_t*>(g1_points.data() + i*144), 144}, false, true);
+           bls12_381::g2 p_g2 = bls12_381::g2::fromJacobianBytesLE({reinterpret_cast<const uint8_t*>(g2_points.data() + i*288), 288}, false, true);
+           bls12_381::pairing::add_pair(v, p_g1, p_g2);
+       }
+       bls12_381::fp12 r = bls12_381::pairing::calculate(v);
```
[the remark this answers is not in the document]
mschoenebeck: Again, thanks for taking the time explaining everything so well. I really appreciate it. I already noticed calls to `context.trx_context.checktime()` here and there, but forgot to ask about it.

There are three host functions with loops. According to your rule of thumb (approx. one check per ms) I did some time measuring on my machine to get an idea about how long each one of the loop runs (one iteration). Here the reasoning behind the numbers I chose:

- `g1_exp`: `155ns` per loop `=> times 10 = 1.55ms`
- `g2_exp`: `260ns` per loop `=> times 6 = 1.56ms`
- `pairng`: `395ns` per loop `=> times 4 = 1.58ms`

Assuming significantly better HW on BP nodes I thought aiming for `1.5ms` on my machine should be fine. Also having at least 3 pairs going through without `checktime()` overhead would be a very common case for groth16 proof verification (fwiw).

Update:
- added `context.trx_context.checktime()` calls in reasonable intervals

libraries/chain/webassembly/crypto.cpp
mschoenebeck: On my machine about 20 seconds. I was thinking about it after I read you previous comment: Does that mean, we need to be able to pass the `yield()` function to `pairing::calculate()` and execute it there in reasonable intervals as well?

spoonincode: > since now with the `checktime()` inside the loop it wouldn't be possible to add that many pairs to `v` anymore

yeah, good point. But is there some number that takes, for example, 25ms to make it through the loop, but then takes 100ms to make it through `bls12_381::pairing::calculate()`? That too would be considered a security defect. I'm not sure of the timing for these operations so maybe it's a non issue.

mschoenebeck: Good question. So I did some measuring again:

It takes `v.size() == 58` pairs in order for the entire loop to take `25ms` (on my machine). The call to `pairing::calculate(v)` then takes about an additional `14.7ms`.

I'm not exactly sure how those number would vary on a faster machine but I guess it would be a similar ratio between loop execution and paring calculation.

spoonincode: It's "only" ~10ms but I think we should be prudent and protect against that -- it definitely would fail previous guidelines we had back in the bug bounty days.

Since we control the library, the best choice is to pass in the ability for `bls12_381::pairing::calculate()` to poll a yield function. An example of that is,
https://github.com/AntelopeIO/leap/blob/7c0694b4c965bf12078680cb4062f092554be0c5/libraries/chain/webassembly/crypto.cpp#L137-L139
where the declaration of that is
```c++
/// pairing_check calculates the Optimal Ate pairing for a set of points.
///  @param marshaled_g1g2_pairs marshaled g1 g2 pair sequence
///  @return -1 for unmarshal error, 0 for unsuccessful pairing and 1 for successful pairing
int32_t pairing_check(std::span<const uint8_t> marshaled_g1g2_pairs, std::function<void()> yield);
```
if you wanted to make the yield optional for users, you could potentially do something like
```c++
int32_t pairing_check(std::span<const uint8_t> marshaled_g1g2_pairs, std::function<void()> yield = std::function<void()>());
```
or just require users to pass in a stub function. Either way is fine; importantly either way ultimately still allows the library to used outside of leap.

(there is another option: add a subjective limit to the size of `n`; for example
https://github.com/AntelopeIO/leap/blob/7c0694b4c965bf12078680cb4062f092554be0c5/libraries/chain/webassembly/crypto.cpp#L146-L156
but I can't find any compelling reason to pick this choice due to our control of the bls lib)

mschoenebeck: Agree. Since there is a `leap` branch of the library now, let's try to tailor it perfectly for leap integration.

I added a `yield` function pointer parameter to `pairing::calculate()` (`pairing::miller_loop()` respectively) and `g2::multiExp()` since this one could also run for about `15ms` in the `25ms`-loop worst case. Here are the numbers for all three functions (my machine):

- G1 multi-exponentiation: with `v.size() == 161` the loop takes `25ms` => `g1::multiexp(v)` then takes only `5.1ms`
- G2 multi-exponentiation: with `v.size() == 96` the loop takes `25ms` => `g2::multiexp(v)` then takes `15.3ms`
- Pairing: with `v.size() == 58` the loop takes `25ms` => `pairing::calculate(v)` then takes `14.7ms`

The timings of `g1::multiExp()` should be fine, I guess, since `25ms + 5.1ms ≅ 30ms`. So if `5ms` to `7ms` is an acceptable 'overhead' for the execution time of either `pairing::calculate()` or `g1/g2::multiExp()` after their corresponding loops (assuming the worst case vector sizes listed above with `25ms` preceding loop execution time) then only one `yield()` call approximately in the middle of execution of `pairing` and `multiexp` should be fine, right? Worst case scenario is then about `32ms` total execution time in case the transaction fails.

It was a bit tricky though, to integrate the `yield()` call accurately into the middle of the `pairing` and `mutliexp` functions because there are multiple loops (including nested loops) of which some depend on `v.size()`. However, I believe I found a good way to add it approximately in the middle of each function's execution path, so that the time check is performed once if `v` exceeds a certain size.

mschoenebeck:
- Updated the BLS library to not throw any exceptions anymore but instead return [`optional`](https://en.cppreference.com/w/cpp/utility/optional) values (return `none` in case of error).
- All possible errors are now handled directly inside the host functions. However, I didn't introduce any more error codes yet but instead kept using `return_code::failure`. Please take a look at `bls_g1_add` and the other host functions:
https://github.com/mschoenebeck/leap/blob/233ae01a92caf4160a905ccb00ac3c7bc55091df/libraries/chain/webassembly/crypto.cpp#L258C1-L262C1

- Fixed the wrong comment in BLS lib (that was a copy&paste error from the reference lib):
https://github.com/mschoenebeck/bls12-381/blob/71a180eae5b3e40ba5f9fa67d9d2998408640306/src/g.cpp#L499C1-L499C1

- Added unit tests for garbage i/o in `add`, `mul` and `exp` functions:
https://github.com/mschoenebeck/leap/blob/233ae01a92caf4160a905ccb00ac3c7bc55091df/libraries/libfc/test/crypto/test_bls.cpp#L135-L217

- Added unit tests for error codes of `pairing` & `map` host functions:
https://github.com/mschoenebeck/aggsigtest/blob/c927ce50a89fe4d6d1b123ee97bbf039fc90625d/aggsigtest.cpp#L204C1-L229C2

mschoenebeck:
I have one question about the correct place to link the bls library. I noticed that `bn256` is actually linked to `chain` library despite being a submodule of `fc`. It is not linked against `fc` though.

https://github.com/AntelopeIO/leap/blob/7c0694b4c965bf12078680cb4062f092554be0c5/libraries/chain/CMakeLists.txt#L132C1-L134C23

I linked the bls lib to `fc` since it is a submodule of that library. Probably a minor thing but what is the correct way?

https://github.com/mschoenebeck/leap/blob/90071617799738820350538f632ca5d5fd54e5e8/libraries/libfc/CMakeLists.txt#L121C1-L122C174

libraries/chain/webassembly/crypto.cpp
arhag: @mschoenebeck: I think we should pass the yield function into `g1::multiExp()` as well for consistency.

We should not assume that there will always be a limit of 30 ms in a transaction. We are proposing that the network upgrade to a higher max transaction time. With a max transaction time of 150 ms, the user could pass a 5x larger array and thus increase the time of the `g1::multiexp` function call by 5x as well.

libraries/libfc/test/crypto/test_bls.cpp
```diff
@@ -0,0 +1,219 @@
...
+    BOOST_CHECK_EQUAL(x.isValid(), false);
+    BOOST_CHECK_EQUAL(p.isOnCurve(), false);
+    BOOST_CHECK_EQUAL(p.inCorrectSubgroup(), false);
+    array<uint64_t, 4> s = {0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF, 0xFFFFFFFFFFFFFFFF};
+    p = g2::multiExp({p}, {s}).value();
+    BOOST_CHECK_EQUAL(p.isOnCurve(), false);
+    BOOST_CHECK_EQUAL(p.inCorrectSubgroup(), false);
+    g2 p_res = g2::fromJacobianBytesBE(hexToBytes<288>("0x158a2a1e3ce68c49f9795908aa3779c6919ed5de5cbcd1d2a331d0742d1eb3cb28014006b5f686204adb5fdca73aea570ee0f0d58880907c8de5867dd99b6b7306b2c3de4a1537e6d042f2b8e44c8086853728cc246726016b0fcf993db3d759005f8ac0cb55113c857c5cf3f83d9b624ce9a2a0a00a1206777cf935721c857b322a611ed0703cf3e922bfb8b19a1f5e10a341b2191ab5a15d35f69850d2adb633e5425eecb7f38dd486a95b3f74d60f3ee6cf692b3c76813407710630763f7605b3828c19203f661732a02f7f546ab354694128bbe5a792a9db4a443c0fe10af0df2bc1b8d07aee99bd6f8c6b26847011aa31634f42f722d52022c736369db470576687fdf819cf15a0db4c01a0bd7028ee17cefdf6d66557d47fb725b6d00f"), false, true).value();
+    BOOST_CHECK_EQUAL(p.equal(p_res), true);
+} FC_LOG_AND_RETHROW();
```
arhag, line 216: These tests of garbage input are good, but I would like to see unit tests that directly test the code paths of the host functions for: these garbage cases; failure cases that return -1; and even one successful case for each function (taken from the libraries test vectors).

This will require adding a new test contract.

See how it was done for the alt_bn_128 host functions: https://github.com/eosnetworkfoundation/mandel/pull/316/files#diff-87c8848613f6d5f711354edd61feeebcaae9b6acd4f5f16ca651075b989dc929

mschoenebeck:
I synced with main. However, I had to resolve a merge conflict in [`unittests/deep-mind/deep-mind.log`](https://github.com/AntelopeIO/leap/compare/main...mschoenebeck:leap:main#diff-eec56ede5875e46c1a6bf5e7c33de5320e88a8cb3ec63e4c369e4a2d1314f5a6). I'm pretty sure it is not correctly resolved since it shows a ton of changes to the main branch. Please take a look.

libraries/libfc/test/crypto/test_bls.cpp
spoonincode: yeah, that failure is because the host function is named `bls_pairing` which doesn't match,
https://github.com/AntelopeIO/leap/blob/712b1a0faa2438dcc28b77bd807c0f1b99b02dd8/libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/intrinsic_mapping.hpp#L277
Fixing the entry in intrinsic_mapping.hpp will resolve it.

(in 2.0 this mismatch would have ended up as a compile error, we lost that somewhere between 2.0 & 3.1 -- #621)

mschoenebeck: Thanks, Matt. Fixed in: https://github.com/AntelopeIO/leap/pull/1071/commits/59cc737a42611ede24d423aaf5fff6c578de039e

mschoenebeck:
Quick update: I was able to resolve the `deep-mind.log` conflicts myself. Fixed in: https://github.com/AntelopeIO/leap/pull/1071/commits/59cc737a42611ede24d423aaf5fff6c578de039e

arhag (approved):
Overall this looks good to me. There are some small tweaks the ENF will do including getting the CI to fully pass. But we are good as it is to merge this into the `bls_integraton` branch.

PR #1091, opened 2023-04-26
[4.0 -> main] fix: not fail on signature checks for dry-run txns

heifner:
In this method, `check_but_dont_fail` is equivalent to whether the transaction is `dry-run`. However, when `dry-run` is `true`, authorization should not be checked. 

Backport of https://github.com/AntelopeIO/leap/pull/1081 already merged into `main` which is why this is an empty diff.
Merges #1089 & #1090 of `release/4.0` into `main`
Thanks to @PinelliaC for the fix.


greg7mdp:
Changes are already in main. Do we still need this PR?

greg7mdp (approved):

heifner:
> Changes are already in main. Do we still need this PR?

Yes, it tells git that these changes have already been merged, so future merges do not pull them in.

spoonincode (approved):

Issue #641, opened 2023-01-17
clean function for escaping control characters in a string and truncating it


spoonincode:
When we log an exception from a contract, we want to remove any new lines put in by the contract writer and limit the size of the message. As suggested in the review of eosnetworkfoundation/mandel#733, use of `fc::format_string` is inefficient; a general purpose `clean` method that optionally escapes control characters (specifically newline) and truncates can be created for such a purpose.

This is a transfer of eosnetworkfoundation/mandel#803 originally entered by @linh2931, it is being moved to `leap` for easier visibility.

heifner:
Specific suggestion: https://github.com/eosnetworkfoundation/mandel/pull/733#discussion_r939331290

Issue #586, opened 2022-12-20
Reorganize tests for plugins


arhag:
Currently we have different types of tests stored under a `test` folder for some plugins, e.g. chain_plugin and producer_plugin. 

Some of these are proper unit tests of the plugin code in isolation. Those can remain in the folder but we should group them all together in a single executable rather than using different executable for each.

Some of these are bringing in other libraries and/or plugins and really should be considered integration tests (not to be confused with the Python end-to-end tests). These belong somewhere outside of the directories for the plugins. I suggest putting them in a sub-directory within the `tests` folder. It would also seem appropriate to bundle these tests together into one executable rather than linking separate executable for each test suite. (~~Regarding any such tests that use appbase, note that there is currently a limitation that prevents multiple testcases using appbase to exist in the same executable. That limitation should first be fixed; see https://github.com/AntelopeIO/leap/issues/585. - fixed [here](https://github.com/AntelopeIO/leap/pull/636)~~)

spoonincode:
>  It would also seem appropriate to bundle these tests together into one executable rather than linking separate executable for each test suite

fwiw, one benefit to separate executables is that it becomes trivially parallelizeable with `ctest -j`

heifner:
`unittests/plugin_tests.cpp` should be moved from `unittests/` to `tests/` and `unittests/CMakeLists.txt` should not include target include of `${CMAKE_SOURCE_DIR}/plugins/http_plugin/include` or `${CMAKE_SOURCE_DIR}/plugins/chain_interface/include`

PR #1095, opened 2023-04-27
Refactor libfc tests into one executable

heifner:
In preparation of adding some new tests for #641, a bit of refactoring of `libfc` tests.
Combine all tests into one executable. Combined they run in a fraction of a second.
Combining them seems in line with #586.


libraries/libfc/test/CMakeLists.txt
```diff
@@ -1,12 +1,13 @@
-add_subdirectory( crypto )
-add_subdirectory( io )
-add_subdirectory( network )
-add_subdirectory( scoped_exit )
-add_subdirectory( static_variant )
-add_subdirectory( variant )
-add_subdirectory( variant_estimated_size )
+file(GLOB UNIT_TESTS
```
spoonincode, line 1: Not a huge fan of `file(GLOB)` since it's easy to be confused why a new .cpp file isn't being compiled. But now that we're on cmake 3.12+ we can use `CONFIGURE_DEPENDS` which makes it less of a trap.

Also, I wonder about using `GLOB_RECURSE` here so we don't need to specify each sub directory?

spoonincode: I think it would have been interesting to try `CONFIGURE_DEPENDS` + `GLOB_RECURSE` and see how it goes. Up to you.

spoonincode (approved):

libraries/libfc/test/test_base64.cpp
```diff
@@ -1,4 +1,4 @@
-#define BOOST_TEST_MODULE base64
+#define BOOST_TEST_MODULE libfc
 #include <boost/test/included/unit_test.hpp>
```
greg7mdp, line 2: Did you leave `included` on purpose here?

libraries/libfc/test/crypto/test_utils.hpp
```diff
@@ -1,9 +1,9 @@
-uint32_t to_uint32(const std::string& s) {
+inline uint32_t to_uint32(const std::string& s) {
```
greg7mdp, line 1: Just our of curiosity, did you need to add `inline` here for a specific reason?

greg7mdp (approved):

libraries/libfc/test/test_base64.cpp
greg7mdp: Hum I think the answer is yes!

spoonincode: Yeah I thought that was a little sneaky how "the magic include" was simply buried in one of the test files like this. It might be clearer to have a discrete `main.cpp` with simply
```c++
#define BOOST_TEST_MODULE libfc
#include <boost/test/included/unit_test.hpp>
```

libraries/libfc/test/crypto/test_utils.hpp
heifner: Yes, it is included in 3 of the test files.

libraries/libfc/test/test_base64.cpp
heifner: Yes on purpose. I agree that is too sneaky, so added a `main.cpp` as Matt suggested.

libraries/libfc/test/crypto/test_utils.hpp
greg7mdp: Oh, I get it. I probably would have added `static`. I didn't think that `inline` changed the linkage status (external) for non-member functions.
Actually, I don't believe it does, and I'm surprised that adding `inline` fixed the issue.

heifner: It is literally the only use of `inline` these days. Compilers typically ignore `inline` for determining if a function should be inlined.
https://en.cppreference.com/w/cpp/language/inline

greg7mdp: Yes, it is somewhat equivalent now.  Do you feel there is a reason to prefer `inline` to `static`? It may be moot in this case, but if the functions are not actually inlined,  `static` and `inline` would make a difference.

heifner: Maybe more of a preference. https://stackoverflow.com/questions/10876930/should-one-never-use-static-inline-function I stopped using `static` years ago but can't remember the reason why.

greg7mdp: I believe that, for the case where the compiler does *not* inline the functions, `static` would cause each compilation unit to have a private copy of the function, while `inline` would cause each compilation unit to export a function with external linkage, which would then later be merged into one by the linker.
`static`: larger code size, less work for linker, maybe better performance as the code may be closer to where it is used 
`inline`: the opposite of above

greg7mdp (approved):

PR #1103, opened 2023-04-27
Fix g++-10 link issue when building with -std=c++20.

greg7mdp:
Within `libraries/wasm-jit/Source`, `WASM` depends on `IR` and `Logging`, and `IR` depends on `Logging`.
With the current declaration of `WASM::check_limits` in `WASM`, we have the following g++-10 link error when building with `-std=c++20`:

```
_64-linux-gnu/libz.so  -ldl  libraries/libfc/secp256k1/libsecp256k1.a  libraries/chainbase/libchainbase.a  -lpthread  libraries/softfloat/libsoftfloat.a && :
/usr/bin/ld: libraries/wasm-jit/Source/IR/libIR.a(DisassemblyNames.cpp.o): warning: relocation against `_ZN4WASM12check_limitsE' in read-only section `.text'
/usr/bin/ld: libraries/wasm-jit/Source/IR/libIR.a(DisassemblyNames.cpp.o): in function `void Serialization::serialize<Serialization::MemoryInputStream>(Serialization::MemoryInputStream&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&)':
/home/greg/github/enf/leap/libraries/wasm-jit/Include/Inline/Serialization.h:274: undefined reference to `WASM::check_limits'
/usr/bin/ld: warning: creating DT_TEXTREL in a PIE
collect2: error: ld returned 1 exit status
```

Moving the declaration of `WASM::check_limits` in `Logging` (end leaf of dependency graph) fixes the issue.


libraries/wasm-jit/Source/Logging/Logging.cpp
```diff
@@ -5,6 +5,10 @@
 #include <cstdarg>
 #include <cstdlib>
 
+namespace WASM
+{
+	bool check_limits = true;
+}
```
heifner, line 11: Did you consider using an `inline static` of a struct instead? Seems like that would be cleaner than `extern`.

greg7mdp: I do like the idea of an `inline static` struct member, as it relies on a C++ feature that I feel should be more foolproof. Still since it is not my code, and this is a workaround for a specific compiler issue, I'd prefer to keep the change as minimal as possible.

heifner (approved):

libraries/wasm-jit/Source/Logging/Logging.cpp
spoonincode: It seems a little strange putting something that has nothing to do with logging in the `Logging` library. Could it work in `IR` instead? The Logging library could easily be on the chopping block whenever someone has time to keep trimming down WAVM.

greg7mdp: Thanks Matt, it looks like it is working fine with the variable in `IR`. I'll push that change.

heifner (approved):

766C6164 (approved):

PR #1108, opened 2023-04-28
Optimization: New escape_str function for sanitizing strings

### LINKED ISSUE #641: clean function for escaping control characters in a string and truncating it

heifner:
Create a new `escape_str` function for use of sanitizing strings without having to use `format_string` with its overhead of `mutable_variant_object`.
Also optimize `print_debug` used for contract console output.

Resolves #641

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -19,15 +19,26 @@ namespace eosio { namespace chain {
 
 static inline void print_debug(account_name receiver, const action_trace& ar) {
    if (!ar.console.empty()) {
-      auto prefix = fc::format_string(
-                                      "\n[(${a},${n})->${r}]",
-                                      fc::mutable_variant_object()
-                                      ("a", ar.act.account)
-                                      ("n", ar.act.name)
-                                      ("r", receiver));
-      dlog(prefix + ": CONSOLE OUTPUT BEGIN =====================\n"
-           + ar.console
-           + prefix + ": CONSOLE OUTPUT END   =====================" );
+      if (fc::logger::get(DEFAULT_LOGGER).is_enabled( fc::log_level::debug )) {
+         std::string prefix;
+         prefix.reserve(3 + 13 + 1 + 13 + 3 + 13 + 1);
+         prefix += "\n[(";
+         prefix += ar.act.account.to_string();
+         prefix += ",";
+         prefix += ar.act.name.to_string();
+         prefix += ")->";
+         prefix += receiver.to_string();
+         prefix += "]";
+
+         std::string output;
+         output.reserve(512);
+         output += prefix;
+         output += ": CONSOLE OUTPUT BEGIN =====================\n";
+         output += ar.console;
+         output += prefix;
+         output += ": CONSOLE OUTPUT END   =====================";
+         dlog( std::move(output) );
+      }
    }
 }
 
--- a/libraries/libfc/include/fc/string.hpp
+++ b/libraries/libfc/include/fc/string.hpp
@@ -1,5 +1,6 @@
 #pragma once
 #include <cstdint>
+#include <limits>
 #include <string>
 
 namespace fc
@@ -20,4 +21,22 @@ namespace fc
   class variant_object;
   std::string format_string( const std::string&, const variant_object&, bool minimize = false );
   std::string trim( const std::string& );
+
+  /**
+   * Convert '\t', '\r', '\n', '\\' and '"'  to "\t\r\n\\\"" if escape_control_chars == true
+   * Convert all other < 32 & 127 ascii to escaped unicode "\u00xx"
+   * Removes invalid utf8 characters
+   * Escapes Control sequence Introducer 0x9b to \u009b
+   * All other characters unmolested.
+   *
+   * @param str input/output string to escape/truncate
+   * @param escape_control_chars if true escapes control chars in str
+   * @param max_len truncate string to max_len
+   * @param add_truncate_str if truncated by max_len, add add_truncate_str to end of any truncated string,
+   *                         new length with be max_len + strlen(add_truncate_str), nullptr allowed if no append wanted
+   * @return pair<reference to possibly modified passed in str, true if modified>
+   */
+  std::pair<std::string&, bool> escape_str( std::string& str, bool escape_control_chars = true,
+                                            std::size_t max_len = std::numeric_limits<std::size_t>::max(),
+                                            const char* add_truncate_str = "..." );
 }
--- a/libraries/libfc/include/fc/string.hpp
+++ b/libraries/libfc/include/fc/string.hpp
@@ -30,13 +30,14 @@ namespace fc
    * All other characters unmolested.
    *
    * @param str input/output string to escape/truncate
-   * @param escape_control_chars if true escapes control chars in str
+   * @param escape_ctrl if on escapes control chars in str
    * @param max_len truncate string to max_len
    * @param add_truncate_str if truncated by max_len, add add_truncate_str to end of any truncated string,
    *                         new length with be max_len + strlen(add_truncate_str), nullptr allowed if no append wanted
    * @return pair<reference to possibly modified passed in str, true if modified>
    */
-  std::pair<std::string&, bool> escape_str( std::string& str, bool escape_control_chars = true,
+  enum class escape_control_chars { off, on };
+  std::pair<std::string&, bool> escape_str( std::string& str, escape_control_chars escape_ctrl = escape_control_chars::on,
                                             std::size_t max_len = std::numeric_limits<std::size_t>::max(),
                                             const char* add_truncate_str = "..." );
 }
--- a/libraries/libfc/include/fc/string.hpp
+++ b/libraries/libfc/include/fc/string.hpp
@@ -23,7 +23,7 @@ namespace fc
   std::string trim( const std::string& );
 
   /**
-   * Convert '\t', '\r', '\n', '\\' and '"'  to "\t\r\n\\\"" if escape_control_chars == true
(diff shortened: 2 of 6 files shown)


libraries/chain/apply_context.cpp
```diff
@@ -19,15 +19,26 @@ namespace eosio { namespace chain {
...
+         prefix += "\n[(";
+         prefix += ar.act.account.to_string();
+         prefix += ",";
+         prefix += ar.act.name.to_string();
+         prefix += ")->";
+         prefix += receiver.to_string();
+         prefix += "]";
+
+         std::string output;
+         output.reserve(512);
```
linh2931, line 34: We didn't have this 512 limit before the change.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2085,23 +2085,22 @@ inline std::string get_detailed_contract_except_info(const packed_transaction_pt
...
       act_name = actions[0].name.to_string();
    }
 
-   details = except_ptr ? except_ptr->top_message() : (trace && trace->except) ? trace->except->top_message() : std::string();
-   if (!details.empty()) {
-      details = fc::format_string("${d}", fc::mutable_variant_object() ("d", details), true);  // true for limiting the formatted string size
-   }
+   std::string details = except_ptr ? except_ptr->top_message()
+                                    : (trace && trace->except) ? trace->except->top_message()
+                                                               : std::string();
```
greg7mdp, line 2101: Maybe parenthesize so precedence is clear
```suggestion
   std::string details = except_ptr ? except_ptr->top_message()
                                    : ((trace && trace->except) ? trace->except->top_message()
                                                               : std::string());
```

heifner: Done.

libraries/libfc/src/string.cpp
```diff
@@ -83,13 +85,42 @@ namespace fc  {
...
       return boost::algorithm::trim_copy(s);
-      /*
-      std::string cpy(s);
-      boost::algorithm::trim(cpy);
-      return cpy;
-      */
   }
 
+  std::pair<std::string&, bool> escape_str( std::string& str, bool escape_control_chars,
+                                            std::size_t max_len, const char* add_truncate_str )
```
greg7mdp, line 91: A suggestion only:

```suggestion
  enum class escape_control_chars { off, on };
  std::pair<std::string&, bool> escape_str( std::string& str, escape_control_chars escape_ctrl,
                                            std::size_t max_len, const char* add_truncate_str )
```

Which allows you to call it as:

`fc::escape_str(details, fc::escape_control_chars::on, 1024);`

which is nicer than having to declare a bool for clarity.

libraries/libfc/src/string.cpp
```diff
@@ -83,13 +85,42 @@ namespace fc  {
...
+     if (itr != str.end() || !fc::is_valid_utf8( str )) {
+        str = escape_string(str, nullptr, escape_control_chars);
+        modified = true;
+        if (str.size() > max_len) {
+           str.resize(max_len);
+           truncated = true;
+        }
+     }
+
+     if (truncated && add_truncate_str != nullptr ) {
```
linh2931, line 116: Just a style preference. `add_truncate_str != nullptr` can just be `add_truncate_str`.

libraries/libfc/include/fc/string.hpp
```diff
@@ -20,4 +21,22 @@ namespace fc
   class variant_object;
   std::string format_string( const std::string&, const variant_object&, bool minimize = false );
   std::string trim( const std::string& );
+
+  /**
+   * Convert '\t', '\r', '\n', '\\' and '"'  to "\t\r\n\\\"" if escape_control_chars == true
```
linh2931, line 26: Should "\t\r\n\\\"" be "\\t\\r\\n\\\\"?

libraries/chain/apply_context.cpp
heifner: This is not a limit, but a hint on memory allocation.

libraries/libfc/src/string.cpp
heifner: I very much prefer the style of explicit `!= nullptr`.

libraries/libfc/src/string.cpp
```diff
@@ -83,13 +85,42 @@ namespace fc  {
...
+     bool modified = false, truncated = false;
+     // truncate early to speed up escape
+     if (str.size() > max_len) {
+        str.resize(max_len);
+        modified = truncated = true;
+     }
+     auto itr = escape_control_chars
+           ? std::find_if(str.begin(), str.end(),
+                          [](const auto& c) {
+              return c == '\x7f' || c == '\\' || c == '\"' ||  (c >= '\x00' && c <= '\x1f'); } )
```
greg7mdp, line 102: Isn't `c >= '\x00'` always true?

libraries/libfc/src/string.cpp
```diff
@@ -83,13 +85,42 @@ namespace fc  {
...
+           : std::find_if(str.begin(), str.end(),
+                          [](const auto& c) {               // x09 = \t, x0a = \n,                   x0d = \r
+              return c == '\x7f' || (c >= '\x00' && c <= '\x08') || c == '\x0b' || c == '\x0c' || (c >= '\x0e' && c <= '\x1f'); } );
```
greg7mdp, line 105:
```suggestion
              return c == '\x7f' || (c <= '\x1f' && c != '\t' && c != '\n' && c != '\r'); } );
```

libraries/libfc/src/string.cpp
heifner: No `c` (char) is signed.

greg7mdp: OK, still in the case that `char` is signed, and that we have some negative ones, then your version does preserve the previous behavior.

libraries/libfc/src/string.cpp
linh2931: ok.

libraries/chain/apply_context.cpp
linh2931: oha,

libraries/libfc/src/string.cpp
heifner: Done.

libraries/libfc/include/fc/string.hpp
linh2931: Github automatically re-formatted two `\\t` into one `\t` above. In the test, `\t` becomes `\\t`.

heifner: Ah, so you think the comment should be:
```
Convert '\t', '\r', '\n', '\\' and '"'  to "\\t\\r\\n\\\\\"" if escape_ctrl == on
```
?

libraries/libfc/src/string.cpp
```diff
@@ -83,13 +85,42 @@ namespace fc  {
...
       return boost::algorithm::trim_copy(s);
-      /*
-      std::string cpy(s);
-      boost::algorithm::trim(cpy);
-      return cpy;
-      */
   }
 
+  std::pair<std::string&, bool> escape_str( std::string& str, escape_control_chars escape_ctrl,
+                                            std::size_t max_len, const char* add_truncate_str )
```
greg7mdp, line 91: I think the more modern version for the last parameter would be to be  be a `std::string_view` instead of a raw pointer.

libraries/libfc/include/fc/string.hpp
heifner: I guess I read the comment as saying that is what the output would look like, not what its internal representation in the string would be.

libraries/libfc/src/string.cpp
heifner: Done.

greg7mdp (approved):

libraries/libfc/src/string.cpp
```diff
@@ -83,13 +85,42 @@ namespace fc  {
...
+     if (itr != str.end() || !fc::is_valid_utf8( str )) {
+        str = escape_string(str, nullptr, escape_ctrl == escape_control_chars::on);
+        modified = true;
+        if (str.size() > max_len) {
+           str.resize(max_len);
+           truncated = true;
+        }
+     }
+
+     if (truncated && !add_truncate_str.empty()) {
```
greg7mdp, line 116:
```suggestion
     if (truncated) {
```

heifner: I thought about that, but the impl of `string.append()` would make additional `if` checks not needed.

heifner: oh, but will almost never be empty(), I'll remove the check.

greg7mdp: [operator+=](https://en.cppreference.com/w/cpp/string/basic_string/operator%2B%3D) has a version that takes a StringViewLike, so the test would de done there instread of in the `if` before. I'm not seeing any drawback.

heifner: The impl of the `operator+=` that takes a `string_view` does:
`append(sv.data(), sv.size())` and then it checks `if (size)` in the impl of `append`. Regardless, I agree and have removed the check.

greg7mdp (approved):

PR #1112, opened 2023-05-01
Tests: misc cleanup

### LINKED ISSUE #259: Use github issue instead of #warning directive in wasm_tests.cpp

heifner:
- Remove unused `NON_VALIDATING_TEST` macro.
- Add `operator<<` for `setup_policy` so `setup_policy` can be used with boost data in tests.
- Allow the `setup_policy` to be provided to `validating_tester` instead of being hard-coded to `full`.
- Update wasm tests to use `validating_tester` with different `setup_policy`.

Resolves #259

### DIFF
--- a/libraries/testing/include/eosio/testing/tester.hpp
+++ b/libraries/testing/include/eosio/testing/tester.hpp
@@ -67,6 +67,8 @@ namespace eosio { namespace testing {
       full
    };
 
+   std::ostream& operator<<(std::ostream& os, setup_policy p);
+
    std::vector<uint8_t> read_wasm( const char* fn );
    std::vector<char>    read_abi( const char* fn );
    std::string          read_wast( const char* fn );
@@ -528,7 +530,7 @@ namespace eosio { namespace testing {
       }
       controller::config vcfg;
 
-      validating_tester(const flat_set<account_name>& trusted_producers = flat_set<account_name>(), deep_mind_handler* dmlog = nullptr) {
+      validating_tester(const flat_set<account_name>& trusted_producers = flat_set<account_name>(), deep_mind_handler* dmlog = nullptr, setup_policy p = setup_policy::full) {
          auto def_conf = default_config(tempdir);
 
          vcfg = def_conf.first;
@@ -538,7 +540,7 @@ namespace eosio { namespace testing {
          validating_node = create_validating_node(vcfg, def_conf.second, true, dmlog);
 
          init(def_conf.first, def_conf.second);
-         execute_setup_policy(setup_policy::full);
+         execute_setup_policy(p);
       }
 
       static void config_validator(controller::config& vcfg) {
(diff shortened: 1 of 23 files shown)


libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -67,6 +67,8 @@ namespace eosio { namespace testing {
       full
    };
 
+   std::ostream& operator<<(std::ostream& os, setup_policy p);
```
linh2931, line 70: Maybe you can add a comment here saying its purpose is for boost data.

libraries/testing/tester.cpp
```diff
@@ -20,6 +20,32 @@ eosio::chain::asset core_from_string(const std::string& s) {
 }
 
 namespace eosio { namespace testing {
+
+   std::ostream& operator<<(std::ostream& os, setup_policy p) {
```
linh2931, line 24: Does os need to be initialized?

libraries/testing/tester.cpp
```diff
@@ -20,6 +20,32 @@ eosio::chain::asset core_from_string(const std::string& s) {
...
+            break;
+         case setup_policy::preactivate_feature_and_new_bios:
+            os << "preactivate_feature_and_new_bios";
+            break;
+         case setup_policy::old_wasm_parser:
+            os << "old_wasm_parser";
+            break;
+         case setup_policy::full:
+            os << "full";
+            break;
```
linh2931, line 43: Nice to have a default here, though it should never happen.

libraries/testing/tester.cpp
heifner: I don't understand the question.

libraries/testing/include/eosio/testing/tester.hpp
heifner: Added comment.

libraries/testing/tester.cpp
heifner: You do get a `warning` if a new value is added, at least with `clang`:
```
tester.cpp:25:14: warning: enumeration value 'next' not handled in switch [-Wswitch]
      switch(p) {
```

Added a `default`.

greg7mdp:
Do we still use the regular `tester` anywhere?

heifner:
> Do we still use the regular `tester` anywhere?

There are many tests that still use it. I believe it is useful outside of validating_tester.

greg7mdp (approved):

libraries/testing/tester.cpp
linh2931: What I meant what happens if `os` has already contained some old value? Should it be set to empty before the first `<<`?

linh2931 (approved):

PR #1115, opened 2023-05-01
Reorganize tests for plugins

766C6164:
Currently we have different types of tests stored under a test folder for some plugins, e.g. chain_plugin and producer_plugin.

Some of these are proper unit tests of the plugin code in isolation. Those can remain in the folder but we should group them all together in a single executable rather than using different executable for each.

Some of these are bringing in other libraries and/or plugins and really should be considered integration tests (not to be confused with the Python end-to-end tests). These belong somewhere outside of the directories for the plugins. I suggest putting them in a sub-directory within the tests folder. It would also seem appropriate to bundle these tests together into one executable rather than linking separate executable for each test suite.

This PR reorganizes tests for the following plugins:
producer_plugin
chain_plugin
trace_api_plugin

Resolves https://github.com/AntelopeIO/leap/issues/586


greg7mdp (approved):
Nice!

spoonincode (approved):

Issue #1062, opened 2023-04-20
Update handling of http-max-response-time-ms & abi-serializer-max-time-ms


heifner:
Proposal https://hackmd.io/@EOSNF/HyFXmJL73

Relevant comments: 
https://github.com/AntelopeIO/leap/issues/677#issuecomment-1416137781
https://github.com/AntelopeIO/leap/issues/679#issue-1566543293
https://github.com/AntelopeIO/leap/issues/723#issuecomment-1440949853
https://github.com/AntelopeIO/leap/issues/679#issuecomment-1412486263
https://github.com/AntelopeIO/leap/issues/608#issuecomment-1444247439

PR #1125, opened 2023-05-03
Remove time_point implicit conversions

heifner:
Some cleanup while working #1062.
Remove implicit conversions of `time_point` and `time_point_sec`. They are error prone and were used in some "interesting" ways.
Removed arithmetic operators of `time_point_sec` as the results could be unexpected.

### DIFF
--- a/libraries/chain/include/eosio/chain/subjective_billing.hpp
+++ b/libraries/chain/include/eosio/chain/subjective_billing.hpp
@@ -8,6 +8,8 @@
 #include <eosio/chain/resource_limits_private.hpp>
 #include <eosio/chain/config.hpp>
 
+#include <fc/time.hpp>
+
 #include <boost/multi_index_container.hpp>
 #include <boost/multi_index/member.hpp>
 #include <boost/multi_index/hashed_index.hpp>
@@ -107,7 +109,7 @@ public:
    void disable_account( chain::account_name a ) { _disabled_accounts.emplace( a ); }
    bool is_account_disabled(const chain::account_name& a ) const { return _disabled || _disabled_accounts.count( a ); }
 
-   void subjective_bill( const chain::transaction_id_type& id, const fc::time_point& expire,
+   void subjective_bill( const chain::transaction_id_type& id, const fc::time_point_sec& expire,
                          const chain::account_name& first_auth, const fc::microseconds& elapsed )
    {
       if( !_disabled && !_disabled_accounts.count( first_auth ) ) {
@@ -116,7 +118,7 @@ public:
                trx_cache_entry{id,
                                first_auth,
                                bill,
-                               expire} );
+                               expire.to_time_point()} );
          if( p.second ) {
             _account_subjective_bill_cache[first_auth].pending_cpu_us += bill;
          }
--- a/libraries/chain/include/eosio/chain/subjective_billing.hpp
+++ b/libraries/chain/include/eosio/chain/subjective_billing.hpp
@@ -109,7 +109,7 @@ public:
    void disable_account( chain::account_name a ) { _disabled_accounts.emplace( a ); }
    bool is_account_disabled(const chain::account_name& a ) const { return _disabled || _disabled_accounts.count( a ); }
 
-   void subjective_bill( const chain::transaction_id_type& id, const fc::time_point_sec& expire,
+   void subjective_bill( const chain::transaction_id_type& id, fc::time_point_sec expire,
                          const chain::account_name& first_auth, const fc::microseconds& elapsed )
    {
       if( !_disabled && !_disabled_accounts.count( first_auth ) ) {
--- a/libraries/libfc/include/fc/time.hpp
+++ b/libraries/libfc/include/fc/time.hpp
@@ -46,7 +46,7 @@ namespace fc {
         static constexpr time_point maximum() { return time_point( microseconds::maximum() ); }
         static constexpr time_point min() { return time_point();                      }
 
-        std::string to_string()const;
+        std::string to_iso_string()const;
         static time_point from_iso_string( const std::string& s );
 
         constexpr const microseconds& time_since_epoch()const { return elapsed; }
--- a/libraries/libfc/include/fc/time.hpp
+++ b/libraries/libfc/include/fc/time.hpp
@@ -87,27 +87,12 @@ namespace fc {
         constexpr time_point to_time_point()const { return time_point( fc::seconds( utc_seconds) ); }
         constexpr uint32_t sec_since_epoch()const { return utc_seconds; }
 
-        constexpr time_point_sec& operator = ( const fc::time_point& t ) {
-          utc_seconds = t.time_since_epoch().count() / 1000000ll;
-          return *this;
-        }
         constexpr friend bool      operator < ( const time_point_sec& a, const time_point_sec& b )  { return a.utc_seconds < b.utc_seconds; }
         constexpr friend bool      operator > ( const time_point_sec& a, const time_point_sec& b )  { return a.utc_seconds > b.utc_seconds; }
         constexpr friend bool      operator <= ( const time_point_sec& a, const time_point_sec& b )  { return a.utc_seconds <= b.utc_seconds; }
         constexpr friend bool      operator >= ( const time_point_sec& a, const time_point_sec& b )  { return a.utc_seconds >= b.utc_seconds; }
         constexpr friend bool      operator == ( const time_point_sec& a, const time_point_sec& b ) { return a.utc_seconds == b.utc_seconds; }
         constexpr friend bool      operator != ( const time_point_sec& a, const time_point_sec& b ) { return a.utc_seconds != b.utc_seconds; }
-        constexpr time_point_sec&  operator += ( uint32_t m ) { utc_seconds+=m; return *this; }
-        constexpr time_point_sec&  operator += ( microseconds m ) { utc_seconds+=m.to_seconds(); return *this; }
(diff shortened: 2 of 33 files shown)


libraries/chain/include/eosio/chain/subjective_billing.hpp
```diff
@@ -107,7 +109,7 @@ class subjective_billing {
    void disable_account( chain::account_name a ) { _disabled_accounts.emplace( a ); }
    bool is_account_disabled(const chain::account_name& a ) const { return _disabled || _disabled_accounts.count( a ); }
 
-   void subjective_bill( const chain::transaction_id_type& id, const fc::time_point& expire,
+   void subjective_bill( const chain::transaction_id_type& id, const fc::time_point_sec& expire,
```
heifner, line 112: `trx_cache_entry` stores the trx expiration as a `time_point`. This could instead be a `time_point_sec` saving some space. I didn't make the change as it likely would not be a benefit as it currently has 64bit alignment. I'm open to changing it to a `time_point_sec` if that would make the code cleaner.

unittests/subjective_billing_tests.cpp
```diff
@@ -96,9 +98,9 @@ BOOST_AUTO_TEST_CASE( subjective_bill_test ) {
    { // expired handling logic, full billing until expiration then failed/decay logic
       subjective_billing sub_bill;
 
-      sub_bill.subjective_bill( id1, now, a, fc::microseconds( 1024 ) );
-      sub_bill.subjective_bill( id2, now + fc::microseconds(1), a, fc::microseconds( 1024 ) );
```
heifner, line 100: This test was a bit nonsensical as a trx expiration is always to a second as it is a `time_point_sec`. Updated to use `fc::second` as that is more valid.

libraries/chain/include/eosio/chain/subjective_billing.hpp
greg7mdp: It looks like getting a value for `expire` is more appropriate here than a `const&`. First it is usually constructed, and second `fc::time_point_sec` is only 8 bytes.

libraries/libfc/include/fc/time.hpp
```diff
@@ -78,17 +78,16 @@ namespace fc {
...
 
-        constexpr time_point_sec( const time_point& t )
+        constexpr explicit time_point_sec( const time_point& t )
         :utc_seconds( t.time_since_epoch().count() / 1000000ll ){}
 
         static constexpr time_point_sec maximum() { return time_point_sec(0xffffffff); }
         static constexpr time_point_sec min() { return time_point_sec(0); }
 
-        constexpr operator time_point()const { return time_point( fc::seconds( utc_seconds) ); }
+        constexpr time_point to_time_point()const { return time_point( fc::seconds( utc_seconds) ); }
```
greg7mdp, line 87: Actually I liked the cast operator better personally. This is more explicit, but whenever we switch a member from `time_point_sec` to `time_point` or vice versa it would require changes throughout. I feel that an implicit conversion between these two time point formats may be warranted.
However, this is just an opinion, not requesting that you change anything.

libraries/libfc/include/fc/time.hpp
```diff
@@ -105,15 +104,14 @@ namespace fc {
...
-        friend constexpr time_point   operator - ( const time_point_sec& t, const microseconds& m )   { return time_point(t) - m;             }
-        friend constexpr microseconds operator - ( const time_point_sec& t, const time_point_sec& m ) { return time_point(t) - time_point(m); }
-        friend constexpr microseconds operator - ( const time_point& t, const time_point_sec& m ) { return time_point(t) - time_point(m); }
+        friend constexpr time_point   operator + ( const time_point_sec& t, const microseconds& m )   { return time_point{fc::seconds(t.utc_seconds) + m}; }
+        friend constexpr time_point   operator - ( const time_point_sec& t, const microseconds& m )   { return time_point{fc::seconds(t.utc_seconds) - m}; }
+        friend constexpr microseconds operator - ( const time_point_sec& t, const time_point_sec& m ) { return fc::seconds(t.utc_seconds) - fc::seconds(m.utc_seconds); }
+        friend constexpr microseconds operator - ( const time_point& t, const time_point_sec& m ) { return t.time_since_epoch() - fc::seconds(m.sec_since_epoch()); }
```
greg7mdp, line 110: I kind of like this better. Again, personal preference.

```suggestion
        constexpr microseconds time_since_epoch()const { return fc::seconds(t.utc_seconds); }
        friend constexpr time_point   operator + ( const time_point_sec& t, const microseconds& m )   { return time_point{t.time_since_epoch() + m}; }
        friend constexpr time_point   operator - ( const time_point_sec& t, const microseconds& m )   { return time_point{t.time_since_epoch() - m}; }
        friend constexpr microseconds operator - ( const time_point_sec& t, const time_point_sec& m ) { return t.time_since_epoch() - m.time_since_epoch(); }
        friend constexpr microseconds operator - ( const time_point& t, const time_point_sec& m )     { return t.time_since_epoch() - m.time_since_epoch(); }
```

libraries/libfc/include/fc/time.hpp
```diff
@@ -46,7 +46,7 @@ namespace fc {
         static constexpr time_point maximum() { return time_point( microseconds::maximum() ); }
         static constexpr time_point min() { return time_point();                      }
 
-        operator std::string()const;
+        std::string to_string()const;
```
greg7mdp, line 49: I think the name should be the same between the two classes `to_string`, `to_iso_string`.

libraries/libfc/include/fc/time.hpp
heifner: In my opinion, having to make those changes should be required. As illustrated in some of these changes, it looks like the wrong type was used simply because the implicit conversion hid the fact that a conversion was happening.

libraries/libfc/include/fc/time.hpp
```diff
@@ -105,15 +104,14 @@ namespace fc {
         constexpr time_point_sec   operator +( uint32_t offset )const { return time_point_sec(utc_seconds + offset); }
         constexpr time_point_sec   operator -( uint32_t offset )const { return time_point_sec(utc_seconds - offset); }
 
-        friend constexpr time_point   operator + ( const time_point_sec& t, const microseconds& m )   { return time_point(t) + m;             }
-        friend constexpr time_point   operator - ( const time_point_sec& t, const microseconds& m )   { return time_point(t) - m;             }
-        friend constexpr microseconds operator - ( const time_point_sec& t, const time_point_sec& m ) { return time_point(t) - time_point(m); }
-        friend constexpr microseconds operator - ( const time_point& t, const time_point_sec& m ) { return time_point(t) - time_point(m); }
+        friend constexpr time_point   operator + ( const time_point_sec& t, const microseconds& m )   { return time_point{fc::seconds(t.utc_seconds) + m}; }
+        friend constexpr time_point   operator - ( const time_point_sec& t, const microseconds& m )   { return time_point{fc::seconds(t.utc_seconds) - m}; }
```
greg7mdp, line 108: This could easily create negative values. Since when converting from `time_point` to `time_point_sec`, we truncate down, any time_point less than 1 million us becomes 0 seconds, and if we sutract any positive number of us we become negative. I don't know if this could be a problem anywhere.

libraries/libfc/include/fc/time.hpp
greg7mdp: Yes, that is certainly a valid point. It would probably be better to use time_point everywhere.

libraries/libfc/include/fc/time.hpp
heifner: These seem rather dangerous actually. I think best to just remove them all.

libraries/chain/include/eosio/chain/subjective_billing.hpp
heifner: Done.

libraries/libfc/include/fc/time.hpp
heifner: Removed all these arithmetic operators as the results could be unexpected. Better to be explicit in my opinion.

libraries/libfc/include/fc/time.hpp
heifner: I held off originally as I was not sure if it was strictly ISO. However, there is an existing `from_iso_string`  which is the same format. I'll rename to `to_iso_string`.

libraries/libfc/include/fc/time.hpp
linh2931: It is better to make the conversion explicit to avoid unintentional conversions.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2229,12 +2229,13 @@ namespace eosio {
 
    void dispatch_manager::expire_txns() {
       size_t start_size = 0, end_size = 0;
+      fc::time_point_sec now{time_point::now()};
 
       std::unique_lock<std::mutex> g( local_txns_mtx );
       start_size = local_txns.size();
       auto& old = local_txns.get<by_expiry>();
       auto ex_lo = old.lower_bound( fc::time_point_sec( 0 ) );
```
linh2931, line 2237: Why was `fc::time_point_sec( 0 )` not moved outside of lock like `fc::time_point_sec now{time_point::now()};`?

heifner: The compiler will likely optimize that away. The `now()` call I thought time consuming enough to move outside.

linh2931 (approved):

greg7mdp (approved):

Issue #1088, opened 2023-04-26
Confusing read-only-read-window-time-us error message


heifner:
With the following config:
```
 max-transaction-time = 10000
 read-only-write-window-time-us = 200000
 read-only-read-window-time-us = 60000
 read-only-threads = 0
```
and when run node on v 4.0.0 - error is
```
plugin_config_exception: Incorrect plugin configuration
--read-only-read-window-time-us (60000) must be greater than --max-transaction-time 10000 ms plus a margin of 10000 us
    {"read":60000,"trx_time":10000,"min":10000}
```

Seems the error would be clearer if the values were all reported in microseconds. As multiple people have reported they don't understand the error. I think people miss the `ms` vs `us`. 

Maybe show the math in the error message. Something like:  
```
--read-only-read-window-time-us (60000 us) must be greater than --max-transaction-time (10000 ms) plus a margin of (10000 us), required: 60000 us < (10000000 us + 10000 us).
    {"read":60000,"trx_time":10000,"min":10000}
```

Also is it better experience to always enforce this or only when `read-only-threads > 0` ?

linh2931:
Another thing. I thought we only validated/initialized read-only trx configuration options when read-only was enabled (even having a comment `// only initialize other read-only options when read-only thread pool is          enabled`). But we missed that. This should be fixed at the same time.

PR #1129, opened 2023-05-04
[4.0] Check read-only trx options only when read-only-thread enabled, and present better error message

linh2931:
- Currently read-only trx options are checked even if `--read-only-thread` is 0. Changed to validate them only when `--read-only-thread` is greater than 0. 
- Current error message for invalid `-read-only-read-window-time-us` is confusing. Changed to use the same time unit and better error explanation. A sample looks like
```
warn  2023-05-04T00:46:28.397 thread-0  producer_plugin.cpp:1216  plugin_initialize  ] 3110006 plugin_config_exception: Incorrect plugin configuration   --read-only-read-window-time-us (20000 us) must be greater than --max-transaction-time  (10000 us) plus 10000 us, required: 20000 us > (10000 us + 10000 us).
```

Resolved https://github.com/AntelopeIO/leap/issues/1088

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -1126,6 +1126,7 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
       }
    }
    EOS_ASSERT( test_mode_ || my->_ro_thread_pool_size == 0 || my->_producers.empty(), plugin_config_exception, "--read-only-threads not allowed on producer node" );
+
    // only initialize other read-only options when read-only thread pool is enabled
    if ( my->_ro_thread_pool_size > 0 ) {
 #ifdef EOSIO_EOS_VM_OC_RUNTIME_ENABLED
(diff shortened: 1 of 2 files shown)


plugins/producer_plugin/test/test_read_only_trx.cpp
```diff
@@ -50,7 +50,8 @@ auto make_unique_trx( const chain_id_type& chain_id ) {
 
 BOOST_AUTO_TEST_SUITE(read_only_trxs)
 
-void error_handling_common(std::vector<const char*>& specific_args) {
+// app_init_status: true -- succeeded, false -- failed
+void test_configs_common(std::vector<const char*>& specific_args, bool app_init_status) {
```
greg7mdp, line 54: Instead of using `bool` which is not very explicit, I prefer using `enum class`, like:

```suggestion
enum class app_init_status { failed, succeeded };
void test_configs_common(std::vector<const char*>& specific_args, app_init_status status) {
```

plugins/producer_plugin/test/test_read_only_trx.cpp
```diff
@@ -59,19 +60,26 @@ void error_handling_common(std::vector<const char*>& specific_args) {
    std::vector<const char*> argv =
       {"test", "--data-dir", temp_dir_str.c_str(), "--config-dir", temp_dir_str.c_str()};
    argv.insert( argv.end(), specific_args.begin(), specific_args.end() );
-   BOOST_CHECK_EQUAL( app->initialize<producer_plugin>( argv.size(), (char**) &argv[0]), false );
+   BOOST_CHECK_EQUAL( app->initialize<producer_plugin>( argv.size(), (char**) &argv[0]), app_init_status );
```
greg7mdp, line 63:
```suggestion
   BOOST_CHECK_EQUAL( app->initialize<producer_plugin>( argv.size(), (char**) &argv[0]), status ==  app_init_status::succeeded);
```

plugins/producer_plugin/test/test_read_only_trx.cpp
```diff
@@ -59,19 +60,26 @@ void error_handling_common(std::vector<const char*>& specific_args) {
...
 }
 
 // --read-only-thread not allowed on producer node
 BOOST_AUTO_TEST_CASE(read_only_on_producer) {
    std::vector<const char*> specific_args = {"-p", "eosio", "-e", "--read-only-threads", "2" };
-   error_handling_common(specific_args);
+   test_configs_common(specific_args, false /* failed */);
```
greg7mdp, line 69:
```suggestion
   test_configs_common(specific_args, app_init_status::failed);
```

greg7mdp (approved):

linh2931:
> ```
> [TRX_TRACE] Speculative execution is REJECTING transient tx: b3c1a9a77f3bb0eef66305ae652aacef888bbae99208e63ce8f1208fd3ffdd88, auth:  : action: test:getage, transaction was executing for too long 6us reached node configured max-transaction-time 0us
> ```
Thanks @heifner for looking into the test failure. I walked away last night after running the unit tests. Should have waited for the integration tests. Fixed.

greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1159,26 +1160,31 @@ void producer_plugin::plugin_initialize(const boost::program_options::variables_
...
    } else {
-      // _max_transaction_time_ms can be set to negative in testing (for unlimited)
-      my->_ro_max_trx_time_us = my->_ro_read_window_effective_time_us;
+      // In integration tests, _max_transaction_time_ms can be set to negative
+      // for unlimited time
+     my->_ro_max_trx_time_us = fc::microseconds::maximum();
    }
-   ilog("ro_thread_pool_size ${s}, ro_write_window_time_us ${ww}, ro_read_window_time_us ${rw}, ro_max_trx_time_us ${t}, ro_read_window_effective_time_us ${w}",
-        ("s", my->_ro_thread_pool_size)("ww", my->_ro_write_window_time_us)("rw", my->_ro_read_window_time_us)("t", my->_ro_max_trx_time_us)("w", my->_ro_read_window_effective_time_us));
+   ilog("ro_thread_pool_size ${s}, ro_max_trx_time_us ${t}}", ("s", my->_ro_thread_pool_size)("t", my->_ro_max_trx_time_us));
```
heifner, line 1187: Seems like for this `ilog` and the one above, we should log from the user perspective instead of the implementation details. Report the values as a user would understand from the command-line/config.ini values.

heifner (approved):

PR #1132, opened 2023-05-04
check for existance of BMI2 & ADX instructions on CI workers

spoonincode:
As we start to add some hand rolled assembly that can detect and utilize optional CPU instruction sets at runtime (#1071), maybe it would be worthwhile checking that the CI runners are able to run all potential implementation combinations.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -141,7 +141,7 @@ jobs:
           cd build
           ctest --output-on-failure -j $(nproc) -LE "(nonparallelizable_tests|long_running_tests)" --timeout 420
       - name: Check CPU Features
-        run: awk 'BEGIN {err = 1} /bmi2/ && /adx/ {err = 0} END {exit err}' /proc/cpuinfo
+        run: awk '/bmi2f/ && /adx/ {exit 0} ENDFILE {exit 1}' /proc/cpuinfo
 
   np-tests:
     name: NP Tests
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -141,7 +141,7 @@ jobs:
           cd build
           ctest --output-on-failure -j $(nproc) -LE "(nonparallelizable_tests|long_running_tests)" --timeout 420
       - name: Check CPU Features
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -140,6 +140,8 @@ jobs:
           zstdcat build.tar.zst | tar x
           cd build
           ctest --output-on-failure -j $(nproc) -LE "(nonparallelizable_tests|long_running_tests)" --timeout 420
+      - name: Check CPU Features
+        run: awk 'BEGIN {err = 1} /bmi2/ && /adx/ {err = 0} END {exit err}' /proc/cpuinfo
```
linh2931, line 144: Run the `awk` on my Ubuntu 20 and 22 boxes; the command returned 0 as expected. Not sure how this check is used in CI. Is it run before or after the unittest?

spoonincode: In this case it's run after the tests, but before or after doesn't really matter: if the CPU the workflow job (which also includes where the tests were run, since it's the same job) are running on doesn't support BMI2 & ADX, it will return a non-zero exit code and thus will fail the job -- CI will fail. If you try changing `/bmi2/` to `/banana/` you can see it returns 1.

For BMI2 & ADX this is probably quite paranoid: github (indirectly) documents their runner CPUs as having these features, and I'm not sure any major cloud provider (AWS, GCP, Azure, etc) even have instances without these CPU features at this point either. But I'm trying to find a pattern we can use as we get more CPU feature-specific code paths implemented and verify those code paths are actually run in CI.

linh2931: Thanks for the detailed explanation!

linh2931 (approved):

heifner (approved):

.github/workflows/build.yaml
spoonincode: (an alternative might be that instead of `BLS_DISABLE_BMI2` we make that in to a tri-state of OFF, AUTO, ON, where AUTO is default, and ON _will error out of BMI2 is not available_. I wasn't really sure I wanted to go that route, for now)

Issue #965, opened 2023-04-04
Nodeos Endpoint IP and Port Configuration


bhazzard:
**Short Description:**
As a node operator, I want to be able to set the IP and port for each endpoint provided by nodeos independently, so that I can optimize my node’s network configuration for my specific needs.

**Longer Description:**
_What_
I want to be able to set the IP and port for each endpoint provided by nodeos independently, including the HTTP endpoints (such as chain api, producer api, net api, db size api, trace api, contracts api, prometheus exporter) and the P2P endpoint.

_Why_
This will allow me to configure my node to use specific network interfaces or ports for each endpoint, which can be particularly useful when running multiple nodes on the same machine or in a network with strict firewall rules.

_How_
- Support ipv4 & ipv6: https://github.com/AntelopeIO/leap/issues/622
- Ability to set ip/port configuration in config file
- Ability to set ip/port configuration in environment variables
- Ability to set ip/port configuration in nodeos cli options
- Handling for missing, invalid, or conflicting configuration
- Documentation in CLI help

heifner:
https://github.com/eosnetworkfoundation/product/blob/huangminghuang/leap-is-965/api-http/proposals/http-plugin.md

heifner:
https://github.com/eosnetworkfoundation/product/pull/143

PR #1137, opened 2023-05-05
Support multiple http endpoints

huangminghuang:
This PR implements the proposal https://github.com/eosnetworkfoundation/product/blob/main/api-http/proposals/http-plugin.md to support multiple endpoints for http plugins.

APIs are categorized into the following groups:
### chain_ro
  - v1/chain/get_activated_protocol_features
  - v1/chain/get_block_info
  - v1/chain/get_block_header_state
  - v1/chain/get_account
  - v1/chain/get_code
  - v1/chain/get_code_hash
  - v1/chain/get_consensus_parameters
  - v1/chain/get_abi
  - v1/chain/get_raw_code_and_abi
  - v1/chain/get_raw_abi
  - v1/chain/get_table_rows
  - v1/chain/get_table_by_scope
  - v1/chain/get_currency_balance
  - v1/chain/get_currency_stats
  - v1/chain/get_producers
  - v1/chain/get_producer_schedule
  - v1/chain/get_scheduled_transactions
  - v1/chain/get_required_keys
  - v1/chain/get_transaction_id
  - v1/chain/send_read_only_transaction
  - v1/chain/get_accounts_by_authorizers
  - v1/chain/get_raw_block
  - v1/chain/get_block
  - v1/chain/get_block_header
  - v1/chain/get_transaction_status

### chain_rw
  - v1/chain/compute_transaction
  - v1/chain/push_transaction
  - v1/chain/push_transactions
  - v1/chain/send_transaction
  - v1/chain/send_transaction2

### db_size
  - v1/db_size/get
### net_ro
  - v1/net/status
  - v1/net/connections

### net_rw
  - v1/net/connect
  - v1/net/disconnect

### producer_ro
  - v1/producer/paused
  - v1/producer/get_runtime_options
  - v1/producer/get_greylist
  - v1/producer/get_whitelist_blacklist
  - v1/producer/get_scheduled_protocol_feature_activations
  - v1/producer/get_supported_protocol_features
  - v1/producer/get_account_ram_corrections
  - v1/producer/get_unapplied_transactions

### producer_rw
  - v1/producer/pause
  - v1/producer/resume
  - v1/producer/update_runtime_options
  - v1/producer/add_greylist_accounts
  - v1/producer/remove_greylist_accounts
  - v1/producer/set_whitelist_blacklist
  - v1/producer/schedule_protocol_feature_activations
  - v1/producer/get_integrity_hash

### snapshot
  - v1/producer/get_snapshot_requests
  - v1/producer/create_snapshot
  - v1/producer/schedule_snapshot
  - v1/producer/unschedule_snapshot

### trace_api
  - v1/trace_api/get_block
  - v1/trace_api/get_transaction_trace

### prometheus
  - v1/prometheus/metrics

### node
  - v1/node/get_supported_apis
  - v1/chain/get_info

## User configuration knob
A single `http-category-address` can be used to configure all addresses in command line and ini file. The option can be used multiple times as needed.

```config.ini
http-server-address   = http-category-address
http-category-address = chain_ro,127.0.0.1:8081
http-category-address = chain_ro,:9090
http-category-address = chain_rw,[::1]:8082
http-category-address = net_ro,127.0.0.1:8081
http-category-address = net_rw,[::]:8083 
http-category-address = producer_ro,/tmp/absolute_unix_path.sock
http-category-address = producer_rw,./relative/unix_path.sock
http-category-address = snapshot,./relative/unix_path.sock
```

A single `hostname:port` specification can be used by multiple categories, as of the case for `127.0.0.1:8081` in above example. However, two specifications having the same port with different hostname strings are always considered as configuration error regardless whether they can be resolved into the same set of IP addresses. For example, the following configuration are invalid.

```
http-category-address = chain_ro,127.0.0.1:8081
http-category-address = chain_rw,localhost:8081 # error, 127.0.0.1 and localhost are different
```

For the node category, it is NOT user configurable because it will be available for all listened http addresses.

For backward compatibility, the HTTP category facility has to be opted in by specifying `http-server-address = http-category-address` explicitly. 
When `http-server-address = http-category-address`, all the unspecified categories are considered disabled.  In addition, `http-server-address = http-category-address`
connot be specified together with a non-empty `unix-socket-path`. 

All existing configuration options in the http plugin apply to all handlers for all addresses; this includes the `http-threads` option. This means all handlers share the same thread pool.

# server address format
* server address can be in one of the following formats
  - ipv4_address:port like 127.0.0.1:8080
  - ipv6_address:port like [2001:db8:3c4d:15::1a2f:1a2b]:8080
  - hostname:port like my.domain.com:8080
  - unix socket path (must starts with '/' or './')

* In the case of the hostname, the IP address is resolved by ARP, *ALL* resolved addresses will be listened to.
* To listen to all interfaces for both IPv4 and IPv6, just use `:8080`. Notice that the behavior of `[::]:8080` is  dependent on whether the IPv4-mapped IPv6 is enabled in the system.

Resolves [IS #965] (https://github.com/AntelopeIO/leap/issues/965)


plugins/http_plugin/http_plugin.cpp
```diff
@@ -243,36 +371,48 @@ namespace eosio {
...
+               auto category_name = spec.substr(0, comma_pos);
+               auto category = to_category(category_name);
+               auto address = spec.substr(comma_pos+1);
+               EOS_ASSERT(category != api_category::unknown, chain::plugin_config_exception, 
+                  "invalid category name `${name}` for http_category_address", ("name", std::string(category_name)));
+               auto [host, port] = split_host_port(address);
+               if (port.size()) {
+                  auto [itr, inserted] = hostnames.try_emplace(port, host);
+                  EOS_ASSERT(inserted || host == itr->second, chain::plugin_config_exception, "unable to listen to port ${port} for both ${host} and ${prev}",
+                  ("port", std::string(port))("host", std::string(host))("prev", std::string(itr->second)));
```
heifner, line 412: These do not need to be wrapped in `string()`. Also fix the indentation.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -243,36 +371,48 @@ namespace eosio {
...
+                chain::plugin_config_exception,
+                "when http-category-address is specified, http-server-address must be set as "
+                "`http-category-address` and `unix-socket-path` must be left unspecified");
+
+            std::map<std::string, std::string> hostnames;
+            auto addresses = options["http-category-address"].as<vector<string>>();
+            for (const auto& spec : addresses) {
+               auto comma_pos = spec.find(',');
+               EOS_ASSERT(comma_pos > 0 && comma_pos != std::string_view::npos, chain::plugin_config_exception,
+               "http-category-address '${spec}' does not contain a required comma to separate the category and address", ("spec", spec));
```
heifner, line 402: Fix indentation.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -362,32 +461,26 @@ namespace eosio {
...
       fc_ilog( logger(), "exit shutdown");
    }
 
-   void http_plugin::add_handler(const string& url, const url_handler& handler, appbase::exec_queue q, int priority, http_content_type content_type) {
-      fc_ilog( logger(), "add api url: ${c}", ("c", url) );
-      auto p = my->plugin_state->url_handlers.emplace(url, my->make_app_thread_url_handler(url, q, priority, handler, my, content_type));
-      EOS_ASSERT( p.second, chain::plugin_config_exception, "http url ${u} is not unique", ("u", url) );
+   void http_plugin::add_handler(api_entry&& entry, appbase::exec_queue q, int priority, http_content_type content_type) {
+      std::string path = entry.path;
+      fc_ilog( logger(), "add api url: ${c}", ("c", path) );
```
heifner, line 474: Would be nice to log the category and address and port (or unix socket) for each one on startup. Doesn't need to be here but seems like we should summarize that all in one log statement.

```
info  2023-05-09T20:14:32.976 nodeos    http_plugin.cpp:474           add_handler          ] add api url: /v1/chain/get_info
info  2023-05-09T20:14:32.976 nodeos    http_plugin.cpp:474           add_handler          ] add api url: /v1/chain/get_activated_protocol_features
info  2023-05-09T20:14:32.976 nodeos    http_plugin.cpp:474           add_handler          ] add api url: /v1/chain/get_block
info  2023-05-09T20:14:32.976 nodeos    http_plugin.cpp:474           add_handler          ] add api url: /v1/chain/get_block_info
```

plugins/http_plugin/include/eosio/http_plugin/http_plugin.hpp
```diff
@@ -88,25 +94,24 @@ namespace eosio {
...
         // standard exception handling for api handlers
         static void handle_exception( const char *api_name, const char *call_name, const string& body, const url_response_callback& cb );
 
         void post_http_thread_pool(std::function<void()> f);
 
-        bool is_on_loopback() const;
-        bool is_secure() const;
+        bool is_on_loopback(api_category category) const;
 
         static bool verbose_errors();
```
heifner, line 117: `get_supported_apis_result get_supported_apis()const;` below is not implemented.

plugins/http_plugin/include/eosio/http_plugin/api_category.hpp
```diff
@@ -0,0 +1,36 @@
...
+};
+
+
+class api_category_set {
+   uint32_t data = {};
+public:
+   constexpr api_category_set() = default;
+   constexpr explicit api_category_set(api_category c) : data(static_cast<uint32_t>(c)){}
+   constexpr bool contains(api_category category) const { return (data & static_cast<uint32_t>(category)) != 0; }
+   constexpr void insert(api_category category) { data |= static_cast<uint32_t>(category); }
```
heifner, line 29: I would recommend using `has_field` and `set_field` of `chain/types.hpp`

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -132,14 +130,16 @@ class beast_http_session : public detail::abstract_conn {
 
       // Request path must be absolute and not contain "..".
       if(req.target().empty() || req.target()[0] != '/' || req.target().find("..") != beast::string_view::npos) {
+         fc_ilog( plugin_state_->logger, "Return bad_reqest:  ${target}",  ("target", std::string(req.target())) );
```
heifner, line 133: If this is kept it should be `fc_dlog`, but does it add anything to the `fc_dlog` in `send_response`?

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -132,14 +130,16 @@ class beast_http_session : public detail::abstract_conn {
...
          error_results results{static_cast<uint16_t>(http::status::bad_request), "Illegal request-target"};
          send_response( fc::json::to_string( results, fc::time_point::maximum() ),
                         static_cast<unsigned int>(http::status::bad_request) );
          return;
       }
 
       try {
-         if(!derived().allow_host(req)) {
+         if(!allow_host(req)) {
+            fc_ilog( plugin_state_->logger, "bad host:  ${HOST}", ("HOST", std::string(req["host"])));
```
heifner, line 142: If this is kept it should be `fc_dlog`, but does it add anything to the `fc_dlog` in `send_response`?

plugins/chain_api_plugin/chain_api_plugin.cpp
```diff
@@ -41,8 +41,9 @@ parse_params<chain_apis::read_only::get_transaction_status_params, http_params_t
    }
 }
 
-#define CALL_WITH_400(api_name, api_handle, api_namespace, call_name, http_response_code, params_type) \
+#define CALL_WITH_400(api_name, category, api_handle, api_namespace, call_name, http_response_code, params_type) \
```
heifner, line 44: Seems like it should be error to configure a `http-category-address` category for a plugin that is not enabled. For example: `--http-category-address "net_ro,./x"` should be an error if `eosio::net_api_plugin` not enabled.

plugins/http_plugin/include/eosio/http_plugin/common.hpp
```diff
@@ -136,6 +137,17 @@ struct http_plugin_state {
 
    explicit http_plugin_state(fc::logger& log)
        : logger(log) {}
+
+   http_plugin::get_supported_apis_result get_supported_apis(api_category_set set) const {
```
heifner, line 141: Seems like this should be implemented in http_plugin.cpp and not here.

tests/CMakeLists.txt
```diff
@@ -241,6 +241,10 @@ add_test(NAME plugin_http_api_test COMMAND tests/plugin_http_api_test.py WORKING
 set_tests_properties(plugin_http_api_test PROPERTIES TIMEOUT 50)
 set_property(TEST plugin_http_api_test PROPERTY LABELS nonparallelizable_tests)
 
+add_test(NAME plugin_http_category_api_test COMMAND tests/plugin_http_api_test.py WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+set_tests_properties(plugin_http_category_api_test PROPERTIES TIMEOUT 50 ENVIRONMENT "PLUGIN_HTTP_TEST_CATEGORY=ON")
+set_property(TEST plugin_http_category_api_test PROPERTY LABELS nonparallelizable_tests)
```
heifner, line 246: How is different then the existing `plugin_http_api_test`, no new args are passed to the test.

heifner (changes requested):

heifner:
Tests failing because keosd not launching:
```
info  2023-05-10T21:38:51.868 keosd     wallet_plugin.cpp:32          plugin_initialize    ] initializing wallet plugin
warn  2023-05-10T21:38:51.869 keosd     http_plugin.cpp:483           plugin_initialize    ] 13 N5boost10wrapexceptINS_12bad_any_castEEE: boost::bad_any_cast: failed conversion using boost::any_cast
rethrow boost::bad_any_cast: failed conversion using boost::any_cast: 
    {"what":"boost::bad_any_cast: failed conversion using boost::any_cast"}
    keosd  http_plugin.cpp:483 plugin_initialize

appbase: exception thrown during plugin "eosio::wallet_api_plugin" initialization.
boost::bad_any_cast: failed conversion using boost::any_cast
```

plugins/http_plugin/http_plugin.cpp
```diff
@@ -59,6 +58,28 @@ namespace eosio {
...
+      if (api_category_set({api_category::chain_ro, api_category::chain_rw}).contains(category))
+         return "eosio::chain_api_plugin";
+      if (api_category_set({api_category::net_ro, api_category::net_rw}).contains(category))
+         return "eosio::net_api_plugin";
+      if (api_category_set({api_category::producer_ro, api_category::producer_rw, api_category::snapshot})
+              .contains(category))
+         return "eosio::producer_api_plugin";
+      // It's a programming error when the control flow reaches this point, 
+      // please make sure all the plugin names are returned from above statements.
+      throw std::runtime_error("No correspding plugin for the category value");
```
heifner, line 80: Seems like this should be `std::logic_error` or `FC_ASSERT` or `EOS_ASSERT` or even `assert()`

greg7mdp: I like `assert()` for things that logically should never happen. No point in adding code in release build for that, and a `throw` can be caught, leaving the programmer unaware that his logic is wrong.

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -217,50 +223,97 @@ void run_test(Protocol& p, size_t max_body_size)
...
+         std::condition_variable& started = plugin->startup_condition();
+         app_thread = std::thread([&]() {
+            try {
+               app->startup();
+               app->exec();
+            } catch (...) {
+               plugin = nullptr;
+               started.notify_all();
+            }
+         });
```
greg7mdp, line 277: This still is not guarantee to work. This `app_thread` could very well execute and do the `notify_all()` **before** the test thread calls `started.wait(lk);`.
I still don't understand why you switched from the promise/future that was used before.

tests/CMakeLists.txt
heifner: I see this passes an ENVIRONMENT variable. Can we change that to an argument to the python test.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -31,6 +32,7 @@ namespace eosio {
 
    static http_plugin_defaults current_http_plugin_defaults;
    static bool verbose_http_errors = false;
+   static std::promise<http_plugin*>* plugin_promise;
```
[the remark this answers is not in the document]
huangminghuang: The problem was the sockets are listened with `app().executor().post()`, which means you cannot guarantee they are listened after `http_plugin::plugin_startup()`.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -123,22 +163,132 @@ namespace eosio {
...
+                  // Therefore, we create the unix socket with the relative path to its parent path to avoid the problem.
+                  std::filesystem::current_path(sock_path.parent_path());
+                  auto restore = fc::make_scoped_exit([cwd]{
+                     std::filesystem::current_path(cwd);
+                  });
+
+                  using stream_protocol = asio::local::stream_protocol;
+                  auto server = std::make_shared<beast_http_listener<stream_protocol::socket>>(
+                        plugin_state, categories, stream_protocol::endpoint{ sock_path.filename().string() });
+                  server->do_accept();
```
greg7mdp, line 210: In the previous version of http_plugin, when we created `beast_http_listener`, we stored the shared_ptr to these, so we could call `stop_listening` and delete the server in `plugin_shutdown`. Why don't we do that anymore?

plugins/http_plugin/http_plugin.cpp
```diff
@@ -123,22 +163,132 @@ namespace eosio {
              };
             return handler;
          }
+         
+         bool is_unix_socket_address(const std::string& address) const {
+            return address[0] == '/' ||
+                   (address[0] == '.' &&
+                    (address[1] == '/' || (address.size() >= 3 && address[1] == '.' && address[2] == '/')));
+         }
```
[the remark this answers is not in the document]
greg7mdp: I like you new version! Nice!

plugins/http_plugin/tests/unit_tests.cpp
```diff
@@ -217,50 +223,98 @@ void run_test(Protocol& p, size_t max_body_size)
...
+               plugin = nullptr;
+               listening_or_failed.store(true);
+            }
+         });
+
+         while (!listening_or_failed.load()) {
+            using namespace std::chrono_literals;
+            std::this_thread::sleep_for(10ms);
+         }
+         return plugin;
```
greg7mdp, line 283: This should work. Maybe appbase infrastructure should hold a `std::promise` for each plugin, which would be provided a value after the plugin successfully completes its startup.

tests/plugin_http_api_test.py
```diff
@@ -58,6 +90,9 @@ def startEnv(self) :
         nodeos_flags = (" --data-dir=%s --config-dir=%s --trace-dir=%s --trace-no-abis --access-control-allow-origin=%s "
                         "--contracts-console --http-validate-host=%s --verbose-http-errors --max-transaction-time -1 --abi-serializer-max-time-ms 30000 --http-max-response-time-ms 30000 "
                         "--p2p-peer-address localhost:9011 --resource-monitor-not-shutdown-on-threshold-exceeded ") % (self.data_dir, self.config_dir, self.data_dir, "\'*\'", "false")
+        nodeos_flags += category_config.nodeosArgs()
+
+        print("nodeos_flags = ", nodeos_flags)
```
jgiszczak, line 95: Remove debug print or convert to Utils.Print.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -451,29 +636,21 @@ namespace eosio {
...
-      return (!my->listen_endpoint || my->listen_endpoint->address().is_loopback());
-   }
-
-   bool http_plugin::is_secure() const {
-      return (!my->listen_endpoint || my->listen_endpoint->address().is_loopback());
+   bool http_plugin::is_on_loopback(api_category category) const {
+      if (my->http_server_address != "http-category-address")
+         return my->http_server_address.empty() || my->on_loopback_only(my->http_server_address);
+      return my->categories_by_address.empty() ||
+             std::all_of(my->categories_by_address.begin(), my->categories_by_address.end(),
```
greg7mdp, line 643:
```suggestion
      return std::all_of(my->categories_by_address.begin(), my->categories_by_address.end(),
```

plugins/http_plugin/http_plugin.cpp
```diff
@@ -123,22 +179,135 @@ namespace eosio {
...
+                  tcp::resolver resolver( plugin_state->thread_pool.get_executor());
+                  auto endpoints = resolver.resolve(host, port, boost::asio::ip::tcp::resolver::passive, ec);
+                  EOS_ASSERT(!ec, chain::plugin_config_exception, "failed to resolve address: ${msg}",
+                             ("msg", ec.message()));
+
+                  int listened = 0;
+                  std::optional<boost::asio::ip::tcp::endpoint> unspecified_ipv4_addr;
+                  bool has_unspecified_ipv6_only = false;
+
+                  auto create_ip_server = [&](auto endpoint) {
```
greg7mdp, line 242:
```suggestion
                  auto create_ip_server = [&](const auto& endpoint) {
```

greg7mdp (changes requested):

greg7mdp:
I'm not understanding where we do the `open()/bind()/listen()` on the acceptor in this new version?

plugins/http_plugin/http_plugin.cpp
```diff
@@ -123,22 +163,132 @@ namespace eosio {
...
-            plugin_state->valid_hosts.emplace(host + ":" + port);
-            plugin_state->valid_hosts.emplace(host + ":" + resolved_port_str);
+         bool on_loopback_only(const std::string& address) {
+            if (is_unix_socket_address(address))
+               return true;
+            auto [host, port] = split_host_port(address);
+            boost::system::error_code ec;
+            tcp::resolver             resolver(plugin_state->thread_pool.get_executor());
+            auto endpoints = resolver.resolve(host, port, boost::asio::ip::tcp::resolver::passive, ec);
+            EOS_ASSERT(!ec, chain::plugin_config_exception, "failed to resolve address: ${msg}", ("msg", ec.message()));
```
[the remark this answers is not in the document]
heifner: Good point, should likely not throw from here.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -123,22 +179,135 @@ namespace eosio {
...
+                        continue;
+                     }
+                     create_ip_server(endpoint);
+                  }
+
+                  if (unspecified_ipv4_addr.has_value() && has_unspecified_ipv6_only) {
+                     create_ip_server(*unspecified_ipv4_addr);
+                  }
+
+                  EOS_ASSERT (listened > 0, chain::plugin_config_exception, "none of the resolved address can be listened" );
```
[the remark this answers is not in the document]
greg7mdp: You added the `to` at the end, but did not fix `addresses`

plugins/http_plugin/http_plugin.cpp
```diff
@@ -123,22 +176,129 @@ namespace eosio {
...
+                        continue;
+                     }
+                     create_ip_server(endpoint);
+                  }
+
+                  if (unspecified_ipv4_addr.has_value() && has_unspecified_ipv6_only) {
+                     create_ip_server(*unspecified_ipv4_addr);
+                  }
+
+                  EOS_ASSERT (listened > 0, chain::plugin_config_exception, "none of the resolved address can be listened to" );
```
greg7mdp, line 278:
```suggestion
                  EOS_ASSERT (listened > 0, chain::plugin_config_exception, "none of the resolved addresses can be listened to" );
```

greg7mdp (changes requested):

tests/plugin_http_api_test.py
huangminghuang: removed

huangminghuang: > I'm not understanding where we do the `open()/bind()/listen()` on the acceptor in this new version?
I guess this is what you don't get.
https://www.boost.org/doc/libs/1_82_0/doc/html/boost_asio/reference/basic_socket_acceptor/basic_socket_acceptor/overload5.html

plugins/http_plugin/include/eosio/http_plugin/common.hpp
```diff
@@ -184,30 +186,32 @@ auto make_http_response_handler(std::shared_ptr<http_plugin_state> plugin_state,
...
-bool host_port_is_valid(const http_plugin_state& plugin_state,
-                        const std::string& header_host_port,
-                        const string& endpoint_local_host_port) {
-   return !plugin_state.validate_host || header_host_port == endpoint_local_host_port || plugin_state.valid_hosts.find(header_host_port) != plugin_state.valid_hosts.end();
+inline std::pair<std::string, std::string> split_host_port(std::string_view endpoint) {
+   std::string::size_type colon_pos = endpoint.rfind(':');
+   if(colon_pos != std::string::npos) {
+      auto port = endpoint.substr(colon_pos + 1);
+      auto hostname = (endpoint[0] == '[' && colon_pos >= 2)  ? endpoint.substr( 1, colon_pos-2 ) : endpoint.substr( 0, colon_pos );
+      return {std::string(hostname), std::string(port)};
```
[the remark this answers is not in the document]
greg7mdp: My bad, sorry, you are correct.

plugins/http_plugin/http_plugin.cpp
```diff
@@ -123,22 +179,135 @@ namespace eosio {
...
+                  std::filesystem::current_path(sock_path.parent_path());
+                  auto restore = fc::make_scoped_exit([cwd]{
+                     std::filesystem::current_path(cwd);
+                  });
+
+                  using stream_protocol = asio::local::stream_protocol;
+                  auto server = std::make_shared<beast_http_listener<stream_protocol::socket>>(
+                        plugin_state, categories, stream_protocol::endpoint{ sock_path.filename().string() });
+                  server->do_accept();
+                  fc_ilog(logger(), "created beast UNIX socket listener at ${addr}", ("addr", sock_path));
```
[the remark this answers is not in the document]
greg7mdp: I still think it would be helpful, as it would confirm that the expected categories are listened to at the expected addresses. 
Otherwise this log entry is not very useful.

tests/plugin_http_api_test.py
greg7mdp: > > I'm not understanding where we do the `open()/bind()/listen()` on the acceptor in this new version?
> > I guess this is what you don't get.
> > https://www.boost.org/doc/libs/1_82_0/doc/html/boost_asio/reference/basic_socket_acceptor/basic_socket_acceptor/overload5.html

Yes, that's it, thanks. This doesn't appear to do the error checking and throws we used to do. Could it be an issue?

plugins/http_plugin/http_plugin.cpp
greg7mdp: @heifner what do you think about this. I still think that explicitly shutting down the http servers in `plugin_shutdown` is cleaner, but if you believe it is equivalent I am fine to go with the simplest solution.

tests/plugin_http_api_test.py
huangminghuang: All asio apis throw std::system_error when there's no ec input parameter.

greg7mdp: Yes, but before we called:
```
void fail(beast::error_code ec, char const* what, fc::logger& logger, char const* action) {
   fc_elog(logger, "${w}: ${m}", ("w", what)("m", ec.message()));
   fc_elog(logger, action);
}
```
which probably logged a more descriptive message describing the issue.

plugins/http_plugin/http_plugin.cpp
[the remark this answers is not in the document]
greg7mdp: Yeah, thank you @huangminghuang !

plugins/http_plugin/http_plugin.cpp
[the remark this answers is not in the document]
heifner: This looks like it is only called from `is_on_loopback` which is only currently called from `plugin_initialize`. However It seems like it would be better for this to only log a warning and return false. Seems like loopback should always be resolvable so unable to resolve likely means not on loopback.

tests/plugin_http_api_test.py
heifner: In general allowing the framework to do more of the work seems like a win. We do need to be careful about exceptions, see https://github.com/AntelopeIO/leap/pull/1175 for example. However, this is only called during startup, so seems safe to allow it to throw and report the error.

However, we do need `http_plugin::plugin_startup` modified as part of this PR to log better error messages than `http_plugin startup fails, shutting down`.

plugins/http_plugin/http_plugin.cpp
heifner: As is I believe is the pattern we want to move to. Just allow it to naturally shutdown.

heifner (approved):

greg7mdp (approved):

Issue #1025, opened 2023-04-13
Improve packed_transaction parsing performance


heifner:
See https://github.com/AntelopeIO/leap/pull/999#discussion_r1165459091
Simple speedup is to create a caching resolver that caches abis as it goes instead of repeatedly retrieving them.

greg7mdp:
@heifner in which context can we optimize this. Is it where we parse a single packed_transaction, or do we sometimes parse a bunch of them?

heifner:
> @heifner in which context can we optimize this. Is it where we parse a single packed_transaction, or do we sometimes parse a bunch of them?

The interface for multiple trxs is rarely if ever used. I was thinking of the single packed_transaction use.

heifner:
> Is it likely that we would retrieve the same abi repeatedly for a single transaction?

Yes, mining trxs often have many of the same input actions.
For example: https://bloks.io/transaction/c65166655b17ca691c8c46f3baf7c40ef90dde5cda16c2a3cb00a6bc93ca0ce7

greg7mdp:
Thanks. All clear now!

greg7mdp:
@heifner , I'm not seeing where it tries to repeatedly retrieve abis when processing packed transactions?

heifner:
> @heifner , I'm not seeing where it tries to repeatedly retrieve abis when processing packed transactions?

https://github.com/AntelopeIO/leap/blob/main/plugins/chain_plugin/chain_plugin.cpp#L2189
The resolver created and used there does a find and creates a new `abi_serializer` for each call to the resolver. It does not cache the `abi_serializer`.

PR #1139, opened 2023-05-05
Update API/serialization time constraints

### LINKED ISSUE #1062: Update handling of http-max-response-time-ms & abi-serializer-max-time-ms

heifner:
# Update of API/serialization time constraints

Requirements: 
- an API node should be able to serve any `atomic` request (such as `get_block`, `send_transaction*` and `push_transaction*`) successfully
- for non `atomic` API calls, such as `get_table_rows`, `get_table_by_scope`, `get_scheduled_transactions` and `get_producers` (where the user is free to request a smaller range if needed), the node operator should be able to specify a max request time.
- some specific (user-provided) ABIs can be crafted to take an excessively long time (abi-bombs), and we must guard against that.

## ABI serialization limit

The ABI deserialization for most ABI intensive requests (including `get_block`, `send_transaction*`, `push_transaction*`, `get_account`, `get_table_rows`) was moved to the HTTP thread pool, so ABI deserialization no longer can affect the responsiveness of nodeos's main thread. 

Note that ABI serialization, when processing the API parameters (for example for `send_transaction2`), still occurs on the main thread and is limited by `abi-serializer-max-time-ms`. In case of serialization failure, the API will fail and return an error.

For the deserialization now occuring on the http thread pool, any single deserialization call will be limited by `abi-serializer-max-time-ms` as well. If that limit is hit for a specific object, the request will not fail, instead that object will be returned in the binary, non-deserialized form. 

So when processing a request such as `get_block`, the request will always succeed (no timeout), but some specific objects may be returned in non-deserialized form. The max time for processing a `get_block` request will only be limited by the maximum number of actions in a block and `abi-serializer-max-time-ms`. 

However because the deserialization is now done off the main thread, the absence of a lower limit should not impact the responsiveness of the node when serving legitimate requests.

## Constraints on the number of items returned for non-atomic APIs

`http_max_response_time` now applies only to requests which specify a range of objects to be retrieved ( `get_table_rows`, `get_table_by_scope`, `get_scheduled_transactions` and `get_producers`), and specifies the maximum time spent on the main thread. The default of `http-max-response-time-ms` has been changed from 30ms to 15ms.

As before, API requests can also specify an optional `time_limit_ms` which now also provides a constraint on the maximum time spent on the main thread (excluding deserialization). If `time_limit_ms` is not specified then the nodeos configuration of `http-max-response-time-ms`  will be used as the limit of time spent on the main thread.

As long as the request is valid and the requested range is not empty, a minimum of one object will always be returned (one row for `get_table_rows` for example). There is also a hard coded maximum of 1000 items returned, unless `http_max_response_time` is set to `-1` in which case there is no limit.

Note: This PR changes `/v1/chain/get_activated_protocol_features` to be considered an atomic request. `limit` and `time_limit_ms` of the request will be ignored and all activated protocol features will always be returned.

So we define:
* **`max_time`** is the minimum of `http_max_response_time` and `time_limit_ms` (negative values are equivalent to infinite time).
* **`max_items`** is the requested range size, limited to 1000 unless `http_max_response_time` is negative.
* **`min_items`** is 1 if the requested range is not empty, 0 otherwise

API processing occurs in two steps:

1. On the main thread, retrieval (in binary format) of as many requested items as possible within the range [**`min_items`**, **`max_items`**], with a **`max_time`** processing limit.
2. On the http thread pool, formatting of the http response, including the ABI deserialization of every object that can be deserialized within`abi-serializer-max-time-ms`, and returning the response. 

> note: if `http_max_response_time` is set to `-1` and the request does not set a `time_limit_ms`, the request will always return the complete requested range.

> note: the protection against ABI bombs described above still applies, so if any specific object deserialization takes longer than `abi-serializer-max-time-ms`, it will be returned in binary form. Unlike prior behavior, this will not cause the request to fail and return an error code.

> note: another change from previous behavior is that `http_max_response_time` now applies exclusively to the time spent on the main thread. If left unchanged, a node running 5.0 will be able to serve larger API requests than a 4.0 node.

### Developer notes
- `safe_add` added to `time_point` that avoids overflow.
- Move constructor added to `mutable_variant_object` for `variant_object` to avoid copy.

Resolves #1062


libraries/chain/abi_serializer.cpp
```diff
@@ -23,7 +23,8 @@ namespace eosio { namespace chain {
 
    template <typename T>
    inline fc::variant variant_from_stream(fc::datastream<const char*>& stream, const abi_serializer::yield_function_t& yield) {
-      fc::yield_function_t y = [&yield](){ yield(0); }; // create yield function matching fc::variant requirements, 0 for recursive depth
+      yield(0); // reset deadline
+      fc::yield_function_t y = [yield](){ yield(1); }; // create yield function matching fc::variant requirements, 0 for recursive depth
```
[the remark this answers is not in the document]
linh2931: OK, I see from create_yield_function below, yield takes recursive depth as an argument. It's still better to add a comment here.

libraries/libfc/include/fc/variant_object.hpp
```diff
@@ -206,13 +206,19 @@ namespace fc
...
       template<typename T,
-               typename = std::enable_if_t<!std::is_base_of<mutable_variant_object,
-                                                            std::decay_t<T>>::value>>
+               typename = std::enable_if_t<!std::is_base_of<mutable_variant_object, std::decay_t<T>>::value &&
+                                           !std::is_base_of<variant, std::decay_t<T>>::value>>
       explicit mutable_variant_object( T&& v )
       :_key_value( new std::vector<entry>() )
       {
-          *this = variant(fc::forward<T>(v)).get_object();
+          *this = std::move(variant(fc::forward<T>(v)).get_object());
```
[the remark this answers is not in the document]
heifner: Why a `move` instead of a `forward`? `v` is `T&&` which is an universal reference.

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -147,7 +149,8 @@ class read_only : public api_base {
    // return deadline for call
    fc::time_point start() const {
```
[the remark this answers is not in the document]
heifner: Because it starts the call and provides the deadline for when the call should be interrupted.

libraries/libfc/include/fc/time.hpp
```diff
@@ -49,6 +49,16 @@ namespace fc {
         std::string to_iso_string()const;
         static time_point from_iso_string( const std::string& s );
 
+        // protect against overflow
+        constexpr time_point& safe_add( const microseconds& m ) {
+           if (m.count() > 0 && elapsed > fc::microseconds::maximum() - m) {
+              elapsed = microseconds::maximum();
+           } else { // does not guard against underflow
```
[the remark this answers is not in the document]
heifner: I could but as the comment says it does not guard against underflow. My thought was that in nominal use m would never be negative so don't pay the cost of checking. Happy to add a true check if you think we should have it. Probably should since caller may expect it.

libraries/libfc/include/fc/variant_object.hpp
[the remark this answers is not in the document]
greg7mdp: Oh my bad, you are right, I misread it (thought that T was not a template parameter of the constructor).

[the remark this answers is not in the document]
greg7mdp: Still, what about adding a `fc::variant::get_object() &&` so get_object() returns a rvalue if called on a rvalue?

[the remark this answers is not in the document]
heifner: I think it is too dangerous. `get_object` returns a `variant_object` which uses an internal `shared_ptr` so even though it is a rvalue it is often not safe to move from. That is why I had to add the overloaded constructor for `variant` and not move from the `get_object` as it is not safe to do so.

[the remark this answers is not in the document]
greg7mdp: I'm having a hard time totally understanding it. `variant` can hold a pointer to a `variant_object`, which itself holds a `shared_ptr` to a `vector`. It looks like this constructor is enabled only if `T` is *not* derived from (or same as)  `variant` or `mutable_variant_object`. What can `T` be then?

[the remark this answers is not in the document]
heifner: `T` an be anything that can be converted to a variant. Which includes every type with a `FC_REFLECT` or an explicit `to_variant()` implementation. Note variant has:
```
   template<typename T>
   variant::variant( const T& val )
   {
      memset( this, 0, sizeof(*this) );
      to_variant( val, *this );
   }
```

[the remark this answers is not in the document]
greg7mdp: Oh, so `T` could be a `variant_object`, and when converted to a variant, its shared_pointer would just be copied, and therefore shared? I see there is a to_variant() which takes a `variant_object`.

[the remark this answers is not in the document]
greg7mdp: But then `to_variant( val, *this );` will call this variant constructor:
```
variant::variant( variant_object obj)
{
   *reinterpret_cast<variant_object**>(this)  = new variant_object(fc::move(obj));
   set_variant_type(this,  object_type );
}
```

which ends up calling the `variant_object::variant_object( variant_object&& obj)` constructor, which ends up stealing the shared_ptr (I think). This is so confusing!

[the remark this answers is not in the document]
greg7mdp: I don't want to waste more of your time on this. I'll keep looking at this until I figure it out.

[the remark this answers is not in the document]
heifner: > Oh, so `T` could be a `variant_object`, and when converted to a variant, its shared_pointer would just be copied, and therefore shared? I see there is a to_variant() which takes a `variant_object`.

It should not be a `variant_object` as there is another constructor for that type. Adding that to the exclude list is probably a good idea to be explicit about it.

libraries/libfc/include/fc/time.hpp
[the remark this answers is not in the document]
linh2931: OK, we don't want to pay the cost of check, if it would never be negative. Maybe the name should be changed bounded_add or something. But you don't need to change.

linh2931 (approved):

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -541,9 +541,11 @@ namespace impl {
          mvo("block_time", act_trace.block_time);
          mvo("producer_block_id", act_trace.producer_block_id);
          mvo("account_ram_deltas", act_trace.account_ram_deltas);
+         ctx.reset_deadline();
```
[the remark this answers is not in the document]
heifner: Removed these calls. Instead attacked the problem directly by honoring a `max_action_data_serialization_time`.

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -120,19 +120,16 @@ struct abi_serializer {
...
-      return [max_serialization_time, deadline](size_t recursion_depth) {
+      return [max_serialization_time, last_call=fc::time_point::now()](size_t recursion_depth) mutable {
          EOS_ASSERT( recursion_depth < max_recursion_depth, abi_recursion_depth_exception,
                      "recursive definition, max_recursion_depth ${r} ", ("r", max_recursion_depth) );
 
-         EOS_ASSERT( fc::time_point::now() < deadline, abi_serialization_deadline_exception,
+         auto now = fc::time_point::now();
+         // recursion_depth 0 is flag used to indicate a reset of deadline
+         if (recursion_depth == 0) last_call = now;
+         EOS_ASSERT( (now - last_call) < max_serialization_time, abi_serialization_deadline_exception,
```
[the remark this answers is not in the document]
heifner: Refactored `abi_serializer` to accept a `max_action_data_serialization_time` which applies only to action data deserialization. Removed this hacky `yield(0)`.

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -183,11 +183,24 @@ namespace impl {
       void check_deadline()const { yield( recursion_depth ); }
       abi_serializer::yield_function_t get_yield_function() { return yield; }
 
+      void start_action_serialization() {
+         if (max_action_serialization_time.count() > 0) {
+            fc::time_point deadline = fc::time_point::now().safe_add(max_action_serialization_time);
+            yield = [deadline, y=yield, max=max_action_serialization_time](size_t depth) {
```
greg7mdp, line 189: I think there is a potential problem here. If we call `start_action_serialization` twice on the same `abi_traverse_context`, the second time `yield` will be the already updated version of yield, so we are creating a cascade of lambda calls. Should we save the original value of `yield` in the constructor?

greg7mdp:
in:

```
fc::scoped_exit<std::function<void()>> abi_traverse_context::enter_scope() {
         std::function<void()> callback = [old_recursion_depth=recursion_depth, this](){
            recursion_depth = old_recursion_depth;
         };
         ...
```

why don't we return:

```
fc::scoped_exit<std::function<void()>> abi_traverse_context::enter_scope() {
         std::function<void()> callback = [this](){
            --recursion_depth;
         };
```

greg7mdp:
Kevin, in your last changes, I think the `abi_traverse_context` constructor should ypdate `yield` to take into account `max_action_data_serialization`.

libraries/libfc/src/variant_object.cpp
```diff
@@ -302,6 +307,12 @@ namespace fc
       return *this;
    }
 
+   mutable_variant_object& mutable_variant_object::operator=( variant_object&& obj )
+   {
+      *_key_value = std::move(*obj._key_value);
+      return *this;
```
greg7mdp, line 313:
```suggestion
      if (obj._key_value.use_count() == 1)
          *_key_value = std::move(*obj._key_value);
      else
          *_key_value = *obj._key_value;
      return *this;
```

libraries/libfc/src/variant_object.cpp
```diff
@@ -286,6 +286,11 @@ namespace fc
    {
    }
 
+   mutable_variant_object::mutable_variant_object( variant_object&& obj )
+      : _key_value( new std::vector<entry>(std::move(*obj._key_value)) )
+   {
+   }
```
greg7mdp, line 292:
```suggestion
      : _key_value( new std::vector<entry> )
   {
      if (obj._key_value.use_count() == 1)
          *_key_value = std::move(*obj._key_value);
      else
          *_key_value = *obj._key_value;
   }
```

greg7mdp (changes requested):

libraries/libfc/src/variant_object.cpp
heifner: I thought about doing that. It is not thread safe which is why I held off. Restricting `variant_object` to use within single threads is probably ok. Another approach would be to remove the use of `shared_ptr` in `variant_object` but that is obviously a much bigger change.

So do you think we should just comment that `variant_object` use across threads is not thread safe and add this?

greg7mdp: I think thread safe, but approximate, and you are right there is a race condition. The doc says:
```
In multithreaded environment, the value returned by use_count is approximate (typical implementations use a [memory_order_relaxed](https://en.cppreference.com/w/cpp/atomic/memory_order) load)
```

So you are right if another thread copies the `variant_object` just after we read the use_count, we would incorrectly steal from the copy.

So really checking the `use_count` may be worse, because bad code would seem to work, only to fail on rare occasions. If we do use some variant_objects across threads, then my suggestion is incorrect I think.

In that case, maybe it is better to just remove these move constructor/assignment operators if we cannot be sure they are safe.

heifner: I think we could `assert(use_count() == 1)`. The constructor version of this is `explicit` so my thought was that if you are moving into this constructor then you should only do so if you know you have exclusive ownership of the `variant_object`.  Maybe we should add a `copy_and_destroy( variant_object&& obj )` method so that is very explicit.

greg7mdp: I vote for adding the `assert()`, and also using my version above to make it safer in release mode just in case.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1348,18 +1337,15 @@ read_only::get_activated_protocol_features( const read_only::get_activated_proto
...
+      // activated protocol features unlikely to ever reach max_return_items
+      for( unsigned int count = 0; count < params.limit && itr != end_itr; ++itr, ++count ) {
          const auto& conv_itr = convert_iterator( itr );
          activation_ordinal_value   = conv_itr.activation_ordinal();
          activation_block_num_value = conv_itr.activation_block_num();
 
          result.activated_protocol_features.emplace_back( conv_itr->to_variant( false, &mvo ) );
-         ++count;
+         if (fc::time_point::now() >= params_deadline)
+            break;
```
greg7mdp, line 1348: Should we even check the deadline in this case, since this is an atomic request and we said the time limit applied only to non-atomic requests.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1608,7 +1594,7 @@ read_only::get_table_by_scope_result read_only::get_table_by_scope( const read_o
                                                                     const fc::time_point& deadline )const {
 
    fc::microseconds params_time_limit = p.time_limit_ms ? fc::milliseconds(*p.time_limit_ms) : fc::milliseconds(10);
```
greg7mdp, line 1596:
```suggestion
   fc::microseconds params_time_limit = p.time_limit_ms ? fc::milliseconds(*p.time_limit_ms) : fc::microseconds::maximum();
```

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -1760,19 +1748,22 @@ read_only::get_producers( const read_only::get_producers_params& params, const f
    }();
 
    fc::microseconds params_time_limit = params.time_limit_ms ? fc::milliseconds(*params.time_limit_ms) : fc::milliseconds(10);
```
greg7mdp, line 1750:
```suggestion
   fc::microseconds params_time_limit = params.time_limit_ms ? fc::milliseconds(*params.time_limit_ms) : fc::microseconds::maximum();
```

greg7mdp (changes requested):

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2231,15 +2217,13 @@ void api_base::send_transaction_gen(API &api, send_transaction_params_t params,
                   }
                   if (!retried) {
                      // we are still on main thread here. The lambda passed to `next()` below will be executed on the http thread pool
-                     auto yield        = abi_serializer::create_yield_function(fc::microseconds::maximum());
-                     auto abi_cache    = abi_serializer_cache_builder(api_base::make_resolver(api.db, std::move(yield))).add_serializers(trx_trace_ptr).get();
+                     auto abi_cache    = abi_serializer_cache_builder(make_resolver(api.db, api.abi_serializer_max_time, throw_on_yield::no)).add_serializers(trx_trace_ptr).get();
```
greg7mdp, line 2220: Did you fix my issue too :-)?
https://github.com/AntelopeIO/leap/issues/1025

plugins/chain_plugin/chain_plugin.cpp
heifner: This is non-atomic because the number of activate protocol features is unbounded. However, unlikely to be very large ever. So yeah, I think we should just remove. If we do remove then we will ignore the user's requested time limit. Should we also ignore the users requested `params.limit` ?

plugins/chain_plugin/chain_plugin.cpp
heifner: The default if not provide by the user is still 10ms I thought. Our proposal I don't think changed that.

plugins/chain_plugin/chain_plugin.cpp
heifner: Only in this one place, I believe there are other locations as well.

plugins/chain_plugin/chain_plugin.cpp
greg7mdp: Yes, I think we should ignore the user's requested time limit. I would consider it atomic because the user cannot specify a range, so if it is less than the configured `http_max_response_time` the user would not be able to get the info.

plugins/chain_plugin/chain_plugin.cpp
greg7mdp: Yes, that's true, we didn't specify that in the proposal,. I think it would be more logical to say that if the user doesn't specify an explicit time limit, the `http_max_response_time` would then be the limit, rather than having a additional hardcoded 10ms here that we didn't explicitly document. But it is not a major issue imo and we can always fine tune that later if we want.

plugins/chain_plugin/chain_plugin.cpp
greg7mdp: well, thanks anyways!

libraries/chain/include/eosio/chain/abi_serializer.hpp
heifner: Updated to call from constructor of a new context type.

libraries/libfc/src/variant_object.cpp
heifner: Added assert()

libraries/libfc/src/variant_object.cpp
heifner: Added.

plugins/chain_plugin/chain_plugin.cpp
heifner: I'm not opposed to making that change. Seems like now is the time to do it along with all the other changes. Should we also reduce the default `http-max-response-time-ms` to say 15ms from 30ms.

greg7mdp: Sounds good to me (reducing to 15ms). This is a value that can easily be changed by the node operator if necessary, and we get rid of this hard-coded 10ms which I think doesn't bring much to the table.

greg7mdp: There are a few spots with this 10ms setting. I like it when we make things simpler :-).

heifner: I agree. I'll add it to the PR description. Seems like the logical way to go.

greg7mdp (approved):

linh2931 (approved):

Issue #1135, opened 2023-05-04
Bug: 4.0.0 GA - SHiP log splitting has a default value set for maximum stride slice retention - expect default to be unlimited


CryptoBloks:
Configuring stride for SHiP on 4.0.0 GA without setting a value for "max-retained-history-files" results in a default value of 10 being set, when the expected value would be "unlimited retention".  With this default value, a SHiP node with stride enabled would start to prune the state-history index once 10 stride slices had been reached, resulting in an unexpected loss of state-history content.

Example configuration:

```
## Retained Blocks Stride
blocks-log-stride = 500000
blocks-retained-dir = retained
#max-retained-block-files =

## Retained State History Stride
state-history-stride = 500000
state-history-retained-dir = retained
#max-retained-history-files =

## Trace API Stride
trace-slice-stride = 500000
```

Temporary workaround for this issues is to set "max-retained-history-files = 4294967295".

PR #1141, opened 2023-05-08
[4.0] change max-retained-history-files default

huangminghuang:
This PR 
- changes the defaults of `max-retained-history-files` to UINT32_MAX when `state-history-retained-dir` is specified,
- adds some unit tests to validate `max-retained-history-files` and `max-retained-block-files`.

Resolves https://github.com/AntelopeIO/leap/issues/1135

### DIFF
--- a/libraries/state_history/include/eosio/state_history/log.hpp
+++ b/libraries/state_history/include/eosio/state_history/log.hpp
@@ -89,7 +89,7 @@ namespace state_history {
       fc::path retained_dir       = "retained";
       fc::path archive_dir        = "archive";
       uint32_t stride             = 1000000;
-      uint32_t max_retained_files = 10;
+      uint32_t max_retained_files = UINT32_MAX;
    };
 } // namespace state_history
 
@@ -129,7 +129,7 @@ struct locked_decompress_stream {
 
 namespace detail {
 
-std::vector<char> zlib_decompress(fc::cfile& file, uint64_t compressed_size) {
+inline std::vector<char> zlib_decompress(fc::cfile& file, uint64_t compressed_size) {
    if (compressed_size) {
       std::vector<char> compressed(compressed_size);
       file.read(compressed.data(), compressed_size);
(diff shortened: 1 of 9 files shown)


libraries/state_history/include/eosio/state_history/log.hpp
```diff
@@ -89,7 +89,7 @@ namespace state_history {
       fc::path retained_dir       = "retained";
       fc::path archive_dir        = "archive";
       uint32_t stride             = 1000000;
-      uint32_t max_retained_files = 10;
+      uint32_t max_retained_files = UINT32_MAX;
```
linh2931, line 92: The PR title and description say "change state-history-stride default", but here max_retained_files default is changed. Please correct the PR title and description.

You can add some explanation of this default to the command line help `max-retained-history-files`.

linh2931 (changes requested):

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -2692,6 +2692,10 @@ fc::variant chain_plugin::get_log_trx(const transaction& trx) const {
     }
     return pretty_output;
 }
+
+const controller::config* chain_plugin::chain_config() const {
```
heifner, line 2696: I think this should return a `const controller::config&` and `EOS_ASSERT` if null.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -512,4 +512,8 @@ void state_history_plugin::handle_sighup() {
    fc::logger::update(logger_name, _log);
 }
 
+const state_history_log_config* state_history_plugin::trace_log_config() const {
```
heifner, line 515: I think this should be removed. Just use `get_trace_log()` and then if not null `->config()`.

heifner (changes requested):

plugins/state_history_plugin/include/eosio/state_history_plugin/state_history_plugin.hpp
```diff
@@ -29,6 +29,8 @@ class state_history_plugin : public plugin<state_history_plugin> {
 
    void handle_sighup() override;
 
+   const state_history_log* trace_log() const;
```
heifner, line 32: Seems odd to provide access to the `trace_log` but not the `chain_state_log`.

heifner (approved):

linh2931 (approved):

PR #1171, opened 2023-05-13
[4.0] Correctly handle close while syncing

### LINKED ISSUE #1170: Getting unlinkable_block_exception when generating a snapshot which is not at current block
Getting unlinkable_block_exception when generating a snapshot which is not at current block.

Start nodeos from an old snapshot
Let nodeos start to sync blocks
Take a snapshot before it gets to live blocks
Notice unlinkable_block_exception

It may be required to have some peers who don't have the blocks after the snapshot date.

image

heifner:
When generating snapshots during syncing, nodeos would timeout a sync and close the connection. The close would cause all queued blocks on that connection to be dropped. Also the `sync_next_expected_num` would not be reset because the `sync_source` had been reset.

Resolves #1170

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1590,7 +1590,7 @@ namespace eosio {
          } );
          sync_known_lib_num = highest_lib_num;
 
-         // if closing the connection we are currently syncing from, then reset our last requested and next expected.
+         // if closing the connection we are currently syncing from or not syncing, then reset our last requested and next expected.
          if( !sync_source || c == sync_source ) {
             reset_last_requested_num(g);
             // if starting to sync need to always start from lib as we might be on our own fork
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3253,18 +3253,6 @@ namespace eosio {
...
-         if( bsp ) {
-            // valid bsp means add_peer_block already called, need to remove it since we are not going to process the block
-            // call on dispatch strand to serialize with the add_peer_block calls
-            my_impl->dispatcher->strand.post( [blk_id]() {
-               my_impl->dispatcher->rm_block( blk_id );
-            } );
-         }
-         return;
-      }
-
```
heifner, line 3267: Note this was already removed on main as an optimization. See #1045.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1591,7 +1591,7 @@ namespace eosio {
          sync_known_lib_num = highest_lib_num;
 
          // if closing the connection we are currently syncing from, then reset our last requested and next expected.
```
linh2931, line 1593: Can you update this comment?

heifner: Updated

linh2931 (approved):

766C6164 (approved):

PR #1183, opened 2023-05-16
Move log line for `onblock` and update saved log file for tests

### LINKED ISSUE #1067: deep mind logging of onblock
Currently `onblock` `TRX_OP CREATE onblock ${id} ${trx}` is logged for `onblock` transactions that could fail as it is logged at the creation of the `onblock` and doesn't verify the `onblock` does not fail before reporting it.

greg7mdp:
Resolves #1067


libraries/chain/controller.cpp
```diff
@@ -402,11 +402,13 @@ struct controller_impl {
       }
    }
 
-   void dmlog_applied_transaction(const transaction_trace_ptr& t) {
+   void dmlog_applied_transaction(const transaction_trace_ptr& t, const signed_transaction *trx = nullptr) {
       // dmlog_applied_transaction is called by push_scheduled_transaction
       // where transient transactions are not possible, and by push_transaction
       // only when the transaction is not transient
       if (auto dm_logger = get_deep_mind_logger(false)) {
+         if (trx && !trx->actions.empty() && trx->actions[0].name == "onblock"_n)
```
heifner, line 410: Use `trace.cpp` `is_onblock`. No need to pass in the `signed_transaction`.
BTW, you would want to use `get_transaction()` of `packed_transaction` not `get_signed_transaction()`.

heifner (changes requested):

libraries/chain/controller.cpp
greg7mdp: Hum, should I also rewrite `dm_logger->on_onblock()` to take a transaction_trace?

heifner: Yes, probably better than doing the `get<packed_transaction>(t.receipt.trx).get_signed_transaction()` with appropriate error checking. Also I missed that `on_onblock` was packing a `signed_transaction`, so ignore my comment about `get_transaction` vs `get_signed_transaction` as signed transaction is desired here apparently.

greg7mdp: Would that be better than passing the `signed_transaction` like today, but using `is_onblock` as you suggested?

greg7mdp: `t.receipt` is just a `transaction_receipt_header`, not a `transaction_receipt`, so it doesn't have the `trx` member.

heifner: ugh, so it is.

greg7mdp: What about using `is_onblock()` like this? Or leave as is?

```
   void dmlog_applied_transaction(const transaction_trace_ptr& t, const signed_transaction *trx = nullptr) {
      // dmlog_applied_transaction is called by push_scheduled_transaction
      // where transient transactions are not possible, and by push_transaction
      // only when the transaction is not transient
      if (auto dm_logger = get_deep_mind_logger(false)) {
         if (is_onblock(*t))
            dm_logger->on_onblock(*trx);
         dm_logger->on_applied_transaction(self.head_block_num() + 1, t);
      }
   }
```

heifner: Yes. `const signed_transaction* trx = nullptr`. Should probably check for or assert trx != nullptr.

libraries/chain/controller.cpp
```diff
@@ -1627,7 +1629,7 @@ struct controller_impl {
                        emit(self.accepted_transaction, trx);
                    }
 
-                   dmlog_applied_transaction(trace);
+                   dmlog_applied_transaction(trace, &trx->packed_trx()->get_signed_transaction());
```
heifner, line 1632: `trn`

greg7mdp: I don't understand what you mean?

heifner: There is already the `trn` variable which is the `signed_transaction` in scope.

greg7mdp: done, thank you Kevin!

greg7mdp:
Oops, this needs to go to 4.0 branch. I'll create another PR.

heifner (approved):

libraries/chain/controller.cpp
```diff
@@ -402,11 +402,13 @@ struct controller_impl {
       }
    }
 
-   void dmlog_applied_transaction(const transaction_trace_ptr& t) {
+   void dmlog_applied_transaction(const transaction_trace_ptr& t, const signed_transaction *trx = nullptr) {
```
linh2931, line 405: Not a big deal, can `const signed_transaction *trx` be changed to `const signed_transaction* trx` to be consistent with existing style?

Better yet, not make this as a default argument. We try to avoid default arguments unless plain arguments cause big changes, so we know every calling place has been looked at.

greg7mdp: Actually, I was trying to keep the changes minimal, otherwise I have to add `nullptr` to the six other call locations, but I'm ok doing it if you think it is preferable.

greg7mdp: I did the whitespace update btw.

PR #1186, opened 2023-05-17
Added cpm-cmake to manage download / build and finding local boost

766C6164:
This is now experimental / work in progress. As @huangminghuang mentioned in original issue in discussion with @greg7mdp, in order to simplify dependency management I tried out cpm-cmake (https://github.com/cpm-cmake) to manage Boost dependency as it pretty much covers all required cases, in particular:
- if system already has boost of such or greater version installed, it will be used
- if system has no boost or older boost installed, it will be downloaded and built from github automatically by CPM

There was one (for now) makefile adjustment has to be done in chainbase submodule to make errorless configuration and build.


766C6164:
Closing that for now, we might want to consider revisiting package management approach in future, which comes with a certain limitations on it's own

Issue #1072, opened 2023-04-24
P2P Improve sync behavior


heifner:
- Track latency of peers and prioritize syncing from those with lower latency
- Add configuration of number of peers to sync from (default 3)
  - This would be different than current `p2p-peer-address` list and `max-clients`
  - Even though many `p2p-peer-address` could be specified and connected to, only the top configured number of peers would be actively synced from.
- Change default `sync-fetch-span` from `100` to `1000`
- Peer should include block start/end range in `notice_message` and peers should only attempt to sync from peers that have range of blocks needed. Currently only HEAD info is included. Especially useful for peers with no block log.
- Do not serve peers until in-sync.

heifner:
https://github.com/eosnetworkfoundation/mandel/issues/792 suggested indicating the block start/end range as part of a `handshake_v2`. I now think using `notice_message` would be cleaner.

heifner:
#1072 was only part of this issue, re-opening.

heifner:
Re-opening as there is a follow on PR to address peer review comment and to sync ahead while applying blocks.

heifner:
Re-open for a follow on PR to address peer review comment of better calculation of peer latency.

PR #1187, opened 2023-05-17
P2P Refactor connections management

heifner:
Refactor connections set into a `connections_manager` class. 
Simplify connections monitor timer implementation.
In preparation of work on #1072.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -315,6 +315,86 @@ namespace eosio {
...
+      void connection_monitor(const std::weak_ptr<connection>& from_connection);
+
+   public:
+      void add_supplied_peers(const vector<string>& peers ) {
+         std::lock_guard g(connections_mtx);
+         supplied_peers.insert( peers.begin(), peers.end() );
+      }
+
+      // not thread safe, only call on startup
+      void init( std::chrono::milliseconds heartbeat_timeout_ms,
```
greg7mdp, line 349: Maybe remove the spaces after parenthesis for consistency (and alignment with next parameters)
```suggestion
      void init(std::chrono::milliseconds heartbeat_timeout_ms,
```

plugins/net_plugin/net_plugin.cpp
```diff
@@ -315,6 +315,86 @@ namespace eosio {
...
+         return fc::microseconds{ connector_period_us.count() };
+      }
+
+      void register_update_p2p_connection_metrics(std::function<void(net_plugin::p2p_connections_metrics)>&& fun){
+         update_p2p_connection_metrics = std::move(fun);
+      }
+
+      void connect_supplied_peers() {
+         std::lock_guard g(connections_mtx);
+         for (const auto& seed_node : supplied_peers) {
```
greg7mdp, line 371: Why call it `seed_node` and not `peer`?
Since we iterate over `supplied_peers`, it is strange that we do `supplied_peers.insert(host);` in `connect_impl`, as it is likely already there!

plugins/net_plugin/net_plugin.cpp
```diff
@@ -315,6 +315,86 @@ namespace eosio {
    constexpr uint32_t signed_block_which       = fc::get_index<net_message, signed_block>();       // see protocol net_message
    constexpr uint32_t packed_transaction_which = fc::get_index<net_message, packed_transaction>(); // see protocol net_message
 
+   class connections_manager {
+      alignas(hardware_destructive_interference_size)
+      mutable std::shared_mutex connections_mtx;
+      std::set<connection_ptr>  connections;
+      chain::flat_set<string>   supplied_peers;
```
greg7mdp, line 322: why is one a `std::set` and the other a `chain::flat_set`? I'd say go flat_set all the way :-).

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1002,22 +1056,17 @@ namespace eosio {
 
 
    template<typename Function>
-   void net_plugin_impl::for_each_connection( Function&& f ) const {
+   void connections_manager::for_each_connection( Function&& f ) const {
```
greg7mdp, line 1059: I'd have these small templates inline in the class. And also `std::forward<Function>(f)` at the call location.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4040,92 +3948,265 @@ namespace eosio {
...
       }
       return result;
    }
 
    // call with connections_mtx
-   connection_ptr net_plugin_impl::find_connection( const string& host )const {
-      for( const auto& c : connections )
-         if( c->peer_address() == host ) return c;
+   connection_ptr connections_manager::find_connection_impl( const string& host )const {
+      for( const auto& c : connections ) {
```
greg7mdp, line 4107: This loop would be faster on a `flat_set`. 

I know this is not related to your change, but instead of having:

```
      std::set<connection_ptr>  connections;
      chain::flat_set<string>   supplied_peers;
```

why don't we have:

```
     std::unordered_map<string, std::optional<connection_ptr>> peers;
```

greg7mdp (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4040,92 +3948,265 @@ namespace eosio {
...
+      connections.clear();
    }
 
-   std::optional<connection_status> net_plugin::status( const string& host )const {
-      std::shared_lock<std::shared_mutex> g( my->connections_mtx );
-      auto con = my->find_connection( host );
-      if( con )
+   connection_ptr connections_manager::round_robin_next( const connection_ptr& current, uint32_t sync_known_lib_num ) const {
+      connection_ptr new_sync_source = current;
+      std::shared_lock g( connections_mtx );
```
linh2931, line 4035: Is shared_lock sufficient? Is the `new_sync_source.reset();` below considered a modify operation?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4040,92 +3948,265 @@ namespace eosio {
...
+      if( connections.empty() ) {
+         new_sync_source.reset();
```
linh2931, line 4037: `current` is passed in as a constant, but here it is modified indirectly by `new_sync_source`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4040,92 +3948,265 @@ namespace eosio {
...
+                  cptr = connections.begin();
+               }
+            }
+         }
+
+         // scan the list of peers looking for another able to provide sync blocks.
+         if( cptr != connections.end() ) {
+            auto cstart_it = cptr;
+            do {
+            // select the first one which is current and has lib above current and break out.
```
linh2931, line 4067: Why no indentation in this do-while body?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4040,92 +3948,265 @@ namespace eosio {
...
-         fc_dlog( logger, "adding new connection to the list: ${host} ${cid}", ("host", host)("cid", c->connection_id) );
-         c->set_heartbeat_timeout( heartbeat_timeout );
-         connections.insert( c );
+   /// RPC API
+   std::optional<connection_status> net_plugin::status( const string& host )const {
+      return my->connections.status(host);
+   }
+
+   /// RPC API
+   vector<connection_status> net_plugin::connections()const {
```
linh2931, line 3967: Can the method name be changed to status_list or some plural?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -315,6 +315,86 @@ namespace eosio {
...
+   private: // must call with held mutex
+      connection_ptr find_connection_impl(const string& host) const;
+      void add_impl(connection_ptr&& c);
+      void connect_impl(const string& peer);
+
+      void connection_monitor(const std::weak_ptr<connection>& from_connection);
+
+   public:
+      void add_supplied_peers(const vector<string>& peers ) {
+         std::lock_guard g(connections_mtx);
```
linh2931, line 344: Why some of methods have implementation in the header file while majority in .cpp?

linh2931 (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4040,92 +3948,265 @@ namespace eosio {
...
+            //try to find it in the list
+            cptr = connections.find( current );
+            cend = cptr;
+            if( cptr == connections.end() ) {
+               // not there - must have been closed! cend is now connections.end, so just flatten the ring.
+               new_sync_source.reset();
+               cptr = connections.begin();
+            } else {
+               // was found - advance the start to the next. cend is the old source.
+               if( ++cptr == connections.end() && cend != connections.end() ) {
```
greg7mdp, line 4057: At this point we know that `cend != connections.end()`:

```suggestion
               if( ++cptr == connections.end() ) {
```

greg7mdp (changes requested):

plugins/net_plugin/tests/auto_bp_peering_unittest.cpp
```diff
@@ -29,9 +30,14 @@ struct mock_net_plugin : eosio::auto_bp_peering::bp_connection_manager<mock_net_
             return;
       }
    }
+};
 
-   std::function<void(std::string)> connect;
-   std::function<void(std::string)> disconnect;
+struct mock_net_plugin : eosio::auto_bp_peering::bp_connection_manager<mock_net_plugin, mock_connection> {
```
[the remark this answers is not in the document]
heifner: Hmm, yeah, I think this more closely matches the actual impl.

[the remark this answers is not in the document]
greg7mdp: totally fair!

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3463,28 +3458,6 @@ namespace eosio {
       }
    }
 
-   // called from any thread
-   void net_plugin_impl::start_conn_timer(boost::asio::steady_timer::duration du, std::weak_ptr<connection> from_connection) {
-      if( in_shutdown ) return;
-      std::lock_guard<std::mutex> g( connector_check_timer_mtx );
-      ++connector_checks_in_flight;
-      connector_check_timer->expires_from_now( du );
-      connector_check_timer->async_wait( [my = shared_from_this(), from_connection{std::move(from_connection)}](boost::system::error_code ec) mutable {
```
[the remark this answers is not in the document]
heifner: Yes, we cancel in shutdown. Cancel means `ec` is true, so `this` is not accessed after the `cancel`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3463,28 +3458,6 @@ namespace eosio {
       }
    }
 
-   // called from any thread
-   void net_plugin_impl::start_conn_timer(boost::asio::steady_timer::duration du, std::weak_ptr<connection> from_connection) {
-      if( in_shutdown ) return;
```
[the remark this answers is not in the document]
heifner: No, because we cancel the timer.

plugins/net_plugin/net_plugin.cpp
heifner: No, because it matches the API name.

plugins/net_plugin/net_plugin.cpp
heifner: `new_sync_source` is a local `shared_ptr`, so that is safe. `shared_ptr` provides thread safety of the shared count.

plugins/net_plugin/net_plugin.cpp
heifner: Because there are incoming connections in `connections` that are not supplied via command-line/config.ini or `/v1/chain/connect`.

plugins/net_plugin/net_plugin.cpp
heifner: Changed to `flat_set`.

plugins/net_plugin/net_plugin.cpp
heifner: Done

plugins/net_plugin/net_plugin.cpp
heifner: `for_each_block_connection` can't be include because it uses `connection` `is_transactions_only_connection`. So moved them both out to be consistent.

plugins/net_plugin/net_plugin.cpp
heifner: Not sure how that happened. Fixed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4040,92 +3948,265 @@ namespace eosio {
...
          result.push_back( c->get_status() );
       }
       return result;
    }
 
    // call with connections_mtx
-   connection_ptr net_plugin_impl::find_connection( const string& host )const {
-      for( const auto& c : connections )
-         if( c->peer_address() == host ) return c;
+   connection_ptr connections_manager::find_connection_impl( const string& host )const {
```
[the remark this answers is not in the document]
heifner: I'll rename to `_i` the name is to indicate it is an private method which requires mutex to be held already.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4040,92 +3948,265 @@ namespace eosio {
...
-      return sync_master->is_in_sync();
+   // call with connections_mtx
+   void connections_manager::add_impl(connection_ptr&& c) {
+      c->set_heartbeat_timeout( heartbeat_timeout );
+      connections.insert( std::move(c) );
    }
 
-   void net_plugin::register_update_p2p_connection_metrics(std::function<void(net_plugin::p2p_connections_metrics)>&& fun){
-      my->update_p2p_connection_metrics = std::move(fun);
+   // called from any thread
```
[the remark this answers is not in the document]
heifner: Added comment

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2987,10 +2978,10 @@ namespace eosio {
             auto c_time = last_handshake_sent.time;
             g_conn.unlock();
             peer_dlog( this, "checking for duplicate" );
-            std::shared_lock<std::shared_mutex> g_cnts( my_impl->connections_mtx );
-            for(const auto& check : my_impl->connections) {
+            bool dup = false;
+            my_impl->connections.for_each_connection([&](const auto& check) {
```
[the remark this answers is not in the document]
heifner: Not sure that a method named `any` is any better than just `for_each_connection`. I Can certainly make an external lambda to avoid the bool dup. I don't see how we can remove the `if( !f( c ) ) return;` as it is used to short circuit out of the for loop.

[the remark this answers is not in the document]
greg7mdp: I am suggesting *adding* the `any` in addition to the `for_each_connection`. If using the `any` in that one location, the short-circuiting of the `for_each` will not be needed anymore.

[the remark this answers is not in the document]
heifner: Oh, I think I misunderstood. See if what I have now matches what you are asking for.

plugins/net_plugin/net_plugin.cpp
heifner: Removed

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3004,31 +2983,33 @@ namespace eosio {
...
-                  go_away_message gam(duplicate);
-                  gam.node_id = conn_node_id;
-                  enqueue(gam);
-                  no_retry = duplicate;
-                  return;
+                  return true;
                }
+               return false;
+            };
+            if (my_impl->connections.any_of_connections(std::move(is_duplicate))) {
```
greg7mdp, line 3006: looks good, but as a matter of preference I'd call the function simply `any_of` because it is always used as:

`connections.any_of_connections`, and I don't think the repetition adds anything.

greg7mdp (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -315,6 +315,75 @@ namespace eosio {
...
+      template <typename Function>
+      void for_each_connection(Function&& f) const;
+
+      template <typename Function>
+      void for_each_block_connection(Function&& f) const;
+
+      template <typename UnaryPredicate>
+      bool any_of_connections(UnaryPredicate&& p) const {
+         std::shared_lock g(connections_mtx);
+         return std::find_if(connections.cbegin(), connections.cend(), std::forward<UnaryPredicate>(p)) != connections.cend();
```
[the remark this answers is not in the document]
heifner: Done

plugins/net_plugin/net_plugin.cpp
heifner: Left it since I added an `any_of_block_connections`

greg7mdp: Cool, but I think your `any_of_block_connections` is missing the `block` check.

heifner: Thanks. Fixed.

greg7mdp (approved):

linh2931 (approved):

Issue #1130, opened 2023-05-04
Bug: 4.0.0 nodeos consumes all available memory during a blocks log replay


CryptoBloks:
While replaying a large blocks log (reproduced on a >270mm block log from Telos mainnet) nodeos consumes all memory available until memory exhaustion, at which point the nodeos process exits.

Steps to reproduce:
- Start nodeos instance with the --replay-blockchain flag, replaying an existing large set of blocks.
- Watch the system memory utilization for the nodeos thread, it will climb very quickly.
- Wait for all available memory to be consumed, process will exit.

Reproduced on Ubuntu 18.04LTS and LEAP 4.0.0 GA.

Starting and stopping the nodeos process releases the consumed memory, so a periodic restart of the process is a valid workaround for now.

Screenshots attached.
![Screenshot 2023-05-03 at 11 26 28 PM](https://user-images.githubusercontent.com/93884436/236127583-9f81a3e0-8341-487b-a63f-fb6fc2c28eef.png)
![Screenshot 2023-05-03 at 11 24 31 PM](https://user-images.githubusercontent.com/93884436/236127586-e5ed001f-6b03-4aa6-a5f0-0a2174c4e1e7.png)
![Screenshot 2023-05-03 at 11 27 13 PM](https://user-images.githubusercontent.com/93884436/236127588-3afe9e72-1ddb-4466-b1cf-e767ebe255cd.png)
![Screenshot 2023-05-03 at 11 23 39 PM](https://user-images.githubusercontent.com/93884436/236127590-4e3eb7e6-3131-4047-86f7-37c069002c74.png)
![Screenshot 2023-05-03 at 11 29 44 PM](https://user-images.githubusercontent.com/93884436/236127593-da6e37aa-f41d-4af4-92b0-6e9190bad9ec.png)

CryptoBloks:
Additional context, this replay is happening on an API node that is also running a full SHiP node.

PR #1191, opened 2023-05-17
[4.0] Fix for nodeos large memory consumption during a blocks log replay

766C6164:
It was reported that during bloks log replay with a large blocks log nodeos process grows in memory to the point of it's exhaustion.

Resolves https://github.com/AntelopeIO/leap/issues/1130

### DIFF
--- a/plugins/state_history_plugin/state_history_plugin.cpp
+++ b/plugins/state_history_plugin/state_history_plugin.cpp
@@ -99,7 +99,8 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
    std::set<acceptor_type>          acceptors;
 
    named_thread_pool<struct ship> thread_pool;
-   bool                           thread_pool_started = false;
+
+   bool  plugin_started = false;
 
    static fc::logger& logger() { return _log; }
 
@@ -295,7 +296,7 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
 
       // avoid accumulating all these posts during replay before ship threads started
       // this is safe as there are no clients connected untill after replay is complete 
-      if (thread_pool_started) {
+      if (plugin_started) {
          boost::asio::post(get_ship_executor(), [self = this->shared_from_this(), block_state]() {
             self->session_mgr.send_update(block_state);
          });
@@ -500,10 +501,8 @@ void state_history_plugin::plugin_startup() {
       my->thread_pool.start( 1, [](const fc::exception& e) {
          fc_elog( _log, "Exception in SHiP thread pool, exiting: ${e}", ("e", e.to_detail_string()) );
          app().quit();
-      },
-      [this]() {
-         my->thread_pool_started = true;
       });
+      my->plugin_started = true; 
    } catch (std::exception& ex) {
       appbase::app().quit();
    }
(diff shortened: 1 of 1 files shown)


plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -292,7 +292,7 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
              "the process");
       }
 
-      boost::asio::post(get_ship_executor(), [self = this->shared_from_this(), block_state]() {
+      boost::asio::post(get_ship_executor(), [self = this->shared_from_this(), &block_state]() {
```
spoonincode, line 295: Does this cause a race where this `block_state` can be destroyed before this lambda is executed? Is there something else that can guarantee `block_state`'s ref count remains alive?

heifner: If this is the memory "leak", then we just need to apply back pressure.

spoonincode: I think you might just be able to skip this post() if the ship thread pool hasn't been started? The problem is replay happens synchronously in chain_plugin's initialize -- before the ship threads are started -- leading to all theses post()s just stacking up until replay is complete.

Since there won't be any ship clients connected until after replay is complete, delaying these posts until the threads are running seems reasonable..

spoonincode: wait.. that can't be right.. since the signal isn't connected until after state_history plugin's initaialize either, hm..

heifner: Replay happens in chain_plugin plugin_startup. We can just skip the post if there are no sessions which is affectively what was done before.

spoonincode: ah okay, then yeah makes sense

766C6164: Tried that and it works, that was definitely it :) Thanks a lot @spoonincode and @heifner!

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -99,6 +99,7 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
    std::set<acceptor_type>          acceptors;
 
    named_thread_pool<struct ship> thread_pool;
+   bool                           thread_pool_started = false;
```
heifner, line 102: As you have it this needs to be an `std::atomic<bool>` as it is accessed from two different threads. You could call it `plugin_started` and set it in `state_history_plugin::plugin_startup` and then it would only be accessed from the main thread.

heifner (changes requested):

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: Thank you, that looks better, done.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -292,9 +294,13 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
              "the process");
       }
 
-      boost::asio::post(get_ship_executor(), [self = this->shared_from_this(), block_state]() {
-         self->session_mgr.send_update(block_state);
-      });
+      // avoid accumulating all these posts during replay before ship threads started
+      // this is safe as there are no clients connected untill after replay is complete
```
heifner, line 298: s/untill/until/`

heifner (approved):

plugins/state_history_plugin/state_history_plugin.cpp
766C6164: thank you, fixed!

greg7mdp (approved):

linh2931 (approved):

PR #1199, opened 2023-05-18
Improve packed_transaction parsing performance

### LINKED ISSUE #1025: Improve packed_transaction parsing performance

greg7mdp:
Resolves #1025.

Introduce a `caching_resolver `class, which provides the same API as a resolver, but which caches `abi_serializers`  as they are retrieved, so we pay the cost of retrieval only once.

### DIFF
--- a/libraries/chain/include/eosio/chain/abi_serializer.hpp
+++ b/libraries/chain/include/eosio/chain/abi_serializer.hpp
@@ -736,7 +736,7 @@ namespace impl {
        * and can be degraded to the normal ::from_variant(...) processing
        */
       template<typename M, typename Resolver, not_require_abi_t<M> = 1>
-      static void extract( const fc::variant& v, M& o, Resolver, abi_traverse_context& ctx )
+      static void extract( const fc::variant& v, M& o, const Resolver&, abi_traverse_context& ctx )
       {
          auto h = ctx.enter_scope();
          from_variant(v, o);
@@ -1084,13 +1084,18 @@ public:
    {
    }
 
+   // make it non-copiable (we should only move it for performance reasons)
+   caching_resolver(const caching_resolver&) = delete;
+   caching_resolver& operator=(const caching_resolver&) = delete;
+
    std::optional<std::reference_wrapper<const abi_serializer>> operator()(const account_name& account) const {
       auto it = abi_serializers.find(account);
       if (it != abi_serializers.end() && it->second)
          return *it->second;
       try {
+         auto serializer = resolver_(account);
          auto& dest = abi_serializers[account];
-         dest = resolver_(account);
+         dest = abi_serializer_cache_t::mapped_type{std::move(serializer)};
          return *dest;
       } catch( ... ) {
          throw; // throw if embedded resolver throws
--- a/libraries/chain/include/eosio/chain/abi_serializer.hpp
+++ b/libraries/chain/include/eosio/chain/abi_serializer.hpp
@@ -1090,13 +1090,20 @@ public:
 
    std::optional<std::reference_wrapper<const abi_serializer>> operator()(const account_name& account) const {
       auto it = abi_serializers.find(account);
-      if (it != abi_serializers.end() && it->second)
-         return *it->second;
+      if (it != abi_serializers.end()) {
+         if (it->second)
+            return *it->second;
+         return {};
+      }
       try {
          auto serializer = resolver_(account);
-         auto& dest = abi_serializers[account];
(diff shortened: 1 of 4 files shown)


libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -1066,8 +1068,39 @@ class abi_serializer_cache_builder {
...
+   std::optional<std::reference_wrapper<const abi_serializer>> operator()(const account_name& account) const {
+      auto it = abi_serializers.find(account);
+      if (it != abi_serializers.end() && it->second)
+         return *it->second;
+      try {
+         auto& dest = abi_serializers[account];
+         dest = resolver_(account);
+         return *dest;
+      } catch( ... ) {
+         throw; // throw if embedded resolver throws
```
heifner, line 1096: Why catch and throw here?

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -825,13 +825,14 @@ namespace impl {
                from_variant(data, act.data);
```
heifner, line 825: The `extract` on line 739 takes `Resolver` by value. I would hope the compiler would remove that copy, but best to change it to a `const&` I think.

heifner (changes requested):

heifner:
> @heifner if I create a `abi_serializer` cache when parsing the `packed_transaction` sent by the user, can I reuse the same cache for serializing the trace result? In my current version we do recreate the cache from the trace.

Short answer:  no. 
The transaction itself or another transaction executed between the time of trx input and trx execution can `setabi` which would change the abi. Your cache could check for a diff `abi_sequence` on the account, but you would have to look up the account.  You would save the creation of the `abi_serializer`, but not sure it is worth the extra complexity.

libraries/chain/include/eosio/chain/abi_serializer.hpp
greg7mdp: yes for sure I'll change it.!

libraries/chain/include/eosio/chain/abi_serializer.hpp
```diff
@@ -1066,8 +1068,39 @@ class abi_serializer_cache_builder {
...
+   resolver_fn_t resolver_;
    abi_serializer_cache_t abi_serializers;
 };
 
+/*
+ * This is equivalent to a resolver, except that everytime the abi_serializer for an account 
+ * is retrieved, it is stored in an unordered_map, so we won't waste time retrieving it again.
+ * This is handy when parsing packed_transactions received in a fc::variant.
+ */
+class caching_resolver {
```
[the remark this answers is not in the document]
greg7mdp: sure.

libraries/chain/include/eosio/chain/abi_serializer.hpp
greg7mdp: done

greg7mdp:
Looks like I have some test failures. I looked at one an it looks like we are trying to retrieve the account before we create it? Am I understanding this right?

image

heifner:
> Looks like I have some test failures. I looked at one an it looks like we are trying to retrieve the account before we create it? Am I understanding this right?

You can ignore that. It is expected.

heifner:
> > Looks like I have some test failures. I looked at one an it looks like we are trying to retrieve the account before we create it? Am I understanding this right?
> 
> You can ignore that. It is expected.

See telegram.

greg7mdp:
Hey, thanks for the tip, I debugged with the core file (super easy)  and it pointed me to the problem. Apparently the cache was returning an incorrect abi_serializer. I'm not 100% sure why, but I think it was because we construct a `std::optional` containing a `std::reference_wrapper`, and when constructing it from an empty `std::optional<abi_serializer>` it didn't work quite as expected. Anyways I tried to be very explicit and it seems to work.

libraries/chain/include/eosio/chain/abi_serializer.hpp
heifner: The catch-throw is still here.

greg7mdp: Excellent point, it is unneeded. Removed it!

heifner (approved):

766C6164 (approved):

PR #1203, opened 2023-05-18
IPv6 support

### LINKED ISSUE #622: IPv6 support
This is a carry over of eosnetworkfoundation/mandel#357 but with the removal of cleos -- as of 4.0 cleos does support IPv6 :tada: 

It's possible as we start development work on IPv6 support we can split this in to separate issues for net_plugin support, state_history_plugin support, and http_plugin support. But all cases will require understanding how to resolve the configuration challenge first.

(possibly non-exhaustive) tasks that need to be accomplished:

* Any configurations that listen on a port (`http-server-address`, `p2p-listen-endpoint`, `state-history-endpoint`, etc) need to be modified to support multiple values.
  * This modification is at odds with the current behavior where one can set these to the empty string to disable listening. Some solution to the configuration scheme needs to be found to accommodate this case.
* For these listen endpoints, they should accept four different kinds of configuration
  1. IPv4 address & port: `10.1.2.3:9876`
  2. IPv6 address & port: `[2607:f8b0:4004:806::2004]:9876` I'm doubtful explicit scope support is needed?
  3. hostname & port: `foo.example.com:9876`, in this case I believe the expectation is that the plugin resolves the hostname and listens on all addresses it resolved. A perfect example is `localhost:9876` which should listen on both `127.0.0.1:9876` and `[::1]:9876`
    It's also possible that instead of "listens on all addresses it resolved" it should listen _only on the first_ IPv6 & IPv4 address it resolved.
    Probably should reflect some on expected behavior here and what other tools do.
  5. wildcard & port (needed? expected?): `*:9876` This is a shorthand for not having to specify both `0.0.0.0:9876` and `[::]:9876` when wanting to listen from everywhere. Implementation can decide whether this warrants two listening sockets or utilization of IPv4 in IPv6 sockets.
* Any defaults of `127.0.0.1:8888` or such need to be changed to `localhost:8888` to ensure both IPv4 & IPv6 connections to it are possible.
* Outgoing connections from p2p plugin that are via hostnames may consider roundrobining through the resolved addresses. This is to ensure if there is both IPv6 & IPv4 records and IPv6 connectivity is busted that eventually the connection will be tried through IPv4.
* http_client_plugin, which expects the entire lookup & HTTP request to complete in a very small amount of time (sub 5ms), probably continues to only try to connect to the first resolved address (whichever it gets: IPv6 or IPv4).

huangminghuang:
This PR refactors net_plugin, http_plugin, and state_history_plugin to support IPv6.

Resolves #622


libraries/libfc/include/fc/network/listener.hpp
```diff
@@ -0,0 +1,211 @@
...
+   typename Protocol::acceptor acceptor_;
+   boost::asio::deadline_timer accept_error_timer_;
+
+   listener(State state, const endpoint_type& endpoint)
+       : state_(state), acceptor_(state->thread_pool.get_executor(), endpoint),
+         accept_error_timer_(state->thread_pool.get_executor()) {}
+
+   void do_accept() {
+      acceptor_.async_accept(state_->thread_pool.get_executor(),
+                             [self = this->shared_from_this()](boost::system::error_code ec, auto&& peer) {
```
heifner, line 62: Name `peer` `peer_socket`

libraries/libfc/include/fc/network/listener.hpp
```diff
@@ -0,0 +1,211 @@
...
+                                        code == EHOSTDOWN || code == EHOSTUNREACH || code == EOPNOTSUPP ||
+                                        code == ENETUNREACH
+#ifdef ENONET
+                                        || code == ENONET
+#endif
+      ) {
+         // according to https://man7.org/linux/man-pages/man2/accept.2.html, reliable application should
+         // retry when these error codes are returned
+         fc_elog(state_->get_logger(), "accept: ${m}", ("m", ec.message()));
+         fc_elog(state_->get_logger(), "closing connection");
```
heifner, line 91: These should be `fc_wlog`

libraries/libfc/include/fc/network/listener.hpp
```diff
@@ -0,0 +1,211 @@
...
+      } else if (int code = ec.value(); code == ENETDOWN || code == EPROTO || code == ENOPROTOOPT ||
+                                        code == EHOSTDOWN || code == EHOSTUNREACH || code == EOPNOTSUPP ||
+                                        code == ENETUNREACH
+#ifdef ENONET
+                                        || code == ENONET
+#endif
+      ) {
+         // according to https://man7.org/linux/man-pages/man2/accept.2.html, reliable application should
+         // retry when these error codes are returned
+         fc_elog(state_->get_logger(), "accept: ${m}", ("m", ec.message()));
```
heifner, line 90: `"closing connection, accept error: ${m}"`
Then you don't need two log lines.

libraries/libfc/include/fc/network/listener.hpp
```diff
@@ -0,0 +1,211 @@
...
+         test_socket.connect(endpoint, ec);
+
+         // looks like a service is already running on that socket, don't touch it... fail out
+         if (ec == boost::system::errc::success) {
+            fc_elog(state->get_logger(), "The unix socket path ${addr} is already in use", ("addr", address));
+            throw std::system_error(std::make_error_code(std::errc::address_in_use));
+         }
+         // socket exists but no one home, go ahead and remove it and continue on
+         else if (ec == boost::system::errc::connection_refused)
+            fs::remove(sock_path);
```
heifner, line 204: Should the destructor remove this path?

libraries/libfc/include/fc/network/listener.hpp
```diff
@@ -0,0 +1,211 @@
...
+   }
+
+   template <typename Socket>
+   void on_accept(boost::system::error_code ec, Socket&& socket) {
+      if (!ec) {
+         static_cast<T*>(this)->create_session(std::forward<Socket>(socket));
+         do_accept();
+      } else if (ec == boost::system::errc::too_many_files_open) {
+         // retry accept() after timeout to avoid cpu loop on accept
+         auto timeout_ms = static_cast<T*>(this)->accept_timeout_ms;
```
heifner, line 74: I this this could either be an argument to the constructor or just normalize everyone to 100ms. My vote would be to just make it a const 100ms for every listener.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -706,6 +704,7 @@ namespace eosio {
    class connection : public std::enable_shared_from_this<connection> {
    public:
       explicit connection( const string& endpoint );
+      explicit connection( tcp::socket&& socket );
       connection();
```
heifner, line 708: This constructor no longer needed, remove it.

heifner (changes requested):

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -98,7 +96,7 @@ class beast_http_session : public detail::abstract_conn,
    // HTTP response object
    std::optional<http::response<http::string_body>> res_;
 
-   std::shared_ptr<http_plugin_state> plugin_state_;
+   http_plugin_state* plugin_state_;
```
spoonincode, line 99: Is there any way to refactor this to not store a naked pointer?

libraries/libfc/include/fc/network/listener.hpp
```diff
@@ -0,0 +1,211 @@
...
+         fc_elog(state_->get_logger(), "accept: too many files open - waiting ${timeout}ms", ("timeout", timeout_ms));
```
spoonincode, line 75: Maybe consider elaborating on this message a little more as I'm not sure everyone understands files == connections in this context. "open file limit reached: not accepting new connections for next Xms" or something like that

I'd like to figure out a way to suggest the action for user to take but I can't think of a concise way to put it.. "open file limit reached: not accepting new connections for next Xms, consider increasing 'ulimit -n'" .. but someone using docker, etc, would adjust this knob differently so don't really like that

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1000,6 +997,28 @@ namespace eosio {
...
+      }
+      string::size_type colon = peer_add.find(':', p);
+      string::size_type colon2 = peer_add.find(':', colon + 1);
+      string::size_type end = colon2 == string::npos
+            ? string::npos : peer_add.find_first_of( " :+=.,<>!$%^&(*)|-#@\t", colon2 + 1 ); // future proof by including most symbols without using regex
+      string host = (p > 0) ? peer_add.substr( 1, p-1 ) : peer_add.substr( 0, colon );
+      string port = peer_add.substr( colon + 1, colon2 == string::npos ? string::npos : colon2 - (colon + 1));
+      string type = colon2 == string::npos ? "" : end == string::npos ?
+         peer_add.substr( colon2 + 1 ) : peer_add.substr( colon2 + 1, end - (colon2 + 1) );
+      return {host, port, type};
```
greg7mdp, line 1019:
```suggestion
      return {std::move(host), std::move(port), std::move(type)};
```

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2434,17 +2445,14 @@ namespace eosio {
...
+         auto [host, port, type] = split_host_port_type(c->peer_address());
          c->set_connection_type( c->peer_address() );
 
          auto resolver = std::make_shared<tcp::resolver>( my_impl->thread_pool.get_executor() );
          connection_wptr weak_conn = c;
          // Note: need to add support for IPv6 too
-         resolver->async_resolve( tcp::v4(), host, port, boost::asio::bind_executor( c->strand,
-            [resolver, weak_conn, host, port]( const boost::system::error_code& err, const tcp::resolver::results_type& endpoints ) {
+         resolver->async_resolve(host, port, boost::asio::bind_executor( c->strand,
+            [resolver, weak_conn, host = host, port = port]( const boost::system::error_code& err, const tcp::resolver::results_type& endpoints ) {
```
greg7mdp, line 2455: I don't see the point of `host = host, port = port`, what am I missing?

```suggestion
            [resolver, weak_conn, host, port]( const boost::system::error_code& err, const tcp::resolver::results_type& endpoints ) {
```

greg7mdp (changes requested):

plugins/net_plugin/net_plugin.cpp
huangminghuang: https://burnicki.pl/en/2021/04/19/capture-structured-bindings.html

plugins/http_plugin/include/eosio/http_plugin/common.hpp
```diff
@@ -148,8 +152,8 @@ struct http_plugin_state {
 * @param session_ptr - beast_http_session object on which to invoke send_response
 * @return lambda suitable for url_response_callback
 */
-inline auto make_http_response_handler(std::shared_ptr<http_plugin_state> plugin_state, detail::abstract_conn_ptr session_ptr, http_content_type content_type) {
-   return [plugin_state{std::move(plugin_state)},
+inline auto make_http_response_handler(http_plugin_state* plugin_state, detail::abstract_conn_ptr session_ptr, http_content_type content_type) {
+   return [plugin_state = plugin_state,
```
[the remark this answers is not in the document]
huangminghuang: The handlers can't be executed after the plugin shutdown, and the state will only be deleted by the destructor of the plugin, so it's safe to do so.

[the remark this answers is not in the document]
spoonincode: Please change to `http_plugin_state&` then.

heifner (approved):

greg7mdp (approved):

Issue #1152, opened 2023-05-10
Unexpected behavior with scheduled snapshots


aaroncox:
The way scheduled snapshots behaves isn't what I'd anticipate, especially from an automation standpoint. 

What I want to accomplish with automated snapshots is to be able to set an interval (without a start/end) and have it take snapshots at the heights which match the `block_spacing`. Currently I'm trying to achieve this by manually defining the `snapshot-schedule.json` as:

```json
{
    "snapshot_requests": [
        {
            "snapshot_request_id": "0",
            "snapshot_description": "Hourly snapshots starting from genesis",
            "block_spacing": "7200",
            "start_block_num": "0",
            "end_block_num": "0"
        }
    ]
}
```

This approach creates an easy, default way for us to automate snapshots at a given height - without the need to make API calls after the node is up and running. With Docker I can simply drop in the above configuration file and know the node is going to take snapshots at a specific block height interval.

What I have discovered though using this configuration is that when nodeos starts from a snapshot of a random height, it almost immediately takes a new snapshot, despite the height not matching the `block_spacing` value. For example using a recent Jungle4 snapshot, the above configuration produces a file named:

`snapshot-04865af1c6ab73887325b198445aad8d9e7d6037db2c307c5fbd7cf0b437975d.bin`

With that block ID hash converting to a block number of `75913969`. 

This also automatically modified my `snapshot-schedule.json` file to have a `start_block_num` value of `75913969`. 

This block height wasn't when I expected a snapshot to occur, since I specified every `7200` blocks with a start height of `0`. I would expect the first snapshot to wait and occur at block number `75909600` (divisible by 7200, no remainder) - and I wouldn't expect the JSON schedule to automatically change based on this startup behavior.

One of the goals with automated backups like this is so that we can have multiple providers creating snapshots at the exact same height, which provides some authenticity to the snapshots since everyone's snapshots at a given height will have the same file hash. My hope was that a simple configuration like the above would allow us to do that.

766C6164:
Thank you for detailed report! start_block_num = 0 is a "special" case which is treated as missing/not specified start_block_num in API call. If it is not specified, it is assumed as "ASAP"/now. That will make a all to api without parameters similar to former create_snapshot call that does it "now" (but async here)
If you set start_block_num to any other value, like start_block_num = 1 you will get behavior closer to what you are expecting.
I  believe the right solution would be to treat cases with "0" and "not specified" separately

BenjaminGormanPMP:
[...]

PR #1220, opened 2023-05-31
Extended snapshot scheduler to better handle missing and zero params

766C6164:
This PR add differentiation between missing and zero-valued parameters in api request for start, end and spacing of blocks, in particular, as it was observed in original issue making start_block_num assignable to "0" in request or directly in snapshot-schedule.json file, such as:

```json
{
    "snapshot_requests": [
        {
            "snapshot_request_id": "0",
            "snapshot_description": "Hourly snapshots starting from genesis",
            "block_spacing": "7200",
            "start_block_num": "0",
            "end_block_num": "0"
        }
    ]
}
```

Prior implementation treated zero start as missing value and replaced it with "now" block num

Resolves https://github.com/AntelopeIO/leap/issues/1152

### DIFF
--- a/libraries/chain/include/eosio/chain/snapshot_scheduler.hpp
+++ b/libraries/chain/include/eosio/chain/snapshot_scheduler.hpp
@@ -20,6 +20,8 @@
 #include <boost/property_tree/json_parser.hpp>
 #include <boost/property_tree/ptree.hpp>
 
+#include <limits>
+
 namespace eosio::chain {
 
 namespace bmi = boost::multi_index;
@@ -38,7 +40,7 @@ public:
    struct snapshot_request_information {
       uint32_t block_spacing = 0;
       uint32_t start_block_num = 0;
-      uint32_t end_block_num = UINT32_MAX - 1;
+      uint32_t end_block_num = std::numeric_limits<uint32_t>::max();
       std::string snapshot_description = "";
    };
 
--- a/libraries/chain/include/eosio/chain/snapshot_scheduler.hpp
+++ b/libraries/chain/include/eosio/chain/snapshot_scheduler.hpp
@@ -38,8 +38,17 @@ public:
    struct snapshot_request_information {
       uint32_t block_spacing = 0;
       uint32_t start_block_num = 0;
-      uint32_t end_block_num = 0;
+      uint32_t end_block_num = UINT32_MAX - 1;
       std::string snapshot_description = "";
+   };
+
+   // this struct used to hold request params in api call
+   // it is differentiate between 0 and empty values
+   struct snapshot_request_params {
+      std::optional<uint32_t> block_spacing;
+      std::optional<uint32_t> start_block_num;
+      std::optional<uint32_t> end_block_num;
+      std::optional<std::string> snapshot_description;
    };
 
    struct snapshot_request_id_information {
@@ -205,6 +214,7 @@ public:
 
 FC_REFLECT(eosio::chain::snapshot_scheduler::snapshot_information, (head_block_id) (head_block_num) (head_block_time) (version) (snapshot_name))
 FC_REFLECT(eosio::chain::snapshot_scheduler::snapshot_request_information, (block_spacing) (start_block_num) (end_block_num) (snapshot_description))
+FC_REFLECT(eosio::chain::snapshot_scheduler::snapshot_request_params, (block_spacing) (start_block_num) (end_block_num) (snapshot_description))
 FC_REFLECT(eosio::chain::snapshot_scheduler::snapshot_request_id_information, (snapshot_request_id))
 FC_REFLECT(eosio::chain::snapshot_scheduler::get_snapshot_requests_result, (snapshot_requests))
 FC_REFLECT_DERIVED(eosio::chain::snapshot_scheduler::snapshot_schedule_information, (eosio::chain::snapshot_scheduler::snapshot_request_id_information)(eosio::chain::snapshot_scheduler::snapshot_request_information), (pending_snapshots))
--- a/libraries/chain/snapshot_scheduler.cpp
+++ b/libraries/chain/snapshot_scheduler.cpp
@@ -24,17 +24,18 @@ void snapshot_scheduler::on_start_block(uint32_t height, chain::controller& chai
    std::vector<uint32_t> unschedule_snapshot_request_ids;
    for(const auto& req: _snapshot_requests.get<0>()) {
       // -1 since its called from start block
-      bool recurring_snapshot = req.block_spacing && (height >= req.start_block_num + 1) && (!((height - req.start_block_num - 1) % req.block_spacing));
-      bool onetime_snapshot = (!req.block_spacing) && (height == req.start_block_num + 1);
+      bool recurring_snapshot  = req.block_spacing && (height >= req.start_block_num + 1) && (!((height - req.start_block_num - 1) % req.block_spacing));
+      bool onetime_snapshot    = (!req.block_spacing) && (height == req.start_block_num + 1);
+      
+      bool marked_for_deletion = ((!req.block_spacing) && (height >= req.start_block_num + 1)) || // if one time snapshot executed or scheduled for the past, it should be gone
+                                 (height > 0 && ((height-1) >= req.end_block_num));               // any snapshot can expire by end block num (end_block_num can be max value)
 
       if(recurring_snapshot || onetime_snapshot) {
          execute_snapshot_with_log(req);
       }
 
       // cleanup - remove expired (or invalid) request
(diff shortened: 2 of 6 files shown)


libraries/chain/snapshot_scheduler.cpp
```diff
@@ -28,18 +27,7 @@ void snapshot_scheduler::on_start_block(uint32_t height, chain::controller& chai
       bool recurring_snapshot = req.block_spacing && (height >= req.start_block_num + 1) && (!((height - req.start_block_num - 1) % req.block_spacing));
       bool onetime_snapshot = (!req.block_spacing) && (height == req.start_block_num + 1);
```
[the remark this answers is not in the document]
greg7mdp: This:

```
         (!req.block_spacing && height >= (req.start_block_num + 1)) ||
         (req.end_block_num > 0 && height >= (req.end_block_num + 1))) {
         unschedule_snapshot_request_ids.push_back(req.snapshot_request_id);
      }
```

needs to be changed to:

```
      if (onetime_snapshot || (recurring_snapshot && height >= (req.end_block_num + 1))) {
         unschedule_snapshot_request_ids.push_back(req.snapshot_request_id);
      }
```

greg7mdp (changes requested):

libraries/chain/snapshot_scheduler.cpp
```diff
@@ -80,15 +65,8 @@ snapshot_scheduler::snapshot_schedule_result snapshot_scheduler::schedule_snapsh
...
-
-   if(sri.end_block_num > 0) {
-      // if "end" is specified, it should be greater then start
-      EOS_ASSERT(sri.start_block_num <= sri.end_block_num, chain::invalid_snapshot_request, "End block number should be greater or equal to start block number");
-      // if also block_spacing specified, check it
-      if(sri.block_spacing > 0) {
-         EOS_ASSERT(sri.start_block_num + sri.block_spacing <= sri.end_block_num, chain::invalid_snapshot_request, "Block spacing exceeds defined by start and end range");
-      }
-   }
+   EOS_ASSERT(sri.start_block_num <= sri.end_block_num, chain::invalid_snapshot_request, "End block number should be greater or equal to start block number");
```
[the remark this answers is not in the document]
766C6164: Its just adds a bit of granularity to error messages so submitters of start > end will get diff message then start + some_huge_spacing > end. Can be changed to 1 message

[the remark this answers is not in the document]
greg7mdp: Yes, fine to leave as is.

libraries/chain/snapshot_scheduler.cpp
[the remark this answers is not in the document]
greg7mdp: Yes, good point. I agree it would be better to not schedule it when it will never succeed.

libraries/chain/include/eosio/chain/snapshot_scheduler.hpp
```diff
@@ -38,10 +38,19 @@ class snapshot_scheduler {
    struct snapshot_request_information {
       uint32_t block_spacing = 0;
       uint32_t start_block_num = 0;
-      uint32_t end_block_num = 0;
+      uint32_t end_block_num = UINT32_MAX - 1;
```
heifner, line 41: Why `-1` ? Seems like `std::numeric<uint32_t>::max()` would be a better value.

libraries/chain/snapshot_scheduler.cpp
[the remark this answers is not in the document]
greg7mdp: Actually, if you make the change I suggested above with `>=`, ` bool onetime_snapshot = (!req.block_spacing) && (height >= req.start_block_num + 1);`, then this suggestion would work. And if reject the ones lower than the current height when submitted, it should maintain the fact that we don't accept requests with start_block_num earlier then current.

tests/test_snapshot_scheduler.cpp
```diff
@@ -104,11 +100,11 @@ BOOST_AUTO_TEST_CASE(snapshot_scheduler_test) {
...
-         snapshot_request_information sri4 = {.start_block_num = 1, .snapshot_description = "One time snapshot on first block"};
-         snapshot_request_information sri5 = {.block_spacing = 10, .snapshot_description = "Recurring every 10 blocks snapshot starting now"};
+         snapshot_request_params sri1 = {.block_spacing = 8, .start_block_num = 1, .end_block_num = 300000, .snapshot_description = "Example of recurring snapshot 1"};
+         snapshot_request_params sri2 = {.block_spacing = 5000, .start_block_num = 100000, .end_block_num = 300000, .snapshot_description = "Example of recurring snapshot 2 that will never happen"};
+         snapshot_request_params sri3 = {.block_spacing = 2, .start_block_num = 0, .end_block_num = 3, .snapshot_description = "Example of recurring snapshot 3 that will expire"};
+         snapshot_request_params sri4 = {.start_block_num = 1, .snapshot_description = "One time snapshot on first block"};
+         snapshot_request_params sri5 = {.block_spacing = 10, .snapshot_description = "Recurring every 10 blocks snapshot starting now"};
```
spoonincode, line 107: I don't see any new unit tests; do we need to add any more tests to expressly cover the example given in #1152? For example, head block is 6, and then a new request to `.block_spacing = 10, .start_block_num = 0, .end_block_num = 0` is made, and then we should have snapshots on blocks 10 & 20, for example.

libraries/chain/include/eosio/chain/snapshot_scheduler.hpp
766C6164: Changed, thank you! (-1 was because I had comparison with end_block_num+1 in the code, changed that as well)

tests/test_snapshot_scheduler.cpp
766C6164: Thank you, done! Extended tests to run a little longer and check for timings on first 2 snapshots

libraries/chain/snapshot_scheduler.cpp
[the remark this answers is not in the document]
766C6164: I looked in detail and putting restriction on scheduling opening a can of worms considering schedule can be changed in json. Scary to remove these checks from the loop. I tried compromise solution - made a better readable and clear condition for deletion and also simplified it :)

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1511,7 +1511,18 @@ void producer_plugin::create_snapshot(producer_plugin::next_function<chain::snap
...
 chain::snapshot_scheduler::snapshot_schedule_result
-producer_plugin::schedule_snapshot(const chain::snapshot_scheduler::snapshot_request_information& sri) {
+producer_plugin::schedule_snapshot(const chain::snapshot_scheduler::snapshot_request_params& srp) {
+   chain::controller& chain = my->chain_plug->chain();
+   const auto head_block_num = chain.head_block_num();
+
+   // missing start/end is set to head block num, missing end to UINT32_MAX
+   chain::snapshot_scheduler::snapshot_request_information sri = {
+      .block_spacing   = srp.block_spacing ? *srp.block_spacing : 0, 
+      .start_block_num = srp.start_block_num ? *srp.start_block_num : head_block_num + 1,
```
greg7mdp, line 1521: Do we need the `+1` here since we compare with `start_block_num + 1` in `on_start_block`?

766C6164: Test (nodeos_snapshot_forked_test.py) starts to fail as it uses call to schedule with no parameters and compares snapshot to synchronous, create_snapshot - older call to make instant snapshot. I guess it will guarantee better that this "asap" snapshot will be executed and not just purged from queue as old. It's really should not be critical, since our goal when start block is not specified to make snapshot reasonably as fast as possible after API request.

greg7mdp: Sounds good, thanks Vlad. I guess if the test fails without this `+1`, it is probably needed and you are right it doesn't hurt to leave it.

greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
766C6164: I re-tested without +1 and both tests (nodeos_snapshot_forked_test.py and nodeos_snapshot_diff_test.py) pass now. But I guess with cluster configuration and manipulation in forked test it can become more flaky and can fail once in a while, when cluster within test is slow.

greg7mdp: Sounds good, I say leave the +1 for safety.

heifner (approved):

Issue #1216, opened 2023-05-27
Get error: x86_64 GS register is not set as expected. EOS VM OC may not run correctly on this platform


matthewdarwin:
Occasionally nodeos will issue error message "x86_64 GS register is not set as expected. EOS VM OC may not run correctly on this platform".  Seems to happen on next restart after kill -9 of nodeos.

Nodeos issues this message after already loading state (in heap mode) and then exits.  Our systemd unit will restart nodeos and then it runs fine.

What's going on here?  Why doesn't it check this before loading all the state?  Wasting cycles loading RAM and then unloading RAM. What is this test doing?

spoonincode:
> Nodeos issues this message after already loading state (in heap mode) and then exits

This message is probably a red herring then as it's only a informational warning and doesn't cause nodeos to exit per se. When you see this error what kind of exit code does nodeos exit with?

matthewdarwin:
Full log files have been sent privately.

![image](https://github.com/AntelopeIO/leap/assets/26171839/674d6fef-d591-428b-8f6e-290cd66c8f15)

```
May 25 18:16:15 waxtest-ship11 nodeos[429697]: info  2023-05-25T18:16:15.366 nodeos    http_plugin.cpp:378           add_async_handler    ] add api url: /v1/net/connect
May 25 18:16:15 waxtest-ship11 nodeos[429697]: info  2023-05-25T18:16:15.366 nodeos    http_plugin.cpp:378           add_async_handler    ] add api url: /v1/net/connections
May 25 18:16:15 waxtest-ship11 nodeos[429697]: info  2023-05-25T18:16:15.366 nodeos    http_plugin.cpp:378           add_async_handler    ] add api url: /v1/net/disconnect
May 25 18:16:15 waxtest-ship11 nodeos[429697]: info  2023-05-25T18:16:15.366 nodeos    http_plugin.cpp:378           add_async_handler    ] add api url: /v1/net/status
May 25 18:16:15 waxtest-ship11 nodeos[429697]: info  2023-05-25T18:16:15.366 nodeos    producer_plugin.cpp:1222      plugin_startup       ] producer plugin:  plugin_startup() begin
May 25 18:16:19 waxtest-ship11 nodeos[429697]: warn  2023-05-25T18:16:19.003 read-0    executor.cpp:142              executor             ] x86_64 GS register is not set as expected. EOS VM OC may not run correctly on this platform
[...]

PR #1221, opened 2023-05-31
[4.0] allow ROtrx threads unlimited time to start

spoonincode:
This change comes about due to investigation of #1216: for some reason in those logs the ROtrx threads were taking a very long time to start and the previous implementation would error out if the threads weren't started within a second.

afaict the only reason to allow a maximum of 1 second previously was simply a proxy to make sure they all started up successfully (a `chain.init_thread_local_data()` that threw would have not run `++num_threads_started`). 

Refactor `named_thread_pool::start()` to simply wait forever for all the threads to complete their init (no matter if `chain.init_thread_local_data()` threw or not); and then if any init _did_ throw, rethrow one of those exceptions inside of `named_thread_pool::start()`.

Unfortunately it's worth noting during my testing of this change I discovered that the primary/likely reason this thread init will fail is uncatchable,
https://github.com/AntelopeIO/leap/blob/db132c5fd44e0b1c492e46e3f51e185cd5c59ed0/libraries/chain/include/eosio/chain/wasm_interface_private.hpp#L64

### DIFF
--- a/libraries/chain/include/eosio/chain/thread_utils.hpp
+++ b/libraries/chain/include/eosio/chain/thread_utils.hpp
@@ -37,9 +37,13 @@ namespace eosio { namespace chain {
 
       /// Spawn threads, can be re-started after stop().
       /// Assumes start()/stop() called from the same thread or externally protected.
+      /// Blocks until all threads are created and completed their init function, or an exception is thrown
+      ///  during thread startup or an init function. Exceptions thrown during these stages are rethrown from start()
+      ///  but some threads might still have been started. Calling stop() after such a failure is safe.
       /// @param num_threads is number of threads spawned
       /// @param on_except is the function to call if io_context throws an exception, is called from thread pool thread.
-      ///                  if an empty function then logs and rethrows exception on thread which will terminate.
+      ///                  if an empty function then logs and rethrows exception on thread which will terminate. Not called
+      ///                  for exceptions during the init function (such exceptions are rethrown from start())
       /// @param init is an optional function to call at startup to initialize any data.
       /// @throw assert_exception if already started and not stopped.
       void start( size_t num_threads, on_except_t on_except, init_t init = {} ) {
@@ -47,9 +51,17 @@ namespace eosio { namespace chain {
          _ioc_work.emplace( boost::asio::make_work_guard( _ioc ) );
          _ioc.restart();
          _thread_pool.reserve( num_threads );
+
+         std::promise<void> start_complete;
+         std::atomic<uint32_t> threads_remaining = num_threads;
+         std::exception_ptr pending_exception;
+         std::mutex pending_exception_mutex;
+
          for( size_t i = 0; i < num_threads; ++i ) {
-            _thread_pool.emplace_back( std::thread( &named_thread_pool::run_thread, this, i, on_except, init )  );
+            _thread_pool.emplace_back( std::thread( &named_thread_pool::run_thread, this, i, on_except, init, std::ref(start_complete),
+                                                    std::ref(threads_remaining), std::ref(pending_exception), std::ref(pending_exception_mutex) ) );
          }
+         start_complete.get_future().get();
       }
 
       /// destroy work guard, stop io_context, join thread_pool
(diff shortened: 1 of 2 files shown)


heifner:
I like the idea of `thread_pool::start()` blocking for us. I would suggest having it block for us, but only if `init` is provided. The idea being that if you provide an `init` to be called you want that completed before returning.

spoonincode:
yeah, I'm liking that too. let me see what I can do..

libraries/chain/include/eosio/chain/thread_utils.hpp
```diff
@@ -37,19 +37,31 @@ namespace eosio { namespace chain {
...
       void start( size_t num_threads, on_except_t on_except, init_t init = {} ) {
          FC_ASSERT( !_ioc_work, "Thread pool already started" );
          _ioc_work.emplace( boost::asio::make_work_guard( _ioc ) );
          _ioc.restart();
          _thread_pool.reserve( num_threads );
+
+         std::promise<void> start_complete;
+         std::atomic<uint32_t> threads_remaining = num_threads;
+         std::exception_ptr pending_exception;
+         std::mutex pending_exception_mutex;
```
spoonincode, line 58: hm, actually, these probably need to be member variables. Because if the `std::thread` ctor throws, `start()` immediately is unwound but some threads could still be running with references to these

heifner: How about adding a try-catch around the `std::thread` construction. Would be nice for these not to be member variables.

libraries/chain/include/eosio/chain/thread_utils.hpp
```diff
@@ -63,16 +83,42 @@ namespace eosio { namespace chain {
...
+            } FC_LOG_AND_RETHROW()
+         }
+         catch( ... ) {
+            std::lock_guard<std::mutex> l( pending_exception_mutex );
+            pending_exception = std::current_exception();
+            decrement_remaining();
+            return;
+         }
+
+         decrement_remaining();
```
heifner, line 119: `pending_exception` is accessed outside the `pending_exception_mutex` here.

libraries/chain/include/eosio/chain/thread_utils.hpp
```diff
@@ -37,19 +37,39 @@ namespace eosio { namespace chain {
...
       void start( size_t num_threads, on_except_t on_except, init_t init = {} ) {
          FC_ASSERT( !_ioc_work, "Thread pool already started" );
          _ioc_work.emplace( boost::asio::make_work_guard( _ioc ) );
          _ioc.restart();
          _thread_pool.reserve( num_threads );
-         for( size_t i = 0; i < num_threads; ++i ) {
-            _thread_pool.emplace_back( std::thread( &named_thread_pool::run_thread, this, i, on_except, init )  );
+
+         std::promise<void> start_complete;
+         std::atomic<uint32_t> threads_remaining = num_threads;
```
heifner, line 56: This could be non-atomic if you protect it always via the mutex. No big deal either way.

heifner (changes requested):

libraries/chain/include/eosio/chain/thread_utils.hpp
spoonincode: The only way `pending_exception` is used in `decrement_remaining()` is if `threads_remaining == 0`, ergo no other threads could be writing to it when it is used here.

heifner: But there is not necessarily a memory barrier requiring change to be visible to other thread, right?

spoonincode: `threads_remaining` is `std::atomic<uint32_t>`, and `operator--()` is defined as equivalent to `fetch_sub(1)-1`, and `fetch_sub()` is `std::memory_order_seq_cst` by default which afaict guarantees the visibility of the change in this case.

heifner (approved):

766C6164 (approved):

PR #1224, opened 2023-06-02
Cleanup pimpl idiom for ship.

greg7mdp:
Normally, for the pimpl idiom, the implementation resides in the private class, in our case state_history_plugin_impl.

Before this change, many functions were defined in state_history_plugin, directly accessing fields from state_history_plugin_impl through the `my` pointer.

This PR does not change what the code does, it should be 100% equivalent to what was there before, however it moves most of the member functions implementation within the state_history_plugin_impl class.

I believe the only behavior change is that handle_sighup() was previously called within the try/catch block of plugin_initialize, and now is called outside a try/catch block, but I don't think this matters.

### DIFF
--- a/plugins/state_history_plugin/state_history_plugin.cpp
+++ b/plugins/state_history_plugin/state_history_plugin.cpp
@@ -76,7 +76,7 @@ public:
    void plugin_initialize(const variables_map& options);
    void plugin_startup();
    void plugin_shutdown();
-   session_manager& session_manager() { return session_mgr; }
+   session_manager& get_session_manager() { return session_mgr; }
 
    static fc::logger& get_logger() { return _log; }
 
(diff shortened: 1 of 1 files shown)


plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -65,12 +68,16 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
...
-
    named_thread_pool<struct ship> thread_pool;
 
    bool  plugin_started = false;
 
+public:
+   void plugin_initialize(const variables_map& options);
+   void plugin_startup();
+   void plugin_shutdown();
+   session_manager& get_session_manager() { return session_mgr; }
```
heifner, line 79: I find it so very annoying that gcc does not like `session_manager`. I have ran into that so many times. I like `session_manager()` over `get_`. Oh well :(

heifner (approved):

plugins/state_history_plugin/state_history_plugin.cpp
greg7mdp: haha yes! I was surprised as well.

766C6164 (approved):

PR #1225, opened 2023-06-02
P2P sync ahead while applying received blocks

### LINKED ISSUE #1072: P2P Improve sync behavior

heifner:
Continue syncing the next `sync-fetch-span` while apply current range of synced blocks. It is prevented from running too far ahead of `head` by `sync-fetch-span`. It does queue up to double the amount of blocks during syncing than before this PR.

Before this PR:
```
[ sync blks A ]             [ sync blks B ]             [ sync blks C ]
    [ apply blocks A.. ... ]    [ apply blocks B.. ... ]    [ apply blocks C.. ... ]
```
After this PR:
```
[ sync blks A ][ sync blks B ][ sync blks C ]       [ sync blks D ]         [ sync blks E ]
    [ apply blocks A.. ... ][ apply blocks B.. ... ][ apply blocks C.. ... ][ apply blocks D.. ... ]
```
On my machine syncing into EOS mainnet, I was able to maintain 100% CPU while syncing.

Includes a fix for excessive handshake messages during syncing.

Resolves #1072

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -244,10 +244,6 @@ namespace eosio {
       void sync_recv_block( const connection_ptr& c, const block_id_type& blk_id, uint32_t blk_num, bool blk_applied );
       void recv_handshake( const connection_ptr& c, const handshake_message& msg, uint32_t nblk_combined_latency );
       void sync_recv_notice( const connection_ptr& c, const notice_message& msg );
-      inline void reset_last_requested_num() {
-         std::lock_guard g(sync_mtx);
-         sync_last_requested_num = 0;
-      }
    };
 
    class dispatch_manager {
@@ -823,6 +819,8 @@ namespace eosio {
       uint32_t                peer_lib_num = 0;
 
       std::atomic<uint32_t>   sync_ordinal{0};
+      // when syncing from a peer, the last block expected of the current range
+      uint32_t                sync_last_requested_block{0};
 
       alignas(hardware_destructive_interference_size)
       std::atomic<uint32_t>   trx_in_progress_size{0};
@@ -1204,7 +1202,7 @@ namespace eosio {
          return;
       if (s == connection_state::connected && curr != connection_state::connecting)
          return;
-      peer_dlog(this, "old connection state ${os} becoming ${ns}", ("os", state_str(curr))("ns", state_str(s)));
+      fc_dlog(logger, "old connection ${id} state ${os} becoming ${ns}", ("id", connection_id)("os", state_str(curr))("ns", state_str(s)));
 
       conn_state = s;
    }
@@ -1316,6 +1314,7 @@ namespace eosio {
       if( !shutdown) my_impl->sync_master->sync_reset_lib_num( self->shared_from_this(), true );
       peer_ilog( self, "closing" );
       self->cancel_wait();
+      self->sync_last_requested_block = 0;
       self->set_state(connection_state::closed);
 
       if( reconnect && !shutdown ) {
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2306,17 +2294,40 @@ namespace eosio {
...
+                  peer_dlog(c, "calling cancel_wait, block ${b}", ("b", blk_num));
+                  c->cancel_wait();
+               } else {
+                  peer_dlog(c, "calling sync_wait, block ${b}", ("b", blk_num));
+                  c->sync_wait();
+               }
+
+               sync_next_expected_num = blk_num + 1;
+            } else {
+               if (sync_last_requested_num != 0 && blk_num >= sync_last_requested_num) {
```
greg7mdp, line 2315: I wonder if this is correct. For example, in the diagram of the PR comment, when we apply the last block of group B, we should compare this `block_num` to the last block of `B`, but won't `sync_last_requested_num` be the last block of `C`?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2306,17 +2294,40 @@ namespace eosio {
...
+            }
+
+            uint32_t head = my_impl->get_chain_head_num();
+            if (head + sync_req_span > sync_last_requested_num) { // don't allow to get too far head (one sync_req_span)
+               if (sync_next_expected_num > sync_last_requested_num && sync_last_requested_num < sync_known_lib_num) {
+                  fc_dlog(logger, "Requesting range ahead, head: ${h} blk_num: ${bn} sync_next_expected_num ${nen} sync_last_requested_num: ${lrn}",
+                          ("h", head)("bn", blk_num)("nen", sync_next_expected_num)("lrn", sync_last_requested_num));
+                  request_next_chunk(std::move(g_sync));
+               }
+            }
```
greg7mdp, line 2329: Probably this block is taking care of my comment above, but then do we need the `else` case above?

greg7mdp:
Why do we use `c->peer_lib_num` at line 1851, and `last_handshake_recv.last_irreversible_block_num` 8 lines below?

heifner:
> Why do we use `c->peer_lib_num` at line 1851, and `last_handshake_recv.last_irreversible_block_num` 8 lines below?

`peer_lib_num` is only valid to access from the connection strand, so can't use it on line 1859.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2306,17 +2294,40 @@ namespace eosio {
...
+
```
greg7mdp, line 2330: I see two issues:
- `>=` at line 2315 should be `>`
- missing `sync_next_expected_num = blk_num + 1;` after like 2319

I think this code could be updated to:

```suggestion
            sync_next_expected_num = blk_num + 1;
            if (!blk_applied) {
               g_sync.unlock();
               if (blk_num >= c->sync_last_requested_block) {
                  peer_dlog(c, "calling cancel_wait, block ${b}", ("b", blk_num));
                  c->cancel_wait();
               } else {
                  peer_dlog(c, "calling sync_wait, block ${b}", ("b", blk_num));
                  c->sync_wait();
               }
            } else {
               if (sync_last_requested_num != 0) {
                  if (blk_num + sync_req_span > sync_last_requested_num &&
                      sync_next_expected_num > sync_last_requested_num &&
                      sync_last_requested_num < sync_known_lib_num) {
                     fc_dlog(logger, "Requesting range ahead, head: ${h} blk_num: ${bn} sync_next_expected_num ${nen} sync_last_requested_num: ${lrn}",
                             ("h", head)("bn", blk_num)("nen", sync_next_expected_num)("lrn", sync_last_requested_num));
                     request_next_chunk(std::move(g_sync));
                  }
               }
            }```

**Review by greg7mdp — CHANGES_REQUESTED** [review]

File: plugins/net_plugin/net_plugin.cpp (hunk 3 above)
Comment (heifner) [reply plugins/net_plugin/net_plugin.cpp:2330]: `sync_recv_block` is called when a block is received (`blk_applied` == `false`) and when a block is applied (`blk_applied` == `true`). 

We should probably have better handling of `sync_last_requested_num == 0` in all cases; I'll fix that up better.

As for the `>=` on line 2315. If it was `>` only then it would not be hit. You will not receive the block after `sync_last_requested_num` unless you request it.

**Comment by greg7mdp** [comment]
Also could we add tese comments if accurate:
```
      uint32_t       sync_known_lib_num{0};      // highest lib num from our currently connected peers
      uint32_t       sync_last_requested_num{0}; // end block number of last requested chunk
      uint32_t       sync_next_expected_num{0};  // next block number we need from peer
      connection_ptr sync_source;                // connection we are currently syncing from
```

plugins/net_plugin/net_plugin.cpp
heifner: You don't want to `sync_next_expected_num = blk_num + 1` when a block is applied because that is behind where you are receiving blocks. So you only increment that when receiving blocks.

I do have what might be nothing more than a safety net in if the applied block is the `sync_last_requested_num` to attempt to request another range.

766C6164 (approved):

PR #1232, opened 2023-06-05
Avoid trying to kill a node twice to prevent orphaned processes.

jgiszczak:
Some tests terminate some nodes but not all.  The `atexit` routine intended to be the final catchall for cleaning up a cluster would fail to terminate all nodes if it tried to terminate a node which had been shut down via `interruptAndVerifyExitStatus` first.  Tolerate terminated nodes and if the global `Debug` flag is set, log double kills. (The log message should be converted to a custom `TRACE` level when the code is converted to the Python `logging` module.)


heifner (approved):

oschwaldp-oci (approved):

ScottBailey:
Accepted.

PR #1234, opened 2023-06-06
Remove fc/string.hpp to_string impls

### LINKED ISSUE #1057: Remove fc/string.hpp to_string impls
Verify `fc/string.hpp` functionality is not used in consensus and change to use `std::to_string` where possible.

greg7mdp:
Resolves #1057.

`fc::to_string` provides the same output as `std::to_string` for integer types, so it can be removed for these types.

However, the version for `double` does not match the `std::to_string` output. See below:

```
Test values: 23.43, 1e-9, 1e40, 1e-40, 123456789.0, std::numeric_limits<double>::quiet_NaN(), std::numeric_limits<double>::infinity()
=====================================================================================================================================

fc::to_string:  23.42999999999999972
std::to_string: 23.430000

fc::to_string:  0.00000000100000000
std::to_string: 0.000000

fc::to_string:  10000000000000000303786028427003666890752.00000000000000000
std::to_string: 10000000000000000303786028427003666890752.000000

fc::to_string:  0.00000000000000000
std::to_string: 0.000000

fc::to_string:  123456789.00000000000000000
std::to_string: 123456789.000000

fc::to_string:  nan
std::to_string: nan

fc::to_string:  inf
std::to_string: inf
```

After replacing the uses of `fc::to_string` for integer types with `std::to_string`, we are left with a single occurence of usage of `fc::to_string(double);` in variant.cpp. Since I am worried about this causing a consensus change, I kept the old implementation as a static function in `variant.cpp`, and I removed all `to_string()` overloads from the fc `string.hpp` header.


libraries/libfc/src/variant.cpp
```diff
@@ -466,18 +466,26 @@ bool  variant::as_bool()const
...
+
 std::string variant::as_string()const
 {
    switch( get_type() )
    {
       case string_type:
           return **reinterpret_cast<const const_string_ptr*>(this);
       case double_type:
-          return to_string(*reinterpret_cast<const double*>(this));
+          return s_fc_to_string(*reinterpret_cast<const double*>(this));
```
heifner, line 484: I don't believe this is used in consensus. Can you attempt to verify that. The difference in generated JSON might cause issues, but if we are planning on moving to a different JSON implementation that difference is likely to be hit anyway.

greg7mdp: > I don't believe this is used in consensus. Can you attempt to verify that. 

How can I check that conclusively?

greg7mdp: Thanks Kevin, will do!

greg7mdp: I verified that this one instance of `variant::as_string()` is never hit with type `double_type` when importing a snapshot, so I assume it means that it is safe to change to std::string.

> Visual code inspection.

I don't quite understand what would cause a change in consensus. I would have thought that only changes in the wasm generated code for contracts or the WASM execution engine could trigger consensus issues, which shows how little I know. So I don't know what to check.

libraries/libfc/src/string.cpp
```diff
@@ -60,28 +60,6 @@ namespace fc  {
...
-  {
-    return boost::lexical_cast<std::string>(d);
-  }
-  std::string to_string( uint16_t d)
-  {
-    return boost::lexical_cast<std::string>(d);
-  }
   std::string trim( const std::string& s )
   {
       return boost::algorithm::trim_copy(s);
```
heifner, line 65: Go ahead and remove `fc::trim` while you are at it.

heifner (changes requested):

libraries/libfc/src/string.cpp
greg7mdp: done

heifner (approved):

PR #1235, opened 2023-06-06
P2P use round trip of time_message to calculate latency

### LINKED ISSUE #1072: P2P Improve sync behavior

heifner:
Address peer review comment: https://github.com/AntelopeIO/leap/pull/1215#discussion_r1210652126

Use round trip time for `time_message` which was already implemented. Since this already did a round trip for previous versions of P2P, no new P2P protocol version is needed. All that was needed was to add the calculation and normalize epoch time. Previous versions of `leap` did not specify duration resolution of `time_since_epoch`. Some clients returned microsecond values and others returned nanosecond values. Included in this PR is a fix so that all `nodeos` going forward will return nanosecond values even if the resolution is only microseconds.

Resolves #1072

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -926,7 +926,7 @@ namespace eosio {
        * day routine and converts to a (at least) 64 bit integer.
        */
       static tstamp get_time() {
-         return std::chrono::system_clock::now().time_since_epoch().count();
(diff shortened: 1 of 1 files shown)


766C6164 (approved):

heifner:
Replaced by #1243. Github somehow became out of sync with this branch. Pushing to the branch did nothing.

PR #1243, opened 2023-06-07
P2P use round trip of time_message to calculate latency

### LINKED ISSUE #1072: P2P Improve sync behavior

heifner:
Address peer review comment: https://github.com/AntelopeIO/leap/pull/1215#discussion_r1210652126

Use round trip time for `time_message` which was already implemented. Since this already did a round trip for previous versions of P2P, no new P2P protocol version is needed. All that was needed was to add the calculation and normalize epoch time. Previous versions of `leap` did not specify duration resolution of `time_since_epoch`. Some clients returned microsecond values and others returned nanosecond values. Included in this PR is a fix so that all `nodeos` going forward will return nanosecond values even if the resolution is only microseconds.

Resolves #1072


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3277,36 +3271,66 @@ namespace eosio {
...
 
       if( msg.org == 0 ) {
          send_time( msg );
          return;  // We don't have enough data to perform the calculation yet.
       }
 
-      double offset = (double(rec - org) + double(msg.xmt - dst)) / 2;
-      double NsecPerUsec{1000};
+      if (org != 0) {
+         int64_t offset = (double(rec - org) + double(msg_xmt - dst)) / 2.0;
```
greg7mdp, line 3327: `msg_xmt - dst` is supposed to be negative, I think. Luckily the tstamp seems to be holding a signed value.

It looks like we assume that the network latency is the same in both directions, which I think may be optimistic for a single round trip. Maybe we should average the offset over time, so it would keep getting more accurate as we exchange messages?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3277,36 +3271,66 @@ namespace eosio {
...
+      int digits = 1;
+      tstamp v = x;
+      while (v/=10)
+         ++digits;
+      // 168 1685 9763 7880 7944 >= 19 is nanoseconds
+      //     1685 9763 7880 7944 >= 16 is microseconds
+      //        1 6859 7637 8807 >= 13 is milliseconds
+      //            16 8597 6378 >= 10 is seconds
+      if (digits >= 19)
+         return x;
```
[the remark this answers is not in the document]
greg7mdp: Actually, why not do comparisons with constants directly, instead of counting the digits and do comparisons with the number of digits?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3277,36 +3271,66 @@ namespace eosio {
...
+      if (digits >= 13)
+         return x*1000*1000;
+      if (digits >= 10)
+         return x*1000*1000*1000;
+      return 0; // unknown or is zero
+   }
+
    void connection::handle_message( const time_message& msg ) {
-      peer_ilog( this, "received time_message" );
+      peer_dlog( this, "received time_message: ${t}", ("t", msg) );
```
linh2931, line 3296: Like the change. I was wondering why time_message deserved info log while looking at log files.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -861,11 +861,11 @@ namespace eosio {
...
-      // Members set from network data
-      tstamp                         org{0};          //!< originate timestamp
-      tstamp                         rec{0};          //!< receive timestamp
-      tstamp                         dst{0};          //!< destination timestamp
-      tstamp                         xmt{0};          //!< transmit timestamp
+      // See NTP protocol. https://datatracker.ietf.org/doc/rfc5905/
+      tstamp                         org{0}; //!< origin timestamp. Time at the client when the request departed for the server.
+      tstamp                         rec{0}; //!< receive timestamp. Time at the server when the request arrived from the client.
+      tstamp                         dst{0}; //!< destination timestamp, Time at the client when the reply arrived from the server.
+      tstamp                         xmt{0}; //!< transmit timestamp, Time at the server when the response left for the client.
```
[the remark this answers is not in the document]
linh2931: Those three letter names org, rec, dst, xmt are confusing when read in isolation. Can they be named more descriptively?

greg7mdp:
Can we rename `net_latency_ns` to `peer_response_time`?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3277,36 +3271,66 @@ namespace eosio {
...
 
       // If the transmit timestamp is zero, the peer is horribly broken.
       if(msg.xmt == 0)
-         return;                 /* invalid timestamp */
+         return; // invalid timestamp
 
-      if(msg.xmt == xmt)
-         return;                 /* duplicate packet */
+      auto msg_xmt = normalize_epoch_to_ns(msg.xmt);
+      auto msg_org = normalize_epoch_to_ns(msg.org);
```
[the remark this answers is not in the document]
heifner: Only new code uses `time_message` for latency calculation. Before this PR, code used time in the handshake message to measure latency.

The `msg.org` can original from a peer. In that case it would not match, so yes there is no reason to normalize it. I'll remove the normalize for `msg.org`.

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
heifner: Done

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
heifner: They match the NTP spec, so there is usefulness in keeping them named as they are in the spec.

plugins/net_plugin/net_plugin.cpp
heifner: The offset is not used outside logging it. I just kept what was already there.

plugins/net_plugin/net_plugin.cpp
heifner: Me too!

heifner:
> Can we rename `net_latency_ns` to `peer_response_time`?

And then not divide by 2?

greg7mdp:
> And then not divide by 2?

sure, makes sense!

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
greg7mdp: let's still do the swap :-)

linh2931 (approved):

greg7mdp:
Kevin, are we doing the handshake (`send_time`) only  once per connection?

heifner:
> Kevin, are we doing the handshake (`send_time`) only once per connection?

No, there is a heartbeat that does it every 5 seconds.

heifner:
> Can we rename `net_latency_ns` to `peer_response_time`?

How about `peer_ping_time`?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1456,17 +1457,21 @@ namespace eosio {
...
    void connection::send_time() {
-      time_message xpkt;
-      xpkt.org = rec;
-      xpkt.rec = dst;
-      xpkt.xmt = get_time();
-      org = xpkt.xmt;
-      enqueue(xpkt);
+      if (org == 0) { // do not send if there is already a time loop in progress
+         time_message xpkt;
+         xpkt.org = rec;
```
greg7mdp, line 1468: I don't understand why we don't do: `xpkt.org  = get_time();` here?

greg7mdp:
> How about `peer_ping_time`?

Sounds good to me. or `peer_ping_time_ns` if you prefer.

plugins/net_plugin/net_plugin.cpp
heifner: It might be clearer to make it `xpkt.org = 0` which is what `rec` is here. The idea is this is the start of the ping.

greg7mdp: I am confused, why would 0 be the start time of the ping, and not the current time?

heifner: Because of this: https://github.com/AntelopeIO/leap/pull/1243/files#diff-fee9397188f7093ef7c6c15c4b268a4866d1c8e9518697ca9edfd2ea9233e5fcL3299-L3301

greg7mdp: OK, but how do we know that `rec == 0` when we call `send_time()` [here](https://github.com/AntelopeIO/leap/blob/aa25f11201cb4350eca3396878fba2c469adafb6/plugins/net_plugin/net_plugin.cpp#L1460-L1461).

If the intent to have `xpkt.org = 0` in that send_time message, why use `rec`?

greg7mdp: I'm not quite sure why we use `rec` as well. Why not?

```
   void connection::send_time() {
      if (org == 0) { // do not send if there is already a time loop in progress
         time_message xpkt;
         xpkt.org = 0; // initiating a ping. Actual org time is in xpkt.xmt 
         xpkt.rec = 0;
         xpkt.xmt = get_time();
         org      = xpkt.xmt;
         peer_dlog(this, "send init time_message: ${t}", ("t", xpkt));
         enqueue(xpkt);
      }
   }
```

greg7mdp: and what about this for the other one:

```
   void connection::send_time(const time_message& msg) {
      assert(org != 0 && org == msg.xmt);
      time_message xpkt { msg.xmt, msg.dst, get_time(), 0 };
      peer_dlog( this, "send time_message: ${t}, org: ${o}", ("t", xpkt)("o", org) );
      enqueue(xpkt);
   }
```

greg7mdp: OK, I give up, the way it is coded is beyond my brain capacity to fully reason on it. If it was up to me, I'd completely rewrite it, but for now I'll just trust you and approve this PR.

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
heifner: I agree it could be clearer. For example, `rec` isn't even needed as a member variable. I'll take a swing at simplifying it a bit.

greg7mdp: The code that I think is hard to reason with is `handle_message(const time_message&)`. if it was me I would have two fully distinct  cases, either we receive a ping message with `msg.org == 0` (initiated from peer), or with `msg.org != 0` (a response to out ping.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1464,22 +1464,25 @@ namespace eosio {
...
-         time_message xpkt;
-         xpkt.org = rec;
-         xpkt.rec = dst;
-         xpkt.xmt = get_time();
-         org      = xpkt.xmt;
+         org = get_time();
+         // xpkt.org == 0 means we are initiating a ping. Actual origin time is in xpkt.xmt.
+         time_message xpkt{
+            .org = 0,
+            .rec = dst,
```
greg7mdp, line 1471: Much nicer, thanks!
Why `.rec = dst` instead of 0?

greg7mdp (approved):

greg7mdp (approved):

PR #1249, opened 2023-06-07
Use retry-num-blocks where possible to avoid polling in tests

jgiszczak:
Enable `--transaction-retry-max-storage-size-gb` on all non-producer nodes in regression tests and make use of `--retry-num-blocks` in `createAccount`, `createInitializeAccount`, `transferFunds`, `regproducer`, `voteproducer`,  and `delegatebw` with at least one attempt to `waitForTransBlock`.  In addition, `undelegatebw` is now tested in a regression test.  It was previously only exercised in unit tests.

Further, `exitOnError` only happens when `silentErrors` is `False`.  There is now at least one attempt to use the above APIs with `silentErrors=False`, which required exposing it in the higher level Test Harness API.


tests/TestHarness/transactions.py
```diff
@@ -37,9 +39,10 @@ def createAccount(self, account, creatorAccount, stakedDeposit=1000, waitForTran
         """Create account and return creation transactions. Return transaction json object.
         waitForTransBlock: wait on creation transaction id to appear in a block."""
         signStr = NodeosQueries.sign_str(sign, [ creatorAccount.activePublicKey ])
+        retryStr = f"--retry-num-blocks {self.retry_num_blocks_default}" if waitForTransBlock else ""
```
oschwaldp-oci, line 42: Why not add `retry_num_blocks` to the `createAccount` argument list that defaults to the `retry_num_blocks_default` if not configured directly for the method call?  That way specific tests could control directly how much to retry?

tests/TestHarness/transactions.py
```diff
@@ -51,7 +54,7 @@ def createAccount(self, account, creatorAccount, stakedDeposit=1000, waitForTran
             self.trackCmdTransaction(trans)
             transId=NodeosQueries.getTransId(trans)
 
-        return self.waitForTransBlockIfNeeded(trans, waitForTransBlock, exitOnError=exitOnError)
+        return trans
```
oschwaldp-oci, line 57: In the case where the `waitForTransBlockIfNeeded` failed, we used to get log messages and exit if `exitOnError` was set:

```
Utils.cmdError("transaction with id %s never made it into a block" % (transId))
Utils.errorExit("Failed to find transaction with id %s in a block before timeout" % (transId))
```
Seems like these features are gone with the new implementation and may want to consider including them if it is common practice to search for those in logs when something goes bad? And no longer exiting with error code.

ScottBailey (approved):
f-strings make code so much clearer, I think.

tests/TestHarness/transactions.py
jgiszczak: Good advice, and it should definitely be done that way when it's possible to do so.

tests/TestHarness/transactions.py
jgiszczak: `exitOnError` was being handled by `processCleosCommand` instead of `waitForTransBlockIfNeeded`.  Unfortunately it doesn't do anything if `silentErrors=True`.  This question prompted rewriting this entire PR, so thanks for that.  (No sarcasm.  The tests are better now.)

jgiszczak:
@ScottBailey Unfortunately only one f-string has survived, and really I should revert that change too.  Or take this opportunity to do a wholesale conversion of `transactions.py`, since this turned out to be only a test improvement PR.

tests/TestHarness/transactions.py
jgiszczak: Added.

tests/TestHarness/transactions.py
jgiszczak: Modified some tests to disable `silentErrors` to be sure this code path is used at least once.

jgiszczak:
@ScottBailey The f-strings are back!  This PR is now working as intended. (It was missing a nodeos command line option to enable retry.)  Please re-review.

ScottBailey (approved):

oschwaldp-oci (approved):

tests/TestHarness/launcher.py
```diff
@@ -521,9 +523,34 @@ def construct_command_line(self, instance: nodeDefinition):
...
+                '--contract-whitelist', '--contract-blacklist', '--action-blacklist', '--key-blacklist',
+                '--sender-bypass-whiteblacklist', '--trusted-producer',
+                # http_plugin
+                '--http-alias',
+                # net_plugin
+                '--p2p-peer-address', '--p2p-auto-bp-peer', '--peer-key', '--peer-private-key',
+                # producer_plugin
+                '--producer-name', '--signature-provider', '--greylist-account', '--disable-subjective-account-billing',
+                # trace_api_plugin
+                '--trace-rpc-abi']
```
oschwaldp-oci, line 545: This seems a little kludgy and maintenance/error prone. But I guess it is okay.

oschwaldp-oci (approved):

heifner (approved):

tests/TestHarness/launcher.py
jgiszczak: It's not perfect but I thought parsing the C++ for all the plugins was asking a bit much.  The options don't change too often and when they do the vast majority are not composeable so it shouldn't be too much of a pain.

If boost wasn't so wretched about allowing option overrides this code wouldn't exist...

linh2931 (approved):

Issue #1239, opened 2023-06-06
`output-dir` seems to be required for `leap-util block-log extract-blocks` but isn't enforced


spoonincode:
`leap-util block-log extract-blocks` doesn't document `output-dir` as required and doesn't require it to be specified to run. But not giving this parameter will result in a cryptic `filesystem error: cannot create directories: Invalid argument` error. `output-dir` is indeed required.

PR #1250, opened 2023-06-08
Made output-dir required option for blocklog subcommands

766C6164:
This PR addresses https://github.com/AntelopeIO/leap/issues/1239, in particular fact that multiple sub-commands under "blocklog" produce output in output-dir and it has no default value, so it should be mandatory to request this value from user.


greg7mdp (approved):

greg7mdp (changes requested):
just saw this should go into 3.2.

PR #1252, opened 2023-06-08
fix `i256` key type conversion by initialize buffer before usage

learnforpractice:
fix `i256` key type conversion by initialize buffer before usage

`export_bits` doesn't fill every byte in the buffer, need to ensure that 'buffer' is initialized before usage:

https://github.com/boostorg/multiprecision/blob/5f81a78cdb38ea35f98ef7a655d2535773c1261b/include/boost/multiprecision/cpp_int/import_export.hpp#L236

### DIFF
--- a/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
+++ b/plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
@@ -941,7 +941,7 @@ public:
             // the following will convert the input to array of 2 uint128_t in little endian, i.e. 50f0fa8360ec998f4bb65b00c86282f5 fb54b91bfed2fe7fe39a92d999d002c5
             // which is the format used by secondary index
             chain::key256_t k;
-            uint8_t buffer[32];
+            uint8_t buffer[32] = {};
             boost::multiprecision::export_bits(v, buffer, 8, false);
             memcpy(&k[0], buffer + 16, 16);
             memcpy(&k[1], buffer, 16);


plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
```diff
@@ -942,6 +942,7 @@ class read_write : public api_base {
             // which is the format used by secondary index
             chain::key256_t k;
             uint8_t buffer[32];
+            memset(buffer, 0, 32);
             boost::multiprecision::export_bits(v, buffer, 8, false);
```
greg7mdp, line 946: Thanks for the fix, good catch! It would definitely work fine.
A thought, since `export_bits` returns the pointer to the byte just after the last one written, we can do a small optimization by zeroing only the remaining bytes. 

```suggestion
            auto out = boost::multiprecision::export_bits(v, buffer, 8, false);
            memset(out, 0, 32 - (out - buffer)); // zero out remaining bytes
```

greg7mdp (changes requested):

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
spoonincode: My hunch is the original is faster (although personally I'd prefer the `uint8_t buffer[32] = {}` style) because the call to `memset()` is completely removed (and there isn't a read dependency on the register).

https://godbolt.org/z/axGa6e4bK

greg7mdp: I have no issue with @spoonincode 's suggestion. Hard to know which version would be fastest without testing, but it probably doesn't matter in this instance.

greg7mdp (approved):

plugins/chain_plugin/include/eosio/chain_plugin/chain_plugin.hpp
heifner: I'll add my vote for the more c++ style of `uint8_t buffer[32] = {}`

learnforpractice:
Updated. Thanks for the suggestion!

greg7mdp (approved):

heifner (approved):

greg7mdp:
@learnforpractice you should be able to just push the merge button to merge your PR.

spoonincode:
thanks @learnforpractice

Issue #1104, opened 2023-04-27
test libtester usage as part of CI


spoonincode:
Leap's libtester is used for native contract unit testing of contracts. For example, https://github.com/AntelopeIO/reference-contracts uses libtester for testing eosio.system and eosio.token.

libtester has long been a very persnickety library that inadvertently breaks causing failures downstream when contract unit tests attempt to use it. Breakage typically takes the form of something being wrong in the cmake files (very manually crafted, sometimes unknown to contributors, and easy to miss), or part of the libtester interface unexpectedly changing (some changes are inevitable, and why the contract build files have a version limiter, but sometimes it's simply not obvious a change breaks the interface for code in a separate repo).

To complicate matters further, libtester can potentially be installed four different ways,
1. Used in-place of part of leap's build tree
2. Used after a `make dev-install` which installs the files to `CMAKE_INSTALL_PREFIX`
3. Used after a `DESTDIR= make dev-install` which installs the files to `DESTDIR/CMAKE_INSTALL_PREFIX`
4. Used from a `leap-dev.deb` install, which ought be similar to 2

We do have some test coverage around 4 _in a different repo_ since the CI of reference-contracts installs `leap-dev.deb` and uses it for testing. This other repo's CI has been an indirect way of getting a libtester sniff test for a leap branch in question. It's also worth mentioning that DUNE uses `leap-dev.deb` so that's another data point of it working (and another reason to make sure it keeps working!).

So as of this moment, 4 works. 1 probably works, as at least a few ENF developers are known to use that method. 2 likely works since it's similar to 4. 3 probably has never worked due to how the cmake is structured.

We really need to do a sanity check of libtester for leap CI runs. A sanity check might be to build reference-contracts via a libtester installed via methods 1, 2, and 4, for example.

766C6164:
Something related to be aware of - there is an effort to add C++ 20 support that requires newer boost (1.82) that is currently brought as submodule. To make it work, dependencies on boost labs in makefiles are specified explicitly. If there is a missing dependency, there is a risk that a header from systemwide installed Boost with different version would be used leading to hard to catch issues: https://github.com/AntelopeIO/leap/pull/1142
Good way to detect missing deps is to NOT have libboost-all-dev installed in a docker image as it is suggested in comments in this PR

PR #1259, opened 2023-06-08
test libtester usage as part of CI

oschwaldp-oci:
Resolves: [1104](https://github.com/AntelopeIO/leap/issues/1104)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -260,7 +260,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu20, ubuntu22]
-    runs-on:  ["self-hosted", "enf-x86-lowtier"] # not sure if this warrants a better machine, but start here
+    runs-on: ["self-hosted", "enf-x86-beefy"] # not sure if this warrants a different machine, but start here due to dev-install step
     container: ${{fromJSON(needs.d.outputs.p)[matrix.platform].image}}
     steps:
       - name: Clone leap
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -254,13 +254,14 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu20, ubuntu22]
-    runs-on:  ["self-hosted", "enf-x86-lowtier"] # not sure if this warrants a better machine, but start here
+    runs-on: ["self-hosted", "enf-x86-beefy"]
     container: ${{fromJSON(needs.d.outputs.p)[matrix.platform].image}}
     steps:
-      - name: Download leap builddir
-        uses: actions/download-artifact@v3
+      - name: checkout leap
+        uses: actions/checkout@v3
         with:
-          name: ${{matrix.platform}}-build
+          submodules: recursive
+          path: leap
       - name: checkout cdt
         uses: actions/checkout@v3
         with:
@@ -276,12 +277,16 @@ jobs:
         run: |
           pwd
           ls -l
-      - name: Extract leap build and make dev-install
+      - name: leap build and make dev-install
         run: |
-          zstdcat build.tar.zst | tar x
+          # https://github.com/actions/runner/issues/2033
+          chown -R $(id -u):$(id -g) $PWD
+          cd leap
+          cmake -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -GNinja
+          cmake --build build
           cd build
           ninja dev-install
-          cd ..
+          cd ../..
       - name: make install cdt
         run: |
           cd cdt
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -115,6 +115,7 @@ jobs:
         with:
           name: leap-dev-ubuntu20-amd64
           path: build/leap-dev*.deb
+
   tests:
     name: Tests
     needs: [d, Build]
@@ -234,6 +235,60 @@ jobs:
         with:
           repository: AntelopeIO/reference-contracts
           path: reference-contracts
+      - name: Build & Test reference-contracts
+        run: |
+          cmake -S reference-contracts -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=On
+          cmake --build build -- -j $(nproc)
+          cd build/tests
+          ctest --output-on-failure -j $(nproc)
+
+  libtester-make-dev-install-test:
+    name: libtester make dev-install test
+    needs: [d, Build]
+    if: always() && needs.Build.result == 'success'
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: [ubuntu20, ubuntu22]
+    runs-on:  ["self-hosted", "enf-x86-lowtier"] # not sure if this warrants a better machine, but start here
+    container: ${{fromJSON(needs.d.outputs.p)[matrix.platform].image}}
+    steps:
+      - name: Download leap builddir
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{matrix.platform}}-build
+          path: leap
+      - name: checkout cdt
+        uses: actions/checkout@v3
+        with:
+          repository: AntelopeIO/cdt
+          submodules: recursive
+          path: cdt
+      - name: checkout reference-contracts
+        uses: actions/checkout@v3
+        with:
+          repository: AntelopeIO/reference-contracts
+          path: reference-contracts
+      - name: Check directory structure
+        run: |
+          pwd
+          ls -l
+      - name: Extract leap build and make dev-install
+        run: |
+          cd leap
+          zstdcat build.tar.zst | tar x
+          cd build
+          make dev-install
+          cd ../..
+      - name: make install cdt
+        run: |
+          cd cdt
+          mkdir build
+          cd build
+          cmake -DCMAKE_BUILD_TYPE=Release ..
+          make -j $(nproc)
+          make install
+          cd ../..
       - name: Build & Test reference-contracts
         run: |
           cmake -S reference-contracts -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=On
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -318,12 +318,6 @@ jobs:
         with:
           name: ${{matrix.platform}}-build
           path: leap
-      - name: checkout cdt
-        uses: actions/checkout@v3
-        with:
-          repository: AntelopeIO/cdt
-          submodules: recursive
-          path: cdt
       - name: checkout reference-contracts
         uses: actions/checkout@v3
         with:
@@ -346,26 +340,22 @@ jobs:
           cd ..
           echo "leap_DIR=$PWD/leap/build/lib/cmake/leap" >> "$GITHUB_ENV"
           echo leap_DIR = $leap_DIR
-      - name: Build cdt
+      - name: Download cdt
+        uses: AntelopeIO/asset-artifact-download-action@v2
+        with:
+          owner: AntelopeIO
+          repo: cdt
+          file: 'cdt_.*amd64.deb'
+          target: main
+          artifact-name: cdt_ubuntu_package_amd64
+          token: ${{github.token}}
+      - name: Install cdt Packages
         run: |
-          pwd
-          ls -l
-          cd cdt
-          pwd
-          mkdir build
-          cd build
-          pwd
-          apt-get update && apt-get install -y pkg-config libcurl4-gnutls-dev
-          export leap_DIR="${{ env.leap_DIR }}"
-          echo leap_DIR=$leap_DIR
-          cmake -DCMAKE_BUILD_TYPE=Release ..
-          make -j $(nproc)
-          cd ../..
-          echo "cdt_DIR=$PWD/cdt/build/lib/cmake/cdt" >> "$GITHUB_ENV"
-          echo cdt_DIR = $cdt_DIR
+          apt install -y ./*.deb
+          rm ./*.deb
       - name: Build & Test reference-contracts
         run: |
-          cmake -S reference-contracts -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=On -Dleap_DIR="${{ env.leap_DIR }}" -Dcdt_DIR="${{ env.cdt_DIR }}"
+          cmake -S reference-contracts -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=On -Dleap_DIR="${{ env.leap_DIR }}"
           cmake --build build -- -j $(nproc)
           cd build/tests
           ctest --output-on-failure -j $(nproc)
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -227,8 +227,8 @@ jobs:
           name: leap-dev-ubuntu20-amd64
       - name: Install cdt and leap-dev Packages
         run: |
-          sudo apt install ./*.deb
-          sudo apt-get install cmake
+          apt install ./*.deb
+          apt-get install cmake
           rm ./*.deb
       - name: checkout reference-contracts
         uses: actions/checkout@v3
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -88,8 +88,12 @@ jobs:
     name: Build leap-dev package
     needs: [d, Build]
     if: always() && needs.Build.result == 'success'
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: [ubuntu20, ubuntu22]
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -201,6 +201,46 @@ jobs:
...
+      - name: checkout reference-contracts
+        uses: actions/checkout@v3
+        with:
+          repository: AntelopeIO/reference-contracts
+          path: src
+      - name: Build & Test reference-contracts
+        run: |
+          cmake -S src -B build -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=On
+          cmake --build build -- -j $(nproc)
+          ctest --test-dir build/tests --output-on-failure -j $(nproc)
```
spoonincode, line 242: It seems like the tests aren't running
https://github.com/AntelopeIO/leap/actions/runs/5222841684/jobs/9429031167?pr=1259#step:7:500
What's going on here is that the cmake version included in ubuntu 20.04 doesn't have the `--test-dir` parameter and for some crazy insane reason ctest doesn't error out on unknown command line arguments.

The reason this works over in reference-contracts is because it just uses github's runner image which has [a whole bunch of stuff already built in](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md) including... cmake 3.26

So will need to change this line to `cd build/tests && cmake --output-on-failure -j $(nproc)` or such.

(it also probably means over in reference-contracts [this line](https://github.com/AntelopeIO/reference-contracts/blob/76197b4bc60d8dc91a5d65ecdbf0f785e982e279/.github/workflows/build.yaml#L88) is unnecessary)

.github/workflows/build.yaml
```diff
@@ -201,6 +201,46 @@ jobs:
...
+          artifact-name: cdt_ubuntu_package_amd64
+          token: ${{github.token}}
+      - name: Download leap-dev
+        uses: actions/download-artifact@v3
+        with:
+          name: leap-dev-ubuntu20-amd64
+      - name: Install cdt and leap-dev Packages
+        run: |
+          apt install -y ./*.deb
+          apt-get -y install cmake
```
spoonincode, line 231: you should be able to remove this line. since you're using the cached base image, it already has cmake
https://github.com/AntelopeIO/leap/blob/651d21f2a3a988a019ff58745d125464c310c201/.cicd/platforms/ubuntu20.Dockerfile#L6

.github/workflows/build.yaml
```diff
@@ -201,6 +201,46 @@ jobs:
...
+  libtester-deb-install-test:
+    name: libtester deb install test
+    needs: [d, Build, dev-package]
+    if: always() && needs.dev-package.result == 'success'
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: [ubuntu20] # Assuming want to do both ubuntu20 and ubuntu22 at some point so setting up with matrix
+    runs-on:  ["self-hosted", "enf-x86-lowtier"] # not sure if this warrants a better machine, but start here
+    container: ${{fromJSON(needs.d.outputs.p)['ubuntu20'].image}}
```
spoonincode, line 213: Since you're using the cached based image, this test won't really check that leap-dev.deb's dependency packages , which are _manually_ set here and thus prone to getting messed up,
https://github.com/AntelopeIO/leap/blob/651d21f2a3a988a019ff58745d125464c310c201/package.cmake#L64
are correct/exhaustive because the cached base image already installed those for building leap itself.

**This is probably an okay limitation for now though.**

But if you're curious to explore a solution, I think you could resolve this by just using ubuntu's base image -- `container: ubuntu:focal` or such, but this comes at the cost of being more difficult to use the matrix set up you have for futureproofing. My hunch is you might be able to do something like this since you'd only have 20 & 22:
```yaml
matrix:
container: ${{matrix.platform == 'ubuntu20' && 'ubuntu:focal' || 'ubuntu:jammy'}}

name: leap-dev-${{matrix.platform}-amd64
```

 (you would also need to leave cmake install here too https://github.com/AntelopeIO/leap/pull/1259/files#r1224554620)

.github/workflows/build.yaml
oschwaldp-oci: Thanks.  I was digging into why the tests weren't running and this was a great head start. I appreciate it.

.github/workflows/build.yaml
oschwaldp-oci: Good point.

.github/workflows/build.yaml
oschwaldp-oci: I honestly had this thought, but wasn't sure if we were ready to move toward the base container images and build up from those for this task or if that was planned for a later effort.  With wanting to get something up and testing/working, I chose to start with the cached images knowing it wouldn't be too difficult to then pivot toward the other.

.github/workflows/build.yaml
```diff
@@ -201,6 +206,188 @@ jobs:
...
+      - name: checkout reference-contracts
+        uses: actions/checkout@v3
+        with:
+          repository: AntelopeIO/reference-contracts
+          path: reference-contracts
+      - name: Check directory structure
+        run: |
+          pwd
+          ls -l
+      - name: leap build and make dev-install
```
spoonincode, line 280: we should try to figure out a way to do this without entirely building leap a second & third time.

For the build-dir case it should be easy -- we just package up the builddir from one job and extract it on the next job. We already package up the builddir as `ubuntu20-build` and `ubuntu22-build`.

The `dev-install` case is going to be trickier because the files will be co-mingled in `/usr`, and we don't really want to change `CMAKE_INSTALL_PREFIX` across the board. One option would be to perform the `dev-install` and then use `install_manifest_dev.txt` file as input to `tar -T`. Another option may be to enable `CPACK_ARCHIVE_COMPONENT_INSTALL` so a `leap-dev.tar.gz` is created.

oschwaldp-oci: Yeah I was planning to come back to optimize once I had all three working. Thanks for the suggestions on where/how to start that process.  I'll definitely be making another pass once I get the 3 scenarios passing.

.github/workflows/build.yaml
```diff
@@ -201,6 +206,188 @@ jobs:
...
+          # https://github.com/actions/runner/issues/2033
+          chown -R $(id -u):$(id -g) $PWD
+          cd leap
+          cmake -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -GNinja
+          cmake --build build
+          cd build
+          ninja install
+          ninja dev-install
+          cd ../..
+      - name: make install cdt
```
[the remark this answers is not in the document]
oschwaldp-oci: Done.

.github/workflows/build.yaml
oschwaldp-oci: Addressed.

.github/workflows/build.yaml
oschwaldp-oci: Done

.github/workflows/build.yaml
oschwaldp-oci: Was able to get this to work with a clone of the `leap` repo along with the unpacked `builddir` then using the `install` and `dev-install` targets.

.github/workflows/build.yaml
```diff
@@ -201,6 +206,155 @@ jobs:
...
+        with:
+          name: ${{matrix.platform}}-build
+      - name: Extract leap build
+        run: |
+          zstdcat build.tar.zst | tar x
+      - name: leap dev-install
+        run: |
+          cd build
+          ninja install
+          ninja dev-install
```
spoonincode, line 279: From the logs, it looks as if the entire source tree is getting rebuilt here, which is why you need a high performing runner to get it done in time.

Needs to be refactored somehow to not perform the entire build all over again here.

oschwaldp-oci: Good catch.  I refactored to `cmake` targets and that looks to have done what I was expecting instead of rebuilding.  Just noticed the `dev-install` target does a build first, which I wasn't expecting.  Also explains why this test was taking so long -- as you noted.

```
cmake --install build
cmake --install build --component dev
```
addressed: https://github.com/AntelopeIO/leap/pull/1259/commits/f069ef42df2860ea749e796d97ccc6c693c18ee5

.github/workflows/build.yaml
```diff
@@ -201,6 +206,153 @@ jobs:
...
+          target: main
+          artifact-name: cdt_ubuntu_package_amd64
+          token: ${{github.token}}
+      - name: Install cdt Packages
+        run: |
+          apt install -y ./*.deb
+          rm ./*.deb
+      - name: Build & Test reference-contracts
+        run: |
+          cmake -S reference-contracts -B reference-contracts/build -DCMAKE_BUILD_TYPE=Release -DBUILD_TESTS=On -Dleap_DIR="${{ env.leap_DIR }}" -DCMAKE_PREFIX_PATH="${{ env.LEAP_BUILD_DIR }}"
```
spoonincode, line 351: We're setting up discovery of leap dir several different ways here -- `leap_DIR` environment variable, `leap_DIR` cache variable, `CMAKE_PREFIX_PATH` cache variable. This suggests uncertainly on how it ought to work. Let's try to pick just one single method. Personally I like the idea of using an environment variable since it allows easier commonality for the build step should we eventually choose to matrix them all together.

.github/workflows/build.yaml
```diff
@@ -201,6 +206,154 @@ jobs:
...
+          target: main
+          artifact-name: cdt_ubuntu_package_amd64
+          token: ${{github.token}}
+      - name: Install cdt Packages
+        run: |
+          apt install -y ./*.deb
+          rm ./*.deb
+      - name: Build & Test reference-contracts
+        run: |
+          export leap_DIR="${{ env.leap_DIR }}"
```
spoonincode, line 351: Isn't this just setting the value to what it already has, from here
https://github.com/AntelopeIO/leap/blob/06373d2cfc59bdefb362e9af9259713c963306ac/.github/workflows/build.yaml#L334-L335
my guess is we only need that one single line in its current location
https://github.com/AntelopeIO/leap/blob/06373d2cfc59bdefb362e9af9259713c963306ac/.github/workflows/build.yaml#L334

.github/workflows/build.yaml
oschwaldp-oci: Addressed: [aa5ef2a](https://github.com/AntelopeIO/leap/pull/1259/commits/aa5ef2a6688f39e291c06869a31716c6d4301cfe)

.github/workflows/build.yaml
```diff
@@ -201,6 +206,152 @@ jobs:
...
 
+  libtester-deb-install-test:
+    name: libtester deb install test
+    needs: [d, Build, dev-package]
+    if: always() && needs.dev-package.result == 'success'
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: [ubuntu20, ubuntu22]
+    runs-on:  ["self-hosted", "enf-x86-lowtier"] # not sure if this warrants a better machine, but start here
```
spoonincode, line 217: consider cleaning up these WIP-style comments

spoonincode (approved):
Looks good. In the future it would be interesting to explore:
* does something like a `midtier` runner make them fast enough to become a required test
* folding all 3 tests in to a matrix -- they share a lot of commonality: all of them clone & build contracts the same way, all of them install cdt the same way. But I'm not sure there is an easy way to tackle the difference in container image right now.
* configuration knobs to select release/branch of cdt & reference-contracts

spoonincode (changes requested):
> configuration knobs to select release/branch of cdt & reference-contracts

actually, maybe we ought to pick stable versions/releases of cdt & reference-contracts here instead of simply `main`. leap CI failing because main of cdt becomes incompatible with main of reference-contracts doesn't sound very fun

oschwaldp-oci:
> > configuration knobs to select release/branch of cdt & reference-contracts
> 
> actually, maybe we ought to pick stable versions/releases of cdt & reference-contracts here instead of simply `main`. leap CI failing because main of cdt becomes incompatible with main of reference-contracts doesn't sound very fun

Yeah, can do that. We had a brief discussion about that trying to pose reasons why `main` vs. a `release` and wondered about the case where breaking changes were going into one or the other `leap`, `cdt`, or `reference-contracts` and using `main` would enable that development to move forward together in lock-step.  But definitely see the pros and cons or either solution here.

.github/workflows/build.yaml
oschwaldp-oci: Addressed: [cb61f1c](https://github.com/AntelopeIO/leap/pull/1259/commits/cb61f1caa85d6be7deaa2184dba3b792193ace6c)

spoonincode:
If we don't fix the versions for cdt & contracts, then we run the risk of someone making an innocuous change (like to documentation) and then their CI run fails because cdt main & contracts main have just started to have some problem. And that's really lousy.

Since the purpose of the test is detect goof ups of libtester and its cmake files before they are approved in to the leap repo, it seems reasonable to fix cdt & contract versions to eliminate those other variables from interfering with the primary purpose.

You're right that it won't give visibility in to when main of cdt & reference contracts fail to build for some reason. But it doesn't seem it's the responsibility of leap's CI -- and certainly this workflow -- to detect such a situation. Possibly we should have some other workflow in an independent repo nightly test a more exhaustive combination of
* libtester main
* cdt main
* reference-contracts main
* eos-system-contracts main
* eos-evm main

But again, it just doesn't seem it's the job of leap's repo to be doing that.

oschwaldp-oci:
> If we don't fix the versions for cdt & contracts, then we run the risk of someone making an innocuous change (like to documentation) and then their CI run fails because cdt main & contracts main have just started to have some problem. And that's really lousy.
> 
> Since the purpose of the test is detect goof ups of libtester and its cmake files before they are approved in to the leap repo, it seems reasonable to fix cdt & contract versions to eliminate those other variables from interfering with the primary purpose.
> 
> You're right that it won't give visibility in to when main of cdt & reference contracts fail to build for some reason. But it doesn't seem it's the responsibility of leap's CI -- and certainly this workflow -- to detect such a situation. Possibly we should have some other workflow in an independent repo nightly test a more exhaustive combination of
> 
> * libtester main
> * cdt main
> * reference-contracts main
> * eos-system-contracts main
> * eos-evm main
> 
> But again, it just doesn't seem it's the job of leap's repo to be doing that.

Completely agree. Good argument.  I'll work to target release branches for `cdt` and `reference-contracts` instead of their `main` branches.

spoonincode:
I hesitate to "cross the streams", but since `reference-contracts` intentionally has no tags nor releases, we probablly should use `eos-system-contracts` instead. I think once we set cdt to, say, `4.0`, and eos-system-contracts to, say, `3.1.1` this looks good to go

.github/workflows/build.yaml
```diff
@@ -201,11 +206,158 @@ jobs:
...
   all-passing:
     name: All Required Tests Passed
-    needs: [dev-package, tests, np-tests]
+    needs: [dev-package, tests, np-tests, libtester-build-tree-test, libtester-make-dev-install-test, libtester-deb-install-test]
     if: always()
     runs-on: ubuntu-latest
     steps:
-      - if: needs.dev-package.result != 'success' || needs.tests.result != 'success' || needs.np-tests.result != 'success'
+      - if: needs.dev-package.result != 'success' || needs.tests.result != 'success' || needs.np-tests.result != 'success' ||
+            needs.libtester-build-tree-test.result != 'success' || needs.libtester-make-dev-install-test.result != 'success' || needs.libtester-deb-install-test.result != 'success'
```
spoonincode, line 362: I know I postulated this change in a previous commit, but from what I can tell:
* `midtier` didn't really make the build all that much faster than `lowtier`; the cost difference is very minor between low vs mid, so it's not something I'm going to fret about, but
* it seems like requiring libtester jobs is further pushing the required CI run time out. We're moving in the wrong direction here,
  * 3.2: Tests+NP tests, ~4m
  * main: Tests+NP tests, ~5.5m
  * this PR: Tests+NP tests+libtester, ~7m it seems?
  * and then #1142 would likely push a full Build+Tests+NP tests+libtester out another ~4-6m :disappointed:

I'm a bit in conflict what to do here... it's an important test but we need to starting holding the line on these times somehow.

oschwaldp-oci: So, one of the issues is that there is a new critical path through the tests which adds one step to the dependency tree.  The deb install test is dependent on the leap-dev package.  The long pole right now, is `NP Tests` which take ~5m21s to complete.  In that same time you now have `leap-dev` ~1m55s and then `libtester deb install` at ~4m37s.  That gives you a net increase of ~1m11s in run time and then spin-up time for the jobs.  We could look at only making the success dependent on the libtester scenarios that can be run in parallel.

oschwaldp-oci: image

spoonincode (approved):
nice, looks good

heifner (approved):

PR #1274, opened 2023-06-12
Made snapshot test more robust in case of asap scheduling

766C6164:
This PR addresses instability of snapshot scheduler test and specifically it's part that executes "asap" scheduling with no start block.

### DIFF
--- a/tests/test_snapshot_scheduler.cpp
+++ b/tests/test_snapshot_scheduler.cpp
@@ -83,25 +83,30 @@ BOOST_AUTO_TEST_CASE(snapshot_scheduler_test) {
             if (!pp->get_snapshot_requests().snapshot_requests.empty()) {
                const auto& snapshot_requests = pp->get_snapshot_requests().snapshot_requests;
 
-               auto validate_snapshot_request = [&](uint32_t sid, uint32_t block_num, uint32_t spacing = 0) {                  
+               auto validate_snapshot_request = [&](uint32_t sid, uint32_t block_num, uint32_t spacing = 0, bool fuzzy_start = false) {                  
                   auto it = find_if(snapshot_requests.begin(), snapshot_requests.end(), [sid](const snapshot_scheduler::snapshot_schedule_information& obj) {return obj.snapshot_request_id == sid;});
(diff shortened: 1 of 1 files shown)


linh2931 (commented):
Should this also be fixed in 4.0?

766C6164:
> Should this also be fixed in 4.0? If so, do 4.0 first. Then merge into main.

I think you right! It seems like failures started to happen in main, but 4.0 shares the same code, so I'll put it there and merge to main

766C6164:
> > Should this also be fixed in 4.0? If so, do 4.0 first. Then merge into main.
> 
> I think you right! It seems like failures started to happen in main, but 4.0 shares the same code, so I'll put it there and merge to main

I got confused, sorry @linh2931. In 4.0 no changes needed, since this particular test, which was flaky, was added here:
https://github.com/AntelopeIO/leap/pull/1220
and only for main. This was not a bug, but behavior change

tests/test_snapshot_scheduler.cpp
```diff
@@ -83,25 +83,30 @@ BOOST_AUTO_TEST_CASE(snapshot_scheduler_test) {
...
                      if (pending.size()==1) {
+                        // pending snapshot block number
                         auto pbn = pending.begin()->head_block_num;
-                        pbn = spacing ?  (spacing + (pbn%spacing)) : pbn;
-                        // if snapshot scheduled with empty start_block_num depending on the timing
-                        // it can be scheduled either for block_num or block_num+1
-                        BOOST_CHECK(block_num==pbn || ((block_num+1)==pbn));                       
+
+                        // first pending snapshot
+                        auto ps_start = spacing ?  (spacing + (pbn%spacing)) : pbn;
```
linh2931, line 95: As `spacing` is used in modular division `pbn%spacing`, it is more readable 
```
auto ps_start = (spacing != 0) ? (spacing + (pbn%spacing)) : pbn;
```

766C6164: Done, thank you!

linh2931 (approved):

greg7mdp (approved):

Issue #1197, opened 2023-05-18
get_transaction_id not working from curl but working from cleos


linh2931:
ovi, [May 17, 2023 at 4:31:19 AM (May 17, 2023 at 4:31:59 AM)]:

ok I figured the get_transaction_id requires a transaction json with all its basic fields filled up, therefore I guessed it is the same json which expects in “cleos get transaction_id”; so I tried that, and for cleos it works but for get_transaction_id fails see below:

this succeeds:

dune -- cleos -u https://eos.greymass.com get transaction_id '{"actions":[{"account":"hezdshrynage","authorization":[{"actor":"eosminer1111","permission":"active"}],"data":{"entry":774831,"miner":"eosminer1111","nonce":139429},"hex_data":"104208574d273155a520020000000000afd20b0000000000","name":"mine2"}],"context_free_actions":[],"context_free_data":[],"delay_sec":0,"expiration":"2023-02-01T23:16:42","max_cpu_usage_ms":0,"max_net_usage_words":0,"ref_block_num":63709,"ref_block_prefix":2456934748,"signatures":["SIG_K1_KfxwURnR8zHNjc6RqwNcCQxQGQWeDL3iJVkS6rDYsy35ESbbGjLtR4vWZ5qTjqZNDZoizeAkmk2jxEVz7ZYbVyqnPc1mdb"]}'

outputs valid trx id:

c613b52998278394efbd606f40ced4641a1e01f468e20baad010f407235ebef7

this fails:

curl --request POST \
--url https://eos.greymass.com/v1/chain/get_transaction_id \
--header 'content-type: application/json' \
--data '{"actions":[{"account":"hezdshrynage","authorization":[{"actor":"eosminer1111","permission":"active"}],"data":{"entry":774831,"miner":"eosminer1111","nonce":139429},"hex_data":"104208574d273155a520020000000000afd20b0000000000","name":"mine2"}],"context_free_actions":[],"context_free_data":[],"delay_sec":0,"expiration":"2023-02-01T23:16:42","max_cpu_usage_ms":0,"max_net_usage_words":0,"ref_block_num":63709,"ref_block_prefix":2456934748,"signatures":["SIG_K1_KfxwURnR8zHNjc6RqwNcCQxQGQWeDL3iJVkS6rDYsy35ESbbGjLtR4vWZ5qTjqZNDZoizeAkmk2jxEVz7ZYbVyqnPc1mdb"]}'

outputs parsing error:

[...]

PR #1280, opened 2023-06-13
[4.0] if actions.data & actions.hex_data provided, use the hex_data in api call get_transaction_id

766C6164:
It appears that logic for handling transaction as part of get_transaction_id request was different in cleos and chain_api_plugin. Resolves https://github.com/AntelopeIO/leap/issues/1197

This PR makes a change to a logic of get_transaction_id handler as such:
- if actions.data & actions.hex_data provided, use the hex_data, test added
- if actions not present or not an array, will throw "Invalid transaction", test added
- if transaction is not parse-able, will throw "Invalid transaction" 

caveat here that there is a few lines of code duplication between chain_api_plugin and cleos, but this is probably safest way to address initial issue without dangers of changing abi serializer

### DIFF
--- a/plugins/chain_api_plugin/chain_api_plugin.cpp
+++ b/plugins/chain_api_plugin/chain_api_plugin.cpp
@@ -64,7 +64,14 @@ parse_params<chain_apis::read_only::get_transaction_id_params, http_params_types
                      maction_vo["data"] = maction_vo["hex_data"];
                      action_vo = maction_vo;
                      vo = mvo;
+                  } else if( action_vo.contains( "data" ) ) {
+                     if( !action_vo["data"].is_string() ) {
+                        EOS_THROW(chain::invalid_http_request, "Request supports only un-exploded 'data' (hex form)");
+                     }
                   }
+               }
+               else {
+                  EOS_THROW(chain::invalid_http_request, "Transaction contains invalid or empty action");
                }
             } 
          }
--- a/plugins/chain_api_plugin/chain_api_plugin.cpp
+++ b/plugins/chain_api_plugin/chain_api_plugin.cpp
@@ -41,6 +41,45 @@ parse_params<chain_apis::read_only::get_transaction_status_params, http_params_t
    }
 }
 
+// if actions.data & actions.hex_data provided, use the hex_data since only currently support unexploded data
+template<>
+chain_apis::read_only::get_transaction_id_params
+parse_params<chain_apis::read_only::get_transaction_id_params, http_params_types::params_required>(const std::string& body) {
+   if (body.empty()) {
+      EOS_THROW(chain::invalid_http_request, "A Request body is required");
+   }
+
+   try {
+      fc::variant trx_var =  fc::json::from_string( body );
+      if( trx_var.is_object() ) {
+         fc::variant_object& vo = trx_var.get_object();
+         if( vo.contains("actions") && vo["actions"].is_array() ) {
+            fc::mutable_variant_object mvo = vo;
+            fc::variants& action_variants = mvo["actions"].get_array();
+            for( auto& action_v : action_variants ) {
+               if( action_v.is_object() ) {
+                 fc::variant_object& action_vo = action_v.get_object();
+                  if( action_vo.contains( "data" ) && action_vo.contains( "hex_data" ) ) {
+                     fc::mutable_variant_object maction_vo = action_vo;
+                     maction_vo["data"] = maction_vo["hex_data"];
+                     action_vo = maction_vo;
+                     vo = mvo;
+                  }
+               }
+            } 
+         }
+         else {
+            throw false;
+         }
+      }
+      auto trx = trx_var.as<chain_apis::read_only::get_transaction_id_params>();
+      if( trx.id() == transaction().id() ) throw false; 
+      return trx;
+   } catch( ... ) {
+      EOS_THROW(chain::invalid_http_request, "Invalid transaction");
+   }
+}
+
 #define CALL_WITH_400(api_name, api_handle, api_namespace, call_name, http_response_code, params_type) \
 {std::string("/v1/" #api_name "/" #call_name), \
    [api_handle](string&&, string&& body, url_response_callback&& cb) mutable { \
--- a/plugins/chain_api_plugin/chain_api_plugin.cpp
+++ b/plugins/chain_api_plugin/chain_api_plugin.cpp
@@ -80,11 +80,7 @@ parse_params<chain_apis::read_only::get_transaction_id_params, http_params_types
          EOS_THROW(chain::invalid_http_request, "Invalid transaction object");
       }
       return trx;
-   } catch(chain::invalid_http_request & e) {
-      throw;
-   } catch( ... ) {
-      EOS_THROW(chain::invalid_http_request, "Invalid transaction");
-   }
+   } EOS_RETHROW_EXCEPTIONS(chain::invalid_http_request, "Invalid transaction");
 }
 
 #define CALL_WITH_400(api_name, api_handle, api_namespace, call_name, http_response_code, params_type) \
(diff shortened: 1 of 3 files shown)


plugins/chain_api_plugin/chain_api_plugin.cpp
```diff
@@ -41,6 +41,45 @@ parse_params<chain_apis::read_only::get_transaction_status_params, http_params_t
    }
 }
 
+// if actions.data & actions.hex_data provided, use the hex_data since only currently support unexploded data
```
heifner, line 44: - As part of this effort, I think we should support exploded `data`.
- As part of this effort, I think we should provide optional support for exploded `data` in cleos. For cleos, this should be optional as a user may want to prevent cleos from communicating with nodeos. Could support the option of providing abi. See `--unpack-action-data` option `--abi-file`.
- To increase security, if both "data" and "hex_data" are provide then we should verify they are equivalent.
- Please add some tests.

plugins/chain_api_plugin/chain_api_plugin.cpp
```diff
@@ -41,6 +41,45 @@ parse_params<chain_apis::read_only::get_transaction_status_params, http_params_t
...
+                     fc::mutable_variant_object maction_vo = action_vo;
+                     maction_vo["data"] = maction_vo["hex_data"];
+                     action_vo = maction_vo;
+                     vo = mvo;
+                  }
+               }
+            } 
+         }
+         else {
+            throw false;
```
heifner, line 72: A better error here would be good.

heifner (changes requested):

plugins/chain_api_plugin/chain_api_plugin.cpp
766C6164: Done, thank you

plugins/chain_api_plugin/chain_api_plugin.cpp
766C6164: I added a separate issue to include exploded data handling improvements:
https://github.com/AntelopeIO/leap/issues/1282

plugins/chain_api_plugin/chain_api_plugin.cpp
```diff
@@ -41,6 +41,50 @@ parse_params<chain_apis::read_only::get_transaction_status_params, http_params_t
...
+                     fc::mutable_variant_object maction_vo = action_vo;
+                     maction_vo["data"] = maction_vo["hex_data"];
+                     action_vo = maction_vo;
+                     vo = mvo;
+                  }
+               }
+            } 
+         }
+         else {
+            EOS_THROW(chain::invalid_http_request, "Transaction actions are missing or invalid");
```
heifner, line 72: This throw is caught with the `catch(...)` below, so it is actually never reported anywhere.

plugins/chain_api_plugin/chain_api_plugin.cpp
```diff
@@ -41,6 +41,50 @@ parse_params<chain_apis::read_only::get_transaction_status_params, http_params_t
...
+         }
+      }
+      else {
+         throw false; // invalid transaction
+      }
+      auto trx = trx_var.as<chain_apis::read_only::get_transaction_id_params>();
+      if( trx.id() == transaction().id() ) {
+         throw false; // invalid transaction
```
heifner, line 80: Better errors to the caller for these `throw false` would be good.

plugins/chain_api_plugin/chain_api_plugin.cpp
766C6164: Doh! Fixed. Thank you!

plugins/chain_api_plugin/chain_api_plugin.cpp
```diff
@@ -41,6 +41,48 @@ parse_params<chain_apis::read_only::get_transaction_status_params, http_params_t
    }
 }
 
+// if actions.data & actions.hex_data provided, use the hex_data since only currently support unexploded data
+template<>
+chain_apis::read_only::get_transaction_id_params
+parse_params<chain_apis::read_only::get_transaction_id_params, http_params_types::params_required>(const std::string& body) {
+   if (body.empty()) {
+      EOS_THROW(chain::invalid_http_request, "A Request body is required");
+   }
```
greg7mdp, line 50: Is there a reason why this is not in the `try` block below?

plugins/chain_api_plugin/chain_api_plugin.cpp
```diff
@@ -41,6 +41,48 @@ parse_params<chain_apis::read_only::get_transaction_status_params, http_params_t
...
+
+   try {
+      fc::variant trx_var =  fc::json::from_string( body );
+      if( trx_var.is_object() ) {
+         fc::variant_object& vo = trx_var.get_object();
+         if( vo.contains("actions") && vo["actions"].is_array() ) {
+            fc::mutable_variant_object mvo = vo;
+            fc::variants& action_variants = mvo["actions"].get_array();
+            for( auto& action_v : action_variants ) {
+               if( action_v.is_object() ) {
```
greg7mdp, line 60: cleos returns an error when one of the provided actions is not an object. Should we do the same instead of just skipping  it?

plugins/chain_api_plugin/chain_api_plugin.cpp
766C6164: Mainly to follow similar pattern of parse_params functions in chain_api_pplugin, http_plugin where check for empty body done like that, etc and it's not going to throw

plugins/chain_api_plugin/chain_api_plugin.cpp
766C6164: Thank you, added (2 more exceptions from cleos)! That will definitely make api replies much more useful.

plugins/chain_api_plugin/chain_api_plugin.cpp
greg7mdp: OK, it is just that the other `EOS_THROW` below would be caught by the `EOS_RETHROW_EXCEPTIONS` and annotated with `Invalid transaction`, but not this first one. If this is what we want then I'm fine with that.

766C6164: They should be still similar:
`
curl -X POST http://127.0.0.1:8888/v1/chain/get_transaction_id --header 'content-type: application/json' --data '{}'

{"code":400,"message":"Invalid Request","error":{"code":3200006,"name":"invalid_http_request","what":"invalid http request","details":[{"message":"Transaction actions are missing or invalid","file":"chain_api_plugin.cpp","line_number":79,"method":"parse_params"}]}}

curl -X POST http://127.0.0.1:8888/v1/chain/get_transaction_id --header 'content-type: application/json' --data ''

{"code":400,"message":"Invalid Request","error":{"code":3200006,"name":"invalid_http_request","what":"invalid http request","details":[{"message":"A Request body is required","file":"chain_api_plugin.cpp","line_number":49,"method":"parse_params"}]}}
`

greg7mdp: OK, looks good, I though the message from `EOS_RETHROW_EXCEPTIONS` would be added, but it doesn't seem to be the case.

greg7mdp (approved):

plugins/chain_api_plugin/chain_api_plugin.cpp
```diff
@@ -41,6 +41,55 @@ parse_params<chain_apis::read_only::get_transaction_status_params, http_params_t
...
+// if actions.data & actions.hex_data provided, use the hex_data since only currently support unexploded data
+template<>
+chain_apis::read_only::get_transaction_id_params
+parse_params<chain_apis::read_only::get_transaction_id_params, http_params_types::params_required>(const std::string& body) {
+   if (body.empty()) {
+      EOS_THROW(chain::invalid_http_request, "A Request body is required");
+   }
+
+   try {
+      fc::variant trx_var =  fc::json::from_string( body );
```
greg7mdp, line 53: Is it an issue that we don't use a yield function to limit the time spent in the deserialization?

PR #1285, opened 2023-06-14
[4.0] Fix for loading snapshot with empty block log

### LINKED ISSUE #1228: Starting node from a snapshot fails if the node has no blocks
We just noticed an issue with Leap v4.0.1 when we created a new Telos Testnet BP node. Starting it from a snapshot (without any blocks in the blocks folder) resulted in this error. 

telegram-cloud-photo-size-5-6329874667602556377-y

We tried different snapshots, and different config.ini settings & start parameters (e.g. --delete-all-blocks) but nothing helped.

We then downgraded to Leap v3.2.3 and nodeos started from the same snapshot without blocks without this error.

heifner:
When running with an empty block log, `--block-log-retain-blocks 0`, it was not possible to load a snapshot. Implement `empty_block_log::reset` to correctly handle provided `first_block_num`.

Resolves #1228

### DIFF
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -496,19 +496,21 @@ namespace eosio { namespace chain {
 
       /// Would remove pre-existing block log and index, never write blocks into disk.
       struct empty_block_log final : block_log_impl {
+         uint32_t first_block_number = std::numeric_limits<uint32_t>::max();
+
          explicit empty_block_log(const bfs::path& log_dir) {
             fc::remove(log_dir / "blocks.log");
             fc::remove(log_dir / "blocks.index");
          }
 
-         uint32_t first_block_num() final { return head ? head->block_num() : 1; }
+         uint32_t first_block_num() final { return head ? head->block_num() : first_block_number; }
          void append(const signed_block_ptr& b, const block_id_type& id, const std::vector<char>& packed_block) final {
             update_head(b, id);
          }
 
          uint64_t get_block_pos(uint32_t block_num) final { return block_log::npos; }
          void reset(const genesis_state& gs, const signed_block_ptr& first_block) final { update_head(first_block); }
-         void reset(const chain_id_type& chain_id, uint32_t first_block_num) final {}
+         void reset(const chain_id_type& chain_id, uint32_t first_block_num) final { first_block_number = first_block_num; }
          void flush() final {}
 
          signed_block_ptr read_block_by_num(uint32_t block_num) final { return {}; };
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -423,7 +423,8 @@ struct controller_impl {
          EOS_ASSERT( root_id == log_head_id, fork_database_exception, "fork database root does not match block log head" );
       } else {
          EOS_ASSERT( fork_db.root()->block_num == lib_num, fork_database_exception,
-                     "empty block log expects the first appended block to build off a block that is not the fork database root. root block number: ${block_num}, lib: ${lib_num}", ("block_num", fork_db.root()->block_num) ("lib_num", lib_num) );
+                     "The first block of an empty block log should be the block after fork database root. fork_db.root block number: ${bn}, lib: ${lib_num}",
+                     ("bn", fork_db.root()->block_num)("lib_num", lib_num) );
       }
 
       const auto fork_head = fork_db_head();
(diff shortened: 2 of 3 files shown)


libraries/chain/block_log.cpp
```diff
@@ -496,19 +496,21 @@ namespace eosio { namespace chain {
 
       /// Would remove pre-existing block log and index, never write blocks into disk.
       struct empty_block_log final : block_log_impl {
+         uint32_t first_block_number = std::numeric_limits<uint32_t>::max();
```
linh2931, line 499: Why not initialize to `1`?

libraries/chain/controller.cpp
```diff
@@ -423,7 +423,8 @@ struct controller_impl {
          EOS_ASSERT( root_id == log_head_id, fork_database_exception, "fork database root does not match block log head" );
       } else {
          EOS_ASSERT( fork_db.root()->block_num == lib_num, fork_database_exception,
-                     "empty block log expects the first appended block to build off a block that is not the fork database root. root block number: ${block_num}, lib: ${lib_num}", ("block_num", fork_db.root()->block_num) ("lib_num", lib_num) );
+                     "The first block of an empty block log should be the block after fork database root. fork_db.root block number: ${bn}, lib: ${lib_num}",
```
linh2931, line 426: This error message can be further improved improved. The user might get confused how an empty block log has a first block (or any block).

libraries/chain/block_log.cpp
heifner: Because `first_block_num()` is suppose to report the first available block in the block log. The `net_plugin` for example will report that to peers so they can sync from that block number. So an empty block log should only report what it has which is just the current head or nothing.

libraries/chain/block_log.cpp
```diff
@@ -496,19 +496,21 @@ namespace eosio { namespace chain {
...
          }
 
-         uint32_t first_block_num() final { return head ? head->block_num() : 1; }
+         uint32_t first_block_num() final { return head ? head->block_num() : first_block_number; }
          void append(const signed_block_ptr& b, const block_id_type& id, const std::vector<char>& packed_block) final {
             update_head(b, id);
          }
 
          uint64_t get_block_pos(uint32_t block_num) final { return block_log::npos; }
          void reset(const genesis_state& gs, const signed_block_ptr& first_block) final { update_head(first_block); }
```
greg7mdp, line 512: Do we need to update `first_block_number` here as well?

heifner: No, because `head` will be set, so `first_block_num()` can get it from `head`.

libraries/chain/controller.cpp
linh2931: Not sure if this is better. How about 
```
"The first block ($lib_num}) when starting with an empty block log should be the block after fork database root (${bn}).",
```

heifner: Works for me, thanks!

libraries/chain/controller.cpp
```diff
@@ -423,7 +423,8 @@ struct controller_impl {
          EOS_ASSERT( root_id == log_head_id, fork_database_exception, "fork database root does not match block log head" );
       } else {
          EOS_ASSERT( fork_db.root()->block_num == lib_num, fork_database_exception,
```
greg7mdp, line 425: Do we hit this `else` block *only* when we use the `empty_block_log`?

heifner: Yes. Otherwise blog would have a head.

greg7mdp: Hum, so if the block log doesn't have a head, it means it doesn't have any block, right. So it doesn't have a first block either. I think logically `blog.first_block_num()` should return an empty `std::optional<uint32_t>`. And because we don't have a `lib`, `lib_num` should also be an empty `std::optional`. Here it seems like we are jumping through hoops to create a value matching what `fork_db.root()->block_num` will be, right? Or am I still missing something?

heifner: I think you make an excellent point. We could certainly look into making a better invariant for block log around first block in a separate PR to main. However, as the software currently stands, it supports the idea of a first block even when there is nothing in the block log. That is true of basic block log and pruned block log as well. Changing that here would very likely break a number of things.

linh2931 (approved):

greg7mdp (approved):

PR #1288, opened 2023-06-14
Pinned Builds manual dispatch workflow in CI

oschwaldp-oci:
Resolves: #735


oschwaldp-oci:
Targeting 3.1 instead. Superseded by: https://github.com/AntelopeIO/leap/pull/1291

PR #1290, opened 2023-06-14
Create initial workflow file for Pinned Builds.

oschwaldp-oci:
Creating a shell of the initial pinned_build.yaml workflow to be merged quickly allowing development of actual workflow in follow on branch.

Need to have workflow file merged before github will allow triggering it in a development branch.

Issue #1101, opened 2023-04-27
attach binary packages (.deb, etc) as release assets automatically


spoonincode:
When a release is created, we can automatically grab the built packages from the appropriate CI workflow (the workflow of the git commit of the tag for the release) and attach those as release artifacts. This will reduce a little bit of burden during release for someone to go manually fish those out, and of course eliminates potential error in fishing out the wrong package to attach to a release.

Depends on #735

PR #1291, opened 2023-06-14
[3.1] Pinned Builds manual dispatch workflow in CI

oschwaldp-oci:
Pinned builds run in GitHub Actions

Resolves: https://github.com/AntelopeIO/leap/issues/735

### DIFF
--- a/.github/workflows/pinned_build.yaml
+++ b/.github/workflows/pinned_build.yaml
@@ -43,6 +43,8 @@ jobs:
           chown -R $(id -u):$(id -g) $PWD
           ./scripts/install_deps.sh
       - name: Build Pinned Build
+        env:
+          LEAP_PINNED_INSTALL_PREFIX: /usr
         run: |
           ./scripts/pinned_build.sh deps build "$(nproc)"
       - name: Upload package
--- a/.github/workflows/pinned_build.yaml
+++ b/.github/workflows/pinned_build.yaml
@@ -53,6 +53,8 @@ jobs:
           pwd
           ls -l
           cd build
+          ls -l
+          ls | grep -E 'leap-[^d]+*.deb'
           echo "PINNED_BUILD_ARTIFACT="$(ls | grep -E 'leap-[^d]+*.deb')"" >> "$GIHUB_ENV"
           cd ..
       - name: Upload package
--- a/.github/workflows/pinned_build.yaml
+++ b/.github/workflows/pinned_build.yaml
@@ -22,10 +22,14 @@ jobs:
     runs-on: ubuntu-latest
     container: ${{ matrix.platform == 'ubuntu18' && 'ubuntu:bionic' || matrix.platform == 'ubuntu20' && 'ubuntu:focal' || 'ubuntu:jammy' }}
     steps:
+      - name: Conditionally update git repo
+        if: ${{ matrix.platform == "ubuntu18" }}
+        run: |
+          add-apt-repository ppa:git-core/ppa
       - name: Update and Install git
         run: |
-          apt-get update
-          apt-get install -y git
+          apt update
+          apt install -y git
           git --version
       - name: Clone leap
         uses: actions/checkout@v3
--- a/.github/workflows/pinned_build.yaml
+++ b/.github/workflows/pinned_build.yaml
@@ -1,0 +1,28 @@
+name: "Pinned Build"
+
+on:
+  workflow_dispatch:
+
+permissions:
+  packages: read
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  d:
+    name: Discover Platforms
+    runs-on: ubuntu-latest
+    outputs:
+      missing-platforms: ${{steps.discover.outputs.missing-platforms}}
+      p: ${{steps.discover.outputs.platforms}}
+    steps:
+      - name: Discover Platforms
+        id: discover
+        uses: AntelopeIO/discover-platforms-action@v1
+        with:
+          platform-file: .cicd/platforms.json
+          password: ${{secrets.GITHUB_TOKEN}}
+          package-name: builders
--- a/.github/workflows/pinned_build.yaml
+++ b/.github/workflows/pinned_build.yaml
@@ -25,11 +25,13 @@ jobs:
       - name: Conditionally update git repo
         if: ${{ matrix.platform == 'ubuntu18' }}
         run: |
+          apt-get install software-properties-common
+          apt-get update
           add-apt-repository ppa:git-core/ppa
       - name: Update and Install git
         run: |
-          apt update
-          apt install -y git
+          apt-get update
+          apt-get install -y git
           git --version
       - name: Clone leap
         uses: actions/checkout@v3
--- a/.github/workflows/pinned_build.yaml
+++ b/.github/workflows/pinned_build.yaml
@@ -18,7 +18,6 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu18, ubuntu20, ubuntu22]
-    # runs-on: ["self-hosted", "enf-x86-beefy"] # not sure if this warrants a different machine, but start here due to building
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/pinned_build.yaml
```diff
@@ -26,3 +26,45 @@ jobs:
...
+          ./scripts/pinned_build.sh deps build "$(nproc)"
+          cd build
+          cpack
+          ls -l
+          cd ..
+      - name: Upload package
+        uses: actions/upload-artifact@v3
+        with:
+          name: leap-${{matrix.platform}}-amd64
+          path: build/leap*.deb
```
spoonincode, line 66: `leap-dev.deb` is broken for pinned builds, so I'd suggest only uploading `leap.deb` as part of this artifact somehow.

I'm also unsure whether or not we should have the artifact name explicitly call out pinned (e.g. `leap-${{matrix.platform}}-pinned-amd64`).. but I think it's okay like this for now

.github/workflows/pinned_build.yaml
```diff
@@ -26,3 +26,45 @@ jobs:
           platform-file: .cicd/platforms.json
           password: ${{secrets.GITHUB_TOKEN}}
           package-name: builders
+
+  Build:
+    name: Build
+    needs: [d]
```
spoonincode, line 32: we're not using the cached build environments in this job, so you can completely remove the "Discover Platforms" job

oschwaldp-oci: Yeah, I put that in simply to not have a completely empty initial workflow file in https://github.com/AntelopeIO/leap/pull/1290. I'll remove it here.

oschwaldp-oci: Addressed: [7c10660](https://github.com/AntelopeIO/leap/pull/1291/commits/7c106605cc4ac138c3604c9fc659913a9e28cb4c)

.github/workflows/pinned_build.yaml
```diff
@@ -26,3 +26,45 @@ jobs:
...
+      - name: Install dependencies
+        run: |
+          # https://github.com/actions/runner/issues/2033
+          chown -R $(id -u):$(id -g) $PWD
+          ./scripts/install_deps.sh
+      - name: Build Pinned Build
+        run: |
+          ./scripts/pinned_build.sh deps build "$(nproc)"
+          cd build
+          cpack
```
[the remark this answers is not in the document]
oschwaldp-oci: Addressed: [da6a068](https://github.com/AntelopeIO/leap/pull/1291/commits/da6a0686390f1ac773b7eee833c795aba343e118)

.github/workflows/pinned_build.yaml
```diff
@@ -26,3 +26,45 @@ jobs:
...
+  Build:
+    name: Build
+    needs: [d]
+    if: always() && needs.d.result == 'success'
+    strategy:
+      fail-fast: false
+      matrix:
+        platform: [ubuntu18, ubuntu20, ubuntu22]
+    runs-on: ["self-hosted", "enf-x86-beefy"] # not sure if this warrants a different machine, but start here due to building
+    container: ${{ matrix.platform == 'ubuntu18' && 'ubuntu:bionic' || matrix.platform == 'ubuntu20' && 'ubuntu:focal' || 'ubuntu:jammy' }}
```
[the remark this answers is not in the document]
oschwaldp-oci: Looks like this is working from initial testing of the workflow, I'm seeing `bionic`, `focal`, and `jammy`.

.github/workflows/pinned_build.yaml
```diff
@@ -0,0 +1,63 @@
...
+          chown -R $(id -u):$(id -g) $PWD
+          ./scripts/install_deps.sh
+      - name: Build Pinned Build
+        run: |
+          ./scripts/pinned_build.sh deps build "$(nproc)"
+      - name: Check and select artifact
+        run: |
+          cd build
+          echo "PINNED_BUILD_ARTIFACT="$(ls | grep -E 'leap-[^d]+*.deb')"" >> "$GITHUB_ENV"
+          echo "PINNED_BUILD_NAME="$(ls | grep -E 'leap-[^d]+*.deb' | sed -e 's/-x86_64.deb/-pinned-amd64/')"" >> "$GITHUB_ENV"
```
spoonincode, line 53: What's the need for these shenanigans? I know the 3.1's .deb files are named a little differently which makes the difference between `leap` & `leap-dev` trickier. What stops us from something like this for 3.1,
```yaml
      - name: Upload package
        uses: actions/upload-artifact@v3
        with:
          name: leap-${{matrix.platform}}-pinned-amd64
          path: build/leap-3*.deb
```
and then we can make it more generic on 3.2+ with
```yaml
      - name: Upload package
        uses: actions/upload-artifact@v3
        with:
          name: leap-${{matrix.platform}}-pinned-amd64
          path: build/leap_*.deb
```

oschwaldp-oci: Those shenanigans allow for selecting the `.deb` file a little more generically without having to specify the leading major version number (The wildcard matching for `upload-artifact` didn't seem to support full regex) and then crafting upload names that matched more closely what I saw in the artifacts on our releases which would include the version numbers, but then add in the `pinned` identifier:

Release assets are named as such:
image

The pinned build upload assets are named:
image

spoonincode: I don't think it's necessary for the artifact name to have the version number in it, and quite possibly we _shouldn't_ do that. It makes things harder when we want to make use of a build that isn't a release; a perfect example of that is in the libtester workflow where you simply grab the `cdt.deb` from main: what version number would you expect to find in that situation? It's difficult to know ahead of time, and then coupled with `asset-artifact-download-action` not supporting wildcards for the `artifact-name` name currently, it seems to break the use case of another repo's workflow being able to find and use any pinned build.

Admittedly once we do #1101 we're going to have to iterate over the artifacts in a workflow and look for some pattern.. but that pattern likely wouldn't need the version numbers still.

> selecting the .deb file a little more generically without having to specify the leading major version number (The wildcard matching for upload-artifact didn't seem to support full regex

afaik this is strictly a 3.1 problem, and shouldn't be a problem 3.2+, so I don't see any real problem with hardcoding the "3" in the 3.1 branch's workflow. Would the yaml samples in my above comment work?

oschwaldp-oci: Sure. That's fine. I can do it that way and follow that naming convention if you'd prefer.  Would you like the artifact to have the `.deb` file extension in the name?

spoonincode (changes requested):

.github/workflows/pinned_build.yaml
```diff
@@ -0,0 +1,56 @@
...
+          ./scripts/pinned_build.sh deps build "$(nproc)"
+      - name: Upload package
+        uses: actions/upload-artifact@v3
+        with:
+          name: leap-${{matrix.platform}}-pinned-amd64
+          path: build/leap-3*.deb
+      - name: Run Parallel Tests
+        run: |
+          cd build
+          ctest --output-on-failure -j $(nproc) -LE "(nonparallelizable_tests|long_running_tests)" --timeout 420
```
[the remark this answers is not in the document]
spoonincode: oookay since it just seems to be flaky let's go

.github/workflows/pinned_build.yaml
```diff
@@ -0,0 +1,56 @@
+name: "Pinned Build"
+
+on:
+  workflow_dispatch:
+
+permissions:
+  packages: read
```
spoonincode, line 7: ah, we could remove this permission I think. nbd tho

.github/workflows/pinned_build.yaml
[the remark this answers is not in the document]
oschwaldp-oci: Manual Pinned Build run example here: https://github.com/AntelopeIO/leap/actions/runs/5346832312/jobs/9695391713

.github/workflows/pinned_build.yaml
oschwaldp-oci: image

PR #1299, opened 2023-06-14
[4.0] Fix for loading a snapshot with an empty block log but an existing fork database

heifner:
Additional test cases and fix for loading a snapshot without a block log. When loading a snapshot with a block log, the block log and fork database can be replayed if they start at or before the snapshot. If no block log exists, for example running with `--block-log-retain-blocks 0`, then the fork database can't be consider valid.

Includes a fix for new test introduced in https://github.com/AntelopeIO/leap/pull/1276 which failed during ci/cd. `ship_streamer.cpp` updated to correctly handle forks.

Corner case related to #1228

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -602,13 +602,12 @@ struct controller_impl {
          } else {
             ilog( "Starting initialization from snapshot and no block log, this may take a significant amount of time" );
             read_from_snapshot( snapshot, 0, std::numeric_limits<uint32_t>::max() );
-            const uint32_t lib_num = head->block_num;
-            ilog( "Snapshot loaded, lib: ${lib}", ("lib", lib_num) );
-            EOS_ASSERT( lib_num > 0, snapshot_exception,
+            EOS_ASSERT( head->block_num > 0, snapshot_exception,
                         "Snapshot indicates controller head at block number 0, but that is not allowed. "
                         "Snapshot is invalid." );
-            blog.reset( chain_id, lib_num + 1 );
+            blog.reset( chain_id, head->block_num + 1 );
          }
+         ilog( "Snapshot loaded, lib: ${lib}", ("lib", head->block_num) );
 
          init(check_shutdown);
          ilog( "Finished initialization from snapshot" );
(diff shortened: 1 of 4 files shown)


libraries/chain/controller.cpp
```diff
@@ -590,14 +593,17 @@ struct controller_impl {
    void startup(std::function<void()> shutdown, std::function<bool()> check_shutdown, const snapshot_reader_ptr& snapshot) {
       EOS_ASSERT( snapshot, snapshot_exception, "No snapshot reader provided" );
       this->shutdown = shutdown;
-      ilog( "Starting initialization from snapshot, this may take a significant amount of time" );
       try {
          snapshot->validate();
          if( auto blog_head = blog.head() ) {
+            ilog( "Starting initialization from snapshot and block log ${b}-${e}, this may take a significant amount of time",
+                  ("b", blog.first_block_num())("e", blog_head->block_num()) );
             read_from_snapshot( snapshot, blog.first_block_num(), blog_head->block_num() );
```
linh2931, line 601: Do you want to add a similar log saying snapshot has been loaded?
```
ilog( "Snapshot loaded, lib: ${lib}", ("lib", lib_num) );
```

heifner: Yes, thanks. Done.

tests/ship_streamer.cpp
```diff
@@ -111,8 +113,9 @@ int main(int argc, char* argv[]) {
       stream.write(boost::asio::buffer(request_type.json_to_bin(request_sb.GetString(), [](){})));
       stream.read_message_max(0);
 
-      //      block_num, block_id
-      std::map<uint32_t, std::string> block_ids;
+      // Each block_num can have multiple block_ids since forks are possible
+      //       block_num,         block_id
+      std::map<uint32_t, std::set<std::string>> block_ids;
```
greg7mdp, line 118: why not `std::map<uint32_t, std::vector<std::string>>` since we never do lookups in the `set`?

greg7mdp (approved):

tests/ship_streamer.cpp
heifner: We do look up here: https://github.com/AntelopeIO/leap/pull/1299/files#diff-40f2ef69e45ff8a50b2d87a1f8515c6e6aeeb184ab3e4cc67eb8b62893e65f88R171

PR #1303, opened 2023-06-15
[4.0] P2P: Do not log small clock skew

### LINKED ISSUE #1284: Unexplained occurances of "peer sent a handshake with a timestamp skewed by at least -2ms"
Even after installing patch for https://github.com/AntelopeIO/leap/issues/1240 on top of leap 4.0.1, we still see messages of the type:

```
Peer sent a handshake with a timestamp skewed by at least -2ms
```

The clocks are not skewed by -2ms because we can see the nodes report the similar block arrival time from the same BP, which would not be the case if they were offset by 2ms.  Now maybe something is causing messages to get delayed?

It can bee seen that node clocks are basically in sync in the image here:

image

Let me know if you need me to collect log data on how oven this happens.

I do not recall this issue occurring in earlier versions of leap 3.1 or 3.2, but I wasn't paying attention to it either. (maybe logging level has been changed so now I see it?)

heifner:
P2P attempts to determine clock skew by comparing handshake time. It reports when this time difference is negative as that is an indication in difference in clocks. This PR changes that logging to only log if the difference is greater than 250ms. This will prevent log spam for clock skew that makes no difference.

Resolves #1284

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1777,7 +1777,9 @@ namespace eosio {
       auto current_time_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
       int64_t network_latency_ns = current_time_ns - msg.time; // net latency in nanoseconds
       if( network_latency_ns < 0 ) {
-         peer_wlog(c, "Peer sent a handshake with a timestamp skewed by at least ${t}ms", ("t", network_latency_ns/1000000));
+         if (network_latency_ns < -(block_interval_ns/2)) {
+            peer_wlog(c, "Peer sent a handshake with a timestamp skewed by at least ${t}ms", ("t", network_latency_ns / 1000000));
+         }
          network_latency_ns = 0;
       }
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1777,7 +1777,9 @@ namespace eosio {
       auto current_time_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::system_clock::now().time_since_epoch()).count();
       int64_t network_latency_ns = current_time_ns - msg.time; // net latency in nanoseconds
       if( network_latency_ns < 0 ) {
-         peer_wlog(c, "Peer sent a handshake with a timestamp skewed by at least ${t}ms", ("t", network_latency_ns/1000000));
+         if (network_latency_ns < -(block_interval_ns/2)) {
```
linh2931, line 1780: You can define a constant `half_block_interval_ns` to avoid division every time.

linh2931: But this is a minor suggestion. I am going to approve it.

linh2931 (approved):

PR #1316, opened 2023-06-17
[4.0] P2P: Use magnitude safe time types

### LINKED ISSUE #1315: 4.0.2: p2p constantly disconnects
After upgrading to 4.0.2, p2p connections constantly connect/disconnect. 

Logs sent privately.

Chart showing number of defined peers which are not currently connected.

image

heifner:
Use `std::chrono` types instead of `tstamp` so that magnitude is maintained. Fixes issue with heartbeat calculation being off on some machines causing pre-mature close of what it thought was a idle connection.

Resolves #1315

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -607,8 +607,7 @@ namespace eosio {
       bool is_transactions_only_connection()const { return connection_type == transactions_only; }
       bool is_blocks_only_connection()const { return connection_type == blocks_only; }
       void set_heartbeat_timeout(std::chrono::milliseconds msec) {
-         std::chrono::system_clock::duration dur = msec;
-         hb_timeout = dur.count();
+         hb_timeout = msec;
       }
 
    private:
@@ -688,9 +687,9 @@ namespace eosio {
       tstamp                         xmt{0};          //!< transmit timestamp
       /** @} */
       // timestamp for the lastest message
-      tstamp                         latest_msg_time{0};
-      tstamp                         hb_timeout{std::chrono::milliseconds{def_keepalive_interval}.count()};
-      tstamp                         latest_blk_time{0};
+      std::chrono::system_clock::time_point       latest_msg_time{std::chrono::system_clock::time_point::min()};
+      std::chrono::milliseconds                   hb_timeout{std::chrono::milliseconds{def_keepalive_interval}};
+      std::chrono::system_clock::time_point       latest_blk_time{std::chrono::system_clock::time_point::min()};
 
       bool connected();
       bool current();
@@ -728,7 +727,7 @@ namespace eosio {
        */
       /**  \brief Check heartbeat time and send Time_message
        */
-      void check_heartbeat( tstamp current_time );
+      void check_heartbeat( std::chrono::system_clock::time_point current_time );
       /**  \brief Populate and queue time_message
        */
       void send_time();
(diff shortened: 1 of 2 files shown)


766C6164 (approved):

greg7mdp (approved):
I don't quite understand what was the issue with tstamp. I felt that standardizing on the number of nanoseconds since epoch was nice and simple, but I'm clearly missing what the issue was here.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1165,8 +1164,8 @@ namespace eosio {
    }
 
    // called from connection strand
-   void connection::check_heartbeat( tstamp current_time ) {
-      if( latest_msg_time > 0 ) {
+   void connection::check_heartbeat( std::chrono::system_clock::time_point current_time ) {
+      if( latest_msg_time > std::chrono::system_clock::time_point::min() ) {
          if( current_time > latest_msg_time + hb_timeout ) {
```
heifner, line 1169: The issue was here. `current_time` was in nanoseconds. `lastest_msg_time` was also in nanoseconds, but `hb_timeout` was in `milliseconds`. So when the timer ran the `latest_msg_time` had to be within 10000 nanoseconds (10 microseconds) or it would close the connection.

greg7mdp: So the issue was that `hb_timeout` was in milliseconds. It seems incorrect to me, as it is defined as `std::chrono::system_clock::duration::rep`, so I feel we should consistently use `system_clock` duration in tstamp (on my linux system, `system_clock`'s `duration` is defined as `chrono::nanoseconds`.

I assume that by using `std::chrono` types, we will benefit from an automatic conversion in `if( current_time > latest_msg_time + hb_timeout )`  which will prevent the error. But `tstamp` was already a std::chrono duration, so I wonder if most of the change are really needed.

My feeling is that the change at line 610 (`hb_timeout = msec;`) would have been enough for the fix, as the problem was that we used `count()` which just returned a tick count.

heifner: Could be. Either way, this is much cleaner/safer code.

greg7mdp: I don't see how this makes the code cleaner or safer. I think if you changed line 610 to `hb_timeout = msec.count()` the issue would come back exactly as before.

heifner: I'm confused at what you are suggesting should be done. `time_point`, `milliseconds`, `nanoseconds` types are well defined and arithmetic operations are handled correctly. `tstamp` (`std::chrono::system_clock::duration::rep`) on the other hand is defined as `rep` - `signed arithmetic type representing the number of ticks in the clock's duration` which is system dependent.

linh2931: Is nanoseconds too sensitive in P2P?

linh2931 (approved):

plugins/net_plugin/net_plugin.cpp
heifner: Yes, nanoseconds is kind of ridiculous to use. However, that is what we were already using for `time_message` and `time` of `handshake_message`. Internally now with this change we just use whatever `std::chrono::system_clock::now()` returns.

heifner: Completely removed `tstamp` type and backported the better `time_message` handling from `main`.

greg7mdp: Thanks Kevin for the change. I'm not quite sure is the initialization with `std::chrono::system_clock::time_point::min()` has a specific benefit, I kinda feel  that `0` is more appropriate than a negative value. And I also thing that defining `tstamp` as `std::chrono::nanoseconds` and using it for our `connection` time tracking members would be better. But maybe we can talk about this further for 5.0.

greg7mdp (approved):

PR #1317, opened 2023-06-20
remove `FC_USE_PTHREAD_NAME_NP` & `pthread_getname_np()` usage

spoonincode:
This refactors the thread name stuff a bit to make it simpler (have gotten bitten here a few times lately such as #1283 and #1058)

* remove having both a `set_os_thread_name()` and `set_thread_name()` -- there is no reason setting the thread name shouldn't just set the OS thread name, having both of these (was `set_thread_name()` even used?) is just confusing
* add freebsd & macos thread name support
* remove the `HAVE_PTHREAD_SETNAME_NP` check; I think this was added at some point to resolve building on musl which didn't have `pthread_getname_np()` but while it does now,
* remove usage of `pthread_getname_np()` -- just not necessary to use this instead of referring to the existing `thread_local thread_name`

### DIFF
--- a/libraries/libfc/src/log/logger_config.cpp
+++ b/libraries/libfc/src/log/logger_config.cpp
@@ -10,6 +10,9 @@
 #include <fc/reflect/variant.hpp>
 #include <fc/exception/exception.hpp>
 
+#define BOOST_DLL_USE_STD_FS
+#include <boost/dll/runtime_symbol_info.hpp>
+
 namespace fc {
 
    log_config& log_config::get() {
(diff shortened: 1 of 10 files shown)


libraries/libfc/src/log/logger_config.cpp
```diff
@@ -133,26 +136,22 @@ namespace fc {
...
-            thread_name = thr_name;
+      if(thread_name.empty()) {
+         try {
+            thread_name = boost::dll::program_location().filename().generic_string();
+         } catch (...) {
+            thread_name = "unknown";
          }
-#else
-         static int thread_count = 0;
-         thread_name = std::string( "thread-" ) + std::to_string( thread_count++ );
```
heifner, line 154: On mac we used to always get the `thread-n` version here. Does this now provide the actual thread name on mac?

spoonincode: yes; really for any platform the thread names should be logged identically now since `get_thread_name()` just pulls from the `static thread_local std::string thread_name` -- that is, no platform specific code is invoked for thread names for logging purposes.

But on macos & freebsd we've also gained the thread names being set at the platform level, and I've confirmed on macos they do show in the debugger.

PR #1319, opened 2023-06-20
[3.2] Report transaction failed if trx was exhausted in non-producing mode

### LINKED ISSUE #1247: Exhausted speculative trx not retried
When running in non-default `--read-mode speculative`, an exhausted trx is not retried until a block is received. A test failure where no new blocks arrived illustrated this: https://github.com/AntelopeIO/leap/actions/runs/5202501518

This condition should not be hit for `read-mode=head` as an exhausted trx is an exhausted block as the billed cpu/net are reset at the end of each trx.

Currently we have:
```c++
                                       if (!self->process_incoming_transaction_async(result, api_trx, return_failure_traces, next)) {
                                          if (self->_pending_block_mode == pending_block_mode::producing) {
                                             self->schedule_maybe_produce_block(true);
                                          } else {
                                             self->restart_speculative_block();
                                          }
                                       }
```
Where `process_incoming_transaction_async` returns false for an exhausted block. Instead `process_incoming_transaction_async` should check `self->_pending_block_mode == pending_block_mode::producing` and return false if exhausted block and producing but return false if not producing and either trx is exhausted or block is exhausted.

greg7mdp:
So we restart a speculative block to retry it immediately, instead of waiting to receive a new block.

Resolves #1247.

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -654,7 +654,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                _unapplied_transactions.add_persisted( trx );
             }
 
-            exhausted = pr.block_exhausted; // transaction exhausted if block was exhausted
+            exhausted = pr.block_exhausted;
 
             if ( !in_producing_mode() && pr.trx_exhausted )
                exhausted = true;  // report transaction exhausted if trx was exhausted in non-producing mode (so we will restart
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -616,7 +616,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                                               bool persist_until_expired,
                                               bool return_failure_trace,
                                               next_function<transaction_trace_ptr> next) {
-         bool exhausted = false;
+         bool success = false;
          chain::controller& chain = chain_plug->chain();
          try {
             const auto& id = trx->id();
@@ -648,12 +648,18 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
             const auto block_deadline = calculate_block_deadline( chain.pending_block_time() );
             push_result pr = push_transaction( block_deadline, trx, persist_until_expired, return_failure_trace, next );
 
-            exhausted = pr.block_exhausted;
+            bool exhausted = pr.block_exhausted;
             if( pr.trx_exhausted ) {
                _unapplied_transactions.add_incoming( trx, persist_until_expired, return_failure_trace, next );
             } else if( pr.persist ) {
                _unapplied_transactions.add_persisted( trx );
             }
+
+            success = !exhausted; // transaction failed if block was exhausted
+
+            if (!(_pending_block_mode == pending_block_mode::producing) && pr.trx_exhausted)
+               success = false;  // report transaction failed if trx was exhausted in non-producing mode (so we restart
+                                 // a speculative block to retry it immediately, instead of waiting to receive a new block)
 
          } catch ( const guard_exception& e ) {
             chain_plugin::handle_guard_exception(e);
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -616,7 +616,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
                                               bool persist_until_expired,
                                               bool return_failure_trace,
                                               next_function<transaction_trace_ptr> next) {
-         bool exhausted = false;
+         bool success = false;
```
heifner, line 619: I think something like `blk_exhausted` or `blk_not_exhausted` would be better. `success` is too vague in my opinion. It is not if the trx failed or succeeded, but rather if it ran out of net/time of the block.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -648,13 +648,19 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
 
-            exhausted = pr.block_exhausted;
+            bool exhausted = pr.block_exhausted;
             if( pr.trx_exhausted ) {
                _unapplied_transactions.add_incoming( trx, persist_until_expired, return_failure_trace, next );
             } else if( pr.persist ) {
                _unapplied_transactions.add_persisted( trx );
             }
 
+            success = !exhausted; // transaction failed if block was exhausted
```
heifner, line 658: For the comment, `failed` is too strong of a word. It was interrupted, it didn't fail. We reserve `fail`/`success` for if the trx asserted or not.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -648,13 +648,19 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+
+            if (!(_pending_block_mode == pending_block_mode::producing) && pr.trx_exhausted)
```
heifner, line 660: Is there a reason to write `!(_pending_block_mode == pending_block_mode::producing)` vs `_pending_block_mode != pending_block_mode::producing` or even `_pending_block_mode == pending_block_mode::speculating` ?

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: done

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: done.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -648,13 +648,19 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+               success = false;  // report transaction failed if trx was exhausted in non-producing mode (so we restart
+                                 // a speculative block to retry it immediately, instead of waiting to receive a new block)
```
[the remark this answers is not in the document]
greg7mdp: done.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: I think this read the most as `if not in cprducing mode`, but I have added members to make it clearer.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -648,13 +648,18 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
             push_result pr = push_transaction( block_deadline, trx, persist_until_expired, return_failure_trace, next );
 
-            exhausted = pr.block_exhausted;
             if( pr.trx_exhausted ) {
                _unapplied_transactions.add_incoming( trx, persist_until_expired, return_failure_trace, next );
             } else if( pr.persist ) {
                _unapplied_transactions.add_persisted( trx );
             }
 
+            exhausted = pr.block_exhausted; // transaction exhausted if block was exhausted
```
heifner, line 657: I think remove the comment. `exhausted` here means block is exhausted. That is why the comment below is needed. We say block is exhausted when trx is exhausted below because we want to restart the block when in speculative mode.

heifner (approved):

linh2931 (approved):

PR #1321, opened 2023-06-20
[4.0 -> main] P2P: Use magnitude safe time types

### LINKED ISSUE #1315: 4.0.2: p2p constantly disconnects

heifner:
Use `std::chrono` types instead of `tstamp` so that magnitude is maintained. Fixes issue with heartbeat calculation being off on some machines causing pre-mature close of what it thought was a idle connection.

Merges `release/4.0` into `main` including #1316 

Resolves #1315

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1460,7 +1460,7 @@ namespace eosio {
 
       }
 
-      org = 0;
+      org = std::chrono::nanoseconds{0};
       send_time();
    }
 
@@ -2263,7 +2263,7 @@ namespace eosio {
          c->close( false, true );
          return;
       }
-      c->latest_blk_time = c->get_time();
+      c->latest_blk_time = std::chrono::system_clock::now();
       c->block_status_monitor_.accepted();
       stages state = sync_state;
       peer_dlog( c, "state ${s}", ("s", stage_str( state )) );
@@ -3307,7 +3307,7 @@ namespace eosio {
       msg.dst = get_time().count();
 
       if (msg.org != 0) {
-         if (msg.org == org) {
+         if (msg.org == org.count()) {
             auto ping = msg.dst - msg.org;
             peer_dlog(this, "send_time ping ${p}us", ("p", ping / 1000));
             peer_ping_time_ns = ping;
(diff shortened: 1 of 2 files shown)


linh2931 (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -864,15 +863,15 @@ namespace eosio {
...
+      std::chrono::nanoseconds               xmt{0}; //!< transmit timestamp, Time at the server when the response left for the client.
+      // std::chrono::nanoseconds (not used) dst{0}; //!< destination timestamp, Time at the client when the reply arrived from the server.
       /** @} */
       // timestamp for the lastest message
-      tstamp                         latest_msg_time{0};
-      tstamp                         hb_timeout{std::chrono::milliseconds{def_keepalive_interval}.count()};
-      tstamp                         latest_blk_time{0};
+      std::chrono::system_clock::time_point       latest_msg_time{std::chrono::system_clock::time_point::min()};
+      std::chrono::milliseconds                   hb_timeout{std::chrono::milliseconds{def_keepalive_interval}};
+      std::chrono::system_clock::time_point       latest_blk_time{std::chrono::system_clock::time_point::min()};
```
greg7mdp, line 874: I'd prefer have these all be `tstamp` as before, but with `tstamp` defined as `std::chrono::nanoseconds`. 

If we want to differentiate between a point in time (since the epoch), i.e `time_point`, and a `duration`, then org/rec/xmt/dst should be `time_point`, and not durations.

Also I don't think it is a good idea to use `std::chrono::system_clock::time_point` ourselves because the duration of the system_clock is not specified. 

So I think the most straightforward thing might be to use `tstamp` defined as `std::chrono::nanoseconds` for both actual timestamps (i.e. time since epoch) and durations. Because it is a duration from std::chrono, we benefit from type safety insuring that we don't add `std::chrono::nanoseconds` and `std::chrono::milliseconds`, but we use a consistent internal representation. I think it would make the code simpler and clearer.

heifner: I created #1326 as this is a merge up PR, I think it makes sense to make these changes in a separate PR.

greg7mdp (approved):

PR #1322, opened 2023-06-20
Add eos-vm-oc-enable auto mode

### LINKED ISSUE #1251: EOS VM OC: Implement new Auto OC Mode
Description can be found at: https://github.com/eosnetworkfoundation/product/issues/165

heifner:
### Short Description
In order to enhance the performance of EOS and EOS EVM, we are introducing the use of OC (Optimizing Compiler) for the EOS VM. This initiative aims to improve the overall efficiency and scalability of the platform.

### What
The primary goal of this initiative is to leverage OC for the EOS VM by implementing an Auto OC Mode as the default setting. This new mode, `eos-vm-oc-enable=auto`, will enhance performance. OC will automatically be used based on different contexts, such as building blocks, applying blocks, executing transactions from HTTP or P2P, and various transaction types. OC will also be used for all contracts on `eosio.*` accounts. This includes: `eosio`, `eosio.token`, `eosio.ibc`, and `eosio.evm`.

### Why
By utilizing OC, we can significantly boost the performance of EOS and EOS EVM. With the Auto OC Mode as the default, the system will automatically apply OC when appropriate, optimizing the execution of transactions and improving overall throughput. The introduction of OC for specific contexts, such as building blocks and speculative transaction executions, further enhances the efficiency of the platform. 

#### Auto OC Mode as default
`eos-vm-oc-enable=auto` is the new default. Scripts that use `--eos-vm-oc-enable` will need to be updated to use `--eos-vm-oc-enable all` to enable OC for all contexts. 
- `eos-vm-oc-enable=true` => `eos-vm-oc-enable=all` (`eos-vm-oc-enable=true` continues to work and is mapped to `all`)
- `eos-vm-oc-enable=false` => `eos-vm-oc-enable=none` (`eos-vm-oc-enable=false` continues to work and is mapped to `none`)
- `--eos-vm-oc-enable` will not now work, replace with `--eos-vm-oc-enable all`

#### Baseline / OC selection
In Auto OC Mode, leverage EOS VM OC, as follows:

| Context                               | OC?                                                             |
|---------------------------------------|-----------------------------------------------------------------|
| Building block                        | baseline, but OC if account is eosio.*                          |
| Applying block                        | OC, _or_<br>if producer: baseline, but OC if account is eosio.* |
| Speculatively executing trx from HTTP | baseline, but OC if account is eosio.*                          |
| Speculatively executing trx from p2p  | baseline, but OC if account is eosio.*  :bulb:                         |
| Compute transaction                   | baseline, but OC if account is eosio.*                          |
| Read only transaction                 | OC                                                              |

:bulb: Although enabling OC for P2P would enable quicker validation of speculative trxs as they traverse the network, this PR does not add OC for P2P to help prevent overload on BP nodes. A separate option to enable OC for P2P when in `auto` is being considered as a follow on enhancement; see #1333.

Resolves #1251


libraries/chain/include/eosio/chain/wasm_interface_collection.hpp
```diff
@@ -0,0 +1,78 @@
+#pragma once
+#include <eosio/chain/wasm_interface.hpp>
+
+namespace eosio::chain {
+
+   /**
+    * @class wasm_interface_collection manages the active wasm_interface to use for execution.
+    */
+   class wasm_interface_collection {
```
heifner, line 9: There is no functionality change here, just a refactor out of controller.

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -355,6 +355,9 @@ namespace eosio { namespace chain {
       void replace_producer_keys( const public_key_type& key );
       void replace_account_keys( name account, name permission, const public_key_type& key );
 
+      void set_producer_node(bool is_producer_node);
+      bool is_producer_node()const;
```
heifner, line 359: Kind of sad about this change. Until now, the core of leap had no need to know if it was a producer or not.

libraries/chain/include/eosio/chain/name.hpp
```diff
@@ -75,6 +75,41 @@ namespace eosio::chain {
...
 
       constexpr explicit operator bool()const { return value != 0; }
+
+      /**
+       *  Returns the prefix.
+       *  for exmaple:
+       *    "eosio.any" -> "eosio"
+       *    "eosio" -> "eosio"
+       */
+      constexpr name prefix() const {
```
heifner, line 85: Implementation copied from CDT

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -114,6 +114,31 @@ void validate(boost::any& v,
...
+
+  // Extract the first string from 'values'. If there is more than
+  // one string, it's an error, and exception will be thrown.
+  std::string const& s = validators::get_single_string(values);
+
+  if (s == "auto") {
+     v = boost::any(wasm_interface::vm_oc_enable::oc_auto);
+  } else if (s == "all" || s == "true" || s == "on") {
+     v = boost::any(wasm_interface::vm_oc_enable::oc_all);
+  } else if (s == "none" || s == "false" || s == "off") {
```
spoonincode, line 135: Other previously valid choices were `1`/`0` and `yes`/`no`

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -114,6 +114,31 @@ void validate(boost::any& v,
...
+     v = boost::any(wasm_interface::vm_oc_enable::oc_auto);
+  } else if (s == "all" || s == "true" || s == "on" || s == "yes" || s == "1") {
+     v = boost::any(wasm_interface::vm_oc_enable::oc_all);
+  } else if (s == "none" || s == "false" || s == "off" || s == "no" || s == "0") {
+     v = boost::any(wasm_interface::vm_oc_enable::oc_none);
+  } else {
+     throw validation_error(validation_error::invalid_option_value);
+  }
+}
+
```
linh2931, line 141: To have fewer options in the future, we could add a deprecation note that only `auto`, `all`, and `none` will be allowed, all others could be deprecated in future releases.

libraries/chain/include/eosio/chain/wasm_interface_collection.hpp
```diff
@@ -0,0 +1,78 @@
...
+         const std::thread::id main_thread_id;
+         const wasm_interface::vm_type wasm_runtime;
+         const wasm_interface::vm_oc_enable eosvmoc_tierup;
+
+         wasm_interface wasmif;  // used by main thread and all threads for EOSVMOC
+         std::mutex threaded_wasmifs_mtx;
+         std::unordered_map<std::thread::id, std::unique_ptr<wasm_interface>> threaded_wasmifs; // one for each read-only thread, used by eos-vm and eos-vm-jit
+   };
+
+} // eosio::chain
```
linh2931, line 78: Thanks for this refactoring. It hides those details from the controller.

linh2931: Can we add some tests for `auto`, `all`, and `none` options to make sure OC is enabled/disabled as configured?

heifner: Any ideas on how? Unfortunately, oc is not setup to be used in unittests because the destructor of `tester` -> `controller`  destroys oc and it can't be restarted because it forks the process on start. Seems like it would require a major refactor to allow oc to be used repeatably in unittests.

We could add some logging and look for a particular log in some integration tests. Let me see what I can do for that.

docs/01_nodeos/03_plugins/chain_plugin/index.md
```diff
@@ -179,7 +179,16 @@ Config Options for eosio::chain_plugin:
                                         code cache
   --eos-vm-oc-compile-threads arg (=1)  Number of threads to use for EOS VM OC
                                         tier-up
-  --eos-vm-oc-enable                    Enable EOS VM OC tier-up runtime
+  --eos-vm-oc-enable arg (=auto)        Enable EOS VM OC tier-up runtime
+                                        ('auto', 'all', 'none').
+                                        'auto' - EOS VM OC tier-up is enabled
```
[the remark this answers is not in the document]
heifner: Added applying blocks. Not for producers seems like too much info for the help string.

libraries/chain/apply_context.cpp
```diff
@@ -1094,4 +1094,19 @@ action_name apply_context::get_sender() const {
...
+// Context             |    OC?
+//-------------------------------------------------------------------------------
+// Building block      | baseline, OC for eosio.*
+// Applying block      | OC unless a producer, OC for eosio.* including producers
+// Speculative API trx | baseline, OC for eosio.*
+// Speculative P2P trx | baseline, OC for eosio.*
+// Compute trx         | baseline, OC for eosio.*
+// Read only trx       | OC
+bool apply_context::should_use_eos_vm_oc()const {
+   return trx_context.is_read_only()
```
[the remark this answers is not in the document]
heifner: Done

libraries/chain/controller.cpp
```diff
@@ -239,6 +240,7 @@ struct controller_impl {
    controller::config              conf;
    const chain_id_type             chain_id; // read by thread_pool threads, value will not be changed
    bool                            replaying = false;
+   bool                            producer_node = false; // true if node is configured as a block producer
```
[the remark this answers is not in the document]
heifner: Renamed

unittests/misc_tests.cpp
```diff
@@ -139,6 +139,25 @@ BOOST_AUTO_TEST_CASE(name_suffix_tests)
...
+   BOOST_CHECK_EQUAL("abc.xyz.qrt"_n.prefix(), "abc.xyz"_n);
+
+   BOOST_CHECK_EQUAL("eosio.any"_n.prefix(), "eosio"_n);
+   BOOST_CHECK_EQUAL("eosio"_n.prefix(), "eosio"_n);
+   BOOST_CHECK_EQUAL("eosio"_n.prefix(), config::system_account_name);
+   BOOST_CHECK_EQUAL("eosio."_n.prefix(), "eosio"_n);
+   BOOST_CHECK_EQUAL("eosio.evm"_n.prefix(), "eosio"_n);
+   BOOST_CHECK_NE("eosi"_n.prefix(), "eosio"_n);
+   BOOST_CHECK_NE("eosioeosio"_n.prefix(), "eosio"_n);
+   BOOST_CHECK_NE("eosioe"_n.prefix(), "eosio"_n);
```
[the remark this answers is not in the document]
heifner: Added.

libraries/chain/include/eosio/chain/wasm_interface_collection.hpp
heifner: Integration test added.

libraries/chain/include/eosio/chain/name.hpp
```diff
@@ -75,6 +75,41 @@ namespace eosio::chain {
...
+         uint64_t result                 = value;
+         bool     not_dot_character_seen = false;
+         uint64_t mask                   = 0xFull;
+
+         // Get characters one-by-one in name in order from right to left
+         for (int32_t offset = 0; offset <= 59;) {
+            auto c = (value >> offset) & mask;
+
+            if (!c) {                        // if this character is a dot
```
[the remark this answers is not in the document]
heifner: Would prefer to keep this identical to CDT implementation.

linh2931 (approved):

libraries/chain/include/eosio/chain/wasm_interface_collection.hpp
```diff
@@ -0,0 +1,78 @@
...
+            }
+         }
+
+         // only called from non-main threads (read-only trx execution threads) when producer_plugin starts them
+         void init_thread_local_data(const chainbase::database& d, const std::filesystem::path& data_dir,
+                                     const eosvmoc::config& eosvmoc_config, bool profile) {
+            EOS_ASSERT(!is_on_main_thread(), misc_exception, "init_thread_local_data called on the main thread");
+            if (is_eos_vm_oc_enabled()) {
+               // EOSVMOC needs further initialization of its thread local data
+               wasmif.init_thread_local_data();
```
spoonincode, line 43: lost an `#ifdef EOSIO_EOS_VM_OC_RUNTIME_ENABLED` on these couple lines during the refactor

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -158,6 +162,10 @@ namespace eosio { namespace chain {
          return it->module;
       }
 
+      bool should_always_oc_tierup()const {
+         return wasm_runtime_time == wasm_interface::vm_type::eos_vm_oc || eosvmoc_tierup == wasm_interface::vm_oc_enable::oc_all;
```
spoonincode, line 166: I don't think `wasm_runtime_time == wasm_interface::vm_type::eos_vm_oc` should be in here. `wasm_runtime_time == wasm_interface::vm_type::eos_vm_oc` is meant to be "OC in _non_-tier up mode" and all the code back here
https://github.com/AntelopeIO/leap/blob/6008c728927f524e06a6d77463c5a5c165190c18/libraries/chain/wasm_interface.cpp#L93
should be skipped entirely in non-tier up mode. It's not clear what the ramifications are for having both a synchronous and asynchronous code_cache at the same time (which is what I think might be happening in this case.. not sure.. regardless it def wasn't originally intended to operate this way)

I think the line above should be something more on the order of
```c++
if(my->eosvmoc &&
  (my->eosvmoc_tierup == wasm_interface::vm_oc_enable::oc_all || context.should_use_eos_vm_oc())) { 
```

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -405,6 +405,8 @@ namespace eosio { namespace testing {
             cfg.eosvmoc_config.cache_size = 1024*1024*8;
 
             for(int i = 0; i < boost::unit_test::framework::master_test_suite().argc; ++i) {
+               // don't use auto tier up for tests, since the point is to test diff vms
+               cfg.eosvmoc_tierup = chain::wasm_interface::vm_oc_enable::oc_none;
```
spoonincode, line 409: Not sure why this is inside the `for()` loop instead of just with the `cfg.` items a few lines above.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -114,6 +114,31 @@ void validate(boost::any& v,
...
+  // Extract the first string from 'values'. If there is more than
+  // one string, it's an error, and exception will be thrown.
+  std::string const& s = validators::get_single_string(values);
+
+  if (s == "auto") {
+     v = boost::any(wasm_interface::vm_oc_enable::oc_auto);
+  } else if (s == "all" || s == "true" || s == "on" || s == "yes" || s == "1") {
+     v = boost::any(wasm_interface::vm_oc_enable::oc_all);
+  } else if (s == "none" || s == "false" || s == "off" || s == "no" || s == "0") {
```
spoonincode, line 135: fwiw, it appears boost was doing a case insensitive compare; so `False`, `oN`, etc was also accepted :smiling_imp: depending on how accommodating you want to be...

libraries/chain/include/eosio/chain/controller.hpp
spoonincode: well, I guess depending how much of a purist you want to be here, maybe you could make controller take a `std::function<bool(apply_context)> tier_up_with_oc`, either as a ctor parameter or a config or something. That way `chain_plugin` can decide how to set this up based on the various configs. And libtester just always passes in a `return false;` for example, I guess.

plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -334,7 +360,11 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
                   EOS_ASSERT(false, plugin_exception, "");
                }
          }), "Number of threads to use for EOS VM OC tier-up")
-         ("eos-vm-oc-enable", bpo::bool_switch(), "Enable EOS VM OC tier-up runtime")
+         ("eos-vm-oc-enable", bpo::value<chain::wasm_interface::vm_oc_enable>()->default_value(chain::wasm_interface::vm_oc_enable::oc_auto),
+          "Enable EOS VM OC tier-up runtime ('auto', 'all', 'none').\n"
+          "'auto' - EOS VM OC tier-up is enabled for eosio.* accounts and read-only trxs.\n"
```
spoonincode: I know Lin already touched on it some, but I do feel like we ought to try and find a short concise way to sneak in that it's also enabled for applying blocks unless a BP. That aspect is completely missing in this description even without the extra "non BP rule".

Maybe something like
> 'auto' - EOS VM OC tier-up is enabled for eosio.* accounts, read-only trxs, and applying blocks (excluding on producers).

We already have plenty more wordy config items in `read-mode`, `peer-log-format`, etc.. so I feel like we can spend 5 words.

plugins/state_history_plugin/tests/main.cpp
```diff
@@ -0,0 +1,2 @@
+#define BOOST_TEST_MODULE state_history_plugin
+#include <boost/test/included/unit_test.hpp>
```
spoonincode, line 2: Since you're using the header-only variant now, you should remove linking to the library from here
https://github.com/AntelopeIO/leap/blob/870673840b94532ee2c07bb70d4b1a1b71182b02/plugins/state_history_plugin/tests/CMakeLists.txt#L2

plugins/chain_plugin/chain_plugin.cpp
spoonincode: whoops somehow I was on an older revision when I made that comment. I see now that we did add commentary around applying blocks. Still not sure penny pinching 3 words and this avoiding documenting the producer exclusion is worth it... but what we have in the latest commit is fine to me since at least it includes block application

libraries/chain/include/eosio/chain/controller.hpp
heifner: Yeah, that seems a bit too far.

plugins/chain_plugin/chain_plugin.cpp
heifner: What do you think of:
`'auto' - EOS VM OC tier-up is enabled for eosio.* accounts, read-only trxs, and except on producers applying blocks.`. Putting it as a paratheses at the end made it seem like producer excluded all cases.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
heifner: Thanks. I missed that `eos_vm_oc` didn't flow throw here.

libraries/chain/include/eosio/chain/wasm_interface_collection.hpp
heifner: I wonder if most of the `#ifdef EOSIO_EOS_VM_OC_RUNTIME_ENABLED` could be removed to simplify differences. For now, I'll just revert back to having them.

heifner: Restored `#ifdef`s

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
spoonincode: It seems a non issue from a functional standpoint -- I remember now that `my->eosvmoc` isn't created unless in `oc_auto` or `oc_all` mode,
https://github.com/AntelopeIO/leap/blob/a8f20cebf41e5b0706616d9defd13e5de95a3027/libraries/chain/include/eosio/chain/wasm_interface_private.hpp#L93-L96
so nothing wonky would have happened since `my->eosvmoc &&` would have always failed out. But yeah def better now with the tweak.

libraries/chain/wasm_interface.cpp
```diff
@@ -105,6 +105,8 @@ namespace eosio { namespace chain {
             once_is_enough = true;
          }
          if(cd) {
+            if (!context.is_applying_block())
+               tlog("${a} speculatively executing ${h} with eos vm oc", ("a", context.get_receiver())("h", code_hash));
```
spoonincode, line 109: was this meant to be left in?

libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
```diff
@@ -306,6 +306,12 @@ code_cache_base::code_cache_base(const std::filesystem::path data_dir, const eos
 }
 
 void code_cache_base::set_on_disk_region_dirty(bool dirty) {
+   // tests can remove directory before destructor is called
```
spoonincode, line 309: This feels pretty weird to me.. it makes it sound like there are cases where `fc::temp_dir` is dtor'ed before the `controller` is dtor'ed?

spoonincode (approved):
looks good

libraries/chain/wasm_interface.cpp
heifner: `tlog` is trace log, but no, that was for testing.

heifner: Sorry, yes, it is needed for the read_only_trx_test.py test. I'll add a comment about that.

libraries/chain/webassembly/runtimes/eos-vm-oc/code_cache.cpp
heifner: Good point. I'll fix the tests and remove this.

PR #1351, opened 2023-06-28
Blocklog: Return std::optional for head_id 

### LINKED ISSUE #1289: Update block_log and fork_db to return empty `std::optional` when requested data not present
We can have empty `block_logs` and `fork_db`, and when that happens APIs like `fork_db.root()` and `blog.first_block_num()` make up credible values that are not always consistent. For example, in `controller.cpp::log_irreversible()`, we retrieve the blog head with:

`block_id_type log_head_id = blog.head_id();` which always return a `sha256`, and the we check if it is valid by checking if it is all zeros.

And if it is invalid, then we check that `fork_db.root()->block_num == blog.first_block_num() - 1`.

It seems to me that if we had `blog.head_id()` returning an empty `std::optional` to mean that the blog is empty, then maybe the only think we would need to check in the assert is that `fork_db.root()` also return std::nullopt.

PS: If I messed up the explanation, Kevin understands what I mean and is welcome to fix it.

heifner:
Minor refactor to make code cleaner/clearer. Return empty optional when block log does not have a current head.

Resolves #1289

### DIFF
--- a/libraries/chain/block_log.cpp
+++ b/libraries/chain/block_log.cpp
@@ -462,7 +462,11 @@ namespace eosio { namespace chain {
          inline static uint32_t  default_initial_version = block_log::max_supported_version;
 
          std::mutex       mtx;
-         std::optional<std::pair<signed_block_ptr, block_id_type>> head;
+         struct signed_block_with_id {
+            signed_block_ptr ptr;
+            block_id_type id;
+         };
+         std::optional<signed_block_with_id> head;
 
          virtual ~block_log_impl() = default;
 
@@ -497,7 +501,7 @@ namespace eosio { namespace chain {
             std::filesystem::remove(log_dir / "blocks.index");
          }
 
-         uint32_t first_block_num() final { return head ? head->first->block_num() : first_block_number; }
+         uint32_t first_block_num() final { return head ? head->ptr->block_num() : first_block_number; }
          void append(const signed_block_ptr& b, const block_id_type& id, const std::vector<char>& packed_block) final {
             update_head(b, id);
          }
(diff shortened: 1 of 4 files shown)


greg7mdp (approved):

greg7mdp:
Upon reflection, it seems that we either have both `{ head, head_id }` of they are both null.

Maybe the `optional` should be `std::optional<signed_block_ptr, block_id_type>` head;`, so it is clear that either both are present or none.

heifner:
> Upon reflection, it seems that we either have both `{ head, head_id }`, or they are both null.
> 
> Maybe the `optional` should be `std::optional<signed_block_ptr, block_id_type> head;`, so it is clear that either both are present or none.

That does make it clearer. Done.

libraries/chain/block_log.cpp
```diff
@@ -1252,12 +1244,12 @@ namespace eosio { namespace chain {
...
       std::lock_guard g(my->mtx);
-      return my->head;
+      return my->head ? my->head->first : signed_block_ptr{};
    }
 
-   block_id_type block_log::head_id() const {
+   std::optional<block_id_type> block_log::head_id() const {
       std::lock_guard g(my->mtx);
-      return my->head_id;
+      return my->head ? my->head->second : std::optional<block_id_type>{};
```
greg7mdp, line 1252:
```suggestion
      return my->head ? my->head->second : {};
```

greg7mdp (approved):

libraries/chain/block_log.cpp
```diff
@@ -462,8 +462,7 @@ namespace eosio { namespace chain {
          inline static uint32_t  default_initial_version = block_log::max_supported_version;
 
          std::mutex       mtx;
-         signed_block_ptr head;
-         block_id_type    head_id;
+         std::optional<std::pair<signed_block_ptr, block_id_type>> head;
```
linh2931, line 465: I like using optional. But I am a little hesitate about std::pair, as when you look at head->first or head->second, you need to go back the definition to see what first and second refer to. Can it be made an optional of struct?

heifner: Done.

linh2931 (approved):

Issue #1156, opened 2023-05-11
CICD support for previous versions of nodeos for Performance Harness


oschwaldp-oci:
Looking for options to test against previous versions of nodeos in CI/CD.

oschwaldp-oci:
As the Performance Harness is meant to benchmark and track performance enhancements and safeguard against degradations over the product development lifecycle, it would be good to have a means to run the Performance Harness in an automated and periodic manner.

To enable the historical benchmarking and safeguard against breaking compatibility with past versions of `nodeos` that still maintain performance measurements for historical comparison, it would be good to have access to those previous versions of `nodeos` in CICD to run smoke tests against for maintaining compatibility as the Performance Harness continues to develop.  Currently, Performance Harness development has attempted to maintain the ability to benchmark performance on releases starting at `2.0`, `2.1`, `3.1`, `3.2` up to current.  If the ability to regenerate historical benchmarks with newly developed metrics is not desired then this could become a non-issue.  If it is desired, some additional input into how far back to go would also be valuable.

PR #1356, opened 2023-06-30
[PH] Create workflow to test backward compatibility of performance harness

oschwaldp-oci:
Create workflow to test backward compatibility of performance harness with prior leap releases.

Introduces a reusable workflow `build-base` for the steps common to these workflows.

Resolves: https://github.com/AntelopeIO/leap/issues/1156


.github/workflows/ph_backward_compatibility.yaml
```diff
@@ -12,10 +12,176 @@ defaults:
...
+            # https://github.com/actions/runner/issues/2033
+            chown -R $(id -u):$(id -g) $PWD
+            cmake -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -GNinja
+            cmake --build build
+            tar -pc --exclude "*.o" build | zstd --long -T0 -9 > build.tar.zst
+        - name: Upload builddir
+          uses: AntelopeIO/upload-artifact-large-chunks-action@v1
+          with:
+            name: ${{matrix.platform}}-build
+            path: build.tar.zst
```
spoonincode, line 81: We shouldn't be copy pasting the 3 jobs above. If we need a leap build to be performed as part of this workflow we should refactor the existing workflows so that building leap is a reusable workflow, that way it can be used from the per-PR workflow and from here.

Performing a build here may not strictly be required since you can download artifacts from other workflow runs, but that requires the user to manually perform a build with the other workflow (and waiting for it to complete) before calling this workflow.

.github/workflows/ph_backward_compatibility.yaml
```diff
@@ -12,10 +12,176 @@ defaults:
...
+          file: 'leap_.*-${{matrix.platform}}.04_amd64.deb' # Ex. leap_3.2.3-ubuntu20.04_amd64.deb  leap_4.0.3-ubuntu20.04_amd64.deb
+          target: '${{matrix.release}}'
+          token: ${{github.token}}
+      - if: ${{ matrix.release == '3.1' }}
+        name: Download Prev Leap Version (v3.1.x)
+        uses: AntelopeIO/asset-artifact-download-action@v2
+        with:
+          owner: AntelopeIO
+          repo: leap
+          file: 'leap-.*-${{matrix.platform}}.04-x86_64.deb' # Ex. leap-3.1.4-ubuntu20.04-x86_64.deb
```
spoonincode, line 122: You might be able to combine these two `if` blocks via something like `file: '(leap).*${{matrix.platform}}.04.*(x86_64|amd64).deb'`

.github/workflows/ph_backward_compatibility.yaml
```diff
@@ -12,10 +12,176 @@ defaults:
...
+          target: '${{matrix.release}}'
+          token: ${{github.token}}
+      - name: Extract and Place Rev Leap Version artifacts
+        run: |
+          mkdir tmp
+          dpkg -x leap*.deb tmp
```
spoonincode, line 128: Why extract it instead of installing it? This won't be reliable across versions due to dependency changes. For example, let's say in 5.0 we no longer require gmp: it won't be in the base builder image any longer yet leap 4.0 requires that to be installed to run. Using the leap-4.0.deb resolves these for us.

.github/workflows/ph_backward_compatibility.yaml
```diff
@@ -12,10 +12,176 @@ defaults:
...
+        if: always()
+        run: |
+          cd build
+          ctest --output-on-failure -R performance_test_basic_http --timeout 480
+      - name: Run Read Only Trx Performance Basic Test
+        if: always()
         run: |
-          echo "Workflow Stub"
+          cd build
+          ctest --output-on-failure -R performance_test_basic_read_only_trxs --timeout 480
```
spoonincode, line 187: Having all of these manually defined like this is at risk to forgetting to add one later. Maybe we should have a new label for performance tests, or maybe just `-R performance_test` is enough.

spoonincode (changes requested):
I know this is still in draft but I have some maintenance concerns with where it's going.

.github/workflows/ph_backward_compatibility.yaml
oschwaldp-oci: Collapsing these back down into a regular expression is the plan. Was splitting them out to get clearer information when a test failed during initial debugging.

oschwaldp-oci: I believe I can do this with a mix of `-R` and `-E` depending on `nodeos` version.

oschwaldp-oci: Addressed: 800a37e60

.github/workflows/ph_backward_compatibility.yaml
oschwaldp-oci: Thanks, good call. Addressed: c63042bb9329

.github/workflows/ph_backward_compatibility.yaml
oschwaldp-oci: Good call. Addressed: b7dea9bfde2d

.github/workflows/ph_backward_compatibility.yaml
oschwaldp-oci: @spoonincode - Starting to address this initially with a reusable workflow `build-base`.  Would like to further refine to allow it to either do the build from scratch or, if the build exists, reuse the artifacts.  Still strategizing how to make this work.

.github/workflows/build.yaml
```diff
@@ -23,28 +23,17 @@ on:
...
-    steps:
-      - name: Discover Platforms
-        id: discover
-        uses: AntelopeIO/discover-platforms-action@v1
-        with:
-          platform-file: .cicd/platforms.json
-          password: ${{secrets.GITHUB_TOKEN}}
-          package-name: builders
+  build-base:
+    uses: AntelopeIO/leap/.github/workflows/build_base.yaml@GH-1156-ph-cicd-nodeos-versions
```
spoonincode, line 35: I think this should just be
```yaml
 uses: .github/workflows/build_base.yaml
```
Hardcoding `AntelopeIO` and/or `leap` breaks forks/mirrors, and we shouldn't be hardcoding the branch -- it should just use the same ref as the caller

.github/workflows/build.yaml
```diff
@@ -23,28 +23,17 @@ on:
         type: string
 
 permissions:
-  packages: read
+  packages: write
```
spoonincode, line 26: I would need to refresh my memory on how permissions & reusable workflows work to know for sure.. but can we leave this to `read` and only plumb through the write permission on the resusable workflow call? That is, can we do something like
```yaml
 build-base:
    uses: .github/workflows/build_base.yaml
    permissions:
       packages: write
       contents: read
```

.github/workflows/build.yaml
oschwaldp-oci: Addressed: 477f62c85779 and 0da3e7b36eb

.github/workflows/build.yaml
oschwaldp-oci: Good call. Finally got that to work. Had tried a couple other options before and didn't get the right combination it seems.  Addressed: 495d6eab30fcde

.github/workflows/build_base.yaml
```diff
@@ -0,0 +1,86 @@
...
+on:
+  workflow_dispatch:
+  workflow_call:
+    outputs:
+      p:
+        description: "Discovered Build Platforms"
+        value: ${{ jobs.d.outputs.p }}
+
+permissions:
+  packages: write
```
spoonincode, line 12: I think this should just be `read`, since below the `build-platforms` job has
```yaml
    permissions:
      packages: write
      contents: read
```
So I think as long as `build_base.yaml` is called with `packages: write`, which it seems to be:
https://github.com/AntelopeIO/leap/blob/4e19e3636237596aab66de2f6fceabcf89e0cc3f/.github/workflows/build.yaml#L34-L38
Having `packages: read` in this location should cause all jobs in this workflow other then `build-platform` be read, where `build-platform` is write.

(btw your ph_backward_compatibility.yaml already does this)
https://github.com/AntelopeIO/leap/blob/4e19e3636237596aab66de2f6fceabcf89e0cc3f/.github/workflows/ph_backward_compatibility.yaml#L6-L8

spoonincode (commented):
just one permission nit, otherwise it seems good best I can tell

.github/workflows/build_base.yaml
oschwaldp-oci: Good catch. Slipped past when I was changing the others. I've corrected that now.

.github/workflows/build.yaml
```diff
@@ -31,20 +31,12 @@ defaults:
...
-      p: ${{steps.discover.outputs.platforms}}
-    steps:
-      - name: Discover Platforms
-        id: discover
-        uses: AntelopeIO/discover-platforms-action@v1
-        with:
-          platform-file: .cicd/platforms.json
-          password: ${{secrets.GITHUB_TOKEN}}
-          package-name: builders
+  build-base:
```
spoonincode, line 34: Since this job (that uses the reusable workflow) doesn't have a name it seems to show up as simply `build-base` in the GUI in various places. I'm a little surprises it doesn't inherit the callee workflow name. Maybe we could consider giving this a name (even just `Build`) to show up nicer.

Screenshot 2023-07-13 at 22-22-58 PH Create workflow to test backward compatibility of performance harness · AntelopeIO_leap@b59fede

---

Screenshot 2023-07-13 at 22-23-21 AntelopeIO_leap C implementation of the Antelope protocol

spoonincode (approved):

.github/workflows/build.yaml
oschwaldp-oci: Updated with name `Run Build Workflow`:
image
image

heifner (approved):

PR #1362, opened 2023-07-03
Fix incorrect serializing of `std::optional` when value is not provided

### LINKED ISSUE #1189: abi_serializer: Missing field exception thrown when an std::optional is used in a field of an action input parameter
```
struct test_input {
   uint64_t a;
   std::optional<uint64_t> b;
};

[[eosio::action]] void test(const test_input& input);
```

Create a failing test case reproducing this issue. Then fix it.

greg7mdp:
resolves #1189 .

When serializing a `std::optional` field, and a value is not provided for the `std::optional`, we need to add into the serialization the flag specifying that a missing value is serialized. Prior to this PR, the flag was missing.


greg7mdp:
Closing this PR as the fix needs to go into 3.2 first.

PR #1363, opened 2023-07-03
no longer eagerly use tcmalloc if it's found

spoonincode:
Since EOSIO 1.0 tcmalloc (from gperftools) would be used if it was found during cmake configuration. Despite that, we've never once had tcmalloc enabled in any sort of CI. I know in certain historical configurations, such as Ubuntu 16.04, the tcmalloc available from the standard package repo caused a dead lock in one of the unit tests (of course, the fault certainly could be have in the test, hard to say).

The benefit of using tcmalloc is supposed improved performance, but we don't have any recent data (or maybe any data ever) to back up that claim.

It seems a little reckless to eagerly enable features like this that we don't actually test ourselves. So move tcmalloc usage behind a cmake option that is defaulted to `off`. Using tcmalloc is still useful for leap developers due to its profiler, etc.

A nice side effect is this may improve the reliability of the proposed automatic database migration approach too.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,9 +159,10 @@ endif()
 
 message( STATUS "Using '${EOSIO_ROOT_KEY}' as public key for 'eosio' account" )
 
-find_package( Gperftools QUIET )
-if( GPERFTOOLS_FOUND )
-    message( STATUS "Found gperftools; compiling Leap with TCMalloc")
+option(ENABLE_TCMALLOC "use tcmalloc (requires gperftools)" OFF)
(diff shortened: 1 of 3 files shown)


heifner (approved):
Do we want to backport this?

spoonincode:
> Do we want to backport this?

I was tempted to suggest that when it seemed the recent `calloc()` change was causing problems in tcmalloc. But since that doesn't appear to be the case any longer, maybe not really important to backport at this point. (I still would like to take it for 5.0+ due to the reasons originally documented)

linh2931 (approved):

PR #1364, opened 2023-07-03
[3.2] Fix incorrect serializing of std::optional when value is not provided - (GH #1189)

greg7mdp:
When serializing a `std::optional` field, and a value is not provided for the `std::optional`, we need to add into the serialization the flag specifying that a missing value is serialized. Prior to this PR, the flag was missing.

### DIFF
--- a/libraries/libfc/include/fc/time.hpp
+++ b/libraries/libfc/include/fc/time.hpp
@@ -1,6 +1,7 @@
 #pragma once
 #include <stdint.h>
 #include <fc/string.hpp>
+#include <limits>
 
 #ifdef _MSC_VER
   #pragma warning (push)
@@ -11,7 +12,7 @@ namespace fc {
   class microseconds {
     public:
         constexpr explicit microseconds( int64_t c = 0) :_count(c){}
-        static constexpr microseconds maximum() { return microseconds(0x7fffffffffffffffll); }
+        static constexpr microseconds maximum() { return microseconds(std::numeric_limits<int64_t>::max()); }
         friend constexpr microseconds operator + (const  microseconds& l, const microseconds& r ) { return microseconds(l._count+r._count); }
         friend constexpr microseconds operator - (const  microseconds& l, const microseconds& r ) { return microseconds(l._count-r._count); }
 
--- a/libraries/libfc/include/fc/time.hpp
+++ b/libraries/libfc/include/fc/time.hpp
@@ -1,7 +1,6 @@
 #pragma once
 #include <stdint.h>
 #include <fc/string.hpp>
-#include <limits>
 
 #ifdef _MSC_VER
   #pragma warning (push)
@@ -12,7 +11,7 @@ namespace fc {
   class microseconds {
     public:
         constexpr explicit microseconds( int64_t c = 0) :_count(c){}
-        static constexpr microseconds maximum() { return microseconds(std::numeric_limits<int64_t>::max()); }
+        static constexpr microseconds maximum() { return microseconds(0x7fffffffffffffffll); }
         friend constexpr microseconds operator + (const  microseconds& l, const microseconds& r ) { return microseconds(l._count+r._count); }
         friend constexpr microseconds operator - (const  microseconds& l, const microseconds& r ) { return microseconds(l._count-r._count); }
 
--- a/unittests/abi_tests.cpp
+++ b/unittests/abi_tests.cpp
@@ -1937,6 +1937,92 @@ BOOST_AUTO_TEST_CASE(abi_type_loop)
 
 } FC_LOG_AND_RETHROW() }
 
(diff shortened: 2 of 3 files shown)


libraries/libfc/include/fc/time.hpp
```diff
@@ -11,7 +12,7 @@ namespace fc {
   class microseconds {
     public:
         constexpr explicit microseconds( int64_t c = 0) :_count(c){}
-        static constexpr microseconds maximum() { return microseconds(0x7fffffffffffffffll); }
+        static constexpr microseconds maximum() { return microseconds(std::numeric_limits<int64_t>::max()); }
```
heifner, line 15: This seems like unnecessary changes for a backport. I would suggest we only do this for `main`.

unittests/abi_tests.cpp
```diff
@@ -40,7 +40,17 @@ FC_REFLECT(act_sig, (sig) )
 
 BOOST_AUTO_TEST_SUITE(abi_tests)
 
+#ifdef NDEBUG
 fc::microseconds max_serialization_time = fc::seconds(1); // some test machines are very slow
+#else
+fc::microseconds max_serialization_time = fc::microseconds::maximum(); // don't check in debug builds
+#endif
```
heifner, line 47: Was this because it was failing locally on a debug build? I think just increasing the existing to 2 or 3 seconds would be preferred to having diff behavior for debug.

heifner (changes requested):

unittests/abi_tests.cpp
greg7mdp: No, it was because the timeouts were hit when I was stepping through the code. The debug build purpose is for debugging, so having timeouts is unfriendly for that). Is there a reason why timeouts in abi tests would be useful in debug builds?

libraries/libfc/include/fc/time.hpp
greg7mdp: sure, will remove for the backports.

unittests/abi_tests.cpp
heifner: Ah, that makes sense. There is a potential for a hang with a debug build, but one could always attach a debugger in that case.

heifner (approved):

greg7mdp:
@heifner @arhag One concern I have is that this PR changes the serialization output for structs containing `std::optional`, when a value is omitted for one or more `std::optional`.
My first thought is that it may not be an issue, because the deserialization of these would have failed anyways, but I thought it would be worth mentioning.

heifner:
> @heifner @arhag One concern I have is that this PR changes the serialization output for structs containing `std::optional`, when a value is omitted for one or more `std::optional`. My first thought is that it may not be an issue, because the deserialization of these would have failed anyways, but I thought it would be worth mentioning.

I don't anticipate any issue here.

unittests/abi_tests.cpp
linh2931: It might be helpful to add a comment explaining in the test code why #if NDEBUG is required.

unittests/abi_tests.cpp
```diff
@@ -1929,6 +1937,113 @@ BOOST_AUTO_TEST_CASE(abi_type_loop)
...
+        "bridge_fee" : "2"
+      }
+      )=====";
+
+      auto var = fc::json::from_string(test_data);
+      verify_byte_round_trip_conversion(abis, "fees", var);
+   }
+
+   {
+      // check conversion when the first optional members is missing
```
linh2931, line 2009: `first` should be `second` in the comment.

unittests/abi_tests.cpp
```diff
@@ -1929,6 +1937,113 @@ BOOST_AUTO_TEST_CASE(abi_type_loop)
...
+        "bridge_fee" : "2"
+      }
+      )=====";
+
+      auto var = fc::json::from_string(test_data);
+      verify_byte_round_trip_conversion(abis, "fees", var);
+   }
+
+   {
+      // check conversion when all optional members are provided
```
linh2931, line 2022: should be last.

unittests/abi_tests.cpp
```diff
@@ -1929,6 +1937,113 @@ BOOST_AUTO_TEST_CASE(abi_type_loop)
...
+        "miner_cut" : "2",
+      }
+      )=====";
+
+      auto var = fc::json::from_string(test_data);
+      verify_byte_round_trip_conversion(abis, "fees", var);
+   }
+
+   {
+      // check conversion when all optional members are provided
```
linh2931, line 2035: should be none.

linh2931 (changes requested):

unittests/abi_tests.cpp
greg7mdp: Thanks, Lin, I updated all the comments.

linh2931 (approved):

PR #1366, opened 2023-07-03
[4.0] Close connection on aysnc_read with a closed socket

### LINKED ISSUE #1349: 4.0.3: p2p connections get dropped frequently
Looking at nodeos logs on 4.0.3, p2p connections between 4.0.3 nodes are getting dropped frequently.  Downgrading to 3.2.3 on a single node seems to improve things. Exact scenario is not clear.

This was noticed because BP node is producing blocks with low number of transactions.  When downgrading to 3.2.3 on the BP node, the transactions went back to the number expected.

heifner:
More paranoid socket shutdown when closing and `async_read` with a closed socket. Also cleanup logic on connection duplicate check. Reset `org` as it needs to be reset on close so new connection will `send_time` message correctly.

Resolves #1349

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1242,7 +1242,7 @@ namespace eosio {
             try {
                c->buffer_queue.clear_out_queue();
                // May have closed connection and cleared buffer_queue
-               if( !c->socket_is_open() || socket != c->socket ) {
+               if( !c->socket->is_open() || socket != c->socket ) {
                   peer_ilog( c, "async write socket ${r} before callback", ("r", c->socket_is_open() ? "changed" : "closed") );
                   c->close();
                   return;
@@ -2488,7 +2488,7 @@ namespace eosio {
             boost::asio::bind_executor( strand,
               [conn = shared_from_this(), socket=socket]( boost::system::error_code ec, std::size_t bytes_transferred ) {
                // may have closed connection and cleared pending_message_buffer
-               if (!conn->socket_is_open() && conn->socket_open) { // if socket_open then close not called
+               if (!conn->socket->is_open() && conn->socket_is_open()) { // if socket_open then close not called
                   peer_dlog( conn, "async_read socket not open, closing");
                   conn->close();
                   return;
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1050,6 +1050,7 @@ namespace eosio {
       self->cancel_wait();
       self->latest_msg_time = std::chrono::system_clock::time_point::min();
       self->latest_blk_time = std::chrono::system_clock::time_point::min();
+      self->org = std::chrono::nanoseconds{0};
 
       if( reconnect && !shutdown ) {
          my_impl->start_conn_timer( std::chrono::milliseconds( 100 ), connection_wptr() );
@@ -2851,6 +2852,7 @@ namespace eosio {
       peer_lib_num = msg.last_irreversible_block_num;
       std::unique_lock<std::mutex> g_conn( conn_mtx );
       last_handshake_recv = msg;
+      auto c_time = last_handshake_sent.time;
       g_conn.unlock();
 
       connecting = false;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2489,7 +2488,18 @@ namespace eosio {
             boost::asio::bind_executor( strand,
               [conn = shared_from_this(), socket=socket]( boost::system::error_code ec, std::size_t bytes_transferred ) {
                // may have closed connection and cleared pending_message_buffer
-               if( !conn->socket_is_open() || socket != conn->socket ) return;
+               if (!conn->socket_is_open() && conn->socket_open) { // if socket_open then close not called
+                  peer_dlog( conn, "async_read socket not open, closing");
+                  conn->close();
```
greg7mdp, line 2493: This will set a 100ms timer to attempt a reconnection. Is this what we want?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2489,7 +2488,18 @@ namespace eosio {
             boost::asio::bind_executor( strand,
               [conn = shared_from_this(), socket=socket]( boost::system::error_code ec, std::size_t bytes_transferred ) {
                // may have closed connection and cleared pending_message_buffer
-               if( !conn->socket_is_open() || socket != conn->socket ) return;
+               if (!conn->socket_is_open() && conn->socket_open) { // if socket_open then close not called
```
greg7mdp, line 2491: This seems wrong. `conn->socket_is_open()` and `conn->socket_open` return the same `std::atomic` bool. So unless the strand gets interrupted between the two checks, the condition willnever be true.

greg7mdp (changes requested):

plugins/net_plugin/net_plugin.cpp
heifner: Yes.

plugins/net_plugin/net_plugin.cpp
heifner: Thanks. I was reading that as `socket->is_open()`.

greg7mdp:
This seems incorrect (should we call `c->close()` if `!c->socket_is_open()`): https://github.com/AntelopeIO/leap/blob/d46d7e903ad51bca5ec1584b12b1fd33525a5766/plugins/net_plugin/net_plugin.cpp#L1245-L1249

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1243,7 +1242,7 @@ namespace eosio {
             try {
                c->buffer_queue.clear_out_queue();
                // May have closed connection and cleared buffer_queue
-               if( !c->socket_is_open() || socket != c->socket ) {
+               if( !c->socket->is_open() || socket != c->socket ) {
                   peer_ilog( c, "async write socket ${r} before callback", ("r", c->socket_is_open() ? "changed" : "closed") );
                   c->close();
```
greg7mdp, line 1247: Shouldn't we close `socket` here, instead of `c->socket`? as in calling `socket->close(ec);`.

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2842,6 +2859,7 @@ namespace eosio {
       peer_lib_num = msg.last_irreversible_block_num;
       std::unique_lock<std::mutex> g_conn( conn_mtx );
       last_handshake_recv = msg;
+      auto c_time = last_handshake_sent.time;
```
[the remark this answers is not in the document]
heifner: Used on line 2908.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2489,7 +2495,18 @@ namespace eosio {
...
+               if (!conn->socket->is_open() && conn->socket_is_open()) { // if socket_open then close not called
+                  peer_dlog( conn, "async_read socket not open, closing");
+                  conn->close();
+                  return;
+               }
+               if (socket != conn->socket ) { // different socket, conn must have created a new socket, make sure previous is closed
+                  peer_dlog( conn, "async_read diff socket closing");
+                  boost::system::error_code ec;
+                  socket->shutdown( tcp::socket::shutdown_both, ec );
+                  socket->close( ec );
```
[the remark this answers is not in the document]
heifner: The docs say that the destructor calls `cancel()` but it does not specify what if any shutdown is performed. Granted I do think is rather paranoid as it is likely that the destructor does clean up the connection.

dimas1185 (approved):

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
dimas1185: it takes time to find an answer, here is what I found:
https://stackoverflow.com/questions/39813066/boost-asio-socket-destructor-closes-connection

however we need to understand why do they recommend this `shutfdown` to be called. maybe because of reading? but we are not concerned about this. So I'm not hard on that just wanted to save few lines, net plugin code is quite verbose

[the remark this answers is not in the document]
dimas1185: another good solution if you have some time and inspiration is to add custom deleter to shared ptr. that way we can save many duplicate code.

PR #1370, opened 2023-07-04
Make sure all exceptions are caught when executing code in http_plugin's thread pool, so Leap doesn't crash when processing APIs.

### LINKED ISSUE #1372: Guard Against crashes in HTTP when making addition nodeos calls
Don't crash when making calls to nodeos in service on an HTTP request. Interactions should be treated as suspect, and a failure should be returned to the client.

We talked with Guillaume. Turn out he was using a build from a fork which had pulled a version post 4.0.3 by mistake:
image

Checked client [repo](https://github.com/fcecin/leap/tree/ux4.0.3) and indeed they had a [change](https://github.com/fcecin/leap/blob/c868094d8f58c07af8809421a29b607d6865d38b/plugins/chain_api_plugin/chain_api_plugin.cpp#L143) not in our release/4.0 branch, which causes the issue.

They will rebase their repo on the right tag (4.0.3) which will solve the issue for them.

The change causing the leap crash has not been released (it is in our main branch), and I have a fix for it. So in a sense the mistake they made was lucky for us, as it allowed us to catch the leap issue before the code was officially released.

To be 100% clear, no 4.0.x leap release is subject to the crash in http_plugin.

greg7mdp:
Resolves #1372 

Make sure all exceptions are caught when executing code in http_plugin's thread pool.

Add section in `tests/nodeos_run_test.py` to verify that an exception when processing the tail end of `get_table_row` in the http thread pool is caught and doesn't kill nodeos.

Also removed duplicated test (same test run twice) in `tests/CMakeLists.txt`


ericpassmore:
please update resolving issue to #1372

greg7mdp:
OK, I have added the test. It passes with this PR, but without the change in macros.hpp, it fails as shown in the image below: 
image

linh2931 (approved):

PR #1374, opened 2023-07-05
[4.0 -> main] Fix incorrect serializing of std::optional when value is not provided - (GH #1189)

### LINKED ISSUE #1189: abi_serializer: Missing field exception thrown when an std::optional is used in a field of an action input parameter

greg7mdp:
Resolves #1189 .

When serializing a std::optional field, and a value is not provided for the std::optional, we need to add into the serialization the flag specifying that a missing value is serialized. Prior to this PR, the flag was missing.

### DIFF
--- a/unittests/abi_tests.cpp
+++ b/unittests/abi_tests.cpp
@@ -41,7 +41,17 @@ FC_REFLECT(act_sig, (sig) )
 
 BOOST_AUTO_TEST_SUITE(abi_tests)
 
+#ifdef NDEBUG
 fc::microseconds max_serialization_time = fc::seconds(1); // some test machines are very slow
+#else
+fc::microseconds max_serialization_time = fc::microseconds::maximum(); // don't check in debug builds
+#endif
+
+static fc::time_point get_deadline() {
+   if (max_serialization_time == fc::microseconds::maximum())
+      return fc::time_point(fc::microseconds::maximum());
+   return fc::time_point::now() + max_serialization_time;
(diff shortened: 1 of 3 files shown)


unittests/abi_tests.cpp
```diff
@@ -41,7 +41,17 @@ FC_REFLECT(act_sig, (sig) )
...
+#ifdef NDEBUG
 fc::microseconds max_serialization_time = fc::seconds(1); // some test machines are very slow
+#else
+fc::microseconds max_serialization_time = fc::microseconds::maximum(); // don't check in debug builds
+#endif
+
+static fc::time_point get_deadline() {
+   if (max_serialization_time == fc::microseconds::maximum())
+      return fc::time_point(fc::microseconds::maximum());
+   return fc::time_point::now() + max_serialization_time;
```
heifner, line 53: You can use `safe_add` in `main`.

heifner (approved):

unittests/abi_tests.cpp
greg7mdp: Thanks. I also did a small cleanup in `safe_add()`

heifner (approved):

linh2931 (approved):

PR #1378, opened 2023-07-07
[PH] Create workflow for manually dispatching performance harness runs

oschwaldp-oci:
Initial attempt of configurable ph run in cicd.

### `platforms.yaml`

Split out `Discover Platforms`, `Build Platforms`, and `Platform Matrix` out into one reusable workflow.  `Platform Matrix` is a new job that dynamically creates the `matrix: platform:` array from the discovered platforms or can be overridden by the input `override-build-matrix`.  This reduces the maintenance overhead of having to update the `matrix: platform` arrays in the workflows anytime a platform is added or dropped.

### `build_base.yaml`
Removes steps now contained in `platforms.yaml` workflow but takes as input `p`, the discovered platforms, and `platform-matrix`. Typically will be the output of the `platforms.yaml` workflow.

### `build.yaml`
Updated to make use of `platforms.yaml` and `build_base.yaml` workflows.

### `ph_backward_compatibility.yaml`
Updated to make use of `platforms.yaml` and `build_base.yaml` workflows.

### `performance_harness_run.yaml`
A manual workflow to configure and run the Performance Harness in CICD.  This workflow will attempt to reuse a build for the current head commit of the branch selected to run on, if it exists, otherwise it will build it from scratch.  It has the following inputs:

- **Use workflow from**
  - Select branch to run on
- **Override build platform**
  - Drop down selection allows `ubuntu20` or `ubuntu22` as the build platform
- **Override perf harness params**
  - Provide any special command line arguments to pass to `performance_test.py`
  - Default (if input field left blank): `testBpOpMode`
  - e.g.
    - `testBpOpMode --max-tps-to-test 50 --test-iteration-min-step 10 --test-iteration-duration-sec 10 --final-iterations-duration-sec 10 overrideBasicTestConfig -v --tps-limit-per-generator 25 --chain-state-db-size-mb 200`
    - `testBpOpMode  --test-iteration-duration-sec 600 --final-iterations-duration-sec 600`


.github/workflows/performance_harness_run.yaml
```diff
@@ -12,10 +25,131 @@ defaults:
...
+            # https://github.com/actions/runner/issues/2033
+            chown -R $(id -u):$(id -g) $PWD
+            cmake -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -GNinja
+            cmake --build build
+            tar -pc --exclude "*.o" build | zstd --long -T0 -9 > build.tar.zst
+        - name: Upload builddir
+          uses: AntelopeIO/upload-artifact-large-chunks-action@v1
+          with:
+            name: ${{github.event.inputs.platform-choice}}-build
+            path: build.tar.zst
```
spoonincode, line 116: Similar comments as https://github.com/AntelopeIO/leap/pull/1356#discussion_r1253679236: we shouldn't be copy pasting all of this, we need to come up with a better approach.

.github/workflows/performance_harness_run.yaml
```diff
@@ -12,10 +25,131 @@ defaults:
...
     steps:
-      - name: Workflow Stub
+      - name: Download builddir
+        uses: actions/download-artifact@v3
+        with:
+          name: ${{github.event.inputs.platform-choice}}-build
+      - name: Run Performance Test
+        run: |
+          # https://github.com/actions/runner/issues/2033  -- need this because of full version label test looking at git revs
+          chown -R $(id -u):$(id -g) $PWD
```
spoonincode, line 136: Is the 2033 workaround really needed in this case? Unless there is a `git` command being run as part of the test, probably not

.github/workflows/performance_harness_run.yaml
```diff
@@ -12,10 +25,131 @@ defaults:
...
+          zstdcat build.tar.zst | tar x
+          cd build
+          ./tests/performance_tests/performance_test.py ${{needs.v.outputs.test-params}}
+      - name: Prepare results
+        id: prep-results
         run: |
-          echo "Workflow Stub"
+          # https://github.com/actions/runner/issues/2033
+          chown -R $(id -u):$(id -g) $PWD
```
spoonincode, line 144: Similar to above -- is 2033 workaround needed here? Seems unlikely since there is just a `tar`

.github/workflows/performance_harness_run.yaml
```diff
@@ -12,10 +25,131 @@ defaults:
...
+  tests:
+    name: Tests
+    needs: [v, d, Build]
+    if: always() && needs.Build.result == 'success'
+    strategy:
+      fail-fast: false
     runs-on: ubuntu-latest
+    container:
+      image: ${{fromJSON(needs.d.outputs.p)[github.event.inputs.platform-choice].image}}
+      options: --security-opt seccomp=unconfined
```
spoonincode, line 127: Does performance harness make use of `unshare`? (does it come along indirectly via TestHarness or something?)

.github/workflows/performance_harness_run.yaml
```diff
@@ -12,10 +25,131 @@ defaults:
...
+          with:
+            name: ${{github.event.inputs.platform-choice}}-build
+            path: build.tar.zst
+
+  tests:
+    name: Tests
+    needs: [v, d, Build]
+    if: always() && needs.Build.result == 'success'
+    strategy:
+      fail-fast: false
```
spoonincode, line 123: Without being a matrix, this is a noop

.github/workflows/performance_harness_run.yaml
```diff
@@ -12,10 +25,131 @@ defaults:
...
     runs-on: ubuntu-latest
```
spoonincode, line 124: This means it runs on a github owned runner, is that what you wanted?

spoonincode (changes requested):

.github/workflows/performance_harness_run.yaml
oschwaldp-oci: Addressed: 6b4f7452093ca4

.github/workflows/performance_harness_run.yaml
oschwaldp-oci: Addressed: 6b4f7452093ca4

.github/workflows/performance_harness_run.yaml
oschwaldp-oci: Addressed: 6b4f7452093ca4

.github/workflows/performance_harness_run.yaml
oschwaldp-oci: Addressed: 6b4f7452093ca4

.github/workflows/performance_harness_run.yaml
oschwaldp-oci: @spoonincode  - Correct. I was using the github owned runner for initial testing and verification of the workflow.  Now that this is closer to complete, do you have a suggestion for a runner to use (`enf-x86-beefy-long`)?  I know that most of the enf runners have time limits on them which could be problematic for performance runs depending on how a user configures the run.

.github/workflows/performance_harness_run.yaml
oschwaldp-oci: @spoonincode - I have refactored the build to be a reusable workflow in [PR#1356]( https://github.com/AntelopeIO/leap/pull/1356).  Hoping to have that PR go into `main` first, then I will pull those changes into this branch and refactor the workflow to use `build-base` as well.

.github/workflows/build_base.yaml
```diff
@@ -3,6 +3,12 @@ name: "Build leap"
 on:
   workflow_dispatch:
   workflow_call:
+    inputs:
+      override-build-matrix:
+        description: 'Override build matrix'
```
spoonincode, line 8: What's driving the need for adding this? You only want to build leap for a single platform when doing the performance harness test?

oschwaldp-oci: Correct.

spoonincode: Not sure it's worth the added complexity in `build_base.yaml` to accommodate that. Are you doing it just to save a dollar? I'd rather take the cost and not have 30+ more lines of yaml here to maintain forever, especially because PH runs aren't regularly done anyways.

If you want to save cost I'd suggest pushing off the complexity in to the `performance_harness_run.yaml` by doing something like trying to fetch an existing built leap.deb & leap-dev.deb, and then only falling back to building leap if they don't exist.

.github/workflows/build_base.yaml
```diff
@@ -58,19 +67,51 @@ jobs:
...
       fail-fast: false
       matrix:
-        platform: [ubuntu20, ubuntu22]
+        platform: ${{fromJSON(needs.pm.outputs.platform-matrix)}}
     runs-on: ["self-hosted", "enf-x86-beefy"]
     container: ${{fromJSON(needs.d.outputs.p)[matrix.platform].image}}
     steps:
         - uses: actions/checkout@v3
           with:
+            ref: '${{inputs.leap-ref}}'
```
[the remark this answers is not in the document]
oschwaldp-oci: I am not sure I really had this working correctly, as you observed. I have removed this feature.

.github/workflows/build.yaml
```diff
@@ -31,9 +31,22 @@ defaults:
     shell: bash
 
 jobs:
+  platforms:
+    name: Run Platforms Workflow
+    uses: ./.github/workflows/platforms.yaml
+    with:
+      override-build-matrix: ${{github.event.inputs.platform-choice}}
```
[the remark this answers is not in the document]
oschwaldp-oci: I thought I had removed that.  Somehow lost in transition.  It is unnecessary in the `build.yaml` workflow as that workflow should always be expecting to use the full input from `platforms.json` to determine the build platforms to support. Removed here: [bf95111](https://github.com/AntelopeIO/leap/pull/1378/commits/bf95111e0117f1ee653960d7e53855c17d45c01e)

.github/workflows/build_base.yaml
spoonincode: yeah, not having to copy pasta around the `[ubuntuxx..]` is a good improvement. I'm a little worried we're going to have to refactor much of this soon again though because while right now there is a 1:1 between build platforms and test platforms that won't always be the case. That is, right now `ubuntu22` means both building and testing on `ubuntu22` but that may not always be the case. For example, for our reproducible builds we may have a separate `repro` platform that is used for builds, but then tested on _both_ `ubuntu20` and `ubuntu22`. Or ARM builds may have a `crossarm` platform for builds that then runs the tests on `ubuntu20arm` or something. I can't quite wrap my head around the semantics of how that is going to work just yet (besides adding more complexity :disappointed:) so what is in this PR is good for now on that front.

.github/workflows/performance_harness_run.yaml
```diff
@@ -2,6 +2,16 @@ name: "Performance Harness Run"
 
 on:
   workflow_dispatch:
+    inputs:
+      platform-choice:
+        description: 'Override build platform'
```
spoonincode, line 7: I guess this isn't really overriding the platform but rather picking which platform

.github/workflows/performance_harness_run.yaml
```diff
@@ -12,10 +22,102 @@ defaults:
...
+  v:
+    name: Discover Inputs
+    runs-on: ubuntu-latest
+    outputs:
+      test-params: ${{steps.overrides.outputs.test-params}}
+    steps:
+      - name: Setup Input Params
+        id: overrides
+        env:
+          GH_TOKEN: ${{secrets.GITHUB_TOKEN}}
```
spoonincode, line 35: this used any where?

.github/workflows/build_base.yaml
oschwaldp-oci: That's a good point. There is definitely a complexity addition that will come when that occurs.  It could be that the `Platform Matrix` job provides a nice location to do the platform setup though and could be expanded to help with that?  Might be that it breaks out into its own workflow.  I agree that it is hard to say what will work best until some of those decisions are made.  I'll consider this resolved for now per your comment.

.github/workflows/performance_harness_run.yaml
oschwaldp-oci: Addressed: [8c47cce](https://github.com/AntelopeIO/leap/pull/1378/commits/8c47cce233698dd4801896b7727eeb09188057a9)

.github/workflows/platforms.yaml
```diff
@@ -0,0 +1,94 @@
...
+
+          if [[ "${{inputs.override-build-matrix}}" != "" ]]; then
+              echo "Executing override."
+              echo 'platform-matrix=["${{inputs.override-build-matrix}}"]' >> $GITHUB_OUTPUT
+          fi
+      - name: Check matrix      
+        id: check-matrix
+        run: |
+          echo '${{ steps.pm-results.outputs.platform-matrix }}'
+          echo '${{ steps.parse-pm.outputs.result }}'
```
spoonincode, line 94: quite a few `echo`s in this workflow that look kind of like debugging, did you mean to leave them in?

oschwaldp-oci: Good catch. Thought I had caught all of these and removed them. Thanks.

Addressed: [cd872b3](https://github.com/AntelopeIO/leap/pull/1378/commits/cd872b3b27580866ee09c335b1475b018a31b725)

.github/workflows/performance_harness_run.yaml
```diff
@@ -12,10 +22,100 @@ defaults:
...
+          id: downloadBuild
+          uses: AntelopeIO/asset-artifact-download-action@v3
+          with:
+            owner: AntelopeIO
+            repo: leap
+            file: build.tar.zst
+            target: ${{github.sha}}
+            artifact-name: ${{github.event.inputs.platform-choice}}-build
+            fail-on-missing-target: false
+            token: ${{github.token}}
```
spoonincode, line 70: with `v3` you don't need this line, if you want to save a line

.github/workflows/performance_harness_run.yaml
```diff
@@ -12,10 +22,100 @@ defaults:
...
+      p: ${{needs.platforms.outputs.p}}
+      platform-matrix: ${{needs.platforms.outputs.platform-matrix}}
+    permissions:
+      packages: write
+      contents: read
+
+  tests:
+    name: Tests
+    needs: [v, platforms, reuse-build, build-base]
+    if: always() && needs.platforms.result == 'success' && (needs.build-base.result == 'success' || needs.reuse-build.result == 'success')
```
spoonincode, line 94: we really need to find a way to get #283 in good shape to avoid these tricky `always()` chains

spoonincode (approved):

.github/workflows/performance_harness_run.yaml
oschwaldp-oci: Done: [d50381b](https://github.com/AntelopeIO/leap/pull/1378/commits/d50381b037be24b1d66b006ab4eb897729613884)

PR #1386, opened 2023-07-11
Signal accepted_block after it is marked valid

### LINKED ISSUE #1275: accepted_block signals non-validated blocks
`controller` emits the `accepted_block` signal for produced and validated blocks in `controller::commit_block`. For produced blocks the block is [marked valid](https://github.com/AntelopeIO/leap/blob/main/libraries/chain/controller.cpp#L1949)  before the emit of `accepted_block`. For validated blocks, the block is [marked valid](https://github.com/AntelopeIO/leap/blob/main/libraries/chain/controller.cpp#L2339) after `controller::commit_block` in `controller::apply_block`. So for validated blocks, we emit non-validated (from the `forkdb` perspective) blocks.

The marking of the block as valid in `forkdb` after it is emitted can lead to errors and confusion. See, for example, #1272.

If a block is not marked valid in `forkdb` then it can't be found via `controller::fetch_block_state_by_number`. If a plugin attempts to find the `accepted_block` in `forkdb`, it can fail as it has not been marked valid.

We should only signal `accepted_block` after the block has been marked valid in `forkdb`.

heifner:
As indicated in #1275, produced blocks were signaled to plugins after being marked valid and placed in the forkdb but signaled before being marked valid and placed in the forkdb for validated blocks. This PR makes sure that the block is marked valid before being signaled for both produced blocks and validated blocks. Also refactors location of updating `head` for applied blocks, consolidating it in `commit_block`.

Resolves #1275


linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -1940,24 +1938,26 @@ struct controller_impl {
...
-            EOS_ASSERT( bsp == head, fork_database_exception, "committed block did not become the new head in fork database");
+            EOS_ASSERT( bsp == fork_db.head(), fork_database_exception, "committed block did not become the new head in fork database");
+         } else if (s != controller::block_status::irreversible) {
+            fork_db.mark_valid( bsp );
          }
+         head = bsp;
 
          // at block level, no transaction specific logging is possible
-         if (auto dm_logger = get_deep_mind_logger(false)) {
+         if (auto* dm_logger = get_deep_mind_logger(false)) {
```
greg7mdp, line 1954: I don't see the benefit of adding the `*` here? I could see it if it was `const auto*` maybe. What am I missing?

greg7mdp (approved):

libraries/chain/controller.cpp
heifner: `on_accepted_block()` is not `const` although it should be. the `*` was a clang tidy popup recommendation that I just accepted. I think it likes the `auto*` over `auto` as it makes it more explicit that it is a pointer.

PR #1388, opened 2023-07-11
[4.0] Test: Fix ship test to wait on correct transaction

### LINKED ISSUE #1383: Test Failure: `ship_streamer_test`
https://github.com/AntelopeIO/leap/actions/runs/5510592746/jobs/10045134304

heifner:
`state_history_plugin` test `ship_streamer_test.py` was not waiting on the correct transaction to be included in a block.

Resolves #1383

### DIFF
--- a/tests/ship_streamer_test.py
+++ b/tests/ship_streamer_test.py
@@ -137,7 +137,7 @@ try:
     nonProdNode.waitForTransBlockIfNeeded(trans, True, exitOnError=True)
     for account in accounts:
         Print(f"Transfer funds {transferAmount} from account {cluster.eosioAccount.name} to {account.name}")
-        nonProdNode.transferFunds(cluster.eosioAccount, account, transferAmount, "test transfer", waitForTransBlock=False)
+        trans=nonProdNode.transferFunds(cluster.eosioAccount, account, transferAmount, "test transfer", waitForTransBlock=False)
     nonProdNode.waitForTransBlockIfNeeded(trans, True, exitOnError=True)
     for account in accounts:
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

tests/ship_streamer_test.py
```diff
@@ -137,7 +137,7 @@ def getLatestSnapshot(nodeId):
     nonProdNode.waitForTransBlockIfNeeded(trans, True, exitOnError=True)
     for account in accounts:
         Print(f"Transfer funds {transferAmount} from account {cluster.eosioAccount.name} to {account.name}")
-        nonProdNode.transferFunds(cluster.eosioAccount, account, transferAmount, "test transfer", waitForTransBlock=False)
+        trans=nonProdNode.transferFunds(cluster.eosioAccount, account, transferAmount, "test transfer", waitForTransBlock=False)
```
spoonincode, line 140: This clobbers `trans` for each transfer, thus only waiting on the last one. I guess no material problem?

spoonincode (approved):

tests/ship_streamer_test.py
heifner: That is the idea. Only wait on the last one. Although, you could argue that it would be best to verify they all make it, but this is certainly faster than that.

spoonincode: > Although, you could argue that it would be best to verify they all make it,

yeah.. that's kind of where I was thinking. But it seems unnecessary for this test

linh2931 (approved):

PR #1392, opened 2023-07-12
[3.1] Memory issue fix

linh2931:
This PR contains security updates to fix a denial of service vulnerability affecting all previous versions of Leap.


heifner (approved):
Would have expected the version bump in a diff PR.

linh2931:
> Would have expected the version bump in a diff PR.

Should I redo this? Or keep as is, but for 3.2 and 4.0, the version bump is in a different PR?

heifner:
> > Would have expected the version bump in a diff PR.
> 
> Should I redo this? Or keep as is, but for 3.2 and 4.0, the version bump is in a different PR?

I don't think it is a huge deal either way. But in general I think it better to keep it as a diff PR.

linh2931:
> I don't think it is a huge deal either way. But in general I think it better to keep it as a diff PR.

I think I should redo it so to make the history trail cleaner. I am closing this PR and starting a new one. Thanks.

Issue #1417, opened 2023-07-19
Block trx idle not being reported correctly


heifner:
`Block #321172855 trx idle: 699846us out of 705689us, success: 307, 52380us, fail: 29, 17200us, transient: 0, 0us, other: -63737us` clearly something is wrong. None of these values should ever be negative.

PR #1434, opened 2023-07-20
[4.0] Fix block time reporting log statement

### LINKED ISSUE #1360: Add producer name to trx idle log message
In debugging https://github.com/AntelopeIO/leap/issues/1349, it was observed by Kevin that it would be good to log the producer name for the following message:

```
Block #317926278 trx idle: 237us out of 199964us, success: 3, 578us, fail: 230, 197119us, transient: 0, 0us, other: 2030us
```
producer_plugin.cpp:290 in leap 4.0.3

This extra information would allow creating charts per block producer on https://blockchain.stats.eosnation.io

heifner:
Fix issue with the tracking of block idle time.
Add producer name to the block time log statement.

Resolves #1360 
Resoves #1417

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -278,12 +278,11 @@ struct block_time_tracker {
       }
    }
 
-   void report( const fc::time_point& idle_trx_time, uint32_t block_num ) {
+   void report(uint32_t block_num, account_name producer) {
       if( _log.is_enabled( fc::log_level::debug ) ) {
          auto now = fc::time_point::now();
-         add_idle_time( now - idle_trx_time );
-         fc_dlog( _log, "Block #${n} trx idle: ${i}us out of ${t}us, success: ${sn}, ${s}us, fail: ${fn}, ${f}us, transient: ${trans_trx_num}, ${trans_trx_time}us, other: ${o}us",
-                  ("n", block_num)
+         fc_dlog( _log, "Block #${n} ${p} trx idle: ${i}us out of ${t}us, success: ${sn}, ${s}us, fail: ${fn}, ${f}us, transient: ${trans_trx_num}, ${trans_trx_time}us, other: ${o}us",
+                  ("n", block_num)("p", producer)
                   ("i", block_idle_time)("t", now - clear_time)("sn", trx_success_num)("s", trx_success_time)
                   ("fn", trx_fail_num)("f", trx_fail_time)
                   ("trans_trx_num", transient_trx_num)("trans_trx_time", transient_trx_time)
@@ -558,11 +557,18 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       void abort_block() {
          auto& chain = chain_plug->chain();
 
+         std::optional<std::tuple<uint32_t, account_name>> block_info;
          if( chain.is_building_block() ) {
-            _time_tracker.report( _idle_trx_time, chain.pending_block_num() );
+            block_info = std::make_tuple(chain.pending_block_num(), chain.pending_block_producer());
          }
          _unapplied_transactions.add_aborted( chain.abort_block() );
          _subjective_billing.abort_block();
+
+         if (block_info) {
+            auto[block_num, block_producer] = *block_info;
+            _time_tracker.report(block_num, block_producer);
+         }
+         _time_tracker.clear();
          _idle_trx_time = fc::time_point::now();
       }
 
@@ -585,6 +591,8 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
          auto now = fc::time_point::now();
          if (now - block->timestamp < fc::minutes(5) || (blk_num % 1000 == 0)) // only log every 1000 during sync
             fc_dlog(_log, "received incoming block ${n} ${id}", ("n", blk_num)("id", id));
+
+         _time_tracker.add_idle_time(now - _idle_trx_time);
 
          EOS_ASSERT( block->timestamp < (now + fc::seconds( 7 )), block_from_the_future,
                      "received a block from the future, ignoring it: ${id}", ("id", id) );
@@ -2028,7 +2036,6 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
 
       try {
          _account_fails.report_and_clear(hbs->block_num);
-         _time_tracker.clear();
 
          if( !remove_expired_trxs( preprocess_deadline ) )
             return start_block_result::exhausted;
@@ -2624,8 +2631,6 @@ void producer_plugin_impl::schedule_production_loop() {
 
    auto result = start_block();
 
-   _idle_trx_time = fc::time_point::now();
-
    if (result == start_block_result::failed) {
       elog("Failed to start a pending block, will try again later");
       _timer.expires_from_now( boost::posix_time::microseconds( config::block_interval_us  / 10 ));
@@ -2661,6 +2666,8 @@ void producer_plugin_impl::schedule_production_loop() {
    } else {
       fc_dlog(_log, "Speculative Block Created");
    }
+
+   _idle_trx_time = fc::time_point::now();
 }
 
 void producer_plugin_impl::schedule_maybe_produce_block( bool exhausted ) {
@@ -2770,6 +2777,8 @@ static auto maybe_make_debug_time_logger() -> std::optional<decltype(make_debug_
 void producer_plugin_impl::produce_block() {
    //ilog("produce_block ${t}", ("t", fc::time_point::now())); // for testing _produce_time_offset_us
    auto start = fc::time_point::now();
+   _time_tracker.add_idle_time( start - _idle_trx_time );
+
    EOS_ASSERT(in_producing_mode(), producer_exception, "called produce_block while not actually producing");
    chain::controller& chain = chain_plug->chain();
    EOS_ASSERT(chain.is_building_block(), missing_pending_block_state, "pending_block_state does not exist but it should, another plugin may have corrupted it");
@@ -2811,8 +2820,6 @@ void producer_plugin_impl::produce_block() {
 
    block_state_ptr new_bs = chain.head_block_state();
 
-   _time_tracker.report(_idle_trx_time, new_bs->block_num);
-
    br.total_time += fc::time_point::now() - start;
 
    ++_metrics.blocks_produced.value;
@@ -2825,6 +2832,11 @@ void producer_plugin_impl::produce_block() {
         ("count",new_bs->block->transactions.size())("lib",chain.last_irreversible_block_num())
         ("net", br.total_net_usage)("cpu", br.total_cpu_usage_us)("et", br.total_elapsed_time)("tt", br.total_time)
         ("confs", new_bs->header.confirmed));
+
+   _time_tracker.report(new_bs->block_num, new_bs->block->producer);
+   _time_tracker.clear();
+
+   _idle_trx_time = fc::time_point::now();
 }
 
 void producer_plugin::received_block(uint32_t block_num) {


linh2931 (approved):

greg7mdp:
In [this location](https://github.com/AntelopeIO/leap/blob/8d09101fcaaac51ce8dae33c0650028e3fabd612/plugins/producer_plugin/producer_plugin.cpp#L729-L733), I think we do `add_idle_time` twice for the same time interval if we get an exception.

I think it would be more logical, everytime we effectively do the equivalent of:

`_time_tracker.add_idle_time( fc::time_point::now() -  self->_idle_trx_time);`

We should immediately afterwards do: `self->_idle_trx_time = fc::time_point::now();`

This way we can ensure we don't count the same interval twice as idle.

And these two statements should be in a `producer_plugin_impl()` member function.

greg7mdp:
Also `_idle_trx_time` is a bad name as it implies a duration, not a time stamp.

Maybe `_unaccounted_idling_tstamp` would be better, as it is the timestamd when we started being idle, and not accounted in time_tracker.

greg7mdp:
I think a better design would be for the time tracker to hold the `_unaccounted_timestamp` and have a single api like:

```c++
class block_time_tracker
{
   enum class time_usage { idle, success, fail, last };

    void add_time(time_usage u)
    {
        _time[static_cast<int>(u)] += now - _unaccounted_tstamp;
        _unaccounted_tstamp = now;
    }

   std::array<fc::microseconds, static_cast<int>(time_usage::last)> _time;
};
```

heifner:
> In [this location](https://github.com/AntelopeIO/leap/blob/8d09101fcaaac51ce8dae33c0650028e3fabd612/plugins/producer_plugin/producer_plugin.cpp#L729-L733), I think we do `add_idle_time` twice for the same time interval if we get an exception.

No, the only exception here (outside alloc exceptions which we can ignore since those are program exit) is via the `future.get()` as 
`process_incoming_transaction_async()` does not throw. So the idle time is only set once.

heifner:
> We should immediately afterwards do: `self->_idle_trx_time = fc::time_point::now();`

The idle time does not start from there. It starts at the end of processing. Adding it here just makes it less likely you notice the logic is wrong.

heifner:
> Also `_idle_trx_time` is a bad name as it implies a duration, not a time stamp.

Interesting. I guess I have always thought of the term `timestamp` (tstamp) as more of an unix epoch term or as an implementation detail. A duration I would expect to be expressed as `elapsed` or similar.  Maybe `_idle_trx_timepoint` ?

heifner:
> I think a better design would be for the time tracker to hold the `_unaccounted_timestamp` and have a single api like:
> 
> ```c++
> class block_time_tracker
> {
>    enum class time_usage { idle, success, fail, last };
> 
>     void add_time(time_usage u)
>     {
>         auto now = fc::time_point::now();
>         _time[static_cast<int>(u)] += now - _unaccounted_tstamp;
>         _unaccounted_tstamp = now;
>     }
> 
>    std::array<fc::microseconds, static_cast<int>(time_usage::last)> _time;
> };
> ```

A bit of Deja Vu on this. Seems like we had a similar discussion already. I'll see if I can clean up the current impl some.

greg7mdp:
> No, the only exception here (outside alloc exceptions which we can ignore since those are program exit) is via the `future.get()` as `process_incoming_transaction_async()` does not throw. So the idle time is only set once.

I'm not following. Either we can't have exceptions (in which case we can remove the try/catch block and the exception handler), or we can have exceptions (and when we do we will call `add_idle_time` twice with the same parameters). What am I missing?

greg7mdp:
> > We should immediately afterwards do: `self->_idle_trx_time = fc::time_point::now();`
> 
> The idle time does not start from there. It starts at the end of processing. Adding it here just makes it less likely you notice the logic is wrong.

Yes, but when we call `add_idle_time` we add the interval [end of processing, now] to the accounted idle time. We should make sure that next time we call `add_idle_time` we don't take the same interval into account again.

greg7mdp:
> A bit of Deja Vu on this. Seems like we had a similar discussion already. I'll see if I can clean up the current impl some.

haha I already forgot all about it! You have a better memory than I do (not a high bar :-). 
I'm just finding it hard to convince myself that we do update the idle time correctly from looking at the code, and I always try to figure out a way to make the code more obviously correct.

heifner:
> > No, the only exception here (outside alloc exceptions which we can ignore since those are program exit) is via the `future.get()` as `process_incoming_transaction_async()` does not throw. So the idle time is only set once.
> 
> I'm not following. Either we can't have exceptions (in which case we can remove the try/catch block and the exception handler), or we can have exceptions (and when we do we will call `add_idle_time` twice with the same parameters). What am I missing?

The main reason the try-catch is here is for the `future.get()`. This can throw if signature validation fails. 
The other reason is if `process_incoming_transaction_async()` throws a `guard_exception` which means `app.quit()` has been called and the program will soon exit, but we would not want an unhandled exception on the `_thread_pool` `executor`. Although, with recent changes to the `named_thread_pool` to handle uncaught exceptions on the `executor` we could likely get away with just a try-catch around the `future.get()`. But it is still nice that we call `next` in that corner case so that an API trx gets the reason for the failure.

greg7mdp:
> The main reason the try-catch is here is for the `future.get()`. This can throw if signature validation fails. 

So if that happens, we'll call `add_idle_time()` twice with the same parameters, right?

heifner:
@greg7mdp Refactored to use RAII for a simpler interface.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -253,58 +253,112 @@ class account_failures {
...
+ private:
+   void add_success_time(bool is_transient) {
+      auto now = fc::time_point::now();
+      if( is_transient ) {
+         transient_trx_time += now - last_trx_time_point;
+         ++transient_trx_num;
+      } else {
+         trx_success_time += now - last_trx_time_point;
+         ++trx_success_num;
+      }
```
greg7mdp, line 337:
```suggestion
      }
      last_trx_time_point = now;
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -253,58 +253,112 @@ class account_failures {
...
+   void add_fail_time(bool is_transient) {
+      auto now = fc::time_point::now();
+      if( is_transient ) {
+         // transient time includes both success and fail time
+         transient_trx_time += now - last_trx_time_point;
+         ++transient_trx_num;
+      } else {
+         trx_fail_time += now - last_trx_time_point;
+         ++trx_fail_num;
+      }
```
greg7mdp, line 349:
```suggestion
      }
      last_trx_time_point = now;
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -253,58 +253,112 @@ class account_failures {
...
+         case trx_status::success:
+            _block_time_tracker.add_success_time(_is_transient);
+            break;
+         case trx_status::fail:
+            _block_time_tracker.add_fail_time(_is_transient);
+            break;
+         case trx_status::other:
+            break; // just reset timer which happens below
+         }
+         _block_time_tracker.start_idle_time();
```
greg7mdp, line 286:
```suggestion
            _block_time_tracker.start_idle_time(); // just reset timer
            break;
         }
```

greg7mdp (commented):
Nice changes.
I'd prefer with my suggestions below, so that everytime we add a time interval to success/fail/idle time, we also reset the `last_trx_time_point`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -775,12 +832,16 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+               trx_tracker.cancel();
                return true;
             }
 
             const auto block_deadline = calculate_block_deadline( chain.pending_block_time() );
             push_result pr = push_transaction( block_deadline, trx, api_trx, return_failure_trace, next );
 
+            if (!pr.failed) {
+               trx_tracker.trx_success();
+            }
```
greg7mdp, line 847: Should we count exhausted transactions as successful?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2661,6 +2702,8 @@ void producer_plugin_impl::schedule_production_loop() {
    } else {
       fc_dlog(_log, "Speculative Block Created");
    }
+
+   _time_tracker.start_idle_time();
```
greg7mdp, line 2709: Hum, I wonder if it shouldn't be `add_idle_time`, with the change I mentioned before in `add_idle_time` resetting `last_trx_time_point = now; `

plugins/producer_plugin/producer_plugin.cpp
heifner: They didn't fail, so don't want to count them as failed. We could count as `other`, but `other` is suppose to be time not spent evaluating trxs. We count add a whole new category, but that seems like it would just muddy the water. I think success is the best option. Time we spent successfully executing transactions; which is true.

plugins/producer_plugin/producer_plugin.cpp
heifner: I don't understand; there is no idle time to add.

greg7mdp: OK, I see that `start_idle_time();` does:

`last_trx_time_point = now;`

So assuming that `now` != `last_trx_time_point`, what was the time spent between `last_trx_time_point` and `now` used for?

heifner: It was used for start block which includes potentially many trx processing.

heifner: So create a `trx_time_tracker` and immediately `cancel()` it?

heifner: I realize this is not counting the time for trxs in start block. I'll fix that.

greg7mdp: > It was used for start block which includes potentially many trx processing.

If we processed trx, they should have updated `last_trx_time_point` at the end.

greg7mdp: > So create a `trx_time_tracker` and immediately `cancel()` it?

No, just call `add_idle_time()` instead of `start_idle_time()`

heifner: But that is not idle time. It is time spent creating a pending block state, activating protocol features, removing expired trxs, removing expired blacklisted trxs, reposting exhausted read-only trxs. Also potentially includes processing trxs but then and only then is the idle time reset.

greg7mdp: So you want it to go to "other". If so then I'm OK with it. 

I think it would be better to have a counter for "other" (as we have for fail/success/idle) so we can check that `fail + success + idle + other` == `wall time`, but maybe this could be done later.

heifner: It would require many additions of `add_other_time` as we would need to do that in many places in start_block and abort_block.

greg7mdp: Not really, it would be replacing `start_idle_time()` with `add_other_time()` I think.

heifner: Would require `add_other_time()` before `process_unapplied_trxs`,  and `process_incoming_trxs` and once in `abort_block` so not too bad. I'll try adding it.

greg7mdp: We create a `trx_time_tracker` only when we start processing a transaction, right? So could it do `add_other_time` in its constructor?

greg7mdp: BTW I really like the RAII implementation!

heifner: > We create a `trx_time_tracker` only when we start processing a transaction, right? So could it do `add_other_time` in its constructor?

But then you are just back to what it does now except you explicitly track the value instead of just calculate it.

greg7mdp: Hum, it is kind of true, especially with the current version of the code which increases the time_point every time we add a duration. 
But still, if we track the "other" value explicitly, we can verify that the sum of all accounted times is the elapsed time, which may be not true if we were to create `trx_time_tracker` objects on more than one thread for example.

heifner: Added explicit tracking of other. Also added a pause/unpause for when in read-window as it didn't seem correct to call that "other" time.

greg7mdp (approved):
Nice! I think the code is easier to follow now.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -253,58 +253,144 @@ class account_failures {
...
+      trx_time_tracker(trx_time_tracker&&) = default;
+
+      trx_time_tracker() = delete;
+      trx_time_tracker(const trx_time_tracker&) = delete;
+      trx_time_tracker& operator=(const trx_time_tracker&) = delete;
+      trx_time_tracker& operator=(trx_time_tracker&&) = delete;
+
+      void trx_success() { _time_status = time_status::success; }
+
+      // Neither success for fail, will be reported as other
```
linh2931, line 271: `for` should be `nor`.

linh2931 (commented):
Not for this PR. We report timing only in `debug` logging, but we track timing no matter whether logging is enabled or not. Not sure how much time is spent on tracking in non-debugging mode.

PR #1436, opened 2023-07-21
Test Fix: Change default max-transaction-cpu-usage for tests

### LINKED ISSUE #1416: Test Failure: performance_test_basic_read_only_trxs
https://github.com/AntelopeIO/leap/actions/runs/5602344355/jobs/10247564180

Set of system contract failing: ` cmd: /__w/leap/leap/build/bin/cleos --url http://localhost:8788  --wallet-url http://localhost:9899 --no-auto-keosd -v set contract -j  eosio /__w/leap/leap/build/unittests/contracts/eosio.system eosio.system.wasm eosio.system.abi`

The `genesis.json` is using 150ms for `max_transaction_cpu_usage` even though the `tests/performance_tests/genesis.json` has the value of 475ms. The value apparently is being overwritten to 150ms somewhere.

Looks like here: `cfg.add_argument('--max-transaction-cpu-usage', type=int, help='the "max-transaction-cpu-usage" value to use in the genesis.json file', default=150000)` in `launcher.py`.

heifner:
Change the default genesis `max_transaction_cpu_usage` from 150ms to 475ms to allow for slow ci/cd. All the integration tests already used 475ms for `max_transaction_cpu_usage`. This PR changes the performance tests to use the same value. Also changed `max_block_cpu_usage` from 200ms to 500ms since it has to be larger than `max_transaction_cpu_usage`. This also matches the value used for integration tests.

Note #1401 was not sufficient because the value in the performance `genesis.json` was being overwritten by `launcher.py`.

Resolves #1416

### DIFF
--- a/tests/TestHarness/launcher.py
+++ b/tests/TestHarness/launcher.py
@@ -359,7 +359,7 @@ class cluster_generator:
                             'net_usage_leeway': 500,
                             'context_free_discount_net_usage_num': 20,
                             'context_free_discount_net_usage_den': 100,
-                            'max_block_cpu_usage': 200000 if self.args.max_block_cpu_usage is None else self.args.max_block_cpu_usage,
+                            'max_block_cpu_usage': 500000 if self.args.max_block_cpu_usage is None else self.args.max_block_cpu_usage,
                             'target_block_cpu_usage_pct': 1000,
                             'max_transaction_cpu_usage': 475000 if self.args.max_transaction_cpu_usage is None else self.args.max_transaction_cpu_usage,
                             'min_transaction_cpu_usage': 100,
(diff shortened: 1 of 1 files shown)


tests/TestHarness/launcher.py
```diff
@@ -375,8 +375,8 @@ def init_genesis(self):
             with open(genesis_path, 'r') as f:
                 genesis = json.load(f)
         genesis['initial_key'] = self.network.nodes['bios'].keys[0].pubkey
-        genesis['initial_configuration']['max_block_cpu_usage'] = self.args.max_block_cpu_usage
-        genesis['initial_configuration']['max_transaction_cpu_usage'] = self.args.max_transaction_cpu_usage
+        genesis['initial_configuration']['max_block_cpu_usage'] = 500000 if self.args.max_block_cpu_usage is None else self.args.max_block_cpu_usage
+        genesis['initial_configuration']['max_transaction_cpu_usage'] = 475000 if self.args.max_transaction_cpu_usage is None else self.args.max_transaction_cpu_usage
```
oschwaldp-oci, line 379: Instead of duplicating the default value here (magic number), I might make the setting of `genesis['initial_configuration']['max_*']` conditional on whether `self.max_* is not None`.  That way you also don't overwrite what is coming in from the loaded genesis file.

heifner: I don't understand. It is conditional on `self.max_* is not None`.

oschwaldp-oci: `if self.args.max_block_cpu_usage is not None: genesis['initial_configuration']['max_block_cpu_usage'] = self.args.max_block_cpu_usage` allows you to not have to respecify the default that we apply when initializing the genesis not from a file.  Otherwise, we are 1) duplicating the specification of a default value as a magic number 2) Possibly overwriting an intentional configuration option in a loaded `genesis.json` file.

heifner: Done

oschwaldp-oci (approved):

greg7mdp (approved):

Issue #1437, opened 2023-07-21
Test failure: libtester test failures


heifner:
![image](https://github.com/AntelopeIO/leap/assets/3933649/59dd0527-a42b-464e-9545-2ac78e679637)

The version requested by libtester is no longer available.

https://github.com/AntelopeIO/leap/actions/runs/5622859188/job/15237949266

PR #1438, opened 2023-07-21
Test Fix: libtester test failures in CICD

oschwaldp-oci:
Resolves: https://github.com/AntelopeIO/leap/issues/1437

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -91,8 +91,8 @@ jobs:
           cpack
       - name: Install dev package
         run: |
-          apt update && apt upgrade -y
-          apt install -y ./build/leap_*.deb ./build/leap-dev*.deb
+          apt-get update
+          apt-get install -y ./build/leap_*.deb ./build/leap-dev*.deb
       - name: Test using TestHarness
         run: |
           python3 -c "from TestHarness import Cluster"
@@ -255,7 +255,8 @@ jobs:
           token: ${{github.token}}
       - name: Install cdt Packages
         run: |
-          apt install -y ./*.deb
+          apt-get update
+          apt-get install -y ./*.deb
           rm ./*.deb
 
       # Reference Contracts
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -91,7 +91,7 @@ jobs:
           cpack
       - name: Install dev package
         run: |
-          apt-get update
+          apt-get update && apt-get upgrade -y
           apt-get install -y ./build/leap_*.deb ./build/leap-dev*.deb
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yaml
```diff
@@ -255,7 +255,8 @@ jobs:
           token: ${{github.token}}
       - name: Install cdt Packages
         run: |
-          apt install -y ./*.deb
+          apt-get update
```
spoonincode, line 258: there is already another `apt-get update` in the same job earlier, but only for the `deb-install` case
https://github.com/AntelopeIO/leap/blob/9ff0583b33259b965bd11f3bc228cd2a30e7c814/.github/workflows/build.yaml#L236-L239
Could consider a single step near the beginning of the job that always does the `apt-get update`, to avoid situations where it's done twice in the same job, and avoid the duplication.

.github/workflows/build.yaml
```diff
@@ -91,8 +91,8 @@ jobs:
           cpack
       - name: Install dev package
         run: |
-          apt update && apt upgrade -y
-          apt install -y ./build/leap_*.deb ./build/leap-dev*.deb
+          apt-get update
```
spoonincode, line 94: reason for removal of `upgrade`? there was some uncertainly around best practices last time this came up. Not doing the `upgrade` means you could end up with "mixed date" dependency installations: if `foo` depends on `bar`, and `bar` is already installed _from a year ago_, doing `apt update && apt install foo` I believe means the latest version of `foo` will be using a one year old `bar` (unless `foo` has an explicit version dependency)

But, I believe Ubuntu tries to maintain ABI compatibility within a release? That seems potentially hard over the course of 4 years. But maybe.

TLDR it was never clear to me the canonical best practice here; but always doing a `upgrade` seems like the safest option.

spoonincode (approved):

.github/workflows/build.yaml
oschwaldp-oci: Good idea. Done.  Only implemented a new step in the `libtester-tests` job as the other looked currently to be and continue to be a single use scenario.

.github/workflows/build.yaml
oschwaldp-oci: Yeah it is difficult to say. Your comment had me go refresh on this as well.  My reasoning here was that the `upgrade` step can force one's hand into installing a host of updated packages that aren't being used by our use case, whereas relying on the deb package install to take care of meeting its own dependency tree would be safe -- so long as the package index had been updated.  Anyway, for now I've added it back in and we can punt this down the road for future discussion...or not.

heifner (approved):

Issue #1435, opened 2023-07-21
Test failure: plugin_test


heifner:
https://github.com/AntelopeIO/leap/actions/runs/5616780429/job/15219829929

```
terminate called after throwing an instance of 'eosio::chain::block_validate_exception'
  what():  Block exception
```

greg7mdp:
I think it may be related to [this](https://github.com/AntelopeIO/appbase/compare/54cc7fb4f9a3dbbe2c71bdf23f9342c9c01b9673...0f959acba5664e0271d3ba6cf3dc1d99abe2663f#diff-b42b62264fe3e09f617aa1e0e1fac4103c2bf1ba63502e47175f84d4ebdb1c29) change (I made) in include/appbase/application_instance.hpp:42, which was applied when you caught appbase to tip. It was part of my plugin startup/shutdown issue which had changes both in appbase and leap, but only the appbase PR was merged to main.

heifner:
Another failure: https://github.com/AntelopeIO/leap/actions/runs/5622859188/job/15236444961
```
warn  2023-07-21T13:33:39.792 plugin_te chain_plugin.cpp:1110         plugin_initialize    ] 3100000 misc_exception: Miscellaneous exception
failed to read response from monitor process
    {}
    plugin_test  compile_monitor.cpp:326 get_connection_to_compile_monitor

appbase: exception thrown during plugin "eosio::chain_plugin" initialization.
Miscellaneous exception
terminate called after throwing an instance of 'eosio::chain::misc_exception'
  what():  Miscellaneous exception
```

heifner:
Looks like the terminate is because we are not joining the `app_thread` in `test_read_only_trx.cpp` on exception.
```
diff --git a/tests/test_read_only_trx.cpp b/tests/test_read_only_trx.cpp
index db36eea1e9..e8f2574fd9 100644
--- a/tests/test_read_only_trx.cpp
+++ b/tests/test_read_only_trx.cpp
@@ -111,6 +111,10 @@ void test_trxs_common(std::vector<const char*>& specific_args, bool test_disable
       plugin_promise.set_value( {app->find_plugin<producer_plugin>(), app->find_plugin<chain_plugin>()} );
       app->exec();
    } );
+   fc::scoped_exit<std::function<void()>> on_except = [&](){
+      if (app_thread.joinable())
+         app_thread.join();
+   };
 
    auto[prod_plug, chain_plug] = plugin_fut.get();
 
```

[...]

PR #1439, opened 2023-07-21
Test Fix: Join thread on exception

heifner:
Avoid test `terminate` due to non-joined thread. I don't think this is the actual problem of #1435 as the test should not be throwing an exception. However, if it does throw we need to join the thread to prevent the test from just terminating.


linh2931 (approved):

greg7mdp (approved):

heifner:
Planning to merge this as it gets ci/cd moving again. However, still working on what is causing the exception which is the real problem that needs fixed for #1435

heifner:
Closing PR, going to create a different one with actual fix.

PR #1442, opened 2023-07-21
Test Fix: Controller calls only valid when building a block

### LINKED ISSUE #1435: Test failure: plugin_test

heifner:
When interacting with `controller` many calls including `push_transaction` & `preactivate_feature` require a pending block. Update `test_utils.hpp` used by `test_read_only_trx.cpp` to only interact with `controller` when it is building a block (has a pending block).

Included in this PR are try-catches around `application` `exec()` since it can throw and if the exceptions are not caught then the test `terminate`s making it very difficult to find the actual issue with the test.

Resolves #1435

### DIFF
--- a/tests/test_read_only_trx.cpp
+++ b/tests/test_read_only_trx.cpp
@@ -87,7 +87,7 @@ BOOST_AUTO_TEST_CASE(not_check_configs_if_no_read_only_threads) {
    test_configs_common(specific_args, app_init_status::succeeded);
 }
 
-void test_trxs_common(std::vector<const char*>& specific_args, bool test_disable_tierup = false) {
+void test_trxs_common(std::vector<const char*>& specific_args, bool test_disable_tierup = false) { try {
    fc::scoped_exit<std::function<void()>> on_exit = []() {
       chain::wasm_interface_collection::test_disable_tierup = false;
    };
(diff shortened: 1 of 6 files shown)


linh2931 (approved):

tests/test_read_only_trx.cpp
```diff
@@ -102,15 +102,23 @@ void test_trxs_common(std::vector<const char*>& specific_args, bool test_disable
...
+         app->find_plugin<chain_plugin>()->chain();
+         app->startup();
+         plugin_promise.set_value({app->find_plugin<producer_plugin>(), app->find_plugin<chain_plugin>()});
+         app->exec();
+         return;
+      } FC_LOG_AND_DROP()
+      BOOST_CHECK(!"app threw exception see logged error");
    } );
+   fc::scoped_exit<std::function<void()>> on_except = [&](){
+      if (app_thread.joinable())
```
greg7mdp, line 119: Instread of having these two `app_thread.join();`, why not put line 102 to 176 in a block and rely on the `scoped_exit` to join the app_thread?

unittests/test_utils.hpp
```diff
@@ -70,13 +74,42 @@ auto push_input_trx(eosio::chain::controller& control, account_name account, sig
...
+                    trx_promise.set_exception(std::current_exception());
+                 }
+              } else {
+                 trx_promise.set_value(std::get<chain::transaction_trace_ptr>(result));
+              }
+           });
+   });
+
+   if (trx_future.wait_for(std::chrono::seconds(5)) == std::future_status::timeout)
+      throw std::runtime_error("failed to execute trx: " + ptrx->get_transaction().actions.at(0).name.to_string() + " to account: " + account.to_string());
```
greg7mdp, line 106: So we posted on the app_thread a lambda containing references to local variables. And, upon a `5s` timer, we can throw an exception and exit the stack frame holding these local variable. I Think it could be an issue if the app thread executes the lambda afterwards.
Maybe it would be better to `make_shared` the promise and pass a `shared_ptr` to it?

unittests/test_utils.hpp
```diff
@@ -85,56 +118,56 @@ auto set_code(eosio::chain::controller& control, account_name account, const vec
...
-         builtin_protocol_feature_t::wtmsig_block_signatures };
-      for (const auto t : pfs) {
-         auto feature_digest = pfm.get_builtin_digest(t);
-         BOOST_CHECK( feature_digest );
-         chain_plug->chain().preactivate_feature( *feature_digest, false );
-      }
-      feature_promise.set_value();
-   });
+   std::atomic<bool> feature_set = false;
+   // has to execute when pending block is not null
```
greg7mdp, line 128: Same here, I think we should pass a `shared_ptr` to ` std::atomic<bool>` to make sure the lambda doesn't reference a non-existing stack frame.

tests/test_read_only_trx.cpp
heifner: Done.

unittests/test_utils.hpp
heifner: Done.

unittests/test_utils.hpp
heifner: Done.

greg7mdp (approved):

unittests/test_utils.hpp
```diff
@@ -70,13 +74,42 @@ auto push_input_trx(eosio::chain::controller& control, account_name account, sig
       trx.sign(testing::tester::get_private_key(account, "active"), control.get_chain_id());
    }
    auto ptrx = std::make_shared<packed_transaction>( trx, packed_transaction::compression_type::zlib );
-   auto fut = transaction_metadata::start_recover_keys( ptrx, control.get_thread_pool(), control.get_chain_id(), fc::microseconds::maximum(), transaction_metadata::trx_type::input );
-   auto r = control.push_transaction( fut.get(), fc::time_point::maximum(), fc::microseconds::maximum(), 0, false, 0 );
-   return r;
+
+   std::shared_ptr<std::promise<transaction_trace_ptr>> trx_promise = std::make_shared<std::promise<transaction_trace_ptr>>();
+   std::future<transaction_trace_ptr> trx_future = trx_promise->get_future();
```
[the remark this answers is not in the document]
heifner: Done.

tests/test_read_only_trx.cpp
```diff
@@ -87,7 +87,7 @@ BOOST_AUTO_TEST_CASE(not_check_configs_if_no_read_only_threads) {
    test_configs_common(specific_args, app_init_status::succeeded);
 }
 
-void test_trxs_common(std::vector<const char*>& specific_args, bool test_disable_tierup = false) {
+void test_trxs_common(std::vector<const char*>& specific_args, bool test_disable_tierup = false) { try {
```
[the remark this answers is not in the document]
heifner: Done.

greg7mdp (approved):

PR #1448, opened 2023-07-25
Disable building leap-dev.deb package by default

spoonincode:
A leap-dev.deb package is currently generated alongside the leap.deb package when performing a `make package` (as our README suggests). This leap-dev.deb, similarly to the `make dev-install` target, contains the files required for "native contract unit testing" a.k.a. libtester. afaik we don’t document this leap-dev.deb, nor do we include it as part of leap’s release assets. Instead, it’s used for punting around libtester in various Antelope CI use cases so that leap doesn’t need to always be rebuilt; for example, eos-system-contracts CI uses leap-dev.deb for its unit tests so that it doesn’t need to build leap just to run the system contract unit tests.

Unfortunately adding boost to leap-dev.deb is increasing the time it takes to perform `make package` significantly: from 30 seconds to 7 minutes. If we leave the README instructions as `make package`, it’ll take some users longer to build the .deb packages than building the software!

Because the sole use case of leap-dev.deb is Antelope CI usage, disable building leap-dev.deb by default, and just enable it in leap’s CI.

### DIFF
--- a/.github/workflows/build_base.yaml
+++ b/.github/workflows/build_base.yaml
@@ -77,7 +77,7 @@ jobs:
           run: |
             # https://github.com/actions/runner/issues/2033
             chown -R $(id -u):$(id -g) $PWD
-            cmake -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/usr -GNinja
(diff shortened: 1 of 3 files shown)


heifner (approved):

greg7mdp (approved):

greg7mdp:
@spoonincode do you want me to merge this commit, and then catchup the `boost_submodule` branch with `main`?

spoonincode:
Either way is/was fine with me; it's more of a "suggest changes" kind of PR that doesn't even require a review for you to decide to merge it in.

PR #1458, opened 2023-07-27
Update C++ standard to C++20

### LINKED ISSUE #1079: Change the default leap C++ compilation flag to `-std=c++20`
#### Prerequisites:

- [x] #1084
- [x] #922
- [ ] #1121 via (#1116 OR #1192)
- [ ] #1142
- [ ] Start new version of nodeos (compiled with C++20) from old version (generated by nodeos compiled with C++17) state, check integrity hash, ensure it can still sync a few blocks afterward. (Run this test with pinned clang-11 and also gcc-10, or whatever minimum versions we decide in #1084.)

#### llvm issue background

This is a known build issue with the LLVM headers (gcc-12, -std=c++20), see [this](https://stackoverflow.com/questions/67698523/error-in-llvm-stlextras-expected-unqualified-id-before-const-with-gcc11).

Manually changing /usr/include/llvm-11/llvm/ADT/STLExtras.h to remove the unneeded `<R>` allows it to build fine.

Issue is fixed in [LLVM-13](https://github.com/llvm/llvm-project/blob/release/13.x/llvm/include/llvm/ADT/STLExtras.h#L1867)

#### Possible way forward

Investigate what part of leap depends on those llvm headers, and possibly keep building these with `-std=c++17`.

greg7mdp:
Resolves #1079.

Interestingly, I tried setting the standard to `23` and I verified that Leap buids with the 3 compilers I tried (gcc-10, gcc-12 and clang-16).

Should we set the standard to c++23? It would not buy us very much if the older compiler we support is gcc-10.


heifner (changes requested):
Looks like need to fix:
```
../libraries/libfc/libraries/bn256/src/array.h:5:13: fatal error: span: No such file or directory
    5 |    #include <span>
```

spoonincode (changes requested):
These need to be changed
https://github.com/AntelopeIO/leap/blob/1f51211475c40b5af9d6e15c82e7f7247117dbdf/CMakeModules/EosioTester.cmake.in#L23
https://github.com/AntelopeIO/leap/blob/1f51211475c40b5af9d6e15c82e7f7247117dbdf/CMakeModules/EosioTesterBuild.cmake.in#L21

spoonincode:
> Looks like need to fix:

tbc this means we need to use `gcc-10` on ubuntu 20

README.md
```diff
@@ -140,6 +139,15 @@ sudo apt-get install -y \
         file \
         zlib1g-dev
 ```
+
+On Ubuntu 20.04, update to gcc-10 which has C++20 support:
+```bash
+sudo apt-get install -y gcc-10 g++-10 cpp-10
+sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-10 100 --slave \
```
spoonincode, line 146: I don't think we should make system-wide global alternations to a user's environment like this. Can we just document it like we did in the past for Ubuntu 18? (`CC=gcc-10 CXX=g++-10 cmake ...`)

greg7mdp: done

README.md
```diff
@@ -140,11 +139,23 @@ sudo apt-get install -y \
         file \
         zlib1g-dev
 ```
+
+On Ubuntu 20.04, install gcc-10 which has C++20 support:
+```bash
+sudo apt-get install -y gcc-10 g++-10 cpp-10
```
spoonincode, line 145: `g++-10` is all you need, since it depends on `gcc-10`, which also then depends on `cpp-10`

spoonincode (approved):

heifner (approved):

PR #1460, opened 2023-07-27
`.gitignore` cleanup - mostly stop ignoring `*.cmake` files.

greg7mdp:
- stop ignoring `*.cmake` files
- remove unneeded exclusion `!.cicd`
- ignore clangd `.cache` directory
- ignore `snapshots` directories


.gitignore
```diff
@@ -94,3 +91,5 @@ CMakeLists.txt.user
 
 node_modules
 package-lock.json
+
+snapshots
```
spoonincode, line 95: when is this one a problem?

.gitignore
```diff
@@ -7,23 +7,20 @@
 *.wast.hpp
 *.dot
 *.abi.hpp
-*.cmake
```
spoonincode, line 10: I don't quite understand the removal of this one...  won't it cause all the various cmake files in the build dir be flagged? Like `CPackSourceConfig.cmake` and `CTestTestfile.cmake` etc etc?

spoonincode (commented):
honestly, imo, this file should be completely removed

.gitignore
greg7mdp: We have this rule in `.gitignore`:

`[Bb]uild*/*`

.gitignore
greg7mdp: Not sure where these come from, I just saw these show up in my "Untracked" files:

```
~/github/enf/leap gitignore_cleanup:main ⇡5 !1 ?14 ❯ ls snapshots plugins/snapshots                                                                                                             
plugins/snapshots:
snapshot-000000062d3e86cb288ecdfd37e4ba5f313b97c04b3c8f47e010275c4a37fe62.bin

snapshots:
snapshot-000000062d3e86cb288ecdfd37e4ba5f313b97c04b3c8f47e010275c4a37fe62.bin  snapshot-00000006c38bac7f934ac5cf688d8530bac0d112e0404548e391aef2b404bda2.bin
```

.gitignore
spoonincode: ah, true. but if the gitignore only supports build directories named `build*` or `cmake-build-{release,debug}`, then quite a bit of other cruft can likely be removed too

greg7mdp:
> honestly, imo, this file should be completely removed

which file? `.gitignore`?

.gitignore
greg7mdp: I'm all for cleaning it up further. I think we should assume that the build dirs are already excluded, as this is normally the case

.gitignore
```diff
@@ -7,23 +7,20 @@
...
-!package.cmake
-!CMakeModules/*.cmake
 *.ninja
 \#*
 \.#*
 CMakeCache.txt
 CMakeFiles
-cmake_install.cmake
 cmake-build-debug/
 cmake-build-release/
```
heifner, line 16: Can we change these to `cmake-build-*/` please. I have for example `cmake-build-debug-clang11/`.

spoonincode (approved):
> > honestly, imo, this file should be completely removed
> 
> which file? `.gitignore`?

yeah, I usually abstain from .gitignore reviews because I'd rather just have the file removed. I'm only here on this one since you tagged me :innocent:

heifner (approved):

PR #1465, opened 2023-07-28
Set `shallow` flag for boost submodule (to speedup cloning)

greg7mdp:


greg7mdp:
Doesn't seem to make a difference in boost submodules update time.

PR #1467, opened 2023-07-28
SHiP: Auto disable-replay-opts instead of exiting with error

### LINKED ISSUE #1403: Disable Replay Opts by default if statehist plugin enabled
Currently, if you have the state history plugin enabled on the node, it requires the additional parameter of --disable-replay-opts to always be passed as an argument.  If you don't pass the argument, the node just errors that it should be disabled.  Can it be just added into the logic that if the statehist plugin is enabled that the "replay opts" are disabled automatically, which would help standardize not having to have the special param for statehist nodes (or forgetting to include it and having to relaunch)

heifner:
- Modified `state_history_plugin` to automatically set `disable-replay-opts=true` instead of indicating an error and exiting.
- Modified the requirement for  `disable-replay-opts=true` only when `chain-state-history=true` is enabled as it is not needed if `state_history_plugin` is running with only `trace-history=true` enabled.
- Modified state history serialization to not require `chainbase` when serializing transaction trace data. This verifies that `disable-replay-opts=true` is not needed if only logging transaction trace data.
- Allow `disable-replay-opts` to be set in `config.ini` instead of only on the command line.

Resolves #1403


plugins/chain_plugin/chain_plugin.cpp
```diff
@@ -403,8 +405,6 @@ void chain_plugin::set_program_options(options_description& cli, options_descrip
           "extract build environment information as JSON, write into specified file, and exit")
          ("force-all-checks", bpo::bool_switch()->default_value(false),
           "do not skip any validation checks while replaying blocks (useful for replaying blocks from untrusted source)")
-         ("disable-replay-opts", bpo::bool_switch()->default_value(false),
```
greg7mdp, line 406: Will removing the command line option cause an error if it is still provided? Probably better than leaving an option doing nothing though.

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -314,12 +314,15 @@ void state_history_plugin::set_program_options(options_description& cli, options
...
 void state_history_plugin_impl::plugin_initialize(const variables_map& options) {
    try {
-      EOS_ASSERT(options.at("disable-replay-opts").as<bool>(), plugin_exception,
-                 "state_history_plugin requires --disable-replay-opts");
-
       chain_plug = app().find_plugin<chain_plugin>();
       EOS_ASSERT(chain_plug, chain::missing_chain_plugin_exception, "");
       auto& chain = chain_plug->chain();
+
+      if (!options.at("disable-replay-opts").as<bool>() && options.at("chain-state-history").as<bool>()) {
```
greg7mdp, line 321: I would have thought it should be: `if (options.at("disable-replay-opts")...` without the `!`.

libraries/state_history/include/eosio/state_history/serialization.hpp
```diff
@@ -697,9 +757,9 @@ datastream<ST>& operator<<(
 template <typename ST>
 datastream<ST>&
 operator<<(datastream<ST>&                                                                         ds,
```
greg7mdp, line 759: indentation of `ds`. Either align, or preferably remove all that white space?

libraries/state_history/include/eosio/state_history/serialization.hpp
```diff
@@ -624,7 +684,7 @@ datastream<ST>& operator<<(datastream<ST>& ds, const history_context_wrapper<boo
 template <typename ST>
 datastream<ST>& operator<<(
     datastream<ST>&                                                                                             ds,
```
greg7mdp, line 686: indentation

libraries/state_history/include/eosio/state_history/serialization.hpp
```diff
@@ -35,11 +49,27 @@ struct history_context_wrapper {
 };
 
 template <typename P, typename T>
-history_context_wrapper<std::decay_t<P>, std::decay_t<T>> make_history_context_wrapper(const chainbase::database& db,
-                                                                                       const P& context, const T& obj) {
+struct history_context_wrapper_stateless {
```
greg7mdp, line 52: I think I would have made `history_context_wrapper_stateless` a base of `history_context_wrapper`. 
the advantage is that the functions accepting a `history_context_wrapper_stateless&` would also take a `history_context_wrapper&`, although I'm not sure if it would have helped a lot.

greg7mdp (approved):

plugins/chain_plugin/chain_plugin.cpp
heifner: Everything in the `cfg` options is allowed on the command line or in the `config.ini`. Everything in the `cli` is only allowed on the command line.

plugins/state_history_plugin/state_history_plugin.cpp
heifner: No need to set it if it is already set.

plugins/chain_plugin/chain_plugin.cpp
greg7mdp: Oh, thanks!

plugins/state_history_plugin/state_history_plugin.cpp
greg7mdp: I thought this was an option passed to ship and I didn't realize it would already be applied to `chain`. Thanks.

linh2931 (commented):
Serialization has undergone significant changes. Do we have sufficient existing tests to cover the changes?

greg7mdp (approved):

linh2931 (approved):

PR #1449, opened 2023-07-25
[3.2] Modify trace_api_plugin to report serialization errors to user

heifner:
The exception handler for the `trace_api_plugin` data handler now rethrows any exceptions so they can be reported to the API caller of `/v1/trace_api/get_block` and `/v1/trace_api/get_transaction_trace`.

`trace_api_plugin` will now report
```
./cleos --verbose --print-response get block_trace 20
RESPONSE:
---------------------
{
  "code": 500,
  "message": "Internal Service Error",
  "error": {
    "code": 2,
    "name": "timeout_exception",
    "what": "deadline 2023-07-25T20:19:49.442 exceeded by 24us ",
    "details": [{
        "message": "deadline 2023-07-25T20:19:49.442 exceeded by 24us ",
        "file": "trace_api_plugin.cpp",
        "line_number": 288,
        "method": "operator()"
      }
    ]
  }
}
---------------------
error 2023-07-25T20:19:49.445 cleos     main.cpp:4498                 operator()           ] Failed with error: 2 timeout_exception: deadline 2023-07-25T20:19:49.442 exceeded by 24us
deadline 2023-07-25T20:19:49.442 exceeded by 24us
    {}
    cleos  trace_api_plugin.cpp:288 operator()
```
Instead of:
```
./cleos --verbose --print-response get block_trace 20
RESPONSE:
---------------------
{
  "code": 404,
  "message": "Trace API: block trace missing",
  "error": {
    "code": 0,
    "name": "",
    "what": "",
    "details": []
  }
}
---------------------
Error 3200005: http request fail
Error Details:
Error code 404
: {"code":404,"message":"Trace API: block trace missing","error":{"code":0,"name":"","what":"","details":[]}}
```

Resolves #1433

PR #1469, opened 2023-07-31
[4.0 -> main] Modify trace_api_plugin to report serialization errors to user

### LINKED ISSUE #1433: trace_api_plugin reports block trace missing when http-max-response-time-ms limit is reached
`{"code":404,"message":"Trace API: block trace missing","error":{"code":0,"name":"","what":"","details":[]}}`

https://github.com/AntelopeIO/leap/blob/v4.0.4/plugins/trace_api_plugin/abi_data_handler.cpp#L40 Calls `except_handler` when `yield()` throw exception and returns `{}`. The `except_handler` is setup to https://github.com/AntelopeIO/leap/blob/v4.0.4/plugins/trace_api_plugin/trace_api_plugin.cpp#L204 only log the error. This means a small `http-max-response-time-ms` causes `trace_api_plugin` to indicate it could not find the block rather than a deadline exception was hit.

heifner:
Merges `release/4.0` into `main` including #1449 & #1468.

The 3.1 & 3.2 version of this PR would send the deadline exception to the user so that they know they hit a deadline. In `main` there is no longer a deadline enforced for `trace_api_plugin` serialization since it is all processed off the main thread. We still report any exceptions to the user if they would happen to hit some kind of serialization error. Also `trace_plugin_test.py` has been updated to expect success even when configured with zero time for `http-max-response-time-ms` and `abi-serializer-max-time-ms`.

Resolves #1433


tests/trace_plugin_test.py
```diff
@@ -95,6 +96,20 @@ def test_TraceApi(self) :
         global testSuccessful
         testSuccessful = True
 
+        # relaunch with no time allocated for http response & abi-serializer. Will fail because get_info fails.
+        node.kill(signal.SIGTERM)
+        Utils.Print("Ignore expected: ERROR: Node relaunch Failed")
+        isRelaunchSuccess = node.relaunch(timeout=10, addSwapFlags={"--http-max-response-time-ms": "0", "--abi-serializer-max-time-ms": "0"})
```
greg7mdp, line 102: Will the relaunch fail as the message indicates?
Also wondering if it is useful to assign to `isRelaunchSuccess`?

greg7mdp (approved):

tests/trace_plugin_test.py
heifner: `get_info` can't fail in `main`, so it actually does work. I'll update the test.

tests/plugin_http_api_test.py
```diff
@@ -1406,6 +1406,22 @@ def test_TraceApi(self) :
...
+        command = "get_transaction_trace"
+        ret_json = self.nodeos.processUrllibRequest(resource, command)
+        self.assertEqual(ret_json["code"], 400)
+        # get_transaction_trace with empty content parameter
+        ret_json = self.nodeos.processUrllibRequest(resource, command, self.empty_content_dict)
+        self.assertEqual(ret_json["code"], 400)
+        # get_transaction_trace with invalid parameter
+        ret_json = self.nodeos.processUrllibRequest(resource, command, self.http_post_invalid_param)
+        self.assertEqual(ret_json["code"], 400)
+        # get_transaction_trace with valid parameter
```
greg7mdp, line 1419:
```suggestion
        # get_transaction_trace with syntactically correct  parameter, but random id, so should return 404 (not found)
```

dimas1185 (approved):

PR #1471, opened 2023-08-01
Fix `-Wsign-compare` warnings

greg7mdp:


libraries/chain/CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ if("eos-vm-oc" IN_LIST EOSIO_WASM_RUNTIMES)
 
    set_source_files_properties(webassembly/runtimes/eos-vm-oc/LLVMJIT.cpp PROPERTIES COMPILE_FLAGS "--std=gnu++17")
    set_source_files_properties(webassembly/runtimes/eos-vm-oc/LLVMEmitIR.cpp PROPERTIES COMPILE_FLAGS "--std=gnu++17")
+   set_source_files_properties(webassembly/runtimes/eos-vm-oc/switch_stack_linux.s PROPERTIES COMPILE_FLAGS "-Wno-unused-command-line-argument")
```
spoonincode, line 39: I think this is a fine fix.. but I wonder why cmake is operating this way :thinking:  it clearly knows that `.s` are assembly files -- not c/c++ files -- so why is it tacking on all the c/c++ options like `-O3 -DNDEBUG` etc

plugins/state_history_plugin/tests/session_test.cpp
```diff
@@ -530,7 +530,7 @@ BOOST_FIXTURE_TEST_CASE(test_session_fork, state_history_test_fixture) {
       std::vector<eosio::state_history::block_position> have_positions;
       eosio::state_history::state_result result;
       // we should get 4 consecutive block result
-      for (int i = 0; i < 4; ++i) {
+      for (uint32_t i = 0; i < 4; ++i) {
```
heifner, line 533: `size_t` is likely better here.

plugins/state_history_plugin/tests/session_test.cpp
```diff
@@ -580,7 +580,7 @@ BOOST_FIXTURE_TEST_CASE(test_session_fork, state_history_test_fixture) {
 
       eosio::state_history::state_result fork_result;
       // we should now get data for fork 3,4
-      for (int i = 2; i < 4; ++i) {
+      for (uint32_t i = 2; i < 4; ++i) {
```
heifner, line 583: `size_t` is likely better here.

heifner (approved):

libraries/chain/CMakeLists.txt
greg7mdp: beats me!

plugins/state_history_plugin/tests/session_test.cpp
greg7mdp: Hum, I'm not sure. The warning was when comparing with `r.this_block->block_num` which is a `uint32_t`.
If it was a comparison with `<container>.size()` then I would have used `size_t`.

plugins/state_history_plugin/tests/session_test.cpp
greg7mdp: same as above.

plugins/state_history_plugin/tests/session_test.cpp
heifner: Ah, I thought it was likely complaining about the `written_data[i]`. `uint32_t` then.

linh2931 (approved):
Thanks for those large amount of edits.

PR #1476, opened 2023-08-02
Tests: Add back old commented out unittests

### LINKED ISSUE #1111: Investigate commented out unittests
There are tests in `api_tests.cpp`, `producer_schedule_tests.cpp`, `resource_limits_tests.cpp`, `wasm_tests.cpp` that are `#if 0` out.
These should either be enabled/fixed or removed.

heifner:
Re-enable unittests which have been commented out since pre-launch of EOS.

Resolves #1111

### DIFF
--- a/unittests/producer_schedule_tests.cpp
+++ b/unittests/producer_schedule_tests.cpp
@@ -22,7 +22,7 @@ account_name get_expected_producer(const vector<producer_authority>& schedule, b
 // Check if two schedule is equal
 bool is_schedule_equal(const vector<producer_authority>& first, const vector<producer_authority>& second) {
    bool is_equal = first.size() == second.size();
-   for (uint32_t i = 0; i < first.size(); i++) {
+   for (uint32_t i = 0; i < first.size() && is_equal; i++) {
       is_equal = is_equal && first.at(i) == second.at(i);
    }
    return is_equal;
--- a/unittests/producer_schedule_tests.cpp
+++ b/unittests/producer_schedule_tests.cpp
@@ -11,187 +11,122 @@ using mvo = fc::mutable_variant_object;
 
 BOOST_AUTO_TEST_SUITE(producer_schedule_tests)
 
-   // Calculate expected producer given the schedule and slot number
-   account_name get_expected_producer(const vector<producer_authority>& schedule, const uint64_t slot) {
-      const auto& index = (slot % (schedule.size() * config::producer_repetitions)) / config::producer_repetitions;
-      return schedule.at(index).producer_name;
-   };
+namespace {
 
-   // Check if two schedule is equal
-   bool is_schedule_equal(const vector<producer_authority>& first, const vector<producer_authority>& second) {
-      bool is_equal = first.size() == second.size();
-      for (uint32_t i = 0; i < first.size(); i++) {
-         is_equal = is_equal && first.at(i) == second.at(i);
-      }
-      return is_equal;
-   };
+// Calculate expected producer given the schedule and slot number
+account_name get_expected_producer(const vector<producer_authority>& schedule, block_timestamp_type t) {
+   const auto& index = (t.slot % (schedule.size() * config::producer_repetitions)) / config::producer_repetitions;
+   return schedule.at(index).producer_name;
+};
 
-   // Calculate the block num of the next round first block
-   // The new producer schedule will become effective when it's in the block of the next round first block
-   // However, it won't be applied until the effective block num is deemed irreversible
-   uint64_t calc_block_num_of_next_round_first_block(const controller& control){
-      auto res = control.head_block_num() + 1;
-      const auto blocks_per_round = control.head_block_state()->active_schedule.producers.size() * config::producer_repetitions;
-      while((res % blocks_per_round) != 0) {
-         res++;
-      }
-      return res;
-   };
-#if 0
-   BOOST_FIXTURE_TEST_CASE( verify_producer_schedule, validating_tester ) try {
+// Check if two schedule is equal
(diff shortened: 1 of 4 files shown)


unittests/producer_schedule_tests.cpp
```diff
@@ -11,187 +11,122 @@ using mvo = fc::mutable_variant_object;
...
+namespace {
+
+// Calculate expected producer given the schedule and slot number
+account_name get_expected_producer(const vector<producer_authority>& schedule, block_timestamp_type t) {
+   const auto& index = (t.slot % (schedule.size() * config::producer_repetitions)) / config::producer_repetitions;
+   return schedule.at(index).producer_name;
+};
+
+// Check if two schedule is equal
+bool is_schedule_equal(const vector<producer_authority>& first, const vector<producer_authority>& second) {
```
greg7mdp, line 23: How is this implementation different from `return first == second;`?

unittests/producer_schedule_tests.cpp
```diff
@@ -11,187 +11,122 @@ using mvo = fc::mutable_variant_object;
...
-      bool is_equal = first.size() == second.size();
-      for (uint32_t i = 0; i < first.size(); i++) {
-         is_equal = is_equal && first.at(i) == second.at(i);
+         // Check if the producer is the same as what we expect
+         const auto block_time = control->head_block_time();
+         const auto& expected_producer = get_expected_producer(current_schedule, block_time);
+         BOOST_TEST(control->head_block_producer() == expected_producer);
+
+         if (scheduled_changed_to_new)
+            break;
```
greg7mdp, line 54: Shouldn't it be like this:

```suggestion
          if (scheduled_changed_to_new) {
            const auto block_time = control->head_block_time();
            const auto& expected_producer = get_expected_producer(new_prod_schd, block_time);
            BOOST_TEST(control->head_block_producer() == expected_producer);
            break;
         }
```

unittests/resource_limits_test.cpp
```diff
@@ -136,33 +136,37 @@ BOOST_AUTO_TEST_SUITE(resource_limits_test)
...
    BOOST_FIXTURE_TEST_CASE(weighted_capacity_cpu, resource_limits_fixture) try {
-      const vector<int64_t> weights = { 234, 511, 672, 800, 1213 };
-      const int64_t total = std::accumulate(std::begin(weights), std::end(weights), 0LL);
+      const vector<int64_t> weights = { 1, 3, 9, 234, 511, 672, 800, 1213, 4242, 77777, 888888, 999999 };
+      const uint128_t total = std::accumulate(std::begin(weights), std::end(weights), 0LL);
       vector<int64_t> expected_limits;
-      std::transform(std::begin(weights), std::end(weights), std::back_inserter(expected_limits), [total](const auto& v){ return v * config::default_max_block_cpu_usage / total; });
+      std::transform(std::begin(weights), std::end(weights), std::back_inserter(expected_limits), [total](const auto& v){
+         uint128_t value = v;
+         uint128_t account_cpu_usage_average_window = config::account_cpu_usage_average_window_ms / config::block_interval_ms;
```
greg7mdp, line 147: Just a remark, `account_cpu_usage_average_window_ms` is a very un-intuitive name for what it really is (`milliseconds_per_day`)

Also, shouldn't  `account_cpu_usage_average_window` be better named `num_slots_per_day`?

unittests/resource_limits_test.cpp
```diff
@@ -136,33 +136,37 @@ BOOST_AUTO_TEST_SUITE(resource_limits_test)
...
+         return (account_cpu_usage_average_window * config::default_max_block_cpu_usage * value) / total;
+      });
 
       for (int64_t idx = 0; idx < weights.size(); idx++) {
```
greg7mdp, line 151: int64_t -> size_t

unittests/resource_limits_test.cpp
```diff
@@ -136,33 +136,37 @@ BOOST_AUTO_TEST_SUITE(resource_limits_test)
...
          const account_name account(idx + 100);
-         initialize_account(account);
-         set_account_limits(account, -1, -1, weights.at(idx));
+         initialize_account(account, false);
+         set_account_limits(account, -1, -1, weights.at(idx), false);
```
greg7mdp, line 154:
```suggestion
         set_account_limits(account, -1, -1, weights[idx], false);
```

unittests/resource_limits_test.cpp
heifner: `account_cpu_usage_average_window_ms` doesn't have to be milliseconds per day.
Renamed `account_cpu_usage_average_window`  to `account_cpu_usage_window `

unittests/producer_schedule_tests.cpp
heifner: lol, it is not different. I'll remove it.

unittests/producer_schedule_tests.cpp
heifner: It is good to test that the `expected_producer` is correct even before the schedule change.

greg7mdp: should you call: get_expected_producer(**new_prod_schd**?

heifner: `new_prod_schd` is equal to `current_schedule` when the schedule changes, which is why the test was working already.

greg7mdp: Hum, I guess. The test is a bit strange because the real test is that we actually break out of the loop before the loop condition `i < check_duration` is hit. And doing `BOOST_TEST(control->head_block_producer() == expected_producer);` before `scheduled_changed_to_new` is true does not really check much.
But still a valid test! Approving the PR.

greg7mdp (approved):

PR #1478, opened 2023-08-03
return `string` from test contract's `_abi()` call instead of `vector` to avoid non-null terminated `char*` access

spoonincode:
When the unittests get the ABI for a test contract,
https://github.com/AntelopeIO/leap/blob/339d98eed107b9fd94736988996082c7002fa52a/unittests/test_contracts.hpp.in#L13-L17
`fc::read_file_contents` will read the file's contents in to `s`. There is no null terminator at the end of the file since it's just a JSON text file, so `s` contains no null terminator in its set data. C++11 of course does define that `s.data()` is null terminated, and `*(s.begin() + s.size()) == '\0'`, but when `std::vector<char>(s.begin(), s.end())` is called, a vector is constructed with `s`'s contents without that implicit null terminator; just the set data of `s` which does not have the null terminator.

Later on, this function may be called like
https://github.com/AntelopeIO/leap/blob/339d98eed107b9fd94736988996082c7002fa52a/tests/chain_plugin_tests.cpp#L46
This is thus passing a non-null terminated `char*` from the `vector` to,
https://github.com/AntelopeIO/leap/blob/339d98eed107b9fd94736988996082c7002fa52a/libraries/testing/tester.cpp#L969-L970
where the call to `fc::json::from_string(abi_json)` constructs a std::string from a non-null terminated `char*`

We might be able to spot fix this creating the `vector` by `std::vector<char>(s.begin(), s.end()+1)`, or by `s.push_back('\0')` before constructing the `vector`. But a better improvement imo, even though it touches a lot of files in clean up, is to just plumb through a `string` everywhere. That type better represents a JSON ABI string.

### DIFF
--- a/libraries/testing/contracts.cpp.in
+++ b/libraries/testing/contracts.cpp.in
@@ -14,8 +14,8 @@ namespace eosio::testing {                                                      
    std::vector<std::uint8_t> contracts:: CN ## _wasm() {                                                                                                                      \
       return std::vector<std::uint8_t>(geosio_testing_contract_ ## CN ## _wasm_data, geosio_testing_contract_ ## CN ## _wasm_data + geosio_testing_contract_ ## CN ## _wasm_size); \
    }                                                                                                                                                                          \
-   std::vector<char> contracts:: CN ## _abi() {                                                                                                                               \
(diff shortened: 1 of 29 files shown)


heifner (approved):
Nice.

linh2931 (approved):

PR #1481, opened 2023-08-07
Smarter block deadlines for speculative blocks

### LINKED ISSUE #1432: Speculative block repeatedly restarted when near block boundary
When speculatively executing trxs if a block is exhausted then it is now immediately restarted via: https://github.com/AntelopeIO/leap/pull/1319. However, when starting a speculative block we set a block deadline according to pending block time minus producer time offset. This can create very small block times. A trx then can exceed that block time but not the configured trx max time resulting in a burst of abort_block/start_block calls.

Example scenario:
- A speculative block is created with a time limit of pending block time minus producer offset. This can be rather small depending on when a block was received and how much time was spent processing the block
- trx exceeds block time and is rescheduled for the next block
- abort block
- new block created with the same time limit as the one just aborted
- trx exceeds block time and is rescheduled for the next block
- abort block
- repeat until block interval is exceeded or an incoming block is applied

When not producing a block, the time limit of a block should be set to `fc::time_point::maximum()` since we want to allow as many speculative trxs as possible until another block is received. If the consensus max cpu or max net is exceeded then the block will be restarted as the trx/block will be considered exhausted.

heifner:
Change: Start speculative blocks every block interval but not more often. Fixes a corner case of rapid start-block/abort block burst for speculative blocks. Also create speculative blocks even when state is stale (over 5 seconds old).

Before this PR, block deadline was set for speculative blocks the same as production blocks for block production time. Because of when blocks are processed this could result in small block speculative windows. When a block came in close to deadline it would cause us to repeatedly start block with a very small deadline, abort block because of deadline, start another block and immediately abort again until the block time window was passed.

In addition, speculative blocks were not started on "stale" state (head block older than 5 seconds). Note this would be the case if a block producer was down and missed a production round (6 seconds). 

The conditions where the stale condition would most likely be triggered (block producer being down or node disconnected from network for some time) are situations in which it most likely does undesirable things for both read-only transactions and non-read-only transactions. It unnecessarily delays a read-only transaction (and dry-run transaction) that may not care that they are potentially getting back stale data. And if they did care, they could protect themselves by setting the expiration appropriately.

It also causes similar problems with write transactions. If the stale condition is due to a block producer being down rather than the node being disconnected from the network, with this PR changes, a successfully executed write transaction can propagate through the network and reach a live block producer that queues up the transaction for when it is its turn to produce a block. Before this PR, the write transaction is simply queued up to be speculatively executed once a new block arrives. If the expiration time of the transaction was set low, there is a good chance the transaction just fails due to expiration by that point.

But it isn't all beneficial. There is a downside with this PR to write transactions, if the stale condition is due to a temporary network disruption (say for 10 seconds). In that case, queuing the write transaction for when the network is restored makes it more likely for that transaction to get into the block (assuming using the normal `send_transaction` endpoint without retries). But with this PR, the network disruption leads to the write transaction being silently dropped despite a successful trace being returned to the client. This is because the transaction is speculatively executed and attempts to send out the transaction which would be dropped if there are no currently available connected peers.

Includes refactor of `producer_watermarks`, `calculate_producer_wake_up_time`, `calculate_next_block_slot` into `block_timing_util`. No change to functionality. Created new tests around the moved functions.

Resolves #1432

### DIFF
--- a/plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
+++ b/plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
@@ -44,7 +44,7 @@ namespace block_timing_util {
    // received it and start producing on schedule. To mitigate the problem, we leave no time gap in block producing. For
    // example, given block_interval=500 ms and cpu effort=400 ms, assuming the our round start at time point 0; in the
    // past, the block start time points would be at time point -500, 0, 500, 1000, 1500, 2000 ....  With this new
-   // approach, the block time points would become -500, -100, 300, 700, 1200 ...
+   // approach, the block time points would become -500, -100, 300, 700, 1100 ...
    inline fc::time_point production_round_block_start_time(uint32_t cpu_effort_us, chain::block_timestamp_type block_time) {
       uint32_t block_slot = block_time.slot;
       uint32_t production_round_start_block_slot =
--- a/plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
+++ b/plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
@@ -1,12 +1,39 @@
 #pragma once
 #include <eosio/chain/block_timestamp.hpp>
 #include <eosio/chain/config.hpp>
+#include <eosio/chain/producer_schedule.hpp>
 
 namespace eosio {
 
 enum class pending_block_mode { producing, speculating };
 
 namespace block_timing_util {
+
+   // Store watermarks
+   class producer_watermarks {
+   public:
+      void consider_new_watermark(chain::account_name producer, uint32_t block_num, chain::block_timestamp_type timestamp) {
+         auto itr = _producer_watermarks.find(producer);
+         if (itr != _producer_watermarks.end()) {
+            itr->second.first  = std::max(itr->second.first, block_num);
+            itr->second.second = std::max(itr->second.second, timestamp);
+         } else {
+            _producer_watermarks.emplace(producer, std::make_pair(block_num, timestamp));
+         }
+      }
+
+      using producer_watermark = std::pair<uint32_t, chain::block_timestamp_type>;
+      std::optional<producer_watermark> get_watermark(chain::account_name producer) const {
+         auto itr = _producer_watermarks.find(producer);
+
+         if (itr == _producer_watermarks.end())
+            return {};
+
+         return itr->second;
+      }
+   private:
+      std::map<chain::account_name, producer_watermark> _producer_watermarks;
+   };
 
    // Calculate when a producer can start producing a given block represented by its block_time
    //
@@ -25,22 +52,95 @@ namespace block_timing_util {
              fc::microseconds(cpu_effort_us * production_round_index);
    }
 
-   inline fc::time_point calculate_block_deadline(uint32_t cpu_effort_us, pending_block_mode mode, chain::block_timestamp_type block_time) {
-      const auto hard_deadline =
-                  block_time.to_time_point() - fc::microseconds(chain::config::block_interval_us - cpu_effort_us);
-      if (mode == pending_block_mode::producing) {
-         auto estimated_deadline = production_round_block_start_time(cpu_effort_us, block_time) + fc::microseconds(cpu_effort_us);
-         auto now                = fc::time_point::now();
-         if (estimated_deadline > now) {
-            return estimated_deadline;
-         } else {
-            // This could only happen when the producer stop producing and then comes back alive in the middle of its own
-            // production round. In this case, we just use the hard deadline.
-            return std::min(hard_deadline, now + fc::microseconds(cpu_effort_us));
-         }
+   inline fc::time_point calculate_producing_block_deadline(uint32_t cpu_effort_us, chain::block_timestamp_type block_time) {
+      auto estimated_deadline = production_round_block_start_time(cpu_effort_us, block_time) + fc::microseconds(cpu_effort_us);
+      auto now                = fc::time_point::now();
+      if (estimated_deadline > now) {
+         return estimated_deadline;
       } else {
-         return hard_deadline;
+         // This could only happen when the producer stop producing and then comes back alive in the middle of its own
+         // production round. In this case, we just use the hard deadline.
+         const auto hard_deadline = block_time.to_time_point() - fc::microseconds(chain::config::block_interval_us - cpu_effort_us);
+         return std::min(hard_deadline, now + fc::microseconds(cpu_effort_us));
       }
    }
-};
+
+   inline uint32_t calculate_next_block_slot(const chain::account_name& producer_name, uint32_t current_block_slot, uint32_t block_num,
+                                             const std::vector<chain::producer_authority>& active_schedule, const producer_watermarks& prod_watermarks) {
+      // determine if this producer is in the active schedule and if so, where
+      auto itr =
+         std::find_if(active_schedule.begin(), active_schedule.end(), [&](const auto& asp) { return asp.producer_name == producer_name; });
+      if (itr == active_schedule.end()) {
+         // this producer is not in the active producer set
+         return UINT32_MAX;
+      }
+
+      size_t   producer_index = itr - active_schedule.begin();
+      uint32_t minimum_offset = 1; // must at least be the "next" block
+
+      // account for a watermark in the future which is disqualifying this producer for now
+      // this is conservative assuming no blocks are dropped.  If blocks are dropped the watermark will
+      // disqualify this producer for longer but it is assumed they will wake up, determine that they
+      // are disqualified for longer due to skipped blocks and re-calculate their next block with better
+      // information then
+      auto current_watermark = prod_watermarks.get_watermark(producer_name);
+      if (current_watermark) {
+         const auto watermark = *current_watermark;
+         if (watermark.first > block_num) {
+            // if I have a watermark block number then I need to wait until after that watermark
+            minimum_offset = watermark.first - block_num + 1;
+         }
+         if (watermark.second.slot > current_block_slot) {
+            // if I have a watermark block timestamp then I need to wait until after that watermark timestamp
+            minimum_offset = std::max(minimum_offset, watermark.second.slot - current_block_slot + 1);
+         }
+      }
+
+      // this producers next opportunity to produce is the next time its slot arrives after or at the calculated minimum
+      uint32_t minimum_slot = current_block_slot + minimum_offset;
+      size_t   minimum_slot_producer_index =
+         (minimum_slot % (active_schedule.size() * chain::config::producer_repetitions)) / chain::config::producer_repetitions;
+      if (producer_index == minimum_slot_producer_index) {
+         // this is the producer for the minimum slot, go with that
+         return minimum_slot;
+      } else {
+         // calculate how many rounds are between the minimum producer and the producer in question
+         size_t producer_distance = producer_index - minimum_slot_producer_index;
+         // check for unsigned underflow
+         if (producer_distance > producer_index) {
+            producer_distance += active_schedule.size();
+         }
+
+         // align the minimum slot to the first of its set of reps
+         uint32_t first_minimum_producer_slot = minimum_slot - (minimum_slot % chain::config::producer_repetitions);
+
+         // offset the aligned minimum to the *earliest* next set of slots for this producer
+         uint32_t next_block_slot = first_minimum_producer_slot + (producer_distance * chain::config::producer_repetitions);
+         return next_block_slot;
+      }
+   }
+
+   // Return the *next* block start time according to its block time slot.
+   // Returns empty optional if no producers are in the active_schedule.
+   // block_num is only used for watermark minimum offset.
+   inline std::optional<fc::time_point> calculate_producer_wake_up_time(uint32_t cpu_effort_us, uint32_t block_num,
+                                                                        const chain::block_timestamp_type& ref_block_time,
+                                                                        const std::set<chain::account_name>& producers,
+                                                                        const std::vector<chain::producer_authority>& active_schedule,
+                                                                        const producer_watermarks& prod_watermarks) {
+      auto ref_block_slot = ref_block_time.slot;
+      // if we have any producers then we should at least set a timer for our next available slot
+      uint32_t wake_up_slot = UINT32_MAX;
+      for (const auto& p : producers) {
+         auto next_producer_block_slot = calculate_next_block_slot(p, ref_block_slot, block_num, active_schedule, prod_watermarks);
+         wake_up_slot                  = std::min(next_producer_block_slot, wake_up_slot);
+      }
+      if (wake_up_slot == UINT32_MAX) {
+         return {};
+      }
+
+      return production_round_block_start_time(cpu_effort_us, chain::block_timestamp_type(wake_up_slot));
+   }
+
+} // namespace block_timing_util
 } // namespace eosio
--- a/plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
+++ b/plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
@@ -10,6 +10,8 @@ enum class pending_block_mode { producing, speculating };
 namespace block_timing_util {
 
    // Store watermarks
+   // Watermarks are recorded times that the specified producer has produced.
+   // Used by calculate_producer_wake_up_time to skip over already produced blocks avoiding duplicate production.
    class producer_watermarks {
    public:
       void consider_new_watermark(chain::account_name producer, uint32_t block_num, chain::block_timestamp_type timestamp) {
--- a/plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
+++ b/plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
@@ -65,58 +65,51 @@ namespace block_timing_util {
       }
    }
 
-   inline uint32_t calculate_next_block_slot(const chain::account_name& producer_name, uint32_t current_block_slot, uint32_t block_num,
-                                             const std::vector<chain::producer_authority>& active_schedule, const producer_watermarks& prod_watermarks) {
-      // determine if this producer is in the active schedule and if so, where
-      auto itr =
-         std::find_if(active_schedule.begin(), active_schedule.end(), [&](const auto& asp) { return asp.producer_name == producer_name; });
-      if (itr == active_schedule.end()) {
-         // this producer is not in the active producer set
-         return UINT32_MAX;
-      }
+   namespace detail {
+      inline uint32_t calculate_next_block_slot(const chain::account_name& producer_name, uint32_t current_block_slot, uint32_t block_num,
+                                                size_t producer_index, size_t active_schedule_size, const producer_watermarks& prod_watermarks) {
+         uint32_t minimum_offset = 1; // must at least be the "next" block
 
-      size_t   producer_index = itr - active_schedule.begin();
-      uint32_t minimum_offset = 1; // must at least be the "next" block
-
-      // account for a watermark in the future which is disqualifying this producer for now
-      // this is conservative assuming no blocks are dropped.  If blocks are dropped the watermark will
-      // disqualify this producer for longer but it is assumed they will wake up, determine that they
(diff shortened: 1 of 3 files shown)


plugins/producer_plugin/test/test_block_timing_util.cpp
```diff
@@ -34,73 +35,214 @@ BOOST_AUTO_TEST_CASE(test_calculate_block_deadline) {
...
+   using namespace eosio;
+   using namespace eosio::chain;
+   using namespace eosio::chain::literals;
+   using namespace eosio::block_timing_util;
+
+   producer_watermarks empty_watermarks;
+   // use full cpu effort for most of these tests since calculate_producing_block_deadline is tested above
+   constexpr uint32_t full_cpu_effort = eosio::chain::config::block_interval_us;
+
+   // no producers
```
linh2931, line 106: You can add a pair of `{` and `}` around this first test to be consistent with other tests.

plugins/producer_plugin/test/test_block_timing_util.cpp
```diff
@@ -34,73 +35,214 @@ BOOST_AUTO_TEST_CASE(test_calculate_block_deadline) {
...
+   BOOST_CHECK_EQUAL(calculate_producer_wake_up_time(full_cpu_effort, 2, chain::block_timestamp_type{}, {}, {}, empty_watermarks), std::optional<fc::time_point>{});
+   { // producers not in active_schedule
+      std::set<chain::account_name>          producers{"p1"_n, "p2"_n};
+      std::vector<chain::producer_authority> active_schedule{{"active1"_n}, {"active2"_n}};
+      BOOST_CHECK_EQUAL(calculate_producer_wake_up_time(full_cpu_effort, 2, chain::block_timestamp_type{}, producers, active_schedule, empty_watermarks), std::optional<fc::time_point>{});
+   }
+   { // Only producer in active_schedule
```
linh2931, line 113: Change to `Only one producer in active_schedule`

plugins/producer_plugin/test/test_block_timing_util.cpp
```diff
@@ -34,73 +35,214 @@ BOOST_AUTO_TEST_CASE(test_calculate_block_deadline) {
...
+      std::set<chain::account_name>          producers{"p1"_n, "p2"_n, "p3"_n};
+      std::vector<chain::producer_authority> active_schedule{{"p1"_n}, {"p2"_n}};
+      const uint32_t prod_round_1st_block_slot = 100 * active_schedule.size() * eosio::chain::config::producer_repetitions - 1;
+      for (uint32_t i = 0; i < static_cast<uint32_t>(config::producer_repetitions * active_schedule.size() * 3); ++i) { // 3 rounds to test boundaries
+         block_timestamp_type block_timestamp(prod_round_1st_block_slot + i);
+         auto block_time = block_timestamp.to_time_point();
+         BOOST_CHECK_EQUAL(calculate_producer_wake_up_time(full_cpu_effort, 2, block_timestamp, producers, active_schedule, empty_watermarks), block_time);
+      }
+   }
+   { // Only producers in active_schedule 21
```
linh2931, line 133: Can you add more comments to the rest of tests on what is intended to test?

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -1,13 +1,40 @@
...
 #include <eosio/chain/config.hpp>
+#include <eosio/chain/producer_schedule.hpp>
 
 namespace eosio {
 
 enum class pending_block_mode { producing, speculating };
 
 namespace block_timing_util {
 
+   // Store watermarks
```
linh2931, line 12: Can you add some comments on the purpose and use of watermarks?

greg7mdp:
> Speculative blocks on non-producer nodes will not restart unless a new block is received or it is exhausted. Speculative blocks on producing nodes will not restart until a production window is reached or it is exhausted.

What is the rationale for these two changes?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1966,19 +1884,31 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
...
+
+      _pending_block_deadline = block_timing_util::calculate_producing_block_deadline(_cpu_effort_us, block_time);
+   } else if (!_producers.empty()) {
+      // cpu effort percent doesn't matter for the first block of the round, use max (block_interval_us) for cpu effort
+      auto wake_time = block_timing_util::calculate_producer_wake_up_time(config::block_interval_us, chain.head_block_num(), chain.head_block_time(),
+                                                                          _producers, chain.head_block_state()->active_schedule.producers,
+                                                                          _producer_watermarks);
+      _pending_block_deadline = wake_time ? *wake_time : fc::time_point::maximum();
+   } else {
+      _pending_block_deadline = fc::time_point::maximum();
```
greg7mdp, line 1907: I wonder if it should be:

` _pending_block_deadline = chain.head_block_time() + fc::seconds(5);`

To be consistent with line 1883:

```
   if (in_speculating_mode()) {
      auto head_block_age = now - chain.head_block_time();
      if (head_block_age > fc::seconds(5))
         return start_block_result::waiting_for_block;
   }
```

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -25,22 +52,95 @@ namespace block_timing_util {
...
+         return estimated_deadline;
+      } else {
+         // This could only happen when the producer stop producing and then comes back alive in the middle of its own
+         // production round. In this case, we just use the hard deadline.
+         const auto hard_deadline = block_time.to_time_point() - fc::microseconds(chain::config::block_interval_us - cpu_effort_us);
+         return std::min(hard_deadline, now + fc::microseconds(cpu_effort_us));
+      }
+   }
+
+   inline uint32_t calculate_next_block_slot(const chain::account_name& producer_name, uint32_t current_block_slot, uint32_t block_num,
```
greg7mdp, line 68: I see why you made the return value `uint32_t`, but I think it would be cleaner if it returned a std::optional

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -25,22 +52,95 @@ namespace block_timing_util {
...
+      auto current_watermark = prod_watermarks.get_watermark(producer_name);
+      if (current_watermark) {
+         const auto watermark = *current_watermark;
+         if (watermark.first > block_num) {
+            // if I have a watermark block number then I need to wait until after that watermark
+            minimum_offset = watermark.first - block_num + 1;
          }
+         if (watermark.second.slot > current_block_slot) {
+            // if I have a watermark block timestamp then I need to wait until after that watermark timestamp
+            minimum_offset = std::max(minimum_offset, watermark.second.slot - current_block_slot + 1);
```
greg7mdp, line 95: Seems inconsistent with line 91? Shouldn't it be:
```suggestion
            minimum_offset = std::max(minimum_offset, watermark.second.slot - block_num + 1);
```

BTW, I don't  understand the difference between `chain.head_block_num()` and `chain.head_block_time().slot`?

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -25,22 +52,95 @@ namespace block_timing_util {
...
+      uint32_t minimum_slot = current_block_slot + minimum_offset;
+      size_t   minimum_slot_producer_index =
+         (minimum_slot % (active_schedule.size() * chain::config::producer_repetitions)) / chain::config::producer_repetitions;
+      if (producer_index == minimum_slot_producer_index) {
+         // this is the producer for the minimum slot, go with that
+         return minimum_slot;
       } else {
-         return hard_deadline;
+         // calculate how many rounds are between the minimum producer and the producer in question
+         size_t producer_distance = producer_index - minimum_slot_producer_index;
```
greg7mdp, line 108: Is it guaranteed that `producer_index >= minimum_slot_producer_index`

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -1,13 +1,40 @@
...
+
+         return itr->second;
+      }
+   private:
+      std::map<chain::account_name, producer_watermark> _producer_watermarks;
+   };
+
    // Calculate when a producer can start producing a given block represented by its block_time
    //
    // In the past, a producer would always start a block `config::block_interval_us` ahead of its block time. However,
```
greg7mdp, line 40: Hum, I don't quite understand this comment. Suppose we have 12 repetitions. given block_interval=500 ms and cpu effort=400 ms, assuming the our round start at time point 0; in the past, the block start time points would be at time point : `-500, 0, ... , 5000, 5500`, so the last block would start as expected in the 12th slot.

With the new approach, the block time points would become `-500, -100, 300, 700, 1100 ..., 3900, 4300`. Does this mean we start our last block at 4.3s and that we process no more transactions for 1.7s until the next producer schedule starts?

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -25,22 +52,95 @@ namespace block_timing_util {
...
+         uint32_t next_block_slot = first_minimum_producer_slot + (producer_distance * chain::config::producer_repetitions);
+         return next_block_slot;
       }
    }
-};
+
+   // Return the *next* block start time according to its block time slot.
+   // Returns empty optional if no producers are in the active_schedule.
+   // block_num is only used for watermark minimum offset.
+   inline std::optional<fc::time_point> calculate_producer_wake_up_time(uint32_t cpu_effort_us, uint32_t block_num,
```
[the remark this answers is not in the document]
heifner: Hmm, I don't think that would be best. These are utility functions.
https://stackoverflow.com/questions/10876930/should-one-never-use-static-inline-function

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
heifner: Yes. See https://github.com/AntelopeIO/leap/pull/867#issuecomment-1481596497

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -25,22 +52,95 @@ namespace block_timing_util {
...
+         if (watermark.second.slot > current_block_slot) {
+            // if I have a watermark block timestamp then I need to wait until after that watermark timestamp
+            minimum_offset = std::max(minimum_offset, watermark.second.slot - current_block_slot + 1);
+         }
+      }
+
+      // this producers next opportunity to produce is the next time its slot arrives after or at the calculated minimum
+      uint32_t minimum_slot = current_block_slot + minimum_offset;
+      size_t   minimum_slot_producer_index =
+         (minimum_slot % (active_schedule.size() * chain::config::producer_repetitions)) / chain::config::producer_repetitions;
```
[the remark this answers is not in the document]
heifner: Currently `producer_authority_schedule` has no knowledge of producer repetitions or block time slots. I don't think I'm following what you are suggesting.

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
heifner: Apparently not, the next line checks:
```
         // check for unsigned underflow
         if (producer_distance > producer_index) {
            producer_distance += active_schedule.size();
         }
```

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
heifner: No. 91 is working with a block_num watermark. Line 95 is working with a block_slot watermark.

Block numbers can never be skipped, but blocks can fail to be produced. When blocks are not produced when they should be block slots are missed. Block slots are just an abbreviated time. 

To illustrate this check out `to_time_point()` of `block_timestamp`:
```
            int64_t msec = slot * (int64_t)IntervalMs;
            msec += EpochMs;
            return fc::time_point(fc::milliseconds(msec));
```
```
Example:

block_slot = 13 - time 2023-08-11T10:00:00.000 - block_num 50
block_slot = 14 - time 2023-08-11T10:00:00.500 - block_num 51
block_slot = 15 - time 2023-08-11T10:00:01.000 - block_num 52
block_slot = 16 - time 2023-08-11T10:00:01.500 - block fails to produce
block_slot = 17 - time 2023-08-11T10:00:02.000 - block fails to produce
block_slot = 18 - time 2023-08-11T10:00:02.500 - block_num 53
```

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
heifner: Refactored to not require an optional.

plugins/producer_plugin/producer_plugin.cpp
heifner: Excellent point. I wrote up a paragraph why it shouldn't correspond to the other 5 seconds and in doing so convinced myself of the opposite. Thanks.

I do wonder the wisdom of the current 5 second limit on speculative block production. The idea is that you don't want to speculative execute on too old of state. However, if a block producer misses a round then this node will stop processing trxs and wait for a block. Since everyone is doing that, then all trxs are held up until a block producer produces a block and sends it along.

I wonder if the new IF provides us with a different liveliness we could use to help make better determination. If we know that a series of blocks were not produced we could safely speculative execute trxs on older state.

heifner:
> > Speculative blocks on non-producer nodes will not restart unless a new block is received or it is exhausted. Speculative blocks on producing nodes will not restart until a production window is reached or it is exhausted.
> 
> What is the rationale for these two changes?

Before this PR, block deadline was set for speculative blocks the same as production blocks for block production time. Because of when blocks are processed this could result in small block speculative windows. When a block came in close to deadline it would cause us to repeatedly start block with a very small deadline, abort block because of deadline, start another block and immediately abort again until the block time window was passed.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: If the node is connected to multiple peers, shouldn't we assume that its state will be up-to-date, as it will receive blocks when they are produced? So I don't see the rationale to assume that our state is old vs the whole chain (again provided that we are connected to multiple peers).

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
greg7mdp: Thanks for the explanation Kevin!

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
greg7mdp: When there is an underflow, I'm not understanding how `producer_distance += active_schedule.size();` addresses the issue?

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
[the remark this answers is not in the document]
greg7mdp: Never mind, no big deal, I just felt that the `schedule` class should be responsible for telling which producer is responsible for a given slot.

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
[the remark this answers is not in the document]
greg7mdp: I guess there are arguments for both. OK.

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
greg7mdp: Sounds odd that we guarantee having 1.1s gaps where no transactions are included in blocks and no blocks produced.

greg7mdp:
Please update line 47:
`// approach, the block time points would become -500, -100, 300, 700, 1200 ...`

`1200` -> `1100`

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
greg7mdp: Oh, never mind, I forgot this was only for non-producer nodes.

heifner: It is for production nodes.

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -1,13 +1,42 @@
...
+   class producer_watermarks {
+   public:
+      void consider_new_watermark(chain::account_name producer, uint32_t block_num, chain::block_timestamp_type timestamp) {
+         auto itr = _producer_watermarks.find(producer);
+         if (itr != _producer_watermarks.end()) {
+            itr->second.first  = std::max(itr->second.first, block_num);
+            itr->second.second = std::max(itr->second.second, timestamp);
+         } else {
+            _producer_watermarks.emplace(producer, std::make_pair(block_num, timestamp));
+         }
```
greg7mdp, line 24:
```suggestion
         auto& w = _producer_watermarks[producer];
         w = std::max(w, std::make_pair(block_num, timestamp));
```

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
greg7mdp: So isn't it odd that we guarantee having 1.1s gaps where no transactions are included in blocks and no blocks produced?

plugins/producer_plugin/producer_plugin.cpp
heifner: See updated PR description.

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
heifner:
```
Example:

block_slot = 13 - time 2023-08-11T10:00:00.000 - block_num 50
block_slot = 14 - time 2023-08-11T10:00:00.500 - block_num 51
block_slot = 15 - time 2023-08-11T10:00:01.000 - block_num 52
!!FORK!!
block_slot = 16 - time 2023-08-11T10:00:01.500 - block_num 50 <- lower block_num, but higher block_slot
block_slot = 17 - time 2023-08-11T10:00:02.000 - block_num 51
```

greg7mdp: Oh my I didn't think of that, you are right. Thanks for explaining.

greg7mdp (approved):

plugins/producer_plugin/test/test_block_timing_util.cpp
```diff
@@ -34,73 +35,219 @@ BOOST_AUTO_TEST_CASE(test_calculate_block_deadline) {
...
+      std::set<chain::account_name>          producers{"p1"_n, "p2"_n, "p3"_n};
+      std::vector<chain::producer_authority> active_schedule{{"p1"_n}, {"p2"_n}};
+      const uint32_t prod_round_1st_block_slot = 100 * active_schedule.size() * eosio::chain::config::producer_repetitions - 1;
+      for (uint32_t i = 0; i < static_cast<uint32_t>(config::producer_repetitions * active_schedule.size() * 3); ++i) { // 3 rounds to test boundaries
+         block_timestamp_type block_timestamp(prod_round_1st_block_slot + i);
+         auto block_time = block_timestamp.to_time_point();
+         BOOST_CHECK_EQUAL(calculate_producer_wake_up_time(full_cpu_effort, 2, block_timestamp, producers, active_schedule, empty_watermarks), block_time);
+      }
+   }
+   { // We have all producers in active_schedule in active_schedule of 21 (plus a couple of extra producers configured), we should produce every block
```
linh2931, line 134: Very minor: change `in active_schedule in active_schedule` to `in active_schedule `

linh2931 (approved):

PR #1482, opened 2023-08-08
Fix base64 encoding

### LINKED ISSUE #1461: Invalid Base64 padding returned by APIs

```
```

heifner:
Remove the invalid addition of "=" to all base64 encoded strings from `fc::variant`.
Also update base64 implementation from: https://github.com/ReneNyffenegger/cpp-base64

Resolves #1461

### DIFF
--- a/libraries/libfc/include/fc/crypto/base64.hpp
+++ b/libraries/libfc/include/fc/crypto/base64.hpp
@@ -1,14 +1,15 @@
 #pragma once
 #include <string>
+#include <string_view>
 
 namespace fc {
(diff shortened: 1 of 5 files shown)


spoonincode (dismissed):
I know we ruled out boost's base64 impl, but can we use openssl's base64 instead of having our own impl?

heifner:
> I know we ruled out boost's base64 impl, but can we use openssl's base64 instead of having our own impl?

Doesn't seem to provide `base64url_decode` url decoding used by webauthn.

linh2931 (approved):
I compared with the upstream version. It looked good. Approve the PR if we decide use this.

Issue #1456, opened 2023-07-27
Do not create/use/cache execution context per contract


linh2931:
A part of https://github.com/eosnetworkfoundation/product/issues/149.

Currently an execution context is created/used/cached per contract. Towards our goal of sharing the same compiled WASM code across all the threads, we need to make an execution context per thread (or wasm interface). A good side effect of this is memory usage reduction since execution context per contract is not cached any more in physical memory in running `nodeos`.

PR #1484, opened 2023-08-08
The companion Leap PR for using a single execution context per wasm interface

linh2931:
The companion PR of EOS VM https://github.com/AntelopeIO/eos-vm/pull/19 for using a single execution context per wasm interface.

Resolves https://github.com/AntelopeIO/leap/issues/1456.


heifner (approved):

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -129,6 +129,12 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
          _instantiated_module(std::move(mod)) {}
 
       void apply(apply_context& context) override {
+         // Reset execution context (reused per thread)
+         _runtime->_exec_ctx.set_module(&(_instantiated_module->get_module()));
+         _instantiated_module->set_context(&_runtime->_exec_ctx);
+         _instantiated_module->reset_max_call_depth();
+         _instantiated_module->reset_max_pages();
```
spoonincode, line 136: I was originally worried about the case where the apply options change between apply() calls right here.

For example `reset_max_call_depth()` sets the context's max call depth to `initial_max_call_depth` which was computed based on the options passed _when the backend is created_,
https://github.com/AntelopeIO/eos-vm/blob/8e1d571db99ff1eecd119c299b42886f72ef40b4/include/eosio/vm/backend.hpp#L177-L180
But the max call depth that should be configured at this point in time is what the current options protocol limits are.

But a few lines down from here,
https://github.com/AntelopeIO/leap/blob/295bb7cf2e3e01c8c920dd4344248853cdfe27b7/libraries/chain/webassembly/runtimes/eos-vm.cpp#L141-L147
The important part is the `initialize()` call that,
https://github.com/AntelopeIO/eos-vm/blob/8e1d571db99ff1eecd119c299b42886f72ef40b4/include/eosio/vm/backend.hpp#L198-L203
seems to set up the context correctly for each apply once `WASM_LIMITS2` is activated.

spoonincode (approved):

libraries/chain/webassembly/runtimes/eos-vm.cpp
linh2931: Thank you for your careful analysis.

heifner (approved):

PR #1489, opened 2023-08-14
Require dry-run trxs to have authorization and add tests

### LINKED ISSUE #1083: Add dry-run unittests
We have no dry-run unittests. See https://github.com/AntelopeIO/leap/pull/1081 for an example of an issue that slipped through. Create corresponding dry-run unittests similar to `read_only_trx_tests.cpp`

Include this test:
```
BOOST_FIXTURE_TEST_CASE(dry_run_auth_test, read_only_trx_tester) { try {
      set_up_test_contract();

      // verify dry-run transaction allows authorizations, but doesn't verify
      send_db_api_transaction("insert"_n, insert_data, {{"alice"_n, config::active_name}}, transaction_metadata::trx_type::dry_run);
} FC_LOG_AND_RETHROW() }
```
where `send_db_api_transaction` has
```
      } else if ( type == transaction_metadata::trx_type::dry_run ) {
         // sign with some irrelevant sigs which should be allowed
         trx.sign(get_private_key("invalid"_n, "active"), control->get_chain_id());
```

heifner:
Add requirement for dry-run/compute transactions to require authorization.
Add tests for dry-run/compute transactions.
Also some minor cleanup to unneeded includes.

Resolves #1083

### DIFF
--- a/libraries/chain/transaction_context.cpp
+++ b/libraries/chain/transaction_context.cpp
@@ -1,4 +1,5 @@
 #include <eosio/chain/apply_context.hpp>
+#include <eosio/chain/account_object.hpp>
 #include <eosio/chain/transaction_context.hpp>
 #include <eosio/chain/authorization_manager.hpp>
 #include <eosio/chain/exceptions.hpp>
@@ -7,16 +8,6 @@
 #include <eosio/chain/transaction_object.hpp>
 #include <eosio/chain/global_property_object.hpp>
 #include <eosio/chain/deep_mind.hpp>
-
-#pragma push_macro("N")
-#undef N
-#include <boost/accumulators/accumulators.hpp>
-#include <boost/accumulators/statistics/stats.hpp>
-#include <boost/accumulators/statistics/min.hpp>
-#include <boost/accumulators/statistics/max.hpp>
-#include <boost/accumulators/statistics/weighted_mean.hpp>
-#include <boost/accumulators/statistics/weighted_variance.hpp>
-#pragma pop_macro("N")
 
 #include <chrono>
 
@@ -828,7 +819,7 @@ namespace eosio { namespace chain {
                actors.insert( auth.actor );
          }
       }
-      EOS_ASSERT( one_auth || is_transient(), tx_no_auths, "transaction must have at least one authorization" );
+      EOS_ASSERT( one_auth || is_read_only(), tx_no_auths, "transaction must have at least one authorization" );
 
       if( enforce_actor_whitelist_blacklist ) {
          control.check_actor_list( actors );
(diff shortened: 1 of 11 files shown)


libraries/chain/transaction_context.cpp
```diff
@@ -828,7 +819,7 @@ namespace eosio { namespace chain {
                actors.insert( auth.actor );
          }
       }
-      EOS_ASSERT( one_auth || is_transient(), tx_no_auths, "transaction must have at least one authorization" );
+      EOS_ASSERT( one_auth || is_read_only(), tx_no_auths, "transaction must have at least one authorization" );
```
linh2931, line 822: Do we want to call out this in 5.0 Release Note as this is a backward incompatible change?

unittests/dry_run_trx_tests.cpp
```diff
@@ -0,0 +1,331 @@
...
+         .name     = "alice"_n,
+         .owner    = authority( get_public_key( "alice"_n, "owner" ) ),
+         .active   = authority( get_public_key( "alice"_n, "active" ) )
+      }
+   };
+
+   send_action(act, false); // should not throw
+   send_action(act, false); // should not throw
+   send_action(act, true); // should not throw
+   BOOST_CHECK_THROW(control->get_account("alice"_n), fc::exception); // not actually created
```
linh2931, line 118: Like this check. Maybe you can wait one block to check to make sure state is really not changed by dry-run.

libraries/chain/transaction_context.cpp
heifner: dry-run trxs would fail when trying to record account cpu/net for first-authorizers. So unless protocol feature `only_bill_first_authorizer` was not enabled (it is enabled on all Antelope chains I know about) then it would fail with an unknown account error. I didn't test without `only_bill_first_authorizer`, so it is possible it would fail somewhere else even if the protocol feature was not enabled. This change just provides a much better error message.

unittests/dry_run_trx_tests.cpp
```diff
@@ -0,0 +1,331 @@
...
+      set_abi( "noauthtable"_n, test_contracts::no_auth_table_abi() );
+      produce_block();
+
+      insert_data = abi_ser.variant_to_binary( "insert", mutable_variant_object()
+         ("user", "alice") ("id", 1) ("age", 10),
+         abi_serializer::create_yield_function( abi_serializer_max_time ) );
+      getage_data = abi_ser.variant_to_binary("getage", mutable_variant_object()
+         ("user", "alice"),
+         abi_serializer::create_yield_function( abi_serializer_max_time ));
+      produce_block();
```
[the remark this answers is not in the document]
heifner: No reason, removed.

PR #1498, opened 2023-08-15
Set specific sliced pages for read-only threads in EOS VM OC Tierup 

linh2931:
https://github.com/AntelopeIO/leap/pull/1488 missed setting specific sliced pages for read-only threads in OC tierup.

This PR
- sets sliced pages for main and read-only threads correctly in OC tierup
- makes sure sliced pages are not set up in OC tierup and OC runtime at the same time by using unique_ptr.

Before the PR, `nodeos` running the integration test `read-only-trx-parallel-eos-vm-oc-test` (with 6 read-only threads) uses 29 TB virtual memory; after the PR, it uses 4.7 TB virtual memory as expected.


heifner:
> Before the PR, `nodeos` running the integration test `read-only-trx-parallel-eos-vm-oc-test` (with 6 read-only threads) uses 29 TB virtual memory; after the PR, it uses 627 GB virtual memory.

Can you add a verification to the test?

libraries/chain/wasm_interface_collection.cpp
```diff
@@ -9,26 +9,29 @@ namespace eosio::chain {
...
       exec = std::make_unique<eosvmoc::executor>(cc);
+      mem  = std::make_unique<eosvmoc::memory>(eosvmoc::memory::sliced_pages_for_ro_thread);
    }
 
    eosvmoc::code_cache_async cc;
 
    // Each thread requires its own exec and mem.
    thread_local static std::unique_ptr<eosvmoc::executor> exec;
-   thread_local static eosvmoc::memory mem;
+   thread_local static std::unique_ptr<eosvmoc::memory>   mem;
```
spoonincode, line 30: (obviously not in this PR) but I'd really like for us to noodle on how to move away from these `thread_local`s, they have dangerous lifetime semantics imo. In this particular example, when `eosvmoc_tier` is destroyed, `cc` is destroyed, but the `exec` created with a reference to that `cc` lives on. So if `executor`'s dtor (which will run well after `eosvmoc_tier`'s dtor) touches the `code_cache_async` that was passed to it via reference: bad news! It's a real trap -- passing a reference to some object to the ctor of an object implies that reference will be valid for the lifetime of the object.. but it's not true here.

linh2931: Thanks. I will be looking into moving away from those thread_locals.

heifner (approved):

spoonincode (approved):

PR #1526, opened 2023-08-17
negate after converting to unsigned to avoid UB on minimum signed value

spoonincode:
Negating the minimum value of a signed integer is undefined behavior since the resulting value cannot be represented by the type (results in a signed overflow). However, casting to an unsigned and then negating that _is_ defined behavior, and I believe the appropriate solution here, as both of those operations are well defined for all inputs.

### DIFF
--- a/libraries/chain/webassembly/console.cpp
+++ b/libraries/chain/webassembly/console.cpp
@@ -47,7 +47,7 @@ namespace eosio { namespace chain { namespace webassembly {
 			unsigned __int128 val_magnitude;
 
 			if( is_negative )
-				val_magnitude = static_cast<unsigned __int128>(-*val); // Works even if val is at the lowest possible value of a int128_t
+				val_magnitude = -static_cast<unsigned __int128>(*val); // Works even if val is at the lowest possible value of a int128_t
 			else
 				val_magnitude = static_cast<unsigned __int128>(*val);
 
--- a/libraries/chain/webassembly/console.cpp
+++ b/libraries/chain/webassembly/console.cpp
@@ -46,6 +46,7 @@ namespace eosio { namespace chain { namespace webassembly {
 			bool is_negative = (*val < 0);
 			unsigned __int128 val_magnitude;
 
+			// negate after conversion to unsigned to avoid possible integer overflow
 			if( is_negative )
 				val_magnitude = -static_cast<unsigned __int128>(*val); // Works even if val is at the lowest possible value of a int128_t
 			else
--- a/libraries/wasm-jit/Source/WAST/ParseNumbers.cpp
+++ b/libraries/wasm-jit/Source/WAST/ParseNumbers.cpp
@@ -239,7 +239,7 @@ bool tryParseInt(ParseState& state,UnsignedInt& outUnsignedInt,I64 minSignedValu
 		return false;
 	};
 
-	outUnsignedInt = isNegative ? -UnsignedInt(I64(u64)) : UnsignedInt(u64);
+	outUnsignedInt = isNegative ? -UnsignedInt(u64) : UnsignedInt(u64);
 		
 	++state.nextToken;
 	WAVM_ASSERT_THROW(nextChar <= state.string + state.nextToken->begin);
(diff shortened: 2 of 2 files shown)


spoonincode:
This isn't consensus, though a change in console output would be.. undesirable. There is an existing test for `std::numeric_limits<int128_t>::lowest()` to give us some confidence that nothing has actually changed on supported platforms,

https://github.com/AntelopeIO/leap/blob/6132e3be2225fe1de088ca31625238bc07f65636/unittests/test-contracts/test_api/test_print.cpp#L35-L48

https://github.com/AntelopeIO/leap/blob/6132e3be2225fe1de088ca31625238bc07f65636/unittests/api_tests.cpp#L2649-L2660

greg7mdp (approved):

libraries/chain/webassembly/console.cpp
```diff
@@ -47,7 +47,7 @@ namespace eosio { namespace chain { namespace webassembly {
 			unsigned __int128 val_magnitude;
 
 			if( is_negative )
-				val_magnitude = static_cast<unsigned __int128>(-*val); // Works even if val is at the lowest possible value of a int128_t
+				val_magnitude = -static_cast<unsigned __int128>(*val); // Works even if val is at the lowest possible value of a int128_t
```
linh2931, line 50: You might want to add a comment here about why `-` is outside.

linh2931 (approved):

spoonincode:
I discovered two more cases of this behavior and decided to fix them as part of this PR in f68377f, I would like re-review when you have a chance @greg7mdp and @linh2931 

WAST parsing is not consensus: it's only used in unit tests. In this particular case the WAST that tripped this error was from the test
https://github.com/AntelopeIO/leap/blob/fb9a0e7ffe3036301d9c2ed7ec90f51f82173250/unittests/wasm_tests.cpp#L351

linh2931 (approved):

libraries/wasm-jit/Source/WAST/ParseNumbers.cpp
```diff
@@ -239,7 +239,7 @@ bool tryParseInt(ParseState& state,UnsignedInt& outUnsignedInt,I64 minSignedValu
 		return false;
 	};
 
-	outUnsignedInt = isNegative ? UnsignedInt(-I64(u64)) : UnsignedInt(u64);
+	outUnsignedInt = isNegative ? -UnsignedInt(I64(u64)) : UnsignedInt(u64);
```
greg7mdp, line 242: Do we still need the `I64` cast anymore?

spoonincode: it does seem completely redundant now, and everything still passes with it removed..

libraries/wasm-jit/Source/WAST/ParseNumbers.cpp
```diff
@@ -229,7 +229,7 @@ bool tryParseInt(ParseState& state,UnsignedInt& outUnsignedInt,I64 minSignedValu
 	{
 	case t_decimalInt:
 		isNegative = parseSign(nextChar);
-		u64 = parseDecimalUnsignedInt(nextChar,state,isNegative ? U64(-minSignedValue) : maxUnsignedValue,"int literal");
+		u64 = parseDecimalUnsignedInt(nextChar,state,isNegative ? -U64(minSignedValue) : maxUnsignedValue,"int literal");
```
greg7mdp, line 232: This is kind of interesting, as the old implementation `U64(-minSignedValue)` is exactly the UB case that we are trying to avoid. So I guess either:
- our compilers were doing what the new implementation does in that case
- or we never parsed values greater that the `maxValue` provided.

spoonincode: > our compilers were doing what the new implementation does in that case

Probably, yeah. I think this is a case of "technically UB because it results in integer overflow.. but likely happens to do what we expect on platforms we care about". (we should still avoid it of course)

> or we never parsed values greater that the `maxValue` provided.

These two lines were hitting the following error so we were actually getting here,
> runtime error: negation of -9223372036854775808 cannot be represented in type 'I64' (aka 'long'); cast to an unsigned type to negate this value to itself

libraries/wasm-jit/Source/WAST/ParseNumbers.cpp
spoonincode: I was a little hesitant because this case is a little unique due to the u64->u32 truncation and wasn't sure if `-UnsignedInt(u64)` vs `UnsignedInt(-u64)` mattered.., but it seems equivalent.

greg7mdp (approved):

PR #1530, opened 2023-08-18
Update gcc version check in CMakeLists.txt.

greg7mdp:
I missed this change with the C++20 switch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,8 +37,8 @@ set( LEAP_UTIL_EXECUTABLE_NAME leap-util )
 
 # http://stackoverflow.com/a/18369825
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 10.5)
-        message(FATAL_ERROR "GCC version must be at least 10.5.0!")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 10.2)
+        message(FATAL_ERROR "GCC version must be at least 10.2.0!")
     endif()
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,8 +37,8 @@ set( LEAP_UTIL_EXECUTABLE_NAME leap-util )
 
 # http://stackoverflow.com/a/18369825
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
-        message(FATAL_ERROR "GCC version must be at least 9.0!")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 10.5)
+        message(FATAL_ERROR "GCC version must be at least 10.5.0!")
     endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


greg7mdp:
@spoonincode it looks like the ubuntu 20.04 runner has gcc 10.3, which surprosed me because the [package](https://packages.ubuntu.com/focal-updates/devel/g++-10-x86-64-linux-gnu) lists 10.5. Do you know how to explain this discrepancy?

spoonincode:
The build platforms are cached to improve CI performance; the platform cache was probably built before Ubuntu 20.04 pushed out the upgrade to 10.5.

(the platform cache doesn't really help all that much for the current Ubuntu 20 & 22 environment since they do nothing more than install packages, but it can really be beneficial for something like the pinned builds or potentially other scenarios where we need to compile a whole toolchain to perform a build)

To investigate this theory more, we can take a look at the sha256 of the `ubuntu20.Dockerfile`: `2e38bdf1373bdc4de194c5a28480726a3b6d74ea3c249369f1fa4e8942f9614b`. You can also find this being used as the docker image tag name in the CI run, https://github.com/AntelopeIO/leap/actions/runs/5904908678/job/16017979866?pr=1530#step:2:19.

Then we can look up that image on github,
https://github.com/orgs/AntelopeIO/packages/container/builders/113490144?tag=2e38bdf1373bdc4de194c5a28480726a3b6d74ea3c249369f1fa4e8942f9614b
> 22 days ago 

I'm not sure how to determine when Ubuntu pushed out a particular package update? The best I could find was an acceptance of the 10.5 package on Jul 31 -- less than 22 days ago.
https://lists.ubuntu.com/archives/focal-changes/2023-July/042080.html

There are a couple things you can do to repopulate the cache. One is to just make a benign modification to `ubuntu20.Dockerfile`: add a comment, some whitespace, whatever, to change the hash of the file. The other is to remove the underlying image which will cause it to be rebuilt next time CI needs it. It seems you need to do that [from this page though](https://github.com/orgs/AntelopeIO/packages/container/builders/versions).
shot-2023-08-18_13-58-31
I'm not entirely sure what the permissions are on doing that (I think you can do that?). Be aware that deleting this while a CI workflow is running may break those running workflows as it pulls a resource out from under it, so maybe just eyeball that nothing else is running before doing that.

Either choice is fine.

CMakeLists.txt
```diff
@@ -37,8 +37,8 @@ set( LEAP_UTIL_EXECUTABLE_NAME leap-util )
 
 # http://stackoverflow.com/a/18369825
 if("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
-    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 9.0)
-        message(FATAL_ERROR "GCC version must be at least 9.0!")
+    if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 10.5)
```
spoonincode, line 40: I think Debian 11 is still on 10.2. Do we really feel we must have 10.5 over 10.x? A bit of a shame to block out Debian 11 unless we have reason to believe we need to.

greg7mdp: Probably 10.2 is fine. I'll update the check for 10.2, and try the `ubuntu20.Dockerfile` to regenerate!

heifner (approved):

PR #1538, opened 2023-08-21
Clean up of bls_private_key, bls_public_key and bls_signature, …

systemzax:
…removed string prefix, added base58 to_string + parsing code for bls_private_key + unit test, removed old unit test

### DIFF
--- a/libraries/libfc/include/fc/crypto/bls_private_key.hpp
+++ b/libraries/libfc/include/fc/crypto/bls_private_key.hpp
@@ -5,14 +5,10 @@
 
 namespace fc::crypto::blslib {
 
-   namespace config {
-      constexpr const char* bls_private_key_base_prefix = "PVT";
-      constexpr const char* bls_private_key_prefix = "BLS";
-   };
-
    class bls_private_key
    {
       public:
+
          bls_private_key() = default;
          bls_private_key( bls_private_key&& ) = default;
          bls_private_key( const bls_private_key& ) = default;
--- a/libraries/libfc/include/fc/crypto/bls_public_key.hpp
+++ b/libraries/libfc/include/fc/crypto/bls_public_key.hpp
@@ -12,13 +12,6 @@ namespace fc { namespace crypto { namespace blslib {
 
    using namespace std;
 
-   namespace config {
-      constexpr const char* bls_public_key_legacy_prefix = "EOS";
-      constexpr const char* bls_public_key_base_prefix = "PUB";
-      constexpr const char* bls_public_key_prefix = "BLS";
-
-   };
-
    class bls_public_key
    {
       public:
@@ -32,24 +25,8 @@ namespace fc { namespace crypto { namespace blslib {
             _pkey = pkey;
          }
 
-/*         bls_public_key( G1Element pkey ){
-            _pkey = pkey.Serialize();
-         }
-*/
-         //bls_public_key( const bls_signature& c, const sha256& digest, bool check_canonical = true );
-
-/*         bls_public_key( storage_type&& other_storage )
-         :_storage(forward<storage_type>(other_storage))
-         {}
-*/
-         bool valid()const;
-
-         size_t which()const;
-
          // serialize to/from string
          explicit bls_public_key(const string& base58str);
-         //std::string to_string() const;
-         //std::string to_string() ;
 
          std::string to_string(const fc::yield_function_t& yield = fc::yield_function_t()) const;
 
@@ -60,9 +37,6 @@ namespace fc { namespace crypto { namespace blslib {
          
 
          friend std::ostream& operator<< (std::ostream& s, const bls_public_key& k);
-         //friend bool operator == ( const bls_public_key& p1, const bls_public_key& p2);
-         //friend bool operator != ( const bls_public_key& p1, const bls_public_key& p2);
-         //friend bool operator < ( const bls_public_key& p1, const bls_public_key& p2);
          friend struct reflector<bls_public_key>;
          friend class bls_private_key;
    }; // bls_public_key
--- a/libraries/libfc/include/fc/crypto/bls_signature.hpp
+++ b/libraries/libfc/include/fc/crypto/bls_signature.hpp
@@ -42,7 +42,7 @@ namespace fc { namespace crypto { namespace blslib {
 
 //         size_t which() const;
 
-         size_t variable_size() const;
+         //size_t variable_size() const;
 
 
          bls12_381::g2 _sig;
@@ -58,7 +58,7 @@ namespace fc { namespace crypto { namespace blslib {
          //friend bool operator == ( const bls_signature& p1, const bls_signature& p2);
          //friend bool operator != ( const bls_signature& p1, const bls_signature& p2);
         //friend bool operator < ( const bls_signature& p1, const bls_signature& p2);
-         friend std::size_t hash_value(const bls_signature& b); //not cryptographic; for containers
+         //friend std::size_t hash_value(const bls_signature& b); //not cryptographic; for containers
          friend bool operator == ( const bls_signature& p1, const bls_signature& p2);
          friend struct reflector<bls_signature>;
          friend class bls_private_key;
--- a/libraries/libfc/src/crypto/bls_private_key.cpp
+++ b/libraries/libfc/src/crypto/bls_private_key.cpp
@@ -25,38 +25,17 @@ namespace fc::crypto::blslib {
       return bls_private_key(v);
    }
 
-   template<typename Data>
-   Data from_wif( const string& wif_key )
-   {
-   /*   auto wif_bytes = from_base58(wif_key);
-      FC_ASSERT(wif_bytes.size() >= 5);
-      auto key_bytes = vector<char>(wif_bytes.begin() + 1, wif_bytes.end() - 4);
-      fc::sha256 check = fc::sha256::hash(wif_bytes.data(), wif_bytes.size() - 4);
-      fc::sha256 check2 = fc::sha256::hash(check);
-
-      FC_ASSERT(memcmp( (char*)&check, wif_bytes.data() + wif_bytes.size() - 4, 4 ) == 0 ||
-                memcmp( (char*)&check2, wif_bytes.data() + wif_bytes.size() - 4, 4 ) == 0 );
-
-      return Data(fc::variant(key_bytes).as<typename Data::data_type>());*/
-   }
-
    static vector<uint8_t> priv_parse_base58(const string& base58str)
    {
-      const auto pivot = base58str.find('_');
-/*
-      if (pivot == std::string::npos) {
-         // wif import
-         using default_type = std::variant_alternative_t<0, bls_private_key::storage_type>;
-         return bls_private_key::storage_type(from_wif<default_type>(base58str));
-      } else {
-         constexpr auto prefix = config::private_key_base_prefix;
-         const auto prefix_str = base58str.substr(0, pivot);
-         FC_ASSERT(prefix == prefix_str, "Private Key has invalid prefix: ${str}", ("str", base58str)("prefix_str", prefix_str));
+      std::vector<char> v1 = fc::from_base58(base58str);
 
-         auto data_str = base58str.substr(pivot + 1);
-         FC_ASSERT(!data_str.empty(), "Private Key has no data: ${str}", ("str", base58str));
-         return base58_str_parser<bls_private_key::storage_type, config::private_key_prefix>::apply(data_str);
-      }*/
+      FC_ASSERT(v1.size() == 32);
+
+      std::vector<uint8_t> v2(32);
+
+      std::copy(v1.begin(), v1.end(), v2.begin());
+
+      return v2;
    }
 
    bls_private_key::bls_private_key(const std::string& base58str)
@@ -66,12 +45,13 @@ namespace fc::crypto::blslib {
    std::string bls_private_key::to_string(const fc::yield_function_t& yield) const
    {
 
-      /*PrivateKey pk = AugSchemeMPL().KeyGen(_seed);
+      std::vector<char> v2(32);
+      std::copy(_seed.begin(), _seed.end(), v2.begin());
 
-      vector<uint8_t> pkBytes   pk.Serialize()
+      std::string data_str = fc::to_base58(v2, yield);
 
-      auto data_str = Util::HexStr(pkBytes);
-      return std::string(config::private_key_base_prefix) + "_" + data_str;*/
+      return data_str;
+
    }
 
 } // fc::crypto::blslib
--- a/libraries/libfc/src/crypto/bls_public_key.cpp
+++ b/libraries/libfc/src/crypto/bls_public_key.cpp
@@ -4,39 +4,10 @@
 
 namespace fc { namespace crypto { namespace blslib {
 
-  /* struct recovery_visitor : fc::visitor<bls_public_key::storage_type> {
-      recovery_visitor(const sha256& digest, bool check_canonical)
-      :_digest(digest)
-      ,_check_canonical(check_canonical)
-      {}
-
-      template<typename SignatureType>
-      bls_public_key::storage_type operator()(const SignatureType& s) const {
-         return bls_public_key::storage_type(s.recover(_digest, _check_canonical));
-      }
-
-      const sha256& _digest;
-      bool _check_canonical;
-   };
-
-   bls_public_key::bls_public_key( const bls_signature& c, const sha256& digest, bool check_canonical )
-   :_storage(std::visit(recovery_visitor(digest, check_canonical), c._storage))
-   {
-   }
-
-   size_t bls_public_key::which() const {
-      return _storage.index();
-   }*/
-
    static bls12_381::g1 parse_base58(const std::string& base58str)
    {  
       
-      constexpr auto prefix = config::bls_public_key_base_prefix;
-      const auto pivot = base58str.find('_');
-      const auto prefix_str = base58str.substr(0, pivot);
-      auto data_str = base58str.substr(pivot + 1);
-   
-      std::vector<char> v1 = fc::from_base58(data_str);
+      std::vector<char> v1 = fc::from_base58(base58str);
 
       FC_ASSERT(v1.size() == 48);
       std::array<uint8_t, 48> v2;
@@ -50,13 +21,6 @@ namespace fc { namespace crypto { namespace blslib {
    :_pkey(parse_base58(base58str))
    {}
 
-
-   bool bls_public_key::valid()const
-   {
-      //return std::visit(is_valid_visitor(), _storage);
-   }
-
-
    std::string bls_public_key::to_string(const fc::yield_function_t& yield)const {
 
       std::vector<char> v2;
@@ -64,10 +28,10 @@ namespace fc { namespace crypto { namespace blslib {
       std::copy(bytes.begin(), bytes.end(), std::back_inserter(v2));
 
       std::string data_str = fc::to_base58(v2, yield);
-
-      //std::string data_str = Util::HexStr(_pkey);
       
-      return std::string(config::bls_public_key_base_prefix) + "_" + data_str;
+      //return std::string(config::bls_public_key_base_prefix) + "_" + data_str;
+      return data_str;
+
 
    }
 
--- a/libraries/libfc/src/crypto/bls_signature.cpp
+++ b/libraries/libfc/src/crypto/bls_signature.cpp
@@ -4,29 +4,11 @@
 
 namespace fc { namespace crypto { namespace blslib {
 
-   struct hash_visitor : public fc::visitor<size_t> {
-/*      template<typename SigType>
-      size_t operator()(const SigType& sig) const {
-         static_assert(sizeof(sig._data.data) == 65, "sig size is expected to be 65");
-         //signatures are two bignums: r & s. Just add up least significant digits of the two
-         return *(size_t*)&sig._data.data[32-sizeof(size_t)] + *(size_t*)&sig._data.data[64-sizeof(size_t)];
-      }
-
-      size_t operator()(const webauthn::bls_signature& sig) const {
-         return sig.get_hash();
-      }*/
-   };
 
    static bls12_381::g2 sig_parse_base58(const std::string& base58str)
    { try {
 
-
-      const auto pivot = base58str.find('_');
-      auto base_str = base58str.substr(pivot + 1);
-      const auto pivot2 = base_str.find('_');
-      auto data_str = base_str.substr(pivot2 + 1);
-
-      std::vector<char> v1 = fc::from_base58(data_str);
+      std::vector<char> v1 = fc::from_base58(base58str);
 
       FC_ASSERT(v1.size() == 96);
       std::array<uint8_t, 96> v2;
@@ -40,25 +22,6 @@ namespace fc { namespace crypto { namespace blslib {
      :_sig(sig_parse_base58(base58str))
    {}
 
-//   size_t bls_signature::which() const {
-//      //return _storage.index();
-//   }
-
-
-   //template<class... Ts> struct overloaded : Ts... { using Ts::operator()...; };
-   //template<class... Ts> overloaded(Ts...) -> overloaded<Ts...>;
-
-   size_t bls_signature::variable_size() const {
-     /* return std::visit(overloaded {
-         [&](const auto& k1r1) {
-            return static_cast<size_t>(0);
-         },
-         [&](const webauthn::bls_signature& wa) {
-            return static_cast<size_t>(wa.variable_size());
-         }
-      }, _storage);*/
-   }
-
    std::string bls_signature::to_string(const fc::yield_function_t& yield) const
    {
 
@@ -68,7 +31,7 @@ namespace fc { namespace crypto { namespace blslib {
 
       std::string data_str = fc::to_base58(v2, yield);
 
-      return std::string(config::bls_signature_base_prefix) + "_" + std::string(config::bls_signature_prefix) + "_" + data_str;
+      return data_str;
 
    }
 
@@ -81,19 +44,6 @@ namespace fc { namespace crypto { namespace blslib {
       return p1._sig == p2._sig;
    }
 
-/*
-   bool operator != ( const bls_signature& p1, const bls_signature& p2) {
-      return !eq_comparator<bls_signature::storage_type>::apply(p1._storage, p2._storage);
-   }
-
-   bool operator < ( const bls_signature& p1, const bls_signature& p2)
-   {
-      return less_comparator<bls_signature::storage_type>::apply(p1._storage, p2._storage);
-   }
-*/
-   size_t hash_value(const bls_signature& b) {
-     //  return std::visit(hash_visitor(), b._storage);
-   }
 } } }  // fc::crypto::blslib
 
 namespace fc
--- a/libraries/libfc/test/test_bls.cpp
+++ b/libraries/libfc/test/test_bls.cpp
@@ -114,40 +114,6 @@ BOOST_AUTO_TEST_CASE(bls_sig_verif_hotstuff_types) try {
 
 } FC_LOG_AND_RETHROW();
 
-#warning test being worked under https://github.com/AntelopeIO/leap/issues/1531
-/*
-//test a aggregate signature from string
-BOOST_AUTO_TEST_CASE(bls_sig_verif_string_multi) try {
-
-  bls_signature test_sig_single = bls_signature("SIG_BLS_23PuSu1B72cPe6wxGkKjAaaZqA1Ph79zSoW7omsKKUrnprbA3cJCJVhT48QKUG6ofjYTTg4BA4TrVENWyrxjTomwLX6TGdVg2RYhKH7Kk9X23K5ohuhKQcWQ6AwJJGVSbSp4");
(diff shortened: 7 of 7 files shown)


libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -25,38 +25,17 @@ namespace fc::crypto::blslib {
...
-         const auto prefix_str = base58str.substr(0, pivot);
-         FC_ASSERT(prefix == prefix_str, "Private Key has invalid prefix: ${str}", ("str", base58str)("prefix_str", prefix_str));
-
-         auto data_str = base58str.substr(pivot + 1);
-         FC_ASSERT(!data_str.empty(), "Private Key has no data: ${str}", ("str", base58str));
-         return base58_str_parser<bls_private_key::storage_type, config::private_key_prefix>::apply(data_str);
-      }*/
+      std::copy(v1.begin(), v1.end(), v2.begin());
+
+      return v2;
```
heifner, line 38: I think we decided to use WIF format.

libraries/libfc/src/crypto/bls_public_key.cpp
```diff
@@ -50,24 +21,17 @@ namespace fc { namespace crypto { namespace blslib {
...
       std::array<uint8_t, 48> bytes = _pkey.toCompressedBytesBE();
       std::copy(bytes.begin(), bytes.end(), std::back_inserter(v2));
 
       std::string data_str = fc::to_base58(v2, yield);
-
-      //std::string data_str = Util::HexStr(_pkey);
       
-      return std::string(config::bls_public_key_base_prefix) + "_" + data_str;
+      //return std::string(config::bls_public_key_base_prefix) + "_" + data_str;
+      return data_str;
```
heifner, line 33: We decided to add a checksum.

libraries/libfc/test/test_bls.cpp
```diff
@@ -280,5 +246,33 @@ BOOST_AUTO_TEST_CASE(bls_bad_sig_verif) try {
...
+
+  //cout << "pk : " << pk.to_string() << "\n";
+  //cout << "signature : " << signature.to_string() << "\n";
+
+  // Verify the signature
+  bool ok = verify(pk, message_1, signature);
+
+  BOOST_CHECK_EQUAL(ok, true);
+
+} FC_LOG_AND_RETHROW();
```
heifner, line 274: Can you add tests for to/from string for public key and signature as well.

heifner (changes requested):

systemzax:
Added checksum + wif import / export based on fc::private_key

libraries/libfc/test/test_bls.cpp
systemzax: There is a test called "bls_serialization_test" which should provide adequate coverage for these, but let me know if it can be improved.

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -25,38 +25,49 @@ namespace fc::crypto::blslib {
...
-   template<typename Data>
-   Data from_wif( const string& wif_key )
+   string to_wif(vector<uint8_t> secret, const fc::yield_function_t& yield )
    {
-   /*   auto wif_bytes = from_base58(wif_key);
+      FC_ASSERT(secret.size() == 32);
+
+      std::array<uint8_t, 32> v2;
+
+      std::copy(secret.begin(), secret.end(), v2.begin());
```
heifner, line 34: I don't see any reason to make this copy. Should be able to use `secret` vector for the `memcpy`.

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -25,38 +25,49 @@ namespace fc::crypto::blslib {
...
+
+      const size_t size_of_data_to_hash = 32 + 1;
+      const size_t size_of_hash_bytes = 4;
+      char data[size_of_data_to_hash + size_of_hash_bytes];
+      data[0] = (char)0x80; // this is the Bitcoin MainNet code
```
heifner, line 39: This seems ok to me, but curious what others think of this? My understanding is Bitcoin uses `0x80` for mainnet and `0xef` for testnets. This just corresponds to what we are doing for other `private_key` WIF encoding.

@arhag @spoonincode

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -25,38 +25,49 @@ namespace fc::crypto::blslib {
...
       auto key_bytes = vector<char>(wif_bytes.begin() + 1, wif_bytes.end() - 4);
       fc::sha256 check = fc::sha256::hash(wif_bytes.data(), wif_bytes.size() - 4);
       fc::sha256 check2 = fc::sha256::hash(check);
 
       FC_ASSERT(memcmp( (char*)&check, wif_bytes.data() + wif_bytes.size() - 4, 4 ) == 0 ||
                 memcmp( (char*)&check2, wif_bytes.data() + wif_bytes.size() - 4, 4 ) == 0 );
 
-      return Data(fc::variant(key_bytes).as<typename Data::data_type>());*/
+      std::vector<uint8_t> v2(32);
+      std::copy(key_bytes.begin(), key_bytes.end(), v2.begin());
```
heifner, line 61: Nothing prevents `key_bytes` from being larger than `32` which would cause a memory corruption here.

heifner: I think you could avoid this copy by using a `std::vector<uint8_t>` internally and casting to `char*`.

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -25,38 +25,49 @@ namespace fc::crypto::blslib {
...
-         constexpr auto prefix = config::private_key_base_prefix;
-         const auto prefix_str = base58str.substr(0, pivot);
-         FC_ASSERT(prefix == prefix_str, "Private Key has invalid prefix: ${str}", ("str", base58str)("prefix_str", prefix_str));
-
-         auto data_str = base58str.substr(pivot + 1);
-         FC_ASSERT(!data_str.empty(), "Private Key has no data: ${str}", ("str", base58str));
-         return base58_str_parser<bls_private_key::storage_type, config::private_key_prefix>::apply(data_str);
-      }*/
+   {  
+      //cout << base58str << "\n";
```
heifner, line 68: remove commented out code

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -66,12 +77,12 @@ namespace fc::crypto::blslib {
    std::string bls_private_key::to_string(const fc::yield_function_t& yield) const
    {
 
-      /*PrivateKey pk = AugSchemeMPL().KeyGen(_seed);
+      FC_ASSERT(_seed.size() == 32);
+      string wif = to_wif(_seed, yield);
 
-      vector<uint8_t> pkBytes   pk.Serialize()
+      //cout << wif << "\n";
```
heifner, line 83: remove commented out code

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -66,12 +77,12 @@ namespace fc::crypto::blslib {
    std::string bls_private_key::to_string(const fc::yield_function_t& yield) const
    {
 
-      /*PrivateKey pk = AugSchemeMPL().KeyGen(_seed);
+      FC_ASSERT(_seed.size() == 32);
```
heifner, line 80: This assert is not needed, `to_wif` has the same assert.

libraries/libfc/include/fc/crypto/bls_signature.hpp
```diff
@@ -42,7 +42,7 @@ namespace fc { namespace crypto { namespace blslib {
 
 //         size_t which() const;
 
-         size_t variable_size() const;
+         //size_t variable_size() const;
```
heifner, line 45: Remove commented out code.

libraries/libfc/include/fc/crypto/bls_signature.hpp
```diff
@@ -58,7 +58,7 @@ namespace fc { namespace crypto { namespace blslib {
          //friend bool operator == ( const bls_signature& p1, const bls_signature& p2);
          //friend bool operator != ( const bls_signature& p1, const bls_signature& p2);
         //friend bool operator < ( const bls_signature& p1, const bls_signature& p2);
-         friend std::size_t hash_value(const bls_signature& b); //not cryptographic; for containers
+         //friend std::size_t hash_value(const bls_signature& b); //not cryptographic; for containers
```
heifner, line 61: Remove commented out code.

libraries/libfc/src/crypto/bls_public_key.cpp
```diff
@@ -4,39 +4,10 @@
...
    static bls12_381::g1 parse_base58(const std::string& base58str)
    {  
       
-      constexpr auto prefix = config::bls_public_key_base_prefix;
-      const auto pivot = base58str.find('_');
-      const auto prefix_str = base58str.substr(0, pivot);
-      auto data_str = base58str.substr(pivot + 1);
-   
-      std::vector<char> v1 = fc::from_base58(data_str);
+      std::vector<char> v1 = fc::from_base58(base58str);
```
heifner, line 10: No checksum verification. See `public_key` implementation.

libraries/libfc/src/crypto/bls_public_key.cpp
```diff
@@ -50,24 +21,17 @@ namespace fc { namespace crypto { namespace blslib {
...
       std::vector<char> v2;
       std::array<uint8_t, 48> bytes = _pkey.toCompressedBytesBE();
       std::copy(bytes.begin(), bytes.end(), std::back_inserter(v2));
 
       std::string data_str = fc::to_base58(v2, yield);
-
-      //std::string data_str = Util::HexStr(_pkey);
       
-      return std::string(config::bls_public_key_base_prefix) + "_" + data_str;
+      //return std::string(config::bls_public_key_base_prefix) + "_" + data_str;
```
heifner, line 32: Seems like we would want to keep a prefix. @arhag ?

libraries/libfc/src/crypto/bls_signature.cpp
```diff
@@ -68,7 +31,7 @@ namespace fc { namespace crypto { namespace blslib {
 
       std::string data_str = fc::to_base58(v2, yield);
 
-      return std::string(config::bls_signature_base_prefix) + "_" + std::string(config::bls_signature_prefix) + "_" + data_str;
+      return data_str;
```
heifner, line 34: I think we would want to keep a prefix for signatures. @arhag ?

libraries/libfc/src/crypto/bls_private_key.cpp
spoonincode: My first thought is whether these should ever be WIF encoded at all. We don't WIF encode R1 private keys -- they're always `PVT_R1...`. WIF encoding of the K1 private keys is legacy I believe (though `cleos create key` still does it :woman_shrugging:)

So would it make more sense to use a `PVT_BLS..` string only?

arhag: I agree with @spoonincode. We do not need legacy encodings for the BLS key pairs. We should follow the encoding we use for R1 public/private keys: `PUB_BLS_...` and `PVT_BLS_...`. We don't even need an encoding for signatures as far as I know; but if we for some reason did, it would also follow the `SIG_BLS_...` encoding.

libraries/libfc/src/crypto/bls_public_key.cpp
arhag: Yes. See https://github.com/AntelopeIO/leap/pull/1538#discussion_r1301981962.

libraries/libfc/src/crypto/bls_signature.cpp
arhag: I don't even think we really need a string encoding for signatures. But if we did, then yes we would want the prefix. See https://github.com/AntelopeIO/leap/pull/1538#discussion_r1301981962.

systemzax:
> Why do we recompute the secret key every time we sign?
> 
> ```
>    bls_signature bls_private_key::sign( const vector<uint8_t>& message ) const
>    {
>       std::array<uint64_t, 4> sk = bls12_381::secret_key(_seed);
>       bls12_381::g2 sig = bls12_381::sign(sk, message);
>       return bls_signature(sig);
>    }
> ```

I changed the storage type from vector<uint8_t> to array<uint64_t, 4>, which allows us to avoid this recomputation on every signature.

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -25,67 +25,51 @@ namespace fc::crypto::blslib {
...
+      FC_ASSERT(config::bls_private_key_base_prefix == base_prefix_str, "BLS Private Key has invalid base prefix: ${str}", ("str", base58str)("base_prefix_str", base_prefix_str));
+      
+      const auto prefix_str = base58str.substr(pivot + 1, 3); //bls
+      FC_ASSERT(config::bls_private_key_prefix == prefix_str, "BLS Private Key has invalid prefix: ${str}", ("str", base58str)("prefix_str", prefix_str));
+
+      auto data_str = base58str.substr(8);
+
+      std::array<uint64_t, 4> bytes = fc::crypto::blslib::serialize_base58<std::array<uint64_t, 4>>(data_str);
+
+      return bytes;
```
greg7mdp, line 44: This is overly complicated and not quite correct. For example the hardcoded values of 3 and 8 could cause issues. What if the string was just `PVT_`?
I suggest that instead of:
```
      constexpr const char* bls_private_key_base_prefix = "PVT";
      constexpr const char* bls_private_key_prefix = "BLS";
```

you just declare:

```
      constexpr const std::string bls_private_key_prefix = "PVT_BLS_";
```

and you check the prefix with something like:

```
      auto res = std::mismatch(config::bls_private_key_base_prefix.begin(), config::bls_private_key_base_prefix.end(),
                               base58str.begin());
      FC_ASSERT(res.first == config::bls_private_key_base_prefix.end(), ...);
```

greg7mdp (changes requested):

libraries/libfc/include/fc/crypto/bls_common.hpp
```diff
@@ -0,0 +1,43 @@
...
+      wrapper wrapped;
+
+      wrapped.data = data;
+      yield();
+      wrapped.check = wrapper::calculate_checksum(wrapped.data, nullptr);
+      yield();
+      auto packed = raw::pack( wrapped );
+      yield();
+      auto data_str = to_base58( packed.data(), packed.size(), yield );
+      yield();
```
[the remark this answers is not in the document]
systemzax: I followed the same pattern than here : https://github.com/AntelopeIO/leap/blob/main/libraries/libfc/include/fc/crypto/common.hpp#L98

But I agree that, at least for the last yield(), it seems unecessary.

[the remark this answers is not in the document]
spoonincode: My guess is those `yield()`s are there in common.hpp so that string/json to bin doesn't get out of control some how on API calls that deal with ABI serialization? Not sure if this concern applies to BLS types or not since, while I'm a little out of the loop now, I didn't think we were wiring them up as ABI types?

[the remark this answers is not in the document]
greg7mdp: I can imagine that we are worried about an attach using a well crafted string which would cause `to_base58()` to take excessively long time, but it is not clear to me what the other `yield()` protect against.

libraries/libfc/include/fc/crypto/bls_signature.hpp
```diff
@@ -32,59 +30,29 @@ namespace fc { namespace crypto { namespace blslib {
...
-         //return fc::crypto::hash_value(k);
-         return 0;
-      }
-   };
-} // std
-
 FC_REFLECT(bls12_381::fp, (d))
 FC_REFLECT(bls12_381::fp2, (c0)(c1))
 FC_REFLECT(bls12_381::g2, (x)(y)(z))
-FC_REFLECT(fc::crypto::blslib::bls_signature, (_sig) )
```
heifner, line 90: Not sure why you have removed `fc::` usage in this header, but this one is outside of the `namespace` `fc` and should be put back.

libraries/libfc/include/fc/crypto/bls_common.hpp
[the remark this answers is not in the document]
heifner: We are not handling them as ABI types; we decided to pass them as strings. Deserialization of these types I believe will all be in initialization of `producer_plugin` or leap util. Likely best to keep the `yield()` calls incase the use is expanded in the future. If we log them somewhere there is the potential for the logger to pause the application.

libraries/libfc/src/crypto/bls_public_key.cpp
```diff
@@ -76,18 +45,18 @@ namespace fc { namespace crypto { namespace blslib {
       return s;
```
greg7mdp, line 45: Should we use a `yield` function in the `to_string()` call in `std::ostream& operator<<(std::ostream& s, const bls_public_key& k)`?

libraries/libfc/src/crypto/bls_signature.cpp
```diff
@@ -1,37 +1,27 @@
...
-      }
-
-      size_t operator()(const webauthn::bls_signature& sig) const {
-         return sig.get_hash();
-      }*/
-   };
+namespace fc::crypto::blslib {
 
    static bls12_381::g2 sig_parse_base58(const std::string& base58str)
    { try {
```
greg7mdp, line 9: `try` should be on a new line.

libraries/libfc/src/crypto/bls_signature.cpp
```diff
@@ -1,37 +1,27 @@
...
-      std::vector<char> v1 = fc::from_base58(data_str);
+      auto data_str = base58str.substr(8);
 
-      FC_ASSERT(v1.size() == 96);
-      std::array<uint8_t, 96> v2;
-      std::copy(v1.begin(), v1.end(), v2.begin());
-      std::optional<bls12_381::g2> g2 = bls12_381::g2::fromCompressedBytesBE(v2);
+      std::array<uint8_t, 96> bytes = fc::crypto::blslib::serialize_base58<std::array<uint8_t, 96>>(data_str);
+
+      std::optional<bls12_381::g2> g2 = bls12_381::g2::fromCompressedBytesBE(bytes);
```
greg7mdp, line 24: same comment as for private and public key `parse_base58` implementations

greg7mdp (changes requested):

libraries/libfc/src/crypto/bls_public_key.cpp
heifner: I think we should remove `operator<<` for all these types. Going to push up a PR to do that for the non-bls public_key/private_key types now.

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -25,67 +25,51 @@ namespace fc::crypto::blslib {
       return bls_private_key(v);
    }
 
-   template<typename Data>
```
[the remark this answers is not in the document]
systemzax: What kind of test do you have in mind? There is a unit test for this function here : https://github.com/AntelopeIO/leap/blob/hotstuff_integration/libraries/libfc/test/test_bls.cpp#L245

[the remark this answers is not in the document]
greg7mdp: 1. for a couple binary keys `k`, check that `bls_private_key(k.to_string()) == k`
2. for a couple of base58 `s`, encodings, check that: `bls_private_key(s).to_string() == s`

libraries/libfc/include/fc/crypto/bls_private_key.hpp
```diff
@@ -17,20 +16,21 @@ namespace fc::crypto::blslib {
...
-         bls_private_key& operator=( const bls_private_key& ) = default;
-         explicit bls_private_key( std::vector<uint8_t> seed ) {
+         explicit bls_private_key(const std::vector<uint8_t>& seed ) {
             _sk = bls12_381::secret_key(seed);
          }
 
-         explicit bls_private_key(const string& base58str);
+         bls_private_key& operator=( const bls_private_key& ) = default;
+
+         explicit bls_private_key(const std::string& base58str);
```
greg7mdp, line 25: This constructor should be with the other constructors.

libraries/libfc/include/fc/crypto/bls_public_key.hpp
```diff
@@ -24,21 +21,19 @@ namespace fc::crypto::blslib {
...
-
-         bls_public_key( bls12_381::g1 pkey ){
+         bls_public_key( const bls12_381::g1& pkey ){
             _pkey = pkey;
          }
 
-         explicit bls_public_key(const string& base58str);
+         bls_public_key& operator= (const bls_public_key& ) = default;
+
+         explicit bls_public_key(const std::string& base58str);
```
greg7mdp, line 30: This constructor should be with the other ones.

libraries/libfc/include/fc/crypto/bls_signature.hpp
```diff
@@ -24,13 +21,14 @@ namespace fc::crypto::blslib {
...
-         bls_signature( bls12_381::g2 sig ){
+         bls_signature( const bls12_381::g2& sig ){
             _sig = sig;
          }
 
-         explicit bls_signature(const string& base58str);
+         bls_signature& operator= (const bls_signature& ) = default;
+
+
+         explicit bls_signature(const std::string& base58str);
```
greg7mdp, line 31: This constructor should be with the other ones.

libraries/libfc/include/fc/crypto/bls_private_key.hpp
```diff
@@ -6,8 +6,7 @@
 namespace fc::crypto::blslib {
 
    namespace config {
-      constexpr const char* bls_private_key_base_prefix = "PVT";
-      constexpr const char* bls_private_key_prefix = "BLS";
+      constexpr std::string_view bls_private_key_prefix = "PVT_BLS_";
```
greg7mdp, line 9: Why `std::string_view` and not `std::string`? With `std::string` you won't have to construct a new `std::string` every time the `to_string()` function is called.

libraries/libfc/test/test_bls.cpp
```diff
@@ -236,4 +236,22 @@ BOOST_AUTO_TEST_CASE(bls_pub_key_sig_serialization) try {
...
+  bool ok1 = bls_private_key(sk.to_string()) == sk;
+
+  std::string str =  sk.to_string();
+
+  bool ok2 = bls_private_key(str).to_string() == str;
+
+  BOOST_CHECK_EQUAL(ok1, true);
+  BOOST_CHECK_EQUAL(ok2, true);
+
+} FC_LOG_AND_RETHROW();
```
greg7mdp, line 253: Great, thanks Guillaume. We'd also need the other test I suggested (for example this does not check that we use the correct prefix in the text representation).

 We'd need the same for the private_key and signature as well.

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -49,6 +49,10 @@ namespace fc::crypto::blslib {
       return std::string(config::bls_private_key_prefix) + data_str;
    }
 
+   bool operator == ( const bls_private_key& pk1, const bls_private_key& pk2) {
+      return std::memcmp(&pk1, &pk2, sizeof(pk1)) == 0;
```
greg7mdp, line 53: I think the comment I added was important, because ideally we shouldn't do `memcmp`.
Also better to compare the member `_pkey` itself in case there was some padding in the struct.

greg7mdp (changes requested):

libraries/libfc/include/fc/crypto/bls_private_key.hpp
```diff
@@ -6,47 +6,46 @@
...
+         std::string to_string(const yield_function_t& yield = yield_function_t()) const;
 
          bls_public_key     get_public_key() const;
-         bls_signature      sign( const vector<uint8_t>& message ) const;
+         bls_signature      sign( const std::vector<uint8_t>& message ) const;
 
          static bls_private_key generate();
 
-         static bls_private_key regenerate( vector<uint8_t> seed ) {
+         static bls_private_key regenerate( std::vector<uint8_t> seed ) {
```
greg7mdp, line 33: Small detail, but I don't get the name, and I don't see why it is needed since we have the  constructor? Why not call the constructor directly?

libraries/libfc/include/fc/crypto/bls_public_key.hpp
```diff
@@ -26,54 +21,30 @@ namespace fc { namespace crypto { namespace blslib {
          bls_public_key() = default;
          bls_public_key( bls_public_key&& ) = default;
          bls_public_key( const bls_public_key& ) = default;
-         bls_public_key& operator= (const bls_public_key& ) = default;
-
-         bls_public_key( bls12_381::g1 pkey ){
+         bls_public_key( const bls12_381::g1& pkey ){
```
greg7mdp, line 24: Please make this explicit and one line, as in:
`explicit bls_public_key( const bls12_381::g1& pkey ) : _pkey(pkey) {}`

remove unnecessary `friend` declarations. Remove extra spaces.

Class can be simplified to:

```
   struct bls_public_key
   {
         bls_public_key() = default;
         bls_public_key(bls_public_key&&) = default;
         bls_public_key(const bls_public_key&) = default;
         explicit bls_public_key(const bls12_381::g1& pkey) : _pkey(pkey) {}
         explicit bls_public_key(const std::string& base58str);

         bls_public_key& operator= (const bls_public_key&) = default;
         std::string to_string(const yield_function_t& yield = yield_function_t()) const;
         friend bool operator==( const bls_public_key& p1, const bls_public_key& p2);

         bls12_381::g1 _pkey;
   }; 
```

libraries/libfc/include/fc/crypto/bls_signature.hpp
```diff
@@ -8,83 +8,49 @@
...
+namespace fc::crypto::blslib {
 
    namespace config {
-      constexpr const char* bls_signature_base_prefix = "SIG";
-      constexpr const char* bls_signature_prefix = "BLS";
+      constexpr std::string_view bls_signature_prefix = "SIG_BLS_";
    };
-
+   
    class bls_signature
```
greg7mdp, line 17: Move constructors together and make the non-default ones `explicit`. Remove unnecessary friend declarations. Remove extra spaces. class can be simplified to:

```
   struct bls_signature
   {
         bls_signature() = default;
         bls_signature(bls_signature&&) = default;
         bls_signature(const bls_signature&) = default;
         explicit bls_signature(const bls12_381::g2& sig) : _sig(sig) {}
         explicit bls_signature(const std::string& base58str);

         bls_signature& operator=(const bls_signature& ) = default;
         std::string to_string(const yield_function_t& yield = yield_function_t()) const;
         friend bool operator==(const bls_signature& p1, const bls_signature& p2);

         bls12_381::g2 _sig;
   };
```

greg7mdp (changes requested):

libraries/libfc/include/fc/crypto/bls_signature.hpp
greg7mdp: An update, I just merged the PR which updates leap/main to use the latest bls12 submodule (which adds the comparison operators), so if you update this catchup from main and update this submodule, the above should now build.

libraries/libfc/include/fc/crypto/bls_signature.hpp
```diff
@@ -8,83 +8,39 @@
...
-
-namespace fc { namespace crypto { namespace blslib {
-
-   using namespace std;
+namespace fc::crypto::blslib {
 
    namespace config {
-      constexpr const char* bls_signature_base_prefix = "SIG";
-      constexpr const char* bls_signature_prefix = "BLS";
+      constexpr std::string_view bls_signature_prefix = "SIG_BLS_";
```
greg7mdp, line 14:
```suggestion
      constexpr std::string bls_signature_prefix = "SIG_BLS_";
```
better to make it a string (like for the key prefixes) and avoid creating a string every time `to_string` is called.

libraries/libfc/src/crypto/bls_signature.cpp
```diff
@@ -1,110 +1,59 @@
...
       std::array<uint8_t, 96> bytes = _sig.toCompressedBytesBE();
-      std::copy(bytes.begin(), bytes.end(), std::back_inserter(v2));
-
-      std::string data_str = fc::to_base58(v2, yield);
 
-      return std::string(config::bls_signature_base_prefix) + "_" + std::string(config::bls_signature_prefix) + "_" + data_str;
+      std::string data_str = fc::crypto::blslib::deserialize_base58<std::array<uint8_t, 96>>(bytes, yield); 
 
-   }
+      return std::string(config::bls_signature_prefix) + data_str;
```
greg7mdp, line 38:
```suggestion
      return config::bls_signature_prefix + data_str;
```

greg7mdp (changes requested):

greg7mdp (approved):

libraries/libfc/include/fc/crypto/bls_signature.hpp
systemzax: Not sure when this was removed but was since added back.

greg7mdp:
@heifner can we merge this one?

heifner (approved):

greg7mdp:
@systemzax please push the `merge` button!

PR #1540, opened 2023-08-22
Support throttling block syncing to peers.

### LINKED ISSUE #1295: P2P: Ability to limit peer bandwidth consumption

jgiszczak:
Peer to peer listen ports may now apply a block sync throttle.  Each occurrence of `p2p-listen-endpoint` has an independent throttle specification.  Here's the updated help text:

```
--p2p-listen-endpoint arg (=0.0.0.0:9876:0)
                                        The actual host:port[:<rate-cap>] used
                                        to listen for incoming p2p connections.
                                        May be used multiple times.   The
                                        optional rate cap will limit block sync
                                        bandwidth to the specified rate.  A
                                        number alone will be interpreted as
                                        bytes per second.  The number may be
                                        suffixed with units.  Supported units
                                        are:   'B/s', 'KB/s', 'MB/s, 'GB/s',
                                        'TB/s', 'KiB/s', 'MiB/s', 'GiB/s',
                                        'TiB/s'.  Transactions and blocks
                                        outside of sync mode are not throttled.
                                          Examples:
                                            192.168.0.100:9876:1MiB/s
                                            node.eos.io:9876:1512KB/s
                                            node.eos.io:9876:0.5GB/s
                                            [2001:db8:85a3:8d3:1319:8a2e:370:7348]:9876:250KB/s
```
Parsing of the rate cap accepts fractional numbers expressed as decimals.  The throttle rate is in bytes per second.  Per the help text, shorthand suffixes are supported, and transactions and blocks being propagated across the peer to peer network between synchronized nodes are not throttled.  Only blocks transmitted to a syncing node are throttled.  A throttle of `0` or `0B/s` means unthrottled, and is the default.

The limit for specifying a rate cap is 30 characters, including the colon field separator.  This is sufficient characters to specify any value in the 64 bit range of `size_t` with room left for a suffix.  If the rate cap does not parse or the value multiplied by the suffix exceeds 18,446,744,073,709,551,615 an exception will be thrown and the node will not start.

IPv6 addresses are supported for listen endpoints.  They must be in square bracket format: `[<ipv6 address>]:port` optionally followed by `:<rate-cap>`.

> [!NOTE]
> Inbound connections from IPs which are configured `p2p-peer-address`es are exempt from throttle rate caps.  Care should be taken in NAT environments to avoid inadvertently exempting connections due to overlapping subnets.

Throttling is stable even at exceptionally low byterates, but of course on a busy network if the throttle is less than the average block size on the network, clients using that listen port will never catch up to the head block.  Such configurations are allowed but not recommended.

### Suggested network topology
Together with #1411 which allows multiple listen endpoints, an edge node for public peering might be configured as follows:
Leap Throttle Sample Network Diagram (truncated)

Resolves #1295.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1195,6 +1206,27 @@ namespace eosio {
...
    {
+      std::istringstream in(limit_str);
+      fc_dlog( logger, "parsing connection endpoint with locale ${l}", ("l", std::locale("").name()));
+      in.imbue(std::locale(""));
+      double limit{0};
+      in >> limit;
+      if( limit > 0.0f ) {
+         std::string units;
+         in >> units;
+         std::regex units_regex{"([KMGT]?[i]?)B/s"};
```
heifner, line 1217: I think we should avoid using regex. This seems like a rather simple format to parse via just `std::string`. Also it is not clear to me what the locale is doing here. Seems like we should simply support [K,M,G,T]B/s.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1195,6 +1206,27 @@ namespace eosio {
...
+         std::regex_match(units, units_match, units_regex);
+         if( units_match.size() == 2 ) {
+            block_sync_rate_limit = static_cast<size_t>(limit * prefix_multipliers.at(units_match[1].str()));
+            fc_dlog( logger, "setting block_sync_rate_limit to ${limit}", ("limit", block_sync_rate_limit));
+         } else {
+            fc_wlog( logger, "listen address ${la} has invalid block sync limit specification, connection will not be throttled", ("la", listen_address));
+            block_sync_rate_limit = 0;
+         }
+      } else {
+         block_sync_rate_limit = static_cast<size_t>(limit);
```
heifner, line 1228: This does not seem what you would want if limit is < 0.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1195,6 +1206,27 @@ namespace eosio {
...
+      }
```
heifner, line 1229: Seems like this section of code should be in `set_connection_type` or at least in its own method.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1195,6 +1206,27 @@ namespace eosio {
...
+      double limit{0};
+      in >> limit;
+      if( limit > 0.0f ) {
+         std::string units;
+         in >> units;
+         std::regex units_regex{"([KMGT]?[i]?)B/s"};
+         std::smatch units_match;
+         std::regex_match(units, units_match, units_regex);
+         if( units_match.size() == 2 ) {
+            block_sync_rate_limit = static_cast<size_t>(limit * prefix_multipliers.at(units_match[1].str()));
```
heifner, line 1221: I think we need to check for overflow.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3927,16 +3975,24 @@ namespace eosio {
    void net_plugin::set_program_options( options_description& /*cli*/, options_description& cfg )
    {
       cfg.add_options()
-         ( "p2p-listen-endpoint", bpo::value< vector<string> >()->default_value( vector<string>(1, string("0.0.0.0:9876")) ), "The actual host:port used to listen for incoming p2p connections. May be used multiple times.")
+         ( "p2p-listen-endpoint", bpo::value< vector<string> >()->default_value( vector<string>(1, string("0.0.0.0:9876:0")) ), "The actual host:port used to listen for incoming p2p connections. May be used multiple times. "
```
heifner, line 3978: What is `0` for? Seems like without `:0` means no throttling.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3927,16 +3975,24 @@ namespace eosio {
...
+           "Block syncing to all peers connected via the port will be throttled to the specified rate. "
```
heifner, line 3979: Need some examples of valid configurations.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3927,16 +3975,24 @@ namespace eosio {
...
+           "See the 'p2p-peer-address' argument for format details.")
          ( "p2p-server-address", bpo::value< vector<string> >(), "An externally accessible host:port for identifying this node. Defaults to p2p-listen-endpoint. May be used as many times as p2p-listen-endpoint. If provided, the first address will be used in handshakes with other nodes. Otherwise the default is used.")
          ( "p2p-peer-address", bpo::value< vector<string> >()->composing(),
            "The public endpoint of a peer node to connect to. Use multiple p2p-peer-address options as needed to compose a network.\n"
-           "  Syntax: host:port[:<trx>|<blk>]\n"
+           "  Syntax: host:port[:<trx>|<blk>][:<rate-cap>]\n"
            "  The optional 'trx' and 'blk' indicates to node that only transactions 'trx' or blocks 'blk' should be sent."
+           "  The optional rate cap will limit block sync bandwidth to the specified rate.  A number alone will be "
+           "  interpreted as bytes per second.  The number may be suffixed with units.  Supported units are: "
```
heifner, line 3987: Unless I'm looking at this wrong a number alone is ignored.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
```diff
@@ -16,9 +16,10 @@ namespace eosio {
 
    // Longest domain name is 253 characters according to wikipedia.
    // Addresses include ":port" where max port is 65535, which adds 6 chars.
+   // Addresses may also include ":bitrate" with suffix and separators, which adds 30 chars.
```
heifner, line 19: Can you explain why 30 characters.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2719,7 +2767,7 @@ namespace eosio {
    }
 
 
-   void net_plugin_impl::create_session(tcp::socket&& socket, const string listen_address) {
+   void net_plugin_impl::create_session(tcp::socket&& socket, const string listen_address, const string limit) {
```
heifner, line 2770: `const string& limit`

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,120 @@
...
+    throttleNode.cmd[i+1] = throttleNode.cmd[i+1] + ':1000B/s'
+
+    cluster.biosNode.kill(signal.SIGTERM)
+    clusterStart = time.time()
+    cluster.launchUnstarted(2)
+
+    syncNode = cluster.getNode(3)
+    time.sleep(15)
+    throttleNode.waitForBlock(endLargeBlocksHeadBlock)
+    endUnthrottledSync = time.time()
```
heifner, line 108: This is end throttled sync time, right?

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,120 @@
...
+    throttleNode = cluster.unstartedNodes[0]
+    i = throttleNode.cmd.index('--p2p-listen-endpoint')
+    throttleNode.cmd[i+1] = throttleNode.cmd[i+1] + ':1000B/s'
+
+    cluster.biosNode.kill(signal.SIGTERM)
+    clusterStart = time.time()
+    cluster.launchUnstarted(2)
+
+    syncNode = cluster.getNode(3)
+    time.sleep(15)
```
heifner, line 106: No reason to sleep right? The wait will wait as long as needed.

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,120 @@
...
+    throttleNode.waitForBlock(endLargeBlocksHeadBlock)
```
heifner, line 107: Seems like you should wait on unthrottled first.

tests/p2p_sync_throttle_test_shape.json
```diff
@@ -0,0 +1,142 @@
+{
```
heifner, line 1: Why is this shape file needed? Can you not configure what you need without it?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1781,14 +1813,30 @@ namespace eosio {
...
+         int sleep_time_us = 100;
+         const int max_sleep_time_us = 100000;
+         while( true) {
+            auto elapsed = std::chrono::duration_cast<std::chrono::seconds>(get_time() - connection_start_time);
+            auto current_rate = double(block_sync_bytes_sent) / elapsed.count();
+            if( current_rate < block_sync_rate_limit ) {
+               enqueue_buffer( sb, no_reason, to_sync_queue);
+               break;
+            }
+            usleep(sleep_time_us);
```
heifner, line 1833: Instead of sleeping can you just return and wait for the next call to `enqueue_block`.

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
jgiszczak: Negative limits are not supported and make no sense at all.

tests/p2p_sync_throttle_test_shape.json
jgiszczak: The line topology has nodes connecting in ascending order and launcher only supports the highest numbered nodes being unstarted.  Since throttling is only applicable to listen ports, I had to use a custom topology so the highest numbered node is connecting into the second-highest node where the throttle is applied.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
jgiszczak: `:1,000,000,000,000` is 18 characters plus up to 5 characters of suffix, plus padding for future expansion.  I wanted to leave room for someone to configure every significant digit of a terabyte per second.  As implemented right now, throttling is very precise, never overshooting, and seldom undershooting by more than 12 bytes per second on a 1000 byte per second test limit.  A sensible person wouldn't write such a config, but a machine-generated one might.

Despite its location, that length is not part of the protocol.  The throttle bytes are always parsed off of the listen address specification before it is used on the network (if p2p-server-address is unspecified).

tests/p2p_sync_throttle_test.py
jgiszczak: Sadly no, and this is a mistake even I keep making.  Down in the depths, `waitForBlock()` uses `get info` with `exitOnFailure` hardcoded to `True`.  If the node hasn't finished enough of its startup to respond to `get info` it will be 404 and the whole test shuts down.  Some amount of sleep is required for the node to finish enough of its startup to respond to `get info`, before invoking `waitForBlock()` (or any of the `wait...` methods).

It would be nice to create an issue to revise the behaviors of the `wait...` methods to tolerate some number of 404 errors instead of throwing in the towel on the very first one, so we can remove the `sleep()` cruft which is scattered throughout the integration tests.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Locale-aware parsing means you can use thousands separators in a long number and it will work.  It also means regional decimal separators are supported.  `0.5GB/s` in `en-GB.UTF8`, `0,5GB/s` in `de-AT.UTF8`, and `500'000B/s` in `de-CH.UTF8` will all parse as 500000 bytes per second.

plugins/net_plugin/net_plugin.cpp
jgiszczak: I have no idea why I didn't just hard code the limit to 0 in that case, as I do on a failed parse.  Changed.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Moved to its own method.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Switched to `boost::numeric_cast`.

plugins/net_plugin/net_plugin.cpp
jgiszczak: I put them all on the wrong option.  Moved from `p2p-peer-address`.

plugins/net_plugin/net_plugin.cpp
jgiszczak: It is redundant but it is a reminder of what's possible when reading the help output.  It's part of the documentation.

plugins/net_plugin/net_plugin.cpp
jgiszczak: It wasn't meant to be.  Unspecified units was meant to allow `:0` as shorthand for explicitly unthrottled.  I got paranoid about `match[1]` but as it turns out if the match count is 1, that index returns an empty string, which then corresponds to the first entry in the `prefix_multipliers` table which multiplies by 1 if units are empty.  Removed check of `count()`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4192,10 +4248,18 @@ namespace eosio {
...
+                  auto last_colon_location = address.rfind(':');
+                  listen_addr = std::string(address, 0, last_colon_location);
+                  limit = std::string(address, last_colon_location+1);
+               }
+
                fc::create_listener<tcp>(
-                     my->thread_pool.get_executor(), logger, accept_timeout, address, extra_listening_log_info,
-                     [my = my, addr = p2p_addr](tcp::socket&& socket) { my->create_session(std::move(socket), addr); });
+                     my->thread_pool.get_executor(), logger, accept_timeout, listen_addr, extra_listening_log_info,
+                     [my = my, addr = p2p_addr, limit = limit](tcp::socket&& socket) { fc_dlog( logger, "start listening on ${addr} with peer sync throttle ${limit}", ("addr", addr)("limit", limit)); my->create_session(std::move(socket), addr, limit); });
```
heifner, line 4262: Since you are only limiting incoming connections, you need to make sure any incoming duplicate connections are also unlimited.

tests/p2p_sync_throttle_test.py
jgiszczak: Fixed confusion in variable names.  Also fixed test, which was timing out in `waitForBlock()` resulting in a spurious success.

tests/p2p_sync_throttle_test.py
jgiszczak: Fixed confusion in variable names.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Added.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Simply returning doesn't work.  Blocks are lost and the receiving node starts getting nothing but unlinkable blocks, after a gap of almost 50 blocks.  At this level in the call stack, it is assumed the method never fails.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Won't incoming duplicates be closed?

heifner: One side is picked to close. So it can be a configured connection but seen as incoming.

plugins/net_plugin/net_plugin.cpp
heifner: This code needs to be moved to `enqueue_sync_block` and then don't update `peer_requested->last` and don't pull the block out of the controller when you throttle.

plugins/net_plugin/net_plugin.cpp
heifner: Can you add these examples to the PR description.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
heifner: Please add this to the comment.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1226,6 +1238,30 @@ namespace eosio {
...
+         std::regex units_regex{"([KMGT]?[i]?)B/s"};
+         std::smatch units_match;
+         std::regex_match(units, units_match, units_regex);
+         try {
+            block_sync_rate_limit = boost::numeric_cast<size_t>(limit * prefix_multipliers.at(units_match[1].str()));
+            fc_dlog( logger, "setting block_sync_rate_limit to ${limit}", ("limit", block_sync_rate_limit));
+         } catch (boost::numeric::bad_numeric_cast&) {
+            fc_wlog( logger, "listen address ${la} block sync limit specification overflowed, connection will not be throttled", ("la", listen_address));
+            block_sync_rate_limit = 0;
+         }
```
heifner, line 1259: Instead of doing this parsing for every connection just do it once in `plugin_initialize` and `EOS_ASSERT` if there is an invalid configuration.

tests/p2p_sync_throttle_test.py
heifner: Please create a GitHub issue for this.

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,120 @@
...
+
+    Print("Configure and launch txn generators")
+    targetTpsPerGenerator = 100
+    testTrxGenDurationSec=60
+    trxGeneratorCnt=1
+    cluster.launchTrxGenerators(contractOwnerAcctName=cluster.eosioAccount.name, acctNamesList=[accounts[0].name,accounts[1].name],
+                                acctPrivKeysList=[account1PrivKey,account2PrivKey], nodeId=prodNode.nodeId, tpsPerGenerator=targetTpsPerGenerator,
+                                numGenerators=trxGeneratorCnt, durationSec=testTrxGenDurationSec, waitToComplete=True)
+
+    endLargeBlocksHeadBlock = nonProdNode.getHeadBlockNum()
```
heifner, line 95: Seems like you should wait a bit before grabbing a block to sync to. The trx generators may not have even started producing trxs yet.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Added.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
jgiszczak: Added.

tests/p2p_sync_throttle_test.py
jgiszczak: Created #1562.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Moved.

(To where it was intended to go in the first place, or why change the function signature... :expressionless:)

plugins/net_plugin/net_plugin.cpp
jgiszczak: Moved.

tests/p2p_sync_throttle_test.py
jgiszczak: `waitToComplete` is set to `True` in the call to launch the transaction generators.  Internally that performs a `wait()` on all generator processes.

heifner: All, missed that.

plugins/net_plugin/include/eosio/net_plugin/protocol.hpp
heifner: I meant in the code.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4039,6 +4015,29 @@ namespace eosio {
...
+         std::regex units_regex{"([KMGT]?[i]?)B/s"};
+         std::smatch units_match;
+         std::regex_match(units, units_match, units_regex);
+         try {
+            block_sync_rate_limit = boost::numeric_cast<size_t>(limit * prefix_multipliers.at(units_match[1].str()));
+            fc_dlog( logger, "setting block_sync_rate_limit to ${limit}", ("limit", block_sync_rate_limit));
+         } catch (boost::numeric::bad_numeric_cast&) {
+            EOS_ASSERT(false, plugin_config_exception, "block sync limit specification overflowed: ${limit}", ("limit", limit_str));
+         }
+      }
```
heifner, line 4037: Should `EOS_ASSERT` if `limit < 0`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4039,6 +4015,29 @@ namespace eosio {
       return fc::json::from_string(s).as<T>();
    }
 
+   size_t net_plugin_impl::parse_connection_limit( const std::string& limit_str) {
```
heifner, line 4018: Maybe rename to `parse_connection_rate_limit`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -475,6 +476,7 @@ namespace eosio {
       std::function<void()> increment_dropped_trxs;
       
    private:
+      static const std::map<std::string, int> prefix_multipliers;
```
heifner, line 479: This can be:
```
      inline static const std::map<std::string, int> prefix_multipliers{
              {"",1},{"K",pow(10,3)},{"M",pow(10,6)},{"G",pow(10, 9)},{"T",pow(10, 12)},
              {"Ki",pow(2,10)},{"Mi",pow(2,20)},{"Gi",pow(2,30)},{"Ti",pow(2,40)}};
```

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3983,6 +4015,29 @@ namespace eosio {
       return fc::json::from_string(s).as<T>();
    }
 
+   size_t net_plugin_impl::parse_connection_limit( const std::string& limit_str) {
+      std::istringstream in(limit_str);
+      fc_dlog( logger, "parsing connection endpoint with locale ${l}", ("l", std::locale("").name()));
```
heifner, line 4020: Also include the `limit_str` in the debug log output.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3983,6 +4015,29 @@ namespace eosio {
...
+      size_t block_sync_rate_limit = 0;
+      if( limit > 0.0f ) {
+         std::string units;
+         in >> units;
+         std::regex units_regex{"([KMGT]?[i]?)B/s"};
+         std::smatch units_match;
+         std::regex_match(units, units_match, units_regex);
+         try {
+            block_sync_rate_limit = boost::numeric_cast<size_t>(limit * prefix_multipliers.at(units_match[1].str()));
+            fc_dlog( logger, "setting block_sync_rate_limit to ${limit}", ("limit", block_sync_rate_limit));
```
heifner, line 4033: Indicate that this is in bytes.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3983,6 +4015,29 @@ namespace eosio {
...
+      in >> limit;
+      size_t block_sync_rate_limit = 0;
+      if( limit > 0.0f ) {
+         std::string units;
+         in >> units;
+         std::regex units_regex{"([KMGT]?[i]?)B/s"};
+         std::smatch units_match;
+         std::regex_match(units, units_match, units_regex);
+         try {
+            block_sync_rate_limit = boost::numeric_cast<size_t>(limit * prefix_multipliers.at(units_match[1].str()));
```
heifner, line 4032: Test cases that fail this:
` --p2p-listen-endpoint 0.0.0.0:9876:8966Tbs` => rate limit of 8966 bytes
`--p2p-listen-endpoint 0.0.0.0:9876:8966Tb/s` => rate limit of 8966 bytes 
`--p2p-listen-endpoint 0.0.0.0:9876:123abc` => rate limit of 123 bytes.
We should generate an error and refuse to start if not a valid type specified.

`--p2p-listen-endpoint 0.0.0.0:9876:8966TB/s` => 
`error 2023-08-26T13:38:40.503 nodeos    net_plugin.cpp:4269           operator()           ] net_plugin::plugin_startup failed to listen on 0.0.0.0:9876:8966TB/s, Incorrect plugin configuration`
Even `1TB/s` and `1TiB/s` do not work.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4192,10 +4247,20 @@ namespace eosio {
 
                std::string extra_listening_log_info =
                      ", max clients is " + std::to_string(my->connections.get_max_client_count());
-                     
+
+               auto listen_addr = address;
+               auto limit = string("0");
+               if( std::count(address.begin(), address.end(), ':') > 1 ) {
```
heifner, line 4253: This doesn't work for IPv6, right? As IPv6 addresses are of the form: `2001:db8:3333:4444:5555:6666:7777:8888`

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3983,6 +4015,29 @@ namespace eosio {
...
+         std::string units;
+         in >> units;
+         std::regex units_regex{"([KMGT]?[i]?)B/s"};
+         std::smatch units_match;
+         std::regex_match(units, units_match, units_regex);
+         try {
+            block_sync_rate_limit = boost::numeric_cast<size_t>(limit * prefix_multipliers.at(units_match[1].str()));
+            fc_dlog( logger, "setting block_sync_rate_limit to ${limit}", ("limit", block_sync_rate_limit));
+         } catch (boost::numeric::bad_numeric_cast&) {
+            EOS_ASSERT(false, plugin_config_exception, "block sync limit specification overflowed: ${limit}", ("limit", limit_str));
```
dimas1185, line 4035: use `EOS_THROW` instead

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,120 @@
...
+
+    throttlingNode = cluster.unstartedNodes[0]
+    i = throttlingNode.cmd.index('--p2p-listen-endpoint')
+    throttlingNode.cmd[i+1] = throttlingNode.cmd[i+1] + ':40000B/s'
```
dimas1185, line 99: Please comment why are you using 40000 bytes here

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,120 @@
...
+
+    throttledNode = cluster.getNode(3)
+    time.sleep(15)
+    assert throttlingNode.waitForBlock(endLargeBlocksHeadBlock), f'wait for block {endLargeBlocksHeadBlock}  on throttled node timed out'
+    endThrottlingSync = time.time()
+    assert throttledNode.waitForBlock(endLargeBlocksHeadBlock, timeout=90), f'Wait for block {endLargeBlocksHeadBlock} on sync node timed out'
+    endThrottledSync = time.time()
+    Print(f'Unthrottled sync time: {endThrottlingSync - clusterStart} seconds')
+    Print(f'Throttled sync time: {endThrottledSync - clusterStart} seconds')
+    assert endThrottledSync - clusterStart > endThrottlingSync - clusterStart + 15, 'Throttled sync time must be at least 15 seconds greater than unthrottled'
```
dimas1185, line 113: please add comment with explanation of how did you calculate those 15 seconds

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,120 @@
...
+
+    cluster.biosNode.kill(signal.SIGTERM)
+    clusterStart = time.time()
+    cluster.launchUnstarted(2)
+
+    throttledNode = cluster.getNode(3)
+    time.sleep(15)
+    assert throttlingNode.waitForBlock(endLargeBlocksHeadBlock), f'wait for block {endLargeBlocksHeadBlock}  on throttled node timed out'
+    endThrottlingSync = time.time()
+    assert throttledNode.waitForBlock(endLargeBlocksHeadBlock, timeout=90), f'Wait for block {endLargeBlocksHeadBlock} on sync node timed out'
```
dimas1185, line 109: please add some general comment about what throttling and throttled nodes are and why later should get the block with delay

plugins/net_plugin/net_plugin.cpp
jgiszczak: Now enforcing the full throttle format.

Added extra listen port to throttle test with 1TB/s limit to verify a non-empty prefix works and net_plugin successfully binds to the address.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Added.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Added.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Added.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Renamed.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Moved, and type changed to prevent overflows.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Now requiring IPV6 square bracket format. (The trailing one, anyway.)

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4015,12 +4014,13 @@ namespace eosio {
...
 
-   size_t net_plugin_impl::parse_connection_limit( const std::string& limit_str) {
+   size_t net_plugin_impl::parse_connection_rate_limit( const std::string& limit_str) {
       std::istringstream in(limit_str);
-      fc_dlog( logger, "parsing connection endpoint with locale ${l}", ("l", std::locale("").name()));
+      fc_dlog( logger, "parsing connection endpoint limit ${limit} with locale ${l}", ("limit", limit_str)("l", std::locale("").name()));
       in.imbue(std::locale(""));
       double limit{0};
       in >> limit;
+      EOS_ASSERT(limit >= 0, plugin_config_exception, "block sync rate limit must be positive: ${limit}", ("limit", limit_str));
```
heifner, line 4023: I think this should be `limit >= 0.0`. Also for below, drop the `f` as that is `float`. `double` is undecorated.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Added throttle exemption for configured p2p-peer-addresses.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Added IPv6 example.  Sample address is taken from the IPv6 address Wikipedia page.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Converted to `EOS_THROW`.

tests/p2p_sync_throttle_test.py
jgiszczak: Added comment.

tests/p2p_sync_throttle_test.py
jgiszczak: Added comments for each node type and an English description of the custom topology.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Updated.

tests/p2p_sync_throttle_test.py
dimas1185: I was expecting something like (avg_block_size * block_amount) / throttle_size_per_sec.

tests/p2p_sync_throttle_test.py
dimas1185: so if avg block size is 10Kb and throttling size is 40Kb/s how do slow down transmission if you just need 20Kb/s for full speed?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1143,6 +1146,11 @@ namespace eosio {
    }
 
 
+   template<typename Function>
+   void connections_manager::for_each_supplied_peer( Function&& f ) const {
+      std::for_each(supplied_peers.begin(), supplied_peers.end(), std::forward<Function>(f));
```
heifner, line 1151: A `shared_lock` on `connections_mtx` is required here.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2768,6 +2776,16 @@ namespace eosio {
                visitors < connections.get_max_client_count())) {
             fc_ilog(logger, "Accepted new connection: " + paddr_str);
 
+            connections.for_each_supplied_peer([&listen_address, &paddr_str, &limit](const string& peer_addr) {
+               auto [host, port, type] = split_host_port_type(peer_addr);
```
heifner, line 2780: The list of `host` strings for comparison with `paddr_str` should be calculated only when `supplied_peers` is modified. Then that new list of host strings can be used here instead of doing this work on every new session. Also you can break out of the for each when a match is found.

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3951,14 +3969,17 @@ namespace eosio {
    {
       cfg.add_options()
          ( "p2p-listen-endpoint", bpo::value< vector<string> >()->default_value( vector<string>(1, string("0.0.0.0:9876:0")) ), "The actual host:port[:<rate-cap>] used to listen for incoming p2p connections. May be used multiple times. "
-           "  The optional rate cap will limit block sync bandwidth to the specified rate.  A number alone will be "
-           "  interpreted as bytes per second.  The number may be suffixed with units.  Supported units are: "
+           "  The optional rate cap will limit per connection block sync bandwidth to the specified rate.  Total "
+           "  allowed bandwidth is the rate-cap multiplied by the connection count limit.  A number alone will be "
+           "  interpreted as bytes per second.  The number is parsed locale-aware and may include thousands and "
```
heifner, line 3974: A number alone generates an error.

tests/p2p_sync_throttle_test.py
jgiszczak: I don't want to give the illusion of precision where there is none.  The time to synchronize varies significantly depending on the machine running the test.  I tried to choose values which are reasonable for a wide range of machines in order to make the test as reliable as possible, but that necessarily involves some guesswork.  Unfortunately the actual block sizes are not available to the test framework so I can't implement my preferred solution and just calculate values.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Fixed.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Added.

plugins/net_plugin/net_plugin.cpp
jgiszczak: At the moment `supplied_peers` is our source of truth for the list of configured peers.  I didn't want to create a parallel data structure for that with peer discovery likely to make it obsolete.

`std::for_each` has no mechanism to break out.  The return value of the UnaryFunction is ignored.

greg7mdp: > `std::for_each` has no mechanism to break out. The return value of the UnaryFunction is ignored

Can you use `std::any_of`?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2683,58 +2718,27 @@ namespace eosio {
...
       set_state(connection_state::connecting);
       pending_message_buffer.reset();
       buffer_queue.clear_out_queue();
       boost::asio::async_connect( *socket, endpoints,
          boost::asio::bind_executor( strand,
-               [resolver, c = shared_from_this(), socket=socket]( const boost::system::error_code& err, const tcp::endpoint& endpoint ) {
+               [c = shared_from_this(), socket=socket, &connections, conn_details]( const boost::system::error_code& err, const tcp::endpoint& endpoint ) {
             if( !err && socket->is_open() && socket == c->socket ) {
-               c->update_endpoints();
+               auto& index = connections.get<by_active_ip>();
```
heifner, line 2737: Need to lock the connections mutex.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -333,9 +336,36 @@ namespace eosio {
...
+               tag<struct by_active_ip>,
+               key<&connection_details::active_ip>
+            >
+         >
+      >;
+   private:
       alignas(hardware_destructive_interference_size)
       mutable std::shared_mutex        connections_mtx;
-      chain::flat_set<connection_ptr>  connections;
+      connection_details_index         connections GUARDED_BY(connections_mtx);
```
heifner, line 368: The GUARDED_BY does not work with `std::shared_lock`. Remove this.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3983,6 +4039,33 @@ namespace eosio {
...
+         in >> units;
+         std::regex units_regex{"([KMGT]?[i]?)B/s"};
+         std::smatch units_match;
+         std::regex_match(units, units_match, units_regex);
+         try {
+            if( units.length() > 0 ) {
+               EOS_ASSERT(units_match.size() == 2, plugin_config_exception, "invalid block sync rate limit specification: ${limit}", ("limit", units));
+            }
+            block_sync_rate_limit = boost::numeric_cast<size_t>(limit * prefix_multipliers.at(units_match[1].str()));
+            fc_dlog( logger, "setting block_sync_rate_limit to ${limit} bytes per second", ("limit", block_sync_rate_limit));
```
heifner, line 4061: Make this a `fc_ilog`.

plugins/net_plugin/net_plugin.cpp
heifner: This still produces odd error messages:
`net_plugin::plugin_startup failed to listen on 0.0.0.0:9876:123gb/s, Incorrect plugin configuration`
It should report that `123gb/s` is not a valid rate limit option.

I think we need a unittest that tests all these various settings.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4320,47 +4422,87 @@ namespace eosio {
...
+      auto endpoint = c->socket->remote_endpoint(ec);
+      connections.insert( connection_details{
+         .host = c->peer_address(), 
+         .c = std::move(c),
+         .active_ip = endpoint} );
    }
 
    // called by API
    string connections_manager::connect( const string& host, const string& p2p_address ) {
+      supplied_peers.insert(host);
```
heifner, line 4442: `connect` can be called from multiple threads from the http thread pool at the same time. This needs to be protected by a mutex.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4320,47 +4422,87 @@ namespace eosio {
...
+            c->set_heartbeat_timeout( heartbeat_timeout );
+            vector<tcp::endpoint> eps{results.begin(), results.end()};
+            std::lock_guard g( connections_mtx );
+            auto [it, inserted] = connections.insert( connection_details{
+               .host = peer_address,
+               .c = std::move(c),
+               .ips = std::move(eps)
+            });
+            if( !err ) {
+               it->c->connect( results, connections, it );
```
heifner, line 4473: This is being called from the http thread pool via `connect` above.

tests/p2p_sync_throttle_test.py
jgiszczak: I don't understand the question.  This is an integration test, not a user scenario.  If the test rate needs to be changed from 40k to 20k for some reason, we simply change it.  It's a test.

dimas1185: question is based on your comment. numbers are from your comment. you chose the numbers so I asked you to explain those.

tests/p2p_sync_throttle_test.py
dimas1185: well, I understand you point but that makes this not a real test but guessed numbers that make this script pass. I understand that you can't do this precisely but if you can think of any approximate formula it would be nice. e.g. you can roughly check blocks.log size and divide this by number of blocks or you can take it from network stats that node supposed to print, to include all messages and serialized data sizes.
Otherwise, imagine this test fails. How do I know if that is error or just block size has changed or network issues? To solve this I can play with numbers to make it pass but I still have no idea if I'm not hiding some bug that way.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Removed.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Yes.  Switched to `std::any_of`.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -182,7 +182,7 @@ class bp_connection_manager {
 
                fc_dlog(self()->get_logger(), "pending_downstream_neighbors: ${pending_downstream_neighbors}",
                        ("pending_downstream_neighbors", to_string(pending_downstream_neighbors)));
-               for (auto neighbor : pending_downstream_neighbors) { self()->connections.connect(config.bp_peer_addresses[neighbor], *self()->p2p_addresses.begin() ); }
+               for (auto neighbor : pending_downstream_neighbors) { self()->connections.resolve_and_connect(config.bp_peer_addresses[neighbor], *self()->p2p_addresses.begin() ); }
```
heifner, line 185: `p2p_addresses` can be empty. Need to check for that here.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4333,47 +4443,105 @@ namespace eosio {
...
+      boost::system::error_code ec;
+      auto endpoint = c->socket->remote_endpoint(ec);
+      connections.insert( connection_details{
+         .host = c->peer_address(), 
+         .c = std::move(c),
+         .active_ip = endpoint} );
    }
 
    // called by API
    string connections_manager::connect( const string& host, const string& p2p_address ) {
```
heifner, line 4462:
```
   string net_plugin::connect( const string& host ) {
      return my->connections.connect( host, *my->p2p_addresses.begin() );
   }
```

`my->p2p_addresses` can be empty. For example:
```
./nodeos --data-dir d --config-dir d -e -p eosio --p2p-listen-endpoint "" --plugin eosio::net_api_plugin
curl http://127.0.0.1:8888/v1/net/connect -d '"127.0.0.1:9876"'
```
Will segfault.

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4333,47 +4443,105 @@ namespace eosio {
    }
 
    void connections_manager::connect_supplied_peers(const string& p2p_address) {
-      std::lock_guard g(connections_mtx);
       for (const auto& peer : supplied_peers) {
```
heifner, line 4446: Missing lock of `connections_mtx`

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2661,56 +2727,33 @@ namespace eosio {
...
+         auto it = my_impl->connections.get_connection_iterator(c);
+         c->connect(it->ips, it);
+      });
       return true;
    }
 
    // called from connection strand
-   void connection::connect( const std::shared_ptr<tcp::resolver>& resolver, const tcp::resolver::results_type& endpoints ) {
+   void connection::connect( const tcp::resolver::results_type& endpoints,
+                             connections_manager::connection_details_index::const_iterator conn_details ) {
```
heifner, line 2747: `conn_details` is an iterator into a collection that requires a lock since it can be accessed from multiple threads. This iterator is being passed off to a connection strand to be accessed via the lambda. That lambda is called without the lock. Even if you acquired the lock the iterator is not necessarily any good. Once the lock is released the iterator should be considered invalid.

Best you could do is re-find the connection via a find with a locked mutex.

plugins/net_plugin/net_plugin.cpp
heifner: This still needs fixed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -339,9 +342,36 @@ namespace eosio {
...
+            ordered_non_unique<
+               tag<struct by_host>,
+               key<&connection_details::host>
+            >,
+            ordered_unique<
+               tag<struct by_connection>,
+               key<&connection_details::c>
+            >,
+            ordered_non_unique<
+               tag<struct by_active_ip>,
```
heifner, line 366: It doesn't look like this index is actually used. The single use could just use `by_host`.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4333,47 +4443,105 @@ namespace eosio {
...
+      index.modify_key(connections.project<by_active_ip>(it), [endpoint](tcp::endpoint& e) {
+         e = endpoint;
+      });
+   }
+
+   connections_manager::connection_details_index::iterator connections_manager::get_connection_iterator(const connection_ptr& c) {
+      std::lock_guard g( connections_mtx );
+      const auto& index = connections.get<by_connection>();
+      auto i = index.find(c);
+      return connections.project<by_host>(i);
```
heifner, line 4521: The lock is released here. The returned iterator can't be trusted after the lock is released.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4333,47 +4443,105 @@ namespace eosio {
...
+      string::size_type colon = peer_address.find(':');
+      if (colon == std::string::npos || colon == 0) {
+         fc_elog( logger, "Invalid peer address. must be \"host:port[:<blk>|<trx>]\": ${p}", ("p", peer_address) );
+         return "invalid peer address";
+      }
+
       std::lock_guard g( connections_mtx );
-      if( find_connection_i( host ) )
+      if( find_connection_i( peer_address ) )
          return "already connected";
```
heifner, line 4479: The `connections_mtx` could be released here.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4333,47 +4443,105 @@ namespace eosio {
...
+         connections.erase(i);
          supplied_peers.erase(host);
          return "connection removed";
       }
       return "no known connection for host";
    }
 
    void connections_manager::close_all() {
-      fc_ilog( logger, "close all ${s} connections", ("s", connections.size()) );
+      auto& index = connections.get<by_host>();
```
heifner, line 4539: I think it would be better to grab the index after acquiring the lock.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4453,63 +4606,71 @@ namespace eosio {
...
             per_connection.peers.push_back(metrics);
          }
 
-         if (!(*it)->socket_is_open() && (*it)->state() != connection::connection_state::connecting) {
-            if (!(*it)->incoming()) {
-               if (!(*it)->resolve_and_connect()) {
-                  it = connections.erase(it);
+         if (!c->socket_is_open() && c->state() != connection::connection_state::connecting) {
+            if (!c->incoming()) {
+               g.unlock();
```
heifner, line 4659: Once the lock is unlocked the iterator can no longer be considered valid.

heifner (changes requested):

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
```diff
@@ -157,7 +157,7 @@ class bp_connection_manager {
    // Only called from connection strand
    // This should only be called after the first handshake message is received to check if an incoming connection
    // has exceeded the pre-configured max_client_count limit.
-   bool exceeding_connection_limit(Connection* new_connection) const {
+   bool exceeding_connection_limit(std::shared_ptr<Connection> new_connection) const {
```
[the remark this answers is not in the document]
jgiszczak: Template type deduction fails unless this type more closely aligns with the type available from the connections multi_index.

tests/p2p_sync_throttle_test_shape.json
```diff
@@ -0,0 +1,142 @@
...
+      ], 
+      "peers": [
+        "testnet_00"
+      ], 
+      "producers": [
+        "eosio"
+      ], 
+      "dont_start": false, 
+      "config_dir_name": "/home/giszczakj/Dev/leap/build/TestLogs/p2p_sync_throttle_test1192292/node_bios", 
+      "data_dir_name": "/home/giszczakj/Dev/leap/build/TestLogs/p2p_sync_throttle_test1192292/node_bios",
```
spoonincode, line 21: are these paths just simply ignored when running the test?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -339,9 +342,36 @@ namespace eosio {
    constexpr uint32_t packed_transaction_which = fc::get_index<net_message, packed_transaction>(); // see protocol net_message
 
    class connections_manager {
+   public:
+      struct connection_details {
+         std::string host;
+         connection_ptr c;
+         tcp::endpoint active_ip;
+         tcp::resolver::results_type ips;
+         operator const connection_ptr&() const { return c; }
```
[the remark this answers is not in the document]
jgiszczak: It's required.  The `std` algorithm templates don't compile without it.

plugins/net_plugin/include/eosio/net_plugin/auto_bp_peering.hpp
jgiszczak: Revised to tolerate node configured with no listen endpoints, here and elsewhere.

plugins/net_plugin/net_plugin.cpp
jgiszczak: As above.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Restored use of mutex.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Reworked how `connection::reconnect()` works to delegate back to `connections_manager`, eliminating this break in encapsulation entirely.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Removed.  I had intended to use this index as part of a more sophisticated response to the `/net/disconnect` API, allowing not just perfect string matches for disconnects but also by IP and port, since that's commonly what's available from, for instance, system firewall state tables.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1138,16 +1183,24 @@ namespace eosio {
    }
 
 
+   template<typename Function>
+   bool connections_manager::for_any_supplied_peer( Function&& f ) const {
```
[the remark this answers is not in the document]
jgiszczak: Renamed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1667,14 +1722,25 @@ namespace eosio {
...
+            if( current_rate < block_sync_rate_limit ) {
+               block_sync_bytes_sent += enqueue_block( sb, true );
+               ++peer_requested->last;
+            } else {
+               return false;
+            }
+         } else {
+            block_sync_bytes_sent += enqueue_block( sb, true );
+            ++peer_requested->last;
+         }
```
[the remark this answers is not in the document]
jgiszczak: Incorporated suggestion.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2661,56 +2727,33 @@ namespace eosio {
...
-         return false;
-      }
-
-      connection_ptr c = shared_from_this();
-
       if( consecutive_immediate_connection_close > def_max_consecutive_immediate_connection_close || no_retry == benign_other ) {
          fc::microseconds connector_period = my_impl->connections.get_connector_period();
          fc::lock_guard g( conn_mtx );
          if( last_close == fc::time_point() || last_close > fc::time_point::now() - connector_period ) {
-            return true; // true so doesn't remove from valid connections
```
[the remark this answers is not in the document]
jgiszczak: The whole method didn't exist for a while.  Its restoration missed including this comment.  Comment now restored.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Call chain reworked to not pass an iterator.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Revised.

tests/p2p_sync_throttle_test_shape.json
jgiszczak: Indeed they are, and no value needs to be provided.  Removed.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3991,6 +4071,54 @@ namespace eosio {
...
+         }
+      }
+      auto block_sync_rate_limit = parse_connection_rate_limit(limit);
+
+      return {listen_addr, block_sync_rate_limit};
+   }
+
+   size_t net_plugin_impl::parse_connection_rate_limit( const std::string& limit_str) const {
+      std::istringstream in(limit_str);
+      fc_dlog( logger, "parsing connection endpoint limit ${limit} with locale ${l}", ("limit", limit_str)("l", std::locale("").name()));
```
spoonincode, line 4097: Wouldn't parsing the config in a locale dependent way mean config files aren't transportable across different systems (when they have a different locale)? I can imagine some confusion from users when copying a sample config file and it errors out on their system.

jgiszczak: Removed locale-awareness.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Reworked to never hold an iterator without the lock.

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
jgiszczak: Mooted by explicit retrieval of the connection instance from `conn_detail` in the relevant `std` algorithm usages.

plugins/net_plugin/net_plugin.cpp
heifner: This still reports the generic error:
`net_plugin::plugin_startup failed to listen on 0.0.0.0:9876:123gb/s, Incorrect plugin configuration` instead of something like `123gb/s` is not a valid rate limit.

jgiszczak: Fixed in https://github.com/AntelopeIO/leap/pull/1540/commits/6db4ad8aa535d0d4aef970ed4b74ef485be537d5.

heifner (approved):

heifner:
Replaced by #1741 & #1742

PR #1549, opened 2023-08-23
Use a single WASM interface per nodeos (instead of per thread) to save WASM compile time and cache

linh2931:
A part of https://github.com/eosnetworkfoundation/product/issues/149.

Currently when read-only threads are enabled, each thread maintains its own WASM interface which in turn keeps its own instantiated module cache. This leads to 
* the sample contract is compiled multiple times by the threads,
* the same instantiated module is cached multiple times,
* extra abstraction layer: `wasm_interface_collection` had to be introduced to hide the complicated logic to manage WASM interface on the main thread and the read-only threads.

This PR changes to use a single WASM interface per nodeos, resolving all the three issues mentioned above.


libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -66,16 +92,30 @@ namespace eosio { namespace chain {
 #endif
          if(!runtime_interface)
             EOS_THROW(wasm_exception, "${r} wasm runtime not supported on this platform and/or configuration", ("r", vm));
+
+#ifdef EOSIO_EOS_VM_OC_RUNTIME_ENABLED
+         if(eosvmoc_tierup != wasm_interface::vm_oc_enable::oc_none) {
+            EOS_ASSERT(vm != wasm_interface::vm_type::eos_vm_oc, wasm_exception, "You can't use EOS VM OC as the base runtime when tier up is    activated");
```
heifner, line 98: Spacing  `"is    activated"`.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -84,48 +124,90 @@ namespace eosio { namespace chain {
...
-                                                      .vm_version = vm_version
-                                                   } ).first;
+         wasm_cache_index::iterator it;
+         if (trx_context.control.is_write_window()) {
+            // When in write window (either read only threads are not enabled or
+            // they are not schedued to run), only main thread is processing
+            // transactions. No need to lock.
+            it = wasm_instantiation_cache.find( boost::make_tuple(code_hash, vm_type, vm_version) );
+         } else {
+            std::shared_lock g(instantiation_cache_mutex);
```
heifner, line 160: I wonder if this is actually better than just acquiring a `scope_lock` on a non-shared mutex. It would simplify the code to just use a normal lock and not acquire it twice when you have to build and add to the cache.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -84,48 +124,90 @@ namespace eosio { namespace chain {
...
-            });
-            trx_context.pause_billing_timer();
-            wasm_instantiation_cache.modify(it, [&](auto& c) {
-               c.module = runtime_interface->instantiate_module(codeobject->code.data(), codeobject->code.size(), code_hash, vm_type, vm_version);
-            });
+         if (trx_context.control.is_write_window()) {
+            return build_and_get_instantiated_module(code_hash, vm_type, vm_version, trx_context);
+         } else {
+            // Use unique lock as we need to modify the cache
+            std::unique_lock g(instantiation_cache_mutex);
```
heifner, line 174: You can use a `std::scope_lock` or `std::lock_guard` here instead. You only need a `unique_lock` if you are going to manually `unlock` it.

libraries/chain/webassembly/runtimes/eos-vm.cpp
```diff
@@ -129,39 +129,41 @@ class eos_vm_instantiated_module : public wasm_instantiated_module_interface {
          _instantiated_module(std::move(mod)) {}
 
       void apply(apply_context& context) override {
-         // Reset execution context (reused per thread)
+         // re-initialize backend from the instantiate module of the contract
+         _runtime->_bkend = *_instantiated_module;
```
heifner, line 133: Add comment about sharing `module`

heifner (changes requested):

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
```diff
@@ -148,6 +230,11 @@ namespace eosio { namespace chain {
 
       const chainbase::database& db;
       const wasm_interface::vm_type wasm_runtime_time;
+      mutable std::shared_mutex instantiation_cache_mutex;
```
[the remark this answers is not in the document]
spoonincode: `mutable` members always stick out a bit to me. It seems like the only reason this is `mutable` is because of `is_code_cached() const`. But maybe this is a good use of `mutable` --  `is_code_cached()` clearly wants to indicate its "presented" (externally accessible) state isn't modified by `is_code_cached()`, so it does seem to make some sense to "exclude" mutexes from `const`ness. Just curious to hear a little more what others think about this pattern... tbh I didn't look around and see if it's well established elsewhere in our code.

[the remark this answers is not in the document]
heifner: I do think mutexes is the perfect example of a valid use of mutable.

libraries/chain/include/eosio/chain/wasm_interface_private.hpp
linh2931: Since an instantiated module is likely in the cache, `find` (read operation)  succeed most of the time. That's why I chose shared mutex. But that made code complicated. Prefer to a simpler code. Thanks.

heifner (approved):

spoonincode (approved):

Issue #1541, opened 2023-08-22
IF: Fix threading issues in qc_chain


heifner:

ericpassmore:
Comments in `qc_chain.cpp`
https://github.com/AntelopeIO/leap/blob/231532316abda2af5308ec90967e2a5f0e71120b/libraries/hotstuff/qc_chain.cpp#L733-L746
Comments in `qc_chain.hpp`
https://github.com/AntelopeIO/leap/blob/231532316abda2af5308ec90967e2a5f0e71120b/libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp#L149-L161
Comments in `qc_pacemaker.hpp`
https://github.com/AntelopeIO/leap/blob/231532316abda2af5308ec90967e2a5f0e71120b/libraries/hotstuff/include/eosio/hotstuff/chain_pacemaker.hpp#L52-L64

fcecin:
Note (1): the upcoming PR for this moves everything in qc_chain that isn't an interface to `private:` as discussed in the last meeting, as that helped with solving threading.

Note (2): the "Non-proposing leaders" / on_beat() comment is not a threading issue so I'm not addressing it in the upcoming PR (this item can be moved to another ticket if needed).

Status: after validating the solution to the get_state cache (the overhead ended up just incrementing an uint64 inside the global engine mutex + qc_chain is now 100% lock-free internally + R/W lock at the cache that is at the chain_pacemaker now + cache update check itself is lock-free), I will work on "do not access chain state from net threads" next.

PR #1550, opened 2023-08-23
IF: Fix threading 1/2 (#1541) & qc_chain private:

fcecin:
- Moved everything in qc_chain.hpp that is not an interface to private:
- Fixed test_hotstuff to use get_state() instead of accessing various (now) private fields
- Fixed test_pacemaker to use get_id() instead of accessing (now) private _id field
- Added get_proposal() helper method to finalizer_state struct in chain/hotstuff.hpp since finalizer_state is now used in tests, etc.
- Made qc_chain class lock-free; all thread synchronization is now done externally
- Solved concurrency get_finalizer_state vs. qc_chain updates by caching chain_pacemaker::get_state(); zero overhead on qc_chain state updating, low overhead on external state read (no global/write locks acquired on cache hits)

This addresses part of #1541 (get_state() caching) and in the process of doing that I had to solve the other issue we discussed in the meeting, which was moving everything that isn't an interface on qc_chain to private:, and that also rippled into the tests and then the finalizer_state struct (which got a new helper method).

### DIFF
--- a/libraries/chain/include/eosio/chain/hotstuff.hpp
+++ b/libraries/chain/include/eosio/chain/hotstuff.hpp
@@ -69,7 +69,7 @@ namespace eosio::chain {
       eosio::chain::extended_schedule schedule;
       map<fc::sha256, hs_proposal_message> proposals;
 
-      hs_proposal_message* get_proposal(const fc::sha256 &id) {
+      hs_proposal_message* get_proposal(const fc::sha256& id) {
          auto it = proposals.find(id);
          if (it == proposals.end())
             return nullptr;
--- a/libraries/chain/include/eosio/chain/hotstuff.hpp
+++ b/libraries/chain/include/eosio/chain/hotstuff.hpp
@@ -69,7 +69,7 @@ namespace eosio::chain {
       eosio::chain::extended_schedule schedule;
       map<fc::sha256, hs_proposal_message> proposals;
 
-      hs_proposal_message* get_proposal(const fc::sha256& id) {
+      const hs_proposal_message* get_proposal(const fc::sha256& id) const {
          auto it = proposals.find(id);
          if (it == proposals.end())
             return nullptr;
--- a/libraries/chain/include/eosio/chain/hotstuff.hpp
+++ b/libraries/chain/include/eosio/chain/hotstuff.hpp
@@ -68,6 +68,13 @@ namespace eosio::chain {
       eosio::chain::quorum_certificate current_qc;
       eosio::chain::extended_schedule schedule;
       map<fc::sha256, hs_proposal_message> proposals;
+
+      hs_proposal_message* get_proposal(const fc::sha256 &id) {
+         auto it = proposals.find(id);
+         if (it == proposals.end())
+            return nullptr;
+         return & it->second;
+      }
    };
 
    using hs_proposal_message_ptr = std::shared_ptr<hs_proposal_message>;
--- a/libraries/hotstuff/chain_pacemaker.cpp
+++ b/libraries/hotstuff/chain_pacemaker.cpp
@@ -112,13 +112,13 @@ namespace eosio { namespace hotstuff {
       return _chain->is_builtin_activated( builtin_protocol_feature_t::instant_finality );
    }
 
-   void chain_pacemaker::get_state(finalizer_state& fs, bool force) const {
+   void chain_pacemaker::get_state(finalizer_state& fs) const {
       if (! enabled())
          return;
 
       // lock-free state version check
       uint64_t current_state_version = _qc_chain.get_state_version();
-      if (force || _state_cache_version != current_state_version) {
+      if (_state_cache_version != current_state_version) {
          finalizer_state current_state;
          {
             csc prof("stat");
--- a/libraries/hotstuff/chain_pacemaker.cpp
+++ b/libraries/hotstuff/chain_pacemaker.cpp
@@ -124,11 +124,12 @@ namespace eosio { namespace hotstuff {
             csc prof("stat");
             std::lock_guard g( _hotstuff_global_mutex ); // lock IF engine to read state
             prof.core_in();
-            _qc_chain.get_state(current_state);
             current_state_version = _qc_chain.get_state_version(); // get potentially fresher version
+            if (_state_cache_version != current_state_version) 
+               _qc_chain.get_state(current_state);
             prof.core_out();
          }
-         {
+         if (_state_cache_version != current_state_version) {
             std::unique_lock ul(_state_cache_mutex); // lock cache for writing
             _state_cache = current_state;
             _state_cache_version = current_state_version;
--- a/libraries/hotstuff/chain_pacemaker.cpp
+++ b/libraries/hotstuff/chain_pacemaker.cpp
@@ -112,9 +112,31 @@ namespace eosio { namespace hotstuff {
       return _chain->is_builtin_activated( builtin_protocol_feature_t::instant_finality );
    }
 
-   void chain_pacemaker::get_state( finalizer_state& fs ) const {
-      if (enabled())
-         _qc_chain.get_state( fs ); // get_state() takes scare of finer-grained synchronization internally
+   void chain_pacemaker::get_state(finalizer_state& fs, bool force) const {
+      if (! enabled())
+         return;
+
+      // lock-free state version check
+      uint64_t current_state_version = _qc_chain.get_state_version();
+      if (force || _state_cache_version != current_state_version) {
+         finalizer_state current_state;
+         {
+            csc prof("stat");
+            std::lock_guard g( _hotstuff_global_mutex ); // lock IF engine to read state
+            prof.core_in();
+            _qc_chain.get_state(current_state);
+            current_state_version = _qc_chain.get_state_version(); // get potentially fresher version
+            prof.core_out();
+         }
+         {
+            std::unique_lock ul(_state_cache_mutex); // lock cache for writing
+            _state_cache = current_state;
+            _state_cache_version = current_state_version;
+         }
+      }
+
+      std::shared_lock sl(_state_cache_mutex); // lock cache for reading
+      fs = _state_cache;
    }
 
    name chain_pacemaker::get_proposer() {
--- a/libraries/hotstuff/include/eosio/hotstuff/chain_pacemaker.hpp
+++ b/libraries/hotstuff/include/eosio/hotstuff/chain_pacemaker.hpp
@@ -23,7 +23,7 @@ namespace eosio::hotstuff {
       void on_hs_new_view_msg(const hs_new_view_message& msg); //new view msg event handler
       void on_hs_new_block_msg(const hs_new_block_message& msg); //new block msg event handler
 
-      void get_state( finalizer_state& fs ) const;
+      void get_state(finalizer_state& fs, bool force = false) const;
 
       //base_pacemaker interface functions
 
@@ -54,7 +54,11 @@ namespace eosio::hotstuff {
       // These requests can come directly from the net threads, or indirectly from a
       //   dedicated finalizer thread (TODO: discuss).
 #warning discuss
-      std::mutex              _hotstuff_global_mutex;
+      mutable std::mutex      _hotstuff_global_mutex;
+
+      mutable finalizer_state     _state_cache;
+      mutable std::shared_mutex   _state_cache_mutex;
+      mutable uint64_t            _state_cache_version = 0;
 
       chain::controller*      _chain = nullptr;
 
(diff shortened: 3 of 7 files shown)


libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -31,65 +33,37 @@ namespace eosio::hotstuff {
...
-         qc = 3,
-         vote = 4
-      };
-
-      bool _chained_mode = false;
-
-      fc::sha256 _b_leaf = NULL_PROPOSAL_ID;
-      fc::sha256 _b_lock = NULL_PROPOSAL_ID;
-      fc::sha256 _b_exec = NULL_PROPOSAL_ID;
+      uint64_t get_state_version() const { return _state_version; } // calling this w/ thread sync is optional
```
heifner, line 45: `_state_version` is not thread-safe. It should be made `std::atomic<>`.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -21,6 +21,8 @@
 #include <exception>
 #include <stdexcept>
 
+#include <mutex>
+#include <shared_mutex>
```
heifner, line 25: These are not used in the header and should be removed.

libraries/hotstuff/chain_pacemaker.cpp
```diff
@@ -112,9 +112,31 @@ namespace eosio { namespace hotstuff {
       return _chain->is_builtin_activated( builtin_protocol_feature_t::instant_finality );
    }
 
-   void chain_pacemaker::get_state( finalizer_state& fs ) const {
-      if (enabled())
-         _qc_chain.get_state( fs ); // get_state() takes scare of finer-grained synchronization internally
+   void chain_pacemaker::get_state(finalizer_state& fs, bool force) const {
+      if (! enabled())
```
heifner, line 116: This is still accessing chain state from non-main thread. I assume you are going to address that in part two of this PR.

libraries/hotstuff/chain_pacemaker.cpp
```diff
@@ -112,9 +112,31 @@ namespace eosio { namespace hotstuff {
       return _chain->is_builtin_activated( builtin_protocol_feature_t::instant_finality );
    }
 
-   void chain_pacemaker::get_state( finalizer_state& fs ) const {
-      if (enabled())
-         _qc_chain.get_state( fs ); // get_state() takes scare of finer-grained synchronization internally
+   void chain_pacemaker::get_state(finalizer_state& fs, bool force) const {
```
heifner, line 115: I don't see that `force` is ever `true`. I think it should be removed if not used.

libraries/hotstuff/include/eosio/hotstuff/chain_pacemaker.hpp
```diff
@@ -54,7 +54,11 @@ namespace eosio::hotstuff {
       // These requests can come directly from the net threads, or indirectly from a
       //   dedicated finalizer thread (TODO: discuss).
 #warning discuss
-      std::mutex              _hotstuff_global_mutex;
+      mutable std::mutex      _hotstuff_global_mutex;
+
+      mutable finalizer_state     _state_cache;
+      mutable std::shared_mutex   _state_cache_mutex;
+      mutable uint64_t            _state_cache_version = 0;
```
heifner, line 61: `_state_cache_version` needs to be `std::atomic<>`

libraries/hotstuff/include/eosio/hotstuff/chain_pacemaker.hpp
```diff
@@ -54,7 +54,11 @@ namespace eosio::hotstuff {
       // These requests can come directly from the net threads, or indirectly from a
       //   dedicated finalizer thread (TODO: discuss).
 #warning discuss
-      std::mutex              _hotstuff_global_mutex;
+      mutable std::mutex      _hotstuff_global_mutex;
+
+      mutable finalizer_state     _state_cache;
+      mutable std::shared_mutex   _state_cache_mutex;
```
heifner, line 60: Please move mutex above data it is protecting and add comment to what it is protecting.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -31,65 +33,37 @@ namespace eosio::hotstuff {
...
-
-      fc::sha256 _b_leaf = NULL_PROPOSAL_ID;
-      fc::sha256 _b_lock = NULL_PROPOSAL_ID;
-      fc::sha256 _b_exec = NULL_PROPOSAL_ID;
+      uint64_t get_state_version() const { return _state_version; } // calling this w/ thread sync is optional
 
-      fc::sha256 _b_finality_violation = NULL_PROPOSAL_ID;
-
-      block_id_type _block_exec = NULL_BLOCK_ID;
+      name get_id() const { return _id; } // so far, only ever relevant in a test environment (no sync)
```
heifner, line 47: Maybe rename to `_get_id()` or `get_id_i()`

heifner (changes requested):

libraries/hotstuff/chain_pacemaker.cpp
fcecin: I was thinking about solving _chain-> in a separate PR, then close #1541 when both PRs are in.

libraries/chain/include/eosio/chain/hotstuff.hpp
```diff
@@ -68,6 +68,13 @@ namespace eosio::chain {
       eosio::chain::quorum_certificate current_qc;
       eosio::chain::extended_schedule schedule;
       map<fc::sha256, hs_proposal_message> proposals;
+
+      hs_proposal_message* get_proposal(const fc::sha256 &id) {
```
heifner, line 72: Not a huge deal, but please to follow current code convention of `const type& arg`

heifner (approved):

fcecin:
For completeness: even if the _state_version counter in qc_chain is ever incremented in non-optimal places inside the core, everything will still work. For optimal result, the idea is to increment it right before a request completes and execution will leave qc_chain, whenever any change to the state was made. There's also no requirement that it has to be incremented only once while serving a request (though it does make sense to increment it only once if there was any change during the request). I scanned the entire qc_chain (all methods, all state changes) and I think I added the increments in the optimal places.

libraries/chain/include/eosio/chain/hotstuff.hpp
```diff
@@ -68,6 +68,13 @@ namespace eosio::chain {
       eosio::chain::quorum_certificate current_qc;
       eosio::chain::extended_schedule schedule;
       map<fc::sha256, hs_proposal_message> proposals;
+
+      hs_proposal_message* get_proposal(const fc::sha256& id) {
+         auto it = proposals.find(id);
+         if (it == proposals.end())
+            return nullptr;
+         return & it->second;
+      }
```
greg7mdp, line 77: Please make the function `const` returning `const hs_proposal_message*`

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -134,31 +103,45 @@ namespace eosio::hotstuff {
...
+      block_id_type _block_exec = NULL_BLOCK_ID;
+      block_id_type _pending_proposal_block = NULL_BLOCK_ID;
+      fc::sha256 _b_leaf = NULL_PROPOSAL_ID;
+      fc::sha256 _b_lock = NULL_PROPOSAL_ID;
+      fc::sha256 _b_exec = NULL_PROPOSAL_ID;
+      fc::sha256 _b_finality_violation = NULL_PROPOSAL_ID;
+      eosio::chain::quorum_certificate _high_qc;
+      eosio::chain::quorum_certificate _current_qc;
+      uint32_t _v_height = 0;
+      eosio::chain::extended_schedule _schedule;
```
greg7mdp, line 137: Why do we duplicate almost all members of `struct finalizer_state` here? It would be better to have the same struct declaration here (maybe creating a base struct for `finalizer_state` with the common members (without the `std::map`). This would also make the `qc_chain::get_state(finalizer_state& fs)` simpler.

libraries/hotstuff/chain_pacemaker.cpp
```diff
@@ -112,9 +112,31 @@ namespace eosio { namespace hotstuff {
...
+      // lock-free state version check
+      uint64_t current_state_version = _qc_chain.get_state_version();
+      if (_state_cache_version != current_state_version) {
+         finalizer_state current_state;
+         {
+            csc prof("stat");
+            std::lock_guard g( _hotstuff_global_mutex ); // lock IF engine to read state
+            prof.core_in();
+            _qc_chain.get_state(current_state);
+            current_state_version = _qc_chain.get_state_version(); // get potentially fresher version
```
greg7mdp, line 128: This is a little awkward. Is there a reason why we don't include the version in `struct finalizer_state`?

libraries/hotstuff/qc_chain.cpp
```diff
@@ -497,6 +491,8 @@ namespace eosio { namespace hotstuff {
       //check for leader change
       leader_rotation_check();
 
+      ++_state_version;
```
greg7mdp, line 494: This is not reliable. For example the call above to ` update(proposal);` contains  EOS_ASSERT macros, which could throw exceptions causing `++_state_version;` not to be executed (even though we modified the state in `insert_proposal` above.
I'd recomment using `auto increment_version = fc::make_scoped_exit([this]() { ++_state_version; };` where the state is modified **first**.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
fcecin: `finalizer_state` was created for the `get_finalizer_state` API. I can create a `finalizer_state  _fs` member of `qc_chain`, but it will leave somethings out since not all state is relevant for the API.

libraries/hotstuff/chain_pacemaker.cpp
fcecin: The version is only relevant for the cache. It's not related to anything outside thread synchronization.

greg7mdp: OK, resolving this for now.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
greg7mdp: I think we may find a somewhat cleaner way to do this eventually, but it is not critical and I don't want to delay the review.

fcecin: `finalizer_state` ended up being useful for the tests as well (once we moved all the qc_chain member vars to private).

libraries/hotstuff/chain_pacemaker.cpp
```diff
@@ -112,9 +112,31 @@ namespace eosio { namespace hotstuff {
...
+            prof.core_out();
+         }
+         {
```
greg7mdp, line 131:
```suggestion
            current_state_version = _qc_chain.get_state_version(); // get potentially fresher version
            if (_state_cache_version != current_state_version) 
               _qc_chain.get_state(current_state);
            prof.core_out();
         }
         if (_state_cache_version != current_state_version) {
```

Reason for this change is that multiple http threads could be calling this at the same time, and end up waiting on the lock at line 125. After the first thread has updated the cache, there is no need for the other threads to do the same thing.

libraries/hotstuff/qc_chain.cpp
```diff
@@ -589,7 +591,7 @@ namespace eosio { namespace hotstuff {
       if (_log) ilog(" === ${id} process_new_view === ${qc}", ("qc", msg.high_qc)("id", _id));
 #endif
       if (update_high_qc(msg.high_qc)) {
-         ++_state_version;
+         ++_state_version; // should be OK; update_high_qc() should not throw
```
greg7mdp, line 594: Maybe OK now, but if someone adds an EOS_ASSERT in there all bets are off. 
Better to add: `auto increment_version = fc::make_scoped_exit([this]() { ++_state_version; });` **before** calling `update_high_qc` and calling `increment_version.cancel()` if the function returns false.

libraries/hotstuff/qc_chain.cpp
```diff
@@ -596,7 +590,9 @@ namespace eosio { namespace hotstuff {
 #ifdef QC_CHAIN_TRACE_DEBUG
       if (_log) ilog(" === ${id} process_new_view === ${qc}", ("qc", msg.high_qc)("id", _id));
 #endif
-      update_high_qc(msg.high_qc);
+      if (update_high_qc(msg.high_qc)) {
+         ++_state_version; // should be OK; update_high_qc() should not throw
+      }
```
greg7mdp, line 595:
```suggestion
      auto increment_version = fc::make_scoped_exit([this]() { ++_state_version; });
      if (!update_high_qc(msg.high_qc)) {
         increment_version.cancel();
      }
```

greg7mdp (changes requested):

greg7mdp (approved):

PR #1552, opened 2023-08-24
IF: Use FC logging framework for better management and consistency

linh2931:
This PR sets up FC logging for hotstuff component such that logging can be controlled dynamically and is consistent with nodeos' other components. More refinements might be needed.

It
* creates a specific logging category for `hotstuff`,
* uses `fc_elog` for errors,
* uses `fc_tlog` for everything was inside `QC_CHAIN_TRACE_DEBUG`
* uses `fc_dlog` for the rest of the original `ilog`.

### DIFF
--- a/libraries/hotstuff/chain_pacemaker.cpp
+++ b/libraries/hotstuff/chain_pacemaker.cpp
@@ -100,9 +100,10 @@ namespace eosio { namespace hotstuff {
 #endif
 //===============================================================================================
 
-   chain_pacemaker::chain_pacemaker(controller* chain, std::set<account_name> my_producers, bool info_logging, bool error_logging)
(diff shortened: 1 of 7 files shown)


greg7mdp:
It looks like with this change, we lost the ability in the test to choose which replicas should log which info (or not log anything). Possibly this will make the log harder to understand.

fcecin:
> It looks like with this change, we lost the ability in the test to choose which replicas should log which info (or not log anything). Possibly this will make the log harder to understand.

I think those controls were more relevant during development of the tests. Since the tests that are in there are pretty much established, they won't suffer from this, as long as the test run has all types of log statements enabled in nodeos. It might actually be good that these custom log mechanisms are all stripped and you get the maximal log every time during test runs. We can add them back later, and possibly in some other form, if needed.

fcecin (approved):

heifner (approved):

Issue #1518, opened 2023-08-17
IF: Address abstraction leakage in HotStuff library


arhag:
Specifically:

* `fc::unsigned_int` usage (which we reserve only for serialization)
* On wire message struct passed throughout the code.

heifner:
From @ericpassmore : chain_plugin.hpp is included in many places carrying hotstuff with it. When I look for the hs_* structures which also include chain_plugin.hpp I find three files net_plugin producer_plugin chain_plugin. An alternative would be having another header with the hotstuff structures and include the new header in the three files.

PR #1560, opened 2023-08-25
IF: Support larger number of finalizers

### LINKED ISSUE #1547: IF: hs_*_message types - review
Type for `quorum_certificate.active_finalizers` https://github.com/AntelopeIO/leap/pull/1536#discussion_r1302380233

> This assumes 21 producers. Also since unsigned_int is used for value this has a upper limit of 32 finalizers (currently producers). Seems like quorum_certificate.active_finalizers should be a std::string instead to support greater than 32 finalizers. Note this affects on wire format so we need to get it right before 5.0.

Do we want to version `hs_*_messages` used by `net_plugin`. Note new ones can always be added by adding to `net_message` `std::variant`, but we might rather create a `std::variant` for each `hs_*_message` or add a `uint16_t version` as first element of each for internal versioning.

heifner:
Change `quorum_certificate` to use a `std::string` bitset representation instead of `unsigned_int` to move from 32 supported finalizers to unlimited.

Resolves #1547

### DIFF
--- a/libraries/chain/include/eosio/chain/hotstuff.hpp
+++ b/libraries/chain/include/eosio/chain/hotstuff.hpp
@@ -13,7 +13,7 @@ namespace eosio::chain {
    const block_id_type NULL_BLOCK_ID = block_id_type("00");
    const fc::sha256 NULL_PROPOSAL_ID = fc::sha256("00");
 
-   using hs_dynamic_bitset = boost::dynamic_bitset<uint32_t>;
+   using hs_bitset = boost::dynamic_bitset<uint32_t>;
 
    inline uint64_t compute_height(uint32_t block_height, uint32_t phase_counter) {
       return (uint64_t{block_height} << 32) | phase_counter;
--- a/libraries/chain/include/eosio/chain/hotstuff.hpp
+++ b/libraries/chain/include/eosio/chain/hotstuff.hpp
@@ -36,11 +36,11 @@ namespace eosio::chain {
          quorum_met = false;
       }
 
-      auto get_active_finalizers() const {
+      boost::dynamic_bitset<> get_active_finalizers() const {
          assert(!active_finalizers.empty());
          return boost::dynamic_bitset(active_finalizers);
       }
-      void set_active_finalizers(const auto& bs) {
+      void set_active_finalizers(const boost::dynamic_bitset<>& bs) {
          assert(!bs.empty());
          boost::to_string(bs, active_finalizers);
       }
--- a/libraries/chain/include/eosio/chain/hotstuff.hpp
+++ b/libraries/chain/include/eosio/chain/hotstuff.hpp
@@ -22,42 +22,10 @@ namespace eosio::chain {
       std::map<name, fc::crypto::blslib::bls_public_key>   bls_pub_keys;
    };
 
-   struct quorum_certificate {
-      explicit quorum_certificate(uint32_t finalizer_size = 0) {
-         reset(NULL_PROPOSAL_ID, finalizer_size);
-      }
-
-      void reset(const fc::sha256& proposal, uint32_t finalizer_size) {
-         proposal_id = proposal;
-         boost::dynamic_bitset b;
-         b.resize(finalizer_size);
-         boost::to_string(b, active_finalizers);
-         active_agg_sig = fc::crypto::blslib::bls_signature();
-         quorum_met = false;
-      }
-
-      boost::dynamic_bitset<> get_active_finalizers() const {
-         assert(!active_finalizers.empty());
-         return boost::dynamic_bitset(active_finalizers);
-      }
-      void set_active_finalizers(const boost::dynamic_bitset<>& bs) {
-         assert(!bs.empty());
-         boost::to_string(bs, active_finalizers);
-      }
-      const std::string& get_active_finalizers_string() const { return active_finalizers; }
-
-      const fc::sha256& get_proposal_id() const { return proposal_id; }
-      const fc::crypto::blslib::bls_signature& get_active_agg_sig() const { return active_agg_sig; }
-      void set_active_agg_sig( const fc::crypto::blslib::bls_signature& sig) { active_agg_sig = sig; }
-      bool is_quorum_met() const { return quorum_met; }
-      void set_quorum_met() { quorum_met = true; }
-
-   private:
-      friend struct fc::reflector<quorum_certificate>;
+   struct quorum_certificate_message {
       fc::sha256                          proposal_id = NULL_PROPOSAL_ID;
       std::string                         active_finalizers; //bitset encoding, following canonical order
       fc::crypto::blslib::bls_signature   active_agg_sig;
-      bool                                quorum_met = false; // not serialized across network
    };
 
    struct hs_vote_message {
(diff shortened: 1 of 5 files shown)


libraries/hotstuff/qc_chain.cpp
```diff
@@ -128,32 +128,22 @@ namespace eosio { namespace hotstuff {
...
-           ("value", value)
-           ("finalizer", finalizer));
+   boost::dynamic_bitset<> qc_chain::update_bitset(const boost::dynamic_bitset<>& finalizer_set, name finalizer ) {
 
-      boost::dynamic_bitset b( 21, value );
+      boost::dynamic_bitset b( finalizer_set );
       vector<name> finalizers = _pacemaker->get_finalizers();
       for (size_t i = 0; i < finalizers.size();i++) {
          if (finalizers[i] == finalizer) {
             b.flip(i);
```
greg7mdp, line 141: should it be `b.set(i)`?

libraries/hotstuff/qc_chain.cpp
```diff
@@ -239,25 +226,25 @@ namespace eosio { namespace hotstuff {
...
          return false;
       }
 
-      boost::dynamic_bitset fb(21, finalizers.value);
       fc::crypto::blslib::bls_public_key agg_key;
 
-      for (boost::dynamic_bitset<>::size_type i = 0; i < fb.size(); i++) {
-         if (fb[i] == 1){
+      for (boost::dynamic_bitset<>::size_type i = 0; i < finalizers.size(); i++) {
+         if (finalizers[i] == 1){
```
greg7mdp, line 240:
```suggestion
         if (finalizers[i]){
```
Comparing with 1 is not quite right since operator[]() returns a `bool`

libraries/hotstuff/qc_chain.cpp
```diff
@@ -498,19 +489,21 @@ namespace eosio { namespace hotstuff {
...
          auto increment_version = fc::make_scoped_exit([this]() { ++_state_version; });
 
-         if (_current_qc.active_finalizers>0)
-            _current_qc.active_agg_sig = fc::crypto::blslib::aggregate({_current_qc.active_agg_sig, vote.sig });
+         boost::dynamic_bitset finalizer_set = _current_qc.get_active_finalizers();
+         if (finalizer_set.any())
+            _current_qc.set_active_agg_sig(fc::crypto::blslib::aggregate({_current_qc.get_active_agg_sig(), vote.sig }));
          else
-            _current_qc.active_agg_sig = vote.sig;
+            _current_qc.set_active_agg_sig(vote.sig);
```
greg7mdp, line 503: I don't think this is ideal. I think we should have a `bls_signature& operator+=(const bls_signature&)` and replace all this with:
```
    _current_qc.active_agg_sig += vote.sig;
```

libraries/chain/include/eosio/chain/hotstuff.hpp
```diff
@@ -21,10 +23,41 @@ namespace eosio::chain {
...
+      const fc::crypto::blslib::bls_signature& get_active_agg_sig() const { return active_agg_sig; }
+      void set_active_agg_sig( const fc::crypto::blslib::bls_signature& sig) { active_agg_sig = sig; }
+      bool is_quorum_met() const { return quorum_met; }
+      void set_quorum_met() { quorum_met = true; }
+
+   private:
+      friend struct fc::reflector<quorum_certificate>;
       fc::sha256                          proposal_id = NULL_PROPOSAL_ID;
-      fc::unsigned_int                    active_finalizers = 0; //bitset encoding, following canonical order
+      std::string                         active_finalizers; //bitset encoding, following canonical order
```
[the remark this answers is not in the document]
heifner: Because this struct is also used for the wire format which will be addressed under #1518.

libraries/chain/include/eosio/chain/hotstuff.hpp
```diff
@@ -21,10 +23,41 @@ namespace eosio::chain {
...
+         boost::to_string(b, active_finalizers);
+         active_agg_sig = fc::crypto::blslib::bls_signature();
+         quorum_met = false;
+      }
+
+      auto get_active_finalizers() const {
+         assert(!active_finalizers.empty());
+         return boost::dynamic_bitset(active_finalizers);
+      }
+      void set_active_finalizers(const auto& bs) {
```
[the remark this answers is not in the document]
heifner: Changed to `boost::dynamic_bitset<>`

libraries/hotstuff/qc_chain.cpp
heifner: One would assume so. @systemzax ?

libraries/hotstuff/qc_chain.cpp
heifner: I would be ok with an `aggregate` method on `bls_signature`, but I don't like `operator+=`.

greg7mdp: `operator+=()` seems like a very natural fit to me, but I'm fine with `aggregate`.

libraries/chain/include/eosio/chain/hotstuff.hpp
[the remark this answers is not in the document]
heifner: I went ahead and addressed this under this PR. There is still work to be done under #1518. For example, making `qc_chain` a pimpl to reduce all the types pulled in by `qc_chain.hpp`.

greg7mdp (approved):

libraries/hotstuff/qc_chain.cpp
```diff
@@ -128,32 +128,22 @@ namespace eosio { namespace hotstuff {
...
-   fc::unsigned_int qc_chain::update_bitset(fc::unsigned_int value, name finalizer ) {
-
-      fc_tlog(_logger, " === update bitset ${value} ${finalizer}",
-           ("value", value)
-           ("finalizer", finalizer));
+   boost::dynamic_bitset<> qc_chain::update_bitset(const boost::dynamic_bitset<>& finalizer_set, name finalizer ) {
 
-      boost::dynamic_bitset b( 21, value );
+      boost::dynamic_bitset b( finalizer_set );
       vector<name> finalizers = _pacemaker->get_finalizers();
```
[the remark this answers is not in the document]
heifner: ? `_pacemaker->get_finalizers()` returns a `vector<name>`. Regardless I expect this to change.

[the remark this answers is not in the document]
greg7mdp: exactly :-)

libraries/hotstuff/qc_chain.cpp
systemzax: Yes, .set(i). Not sure why I had flip there, doesn't make sense.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -38,28 +38,34 @@ namespace eosio::hotstuff {
 
       explicit quorum_certificate(const quorum_certificate_message& msg)
               : proposal_id(msg.proposal_id)
-              , active_finalizers(msg.active_finalizers) // conversion from string
+              , active_finalizers()
```
greg7mdp, line 41: minor... this line is not needed.
Why not initialize directly from the message:
`, active_finalizers(msg.active_finalizers.cbegin(), msg.active_finalizers.cend())`

greg7mdp (approved):

fcecin (approved):

Issue #1511, opened 2023-08-17
IF: Host function to set active finalizer set


arhag:
Depends on https://github.com/AntelopeIO/leap/issues/1509.

One new host function called `set_finalizers` should be added upon the activation of the IF protocol feature.

arhag:
Progress started with code in https://github.com/fcecin/leap/commits/74295f20142ef415d6e118701777f12a7937f187.

arhag:
The structure provided to the host function (typically referred to as the "finalizer set") should include:
* `uint64_t threshold`
* `std::vector` of structs that each contain:
   - `finalizer_key` which is a BLS public key
   - `uint64_t weight` (check that this is not zero)
   - `std::string description` (this is intended to hold the name of the producer using this finalizer key)

The host function should ensure that the `finalizer_key` is unique within the `std::vector`. It should also add all weights and ensure that the threshold is greater than half the total weight.

PR #1561, opened 2023-08-25
set_finalizers host function (#1511)

fcecin:
This PR is *not* ready to be merged; this is not a request for a final review as the code is not complete. I created it to track progress on the set_finalizers branch and hold discussions. At some point this PR will have everything and we will merge.

- Implemented host function (not tested);
- Finalizer set types;
- Placeholders for fc::crypto::blslib::bls_public_key operators == != and <;
- Stub notify_set_finalizers signal to be connected to chain_plugin, to reach chain_pacemaker


libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -356,6 +360,7 @@ namespace eosio { namespace chain {
          signal<void(const hs_vote_message_ptr&)> new_hs_vote_message;
          signal<void(const hs_new_view_message_ptr&)>    new_hs_new_view_message;
          signal<void(const hs_new_block_message_ptr&)> new_hs_new_block_message;
+         signal<void(std::tuple<const uint64_t, const vector<finalizer_authority>&>)> notify_set_finalizers;
```
heifner, line 363: I don't think we want this as a signal. Remove this.

heifner (changes requested):

fcecin:
I'm thinking whether the structs in finalizer_set.hpp (copied from the producer schedule, mostly) are what we want:

* are they versioned?
* do we want the shared_ versions?
* should the set_finalizers host function just instantiate an entire finalized_set (or shared_finalizer_set) and pass it to the controller?

fcecin:
@mschoenebeck is going to give me a hand to solve the operator== != < for bls_public_key, so we will be rolling that into this PR

greg7mdp:
> @mschoenebeck is going to give me a hand to solve the operator== != < for bls_public_key, so we will be rolling that into this PR

Please use the C++20 [spaceship operator](https://en.cppreference.com/w/cpp/language/default_comparisons)

fcecin:
I'll test the host function tomorrow (call it and see if it logs the same info at the pacemaker). After that basic test passes (I will notify) I think it will be good on my side for the final code reviews.

(NOTE: I have no idea why the CI checks are failing right now or if that's actually relevant)

libraries/chain/controller.cpp
```diff
@@ -1990,6 +1994,16 @@ struct controller_impl {
...
 
+   void set_finalizers_impl(uint64_t fthreshold, vector<finalizer_authority> finalizers) {
+      this->fthreshold = fthreshold;
+      this->finalizers = finalizers;
+   }
+
+   void get_finalizers_impl(uint64_t& fthreshold, vector<finalizer_authority>& finalizers) {
+      fthreshold = this->fthreshold;
+      finalizers = this->finalizers;
+   }
```
greg7mdp, line 2005: I think it would be nicer to have:
```
std::pair<uint64_t, vector<finalizer_authority>> get_finalizers_impl() {
   return { fthreshold, finalizers );
}
```

which you can the use as:

```
auto [threshold, finalizers] = controller.get_finalizers();
```

libraries/chain/controller.cpp
```diff
@@ -3313,6 +3327,14 @@ int64_t controller::set_proposed_producers( vector<producer_authority> producers
    return version;
 }
 
+void controller::set_finalizers( uint64_t fthreshold, vector<finalizer_authority> finalizers ) {
+   my->set_finalizers_impl(fthreshold, std::move(finalizers));
+}
+
+void controller::get_finalizers( uint64_t& fthreshold, vector<finalizer_authority>& finalizers ) {
+   my->get_finalizers_impl(fthreshold, finalizers);
+}
```
greg7mdp, line 3336:
```suggestion
std::pair<uint64_t, vector<finalizer_authority>> controller::get_finalizers() const {
   return my->get_finalizers_impl();
}
```

libraries/chain/controller.cpp
```diff
@@ -1990,6 +1994,16 @@ struct controller_impl {
       emit( self.new_hs_new_block_message, msg );
    }
 
+   void set_finalizers_impl(uint64_t fthreshold, vector<finalizer_authority> finalizers) {
+      this->fthreshold = fthreshold;
+      this->finalizers = finalizers;
```
greg7mdp, line 1999:
```suggestion
      this->finalizers = std::move(finalizers);
```

greg7mdp (changes requested):

libraries/chain/controller.cpp
```diff
@@ -1999,9 +1999,8 @@ struct controller_impl {
       this->finalizers = std::move(finalizers);
    }
 
-   void get_finalizers_impl(uint64_t& fthreshold, vector<finalizer_authority>& finalizers) {
-      fthreshold = this->fthreshold;
-      finalizers = this->finalizers;
+   std::pair<uint64_t, vector<finalizer_authority>> get_finalizers_impl() {
```
greg7mdp, line 2002: Thanks for the changes, @fcecin . this function should be `const`.

libraries/chain/include/eosio/chain/finalizer_set.hpp
```diff
@@ -0,0 +1,144 @@
...
+      auto to_shared(chainbase::allocator<char> alloc) const {
+         return shared_finalizer_authority(description, fweight, public_key);
+      }
+
+      static auto from_shared( const shared_finalizer_authority& src ) {
+         finalizer_authority result;
+         result.description = src.description;
+         result.fweight = src.fweight;
+         result.public_key = src.public_key;
+         return result;
```
greg7mdp, line 62:
```suggestion
         return finalizer_authority { src.description, src.fweight, src.public_key };
```

libraries/chain/include/eosio/chain/finalizer_set.hpp
```diff
@@ -0,0 +1,144 @@
...
+       * this method produces an appropriate variant for contracts where the authority field
+       * is correctly formatted
+       */
+      fc::variant get_abi_variant() const;
+
+      friend bool operator == ( const finalizer_authority& lhs, const finalizer_authority& rhs ) {
+         return tie( lhs.description, lhs.fweight, lhs.public_key ) == tie( rhs.description, rhs.fweight, rhs.public_key );
+      }
+      friend bool operator != ( const finalizer_authority& lhs, const finalizer_authority& rhs ) {
+         return tie( lhs.description, lhs.fweight, lhs.public_key ) != tie( rhs.description, rhs.fweight, rhs.public_key );
```
greg7mdp, line 83:
```suggestion
         return !(lhs == rhs);
```

libraries/chain/include/eosio/chain/finalizer_set.hpp
```diff
@@ -0,0 +1,144 @@
...
+#include <eosio/chain/types.hpp>
+#include <chainbase/chainbase.hpp>
+#include <eosio/chain/authority.hpp>
+#include <eosio/chain/snapshot.hpp>
+
+#include <fc/crypto/bls_public_key.hpp>
+
+namespace eosio::chain {
+
+   struct shared_finalizer_authority {
```
greg7mdp, line 13: I don't understand why we have two classes `shared_finalizer_authority` and `finalizer_authority`. They seem identical to me?

fcecin: This is all copy/paste from the producer schedule and producer authority. Finalizer set was the producer schedule, and finalizer authority is from producer authority. I didn't know if the shared versions were needed, so I added them by default. If we are positive we don't need them, then I can just remove them.

heifner: Leave for now until we decide what we are storing in chainbase. The `shared` versions are the chainbase types.

heifner (changes requested):

libraries/chain/include/eosio/chain/finalizer_set.hpp
```diff
@@ -0,0 +1,140 @@
...
+
+      shared_finalizer_authority( const std::string& description, const uint64_t fweight, const fc::crypto::blslib::bls_public_key& public_key )
+      :description(description)
+      ,fweight(fweight)
+      ,public_key(public_key)
+      {}
+
+      std::string      description;
+      uint64_t         fweight;
+      fc::crypto::blslib::bls_public_key   public_key;
```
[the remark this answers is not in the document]
fcecin: It's a `std::array<uint64_t, 6> d;` x 3 (`x`, `y` and `z` members).

libraries/chain/include/eosio/chain/finalizer_set.hpp
```diff
@@ -0,0 +1,140 @@
...
+      shared_finalizer_authority& operator= ( shared_finalizer_authority && ) = default;
+      shared_finalizer_authority& operator= ( const shared_finalizer_authority & ) = default;
+
+      shared_finalizer_authority( const std::string& description, const uint64_t fweight, const fc::crypto::blslib::bls_public_key& public_key )
+      :description(description)
+      ,fweight(fweight)
+      ,public_key(public_key)
+      {}
+
+      std::string      description;
```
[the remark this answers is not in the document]
fcecin: This breaks compilation (don't know why) so leaving for later

fcecin:
I'm thinking this is going to be tricky to test properly: to compile a contract with set_finalizers() in it, wouldn't an updated CDT be needed? But by reviewing the code, it either works or needs minor modifications, so I think it's OK for checking in and push testing for later. Unless someone has a better idea.

The one question I have is whether the bls_public_key inside the finalizer_authority will correctly convert the base58 encoded string of the key at the contract level into the BLS libary g1 object. I don't see why it wouldn't.

greg7mdp (approved):

PR #1563, opened 2023-08-25
Remove some conversion operators

heifner:
This started as a PR to remove `operator<<` from `public_key` as `to_string()` which optionally takes a `yield` function should be used instead. In the process noticed some conversion operators that could be removed. Some of these changes are a bit opinionated, but this PR actually ends up making what I think is a good case for not using conversion operators. Note there are a few optimizations that were not obvious because the conversion operators "magically" provided a less than ideal implementation.

This PR does not remove `operator<<` from `public_key`  as that breaks tests in `libtester`. There are however, enough improvements that this PR is still worthwhile in my opinion.

I didn't add spaceship operators because this PR already changed more than I originally anticipated.

### DIFF
--- a/libraries/chain/include/eosio/chain/authority.hpp
+++ b/libraries/chain/include/eosio/chain/authority.hpp
@@ -212,10 +212,6 @@ struct authority {
 
    friend bool operator == ( const authority& lhs, const shared_authority& rhs );
 
-   friend bool operator != ( const authority& lhs, const authority& rhs ) {
-      return tie( lhs.threshold, lhs.keys, lhs.accounts, lhs.waits ) != tie( rhs.threshold, rhs.keys, rhs.accounts, rhs.waits );
-   }
-
    void sort_fields () {
       std::sort(std::begin(keys), std::end(keys));
       std::sort(std::begin(accounts), std::end(accounts));
--- a/libraries/chain/include/eosio/chain/authority.hpp
+++ b/libraries/chain/include/eosio/chain/authority.hpp
@@ -179,7 +179,7 @@ namespace config {
 struct shared_authority;
 
 struct authority {
-   explicit authority( public_key_type k, uint32_t delay_sec = 0 )
+   authority( public_key_type k, uint32_t delay_sec = 0 )
    :threshold(1),keys({{k,1}})
    {
       if( delay_sec > 0 ) {
@@ -341,7 +341,7 @@ inline bool validate( const Authority& auth ) {
 } } // namespace eosio::chain
 
 namespace fc {
-   void to_variant(const eosio::chain::shared_public_key& var, fc::variant& vo, const fc::yield_function_t& yield);
+   void to_variant(const eosio::chain::shared_public_key& var, fc::variant& vo);
 } // namespace fc
 
 FC_REFLECT(eosio::chain::permission_level_weight, (permission)(weight) )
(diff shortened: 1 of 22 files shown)


greg7mdp (approved):

heifner:
Ugh! `libtester` tests failed due to missing `operator<<` needed by `BOOST_TEST`. Not sure if that means we need to just abandon removing `operator<<`. Thoughts @spoonincode ?

spoonincode:
the nice thing about `operator<<` here and using things like `BOOST_CHECK_EQUAL()` over `BOOST_CHECK()`, is that I believe `BOOST_CHECK_EQUAL()` will print out the two values on failure, making it much easier to diagnose the failure. It is kind of unfortunate to lose that.

heifner:
> the nice thing about `operator<<` here and using things like `BOOST_CHECK_EQUAL()` over `BOOST_CHECK()`, is that I believe `BOOST_CHECK_EQUAL()` will print out the two values on failure, making it much easier to diagnose the failure. It is kind of unfortunate to lose that.

Could change all of them to compare the string version of public key, but figured someone could easily add that if a test fails.

heifner:
> the nice thing about `operator<<` here and using things like `BOOST_CHECK_EQUAL()` over `BOOST_CHECK()`, is that I believe `BOOST_CHECK_EQUAL()` will print out the two values on failure, making it much easier to diagnose the failure. It is kind of unfortunate to lose that.

Added the `operator<<` back as `libtester` uses it for its tests. @spoonincode

greg7mdp (approved):

libraries/libfc/include/fc/crypto/public_key.hpp
```diff
@@ -40,15 +40,15 @@ namespace fc { namespace crypto {
 
          // serialize to/from string
          explicit public_key(const std::string& base58str);
-         std::string to_string(const fc::yield_function_t& yield = fc::yield_function_t()) const;
+         std::string to_string(const fc::yield_function_t& yield) const;
```
dimas1185, line 43: I'm not getting the point of removing this and then use `to_string({})` in so many places. What is the benefit in your opinion?

libraries/chain/include/eosio/chain/authority.hpp
```diff
@@ -200,6 +210,8 @@ struct authority {
       return tie( lhs.threshold, lhs.keys, lhs.accounts, lhs.waits ) == tie( rhs.threshold, rhs.keys, rhs.accounts, rhs.waits );
    }
 
+   friend bool operator == ( const authority& lhs, const shared_authority& rhs );
+
```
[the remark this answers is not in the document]
greg7mdp: You don't need it in C++20 anymore, the language provides it. `(a != b)` is automatically rewritten as `!(a == b)`

libraries/libfc/include/fc/crypto/public_key.hpp
heifner: The yield is there to prevent abuse. By defaulting the parameter, it makes it easy to miss that a yield should be provided if it is used in critical code.

libraries/chain/include/eosio/chain/authority.hpp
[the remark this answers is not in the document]
dimas1185: ok, then please remove `friend bool operator != ( const authority& lhs, const authority& rhs )`

libraries/libfc/include/fc/crypto/public_key.hpp
dimas1185: ok, it still sounds too strict but it seems that this syntax sugar was abused so I trust you.

dimas1185 (approved):

PR #1565, opened 2023-08-26
IF: Remove recursive implementation of commit

### LINKED ISSUE #1553: IF: Remove recursive implementation of commit in qc_chain

heifner:
Remove recursion by using a loop. Accumulate proposals to commit in a `std::vector`.

Resolves #1553


fcecin (approved):

greg7mdp (commented):
I have to say that I disagree with this change. I think the code is less clear and definitely less efficient. The rationale against recursion is not clear to me in this case, as it would probably require a hierarchy depth in the order of 1 million proposals to blow the stack.

heifner:
> I have to say that I disagree with this change. I think the code is less clear and definitely less efficient. The rationale against recursion is not clear to me in this case, as it would probably require a hierarchy depth in the order of 1 million proposals to blow the stack.

@arhag opinion?

libraries/hotstuff/qc_chain.cpp
```diff
@@ -1054,68 +1055,57 @@ namespace eosio { namespace hotstuff {
...
+         // Execute commands [...]
          fc_dlog(_logger, " === ${id} committed proposal #${block_num} phase ${phase_counter} block_id : ${block_id} proposal_id : ${proposal_id}",
-                        ("id", _id)
-                        ("block_num", proposal.block_num())
-                        ("phase_counter", proposal.phase_counter)
-                        ("block_id", proposal.block_id)
-                        ("proposal_id", proposal.proposal_id));
-      }
-      else {
+                 ("id", _id)("block_num", proposal->block_num())("phase_counter", proposal->phase_counter)("block_id", proposal->block_id)("proposal_id", proposal->proposal_id));
```
greg7mdp, line 1103: This log statement will now appear **before** the parent proposals ones, which I think is the wrong order.

Also the `proposal_stack` as it is now is not very useful as we `pop` immediately after `push`, so the size of the stack never exceeds one. We could use a variable instead of a stack. The stack would only be needed if we wanted to preserve the order of the `fc_dlog` statements.

greg7mdp (changes requested):

fcecin (approved):

greg7mdp:
@heifner I can't add you as a reviewer... probably because it is your PR. But I have updated it.

greg7mdp (approved):

heifner (commented):
LGTM

heifner:
> @heifner I can't add you as a reviewer... probably because it is your PR. But I have updated the branch and approved the PR, so I guess I can merge ;-)

I can't approve. So you approve and then merge.

greg7mdp:
> I can't approve. So you approve and then merge.

Thanks @heifner will do when the tests pass.

PR #1566, opened 2023-08-26
IF: Remove NULL_BLOCK_ID & NULL_PROPOSAL_ID and use empty()

### LINKED ISSUE #1545: IF: Use empty() instead of NULL_BLOCK_ID, NULL_PROPOSAL_ID
Needs to be done before release because it affects what is transferred across the network.

heifner:
Resolves #1545


libraries/hotstuff/qc_chain.cpp
```diff
@@ -525,14 +525,14 @@ namespace eosio { namespace hotstuff {
...
-               if (_pending_proposal_block == NULL_BLOCK_ID)
+               if (_pending_proposal_block.empty())
                   proposal_candidate = new_proposal_candidate( p->block_id, p->phase_counter + 1 );
                else
                   proposal_candidate = new_proposal_candidate( _pending_proposal_block, 0 );
 
                reset_qc(proposal_candidate.proposal_id);
                fc_tlog(_logger, " === ${id} setting _pending_proposal_block to null (process_vote)", ("id", _id));
-               _pending_proposal_block = NULL_BLOCK_ID;
+               _pending_proposal_block = block_id_type{};
```
greg7mdp, line 535: This works and I slightly prefer.
```suggestion
               _pending_proposal_block = {};
```

greg7mdp (approved):

libraries/hotstuff/qc_chain.cpp
heifner: I actually went back and forth. I had it as `{}` first. I'll make the change.

fcecin (approved):

PR #1587, opened 2023-08-31
Changed bls private, public key / signature encoding from base58 to base64

systemzax:


libraries/libfc/include/fc/crypto/bls_common.hpp
```diff
@@ -4,25 +4,25 @@
...
+      FC_ASSERT(!unpacker.remaining(), "decoded base64 length too long");
       auto checksum = wrapper::calculate_checksum(wrapped.data, nullptr);
       FC_ASSERT(checksum == wrapped.check);
 
       return wrapped.data;
    }
 
    template <typename Container>
-   static std::string deserialize_base58( Container data, const yield_function_t& yield) {
+   static std::string deserialize_base64( Container data, const yield_function_t& yield) {
```
heifner, line 25: You can remove all the `yield`s including in the function parameter.

libraries/libfc/test/test_bls.cpp
```diff
@@ -261,6 +265,8 @@ BOOST_AUTO_TEST_CASE(bls_binary_keys_encoding_check) try {
 
   std::string sig_str = sig.to_string();
 
+  std::cout << sig_str << "\n";
```
heifner, line 268: Remove the `std::cout` from the test.

heifner (changes requested):

libraries/libfc/include/fc/crypto/bls_common.hpp
systemzax: ah yes.

arhag (approved):

heifner (approved):

PR #1588, opened 2023-08-31
IF: Modify set_finalizer host function struct type

heifner:
- Rename `finalizer_set` `version` to `generation`.
- Changed `finalizer_set` `bls_public_key` to Jacobian little-endian encoding.
- Remove `generation` from `set_finalizers` host function parameter struct type.
- Add `hs_finalizer_set_extension` for block header extension.
- Move implementation of `finalizer_authority` and `finalizer_set` to source file to avoid including `bls12-381` headers into `libtester`.

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1980,6 +1980,7 @@ struct controller_impl {
    void set_finalizers_impl(const finalizer_set& fin_set) {
       // TODO store in chainbase
       current_finalizer_set = fin_set;
+      ++current_finalizer_set.generation;
    }
 
    /**
--- a/libraries/chain/finalizer_set.cpp
+++ b/libraries/chain/finalizer_set.cpp
@@ -13,15 +13,11 @@ namespace eosio::chain {
    finalizer_set::~finalizer_set() = default;
 
    finalizer_set::finalizer_set(const finalizer_set&) = default;
-   finalizer_set::finalizer_set(finalizer_set&&) = default;
+   finalizer_set::finalizer_set(finalizer_set&&) noexcept = default;
 
    finalizer_set& finalizer_set::operator=(const finalizer_set&) = default;
-   finalizer_set& finalizer_set::operator=(finalizer_set&&) = default;
+   finalizer_set& finalizer_set::operator=(finalizer_set&&) noexcept = default;
 
    auto finalizer_set::operator<=>(const finalizer_set&) const = default;
-
-
-   hs_finalizer_set_extension::hs_finalizer_set_extension(const finalizer_set& s)
-           : finalizer_set(s) {}
 
 } /// eosio::chain
--- a/libraries/chain/finalizer_set.cpp
+++ b/libraries/chain/finalizer_set.cpp
@@ -1,0 +1,27 @@
+#include <eosio/chain/finalizer_set.hpp>
+#include <eosio/chain/finalizer_authority.hpp>
+#include <fc/crypto/bls_public_key.hpp>
+
+namespace eosio::chain {
+
+   /**
+    * These definitions are all here to avoid including bls_public_key.hpp which includes <bls12-381/bls12-381.hpp>
+    * and pulls in bls12-381 types. This keeps bls12-381 out of libtester.
+    */
+
+   finalizer_set::finalizer_set() = default;
+   finalizer_set::~finalizer_set() = default;
+
+   finalizer_set::finalizer_set(const finalizer_set&) = default;
+   finalizer_set::finalizer_set(finalizer_set&&) = default;
+
+   finalizer_set& finalizer_set::operator=(const finalizer_set&) = default;
+   finalizer_set& finalizer_set::operator=(finalizer_set&&) = default;
+
+   auto finalizer_set::operator<=>(const finalizer_set&) const = default;
+
+
+   hs_finalizer_set_extension::hs_finalizer_set_extension(const finalizer_set& s)
+           : finalizer_set(s) {}
+
+} /// eosio::chain
--- a/libraries/chain/include/eosio/chain/config.hpp
+++ b/libraries/chain/include/eosio/chain/config.hpp
@@ -133,7 +133,8 @@ static_assert(maximum_tracked_dpos_confirmations >= ((max_producers * 2 / 3) + 1
 /**
  * Maximum number of finalizers in the finalizer set
  */
-const static int max_finalizers = max_producers;
+const static size_t max_finalizers = 64*1024;
+const static size_t max_finalizer_description = 256;
 
 /**
  * The number of blocks produced per round is based upon all producers having a chance
--- a/libraries/chain/include/eosio/chain/config.hpp
+++ b/libraries/chain/include/eosio/chain/config.hpp
@@ -134,7 +134,7 @@ static_assert(maximum_tracked_dpos_confirmations >= ((max_producers * 2 / 3) + 1
  * Maximum number of finalizers in the finalizer set
  */
 const static size_t max_finalizers = 64*1024;
-const static size_t max_finalizer_description = 256;
+const static size_t max_finalizer_description_size = 256;
 
 /**
  * The number of blocks produced per round is based upon all producers having a chance
--- a/libraries/chain/include/eosio/chain/finalizer_set.hpp
+++ b/libraries/chain/include/eosio/chain/finalizer_set.hpp
@@ -1,59 +1,49 @@
 #pragma once
 
-#include <eosio/chain/config.hpp>
 #include <eosio/chain/types.hpp>
-#include <chainbase/chainbase.hpp>
-#include <eosio/chain/authority.hpp>
-#include <eosio/chain/snapshot.hpp>
-
-#include <fc/crypto/bls_public_key.hpp>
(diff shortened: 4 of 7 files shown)


libraries/chain/webassembly/privileged.cpp
```diff
@@ -155,25 +156,27 @@ namespace eosio { namespace chain { namespace webassembly {
...
       std::set<fc::crypto::blslib::bls_public_key> unique_finalizer_keys;
-#warning REVIEW: Is checking for unique finalizer descriptions at all relevant?
-      std::set<std::string> unique_finalizers;
       uint64_t f_weight_sum = 0;
 
       for (const auto& f: finalizers) {
          f_weight_sum += f.fweight;
          unique_finalizer_keys.insert(f.public_key);
-         unique_finalizers.insert(f.description);
+         EOS_ASSERT( f.description.size() <= config::max_finalizer_description, wasm_execution_error, "Finalizer description greater than 256" );
```
linh2931, line 177: You can move this assert to the beginning of this loop body (before `f_weight_sum += f.fweight;`)

libraries/chain/controller.cpp
```diff
@@ -1980,6 +1980,7 @@ struct controller_impl {
    void set_finalizers_impl(const finalizer_set& fin_set) {
       // TODO store in chainbase
       current_finalizer_set = fin_set;
+      ++current_finalizer_set.generation;
```
linh2931, line 1983: This method sets finalizer set and increments the generation. Should its name be changed to reflect the two tasks?

heifner: This will be going away in the next PR.

libraries/chain/webassembly/privileged.cpp
heifner: Done

libraries/chain/finalizer_set.cpp
```diff
@@ -0,0 +1,27 @@
...
+   /**
+    * These definitions are all here to avoid including bls_public_key.hpp which includes <bls12-381/bls12-381.hpp>
+    * and pulls in bls12-381 types. This keeps bls12-381 out of libtester.
+    */
+
+   finalizer_set::finalizer_set() = default;
+   finalizer_set::~finalizer_set() = default;
+
+   finalizer_set::finalizer_set(const finalizer_set&) = default;
+   finalizer_set::finalizer_set(finalizer_set&&) = default;
```
greg7mdp, line 16: move copy constructors and assignment operators should be marked `noexcept` (here and in the header as well)

libraries/chain/include/eosio/chain/config.hpp
```diff
@@ -133,7 +133,8 @@ static_assert(maximum_tracked_dpos_confirmations >= ((max_producers * 2 / 3) + 1
 /**
  * Maximum number of finalizers in the finalizer set
  */
-const static int max_finalizers = max_producers;
+const static size_t max_finalizers = 64*1024;
+const static size_t max_finalizer_description = 256;
```
greg7mdp, line 137: detail: I think I would have named it `finalizer_description_max_size`

libraries/chain/include/eosio/chain/finalizer_set.hpp
```diff
@@ -1,59 +1,49 @@
...
+
+      finalizer_set& operator=(const finalizer_set&);
+      finalizer_set& operator=(finalizer_set&&);
 
-      std::string  description;
-      uint64_t     fweight = 0; // weight that this finalizer's vote has for meeting fthreshold
-      fc::crypto::blslib::bls_public_key  public_key;
+      uint32_t                                       generation = 0; ///< sequentially incrementing version number
+      uint64_t                                       fthreshold = 0;  ///< vote fweight threshold to finalize blocks
+      std::vector<finalizer_authority>               finalizers; ///< Instant Finality voter set
```
greg7mdp, line 21: maybe the indentation is a little excessive?

libraries/chain/include/eosio/chain/finalizer_set.hpp
```diff
@@ -1,59 +1,49 @@
...
 
-      friend bool operator == ( const finalizer_authority& lhs, const finalizer_authority& rhs ) {
-         return tie( lhs.description, lhs.fweight, lhs.public_key ) == tie( rhs.description, rhs.fweight, rhs.public_key );
-      }
-      friend bool operator != ( const finalizer_authority& lhs, const finalizer_authority& rhs ) {
-         return !(lhs == rhs);
-      }
+      auto operator<=>(const finalizer_set&) const;
```
greg7mdp, line 23: move up with other functions.

libraries/chain/include/eosio/chain/finalizer_set.hpp
```diff
@@ -1,59 +1,49 @@
...
+      static constexpr bool     enforce_unique() { return true; }
+
+      hs_finalizer_set_extension() = default;
+      hs_finalizer_set_extension(const hs_finalizer_set_extension&) = default;
+      hs_finalizer_set_extension( hs_finalizer_set_extension&& ) = default;
+
+      hs_finalizer_set_extension& operator=(const hs_finalizer_set_extension&) = default;
+      hs_finalizer_set_extension& operator=(hs_finalizer_set_extension&&) = default;
+
+      hs_finalizer_set_extension(const finalizer_set& s);
```
greg7mdp, line 43: I think we should remove all these and the implementation of `  hs_finalizer_set_extension(const finalizer_set& s)`

libraries/chain/webassembly/privileged.cpp
```diff
@@ -152,28 +153,30 @@ namespace eosio { namespace chain { namespace webassembly {
    }
 
    void interface::set_finalizers(span<const char> packed_finalizer_set) {
```
greg7mdp, line 166: shouldn't we use `std::span` instead of just `span`? Same with `vector` below.

libraries/chain/webassembly/privileged.cpp
```diff
@@ -152,28 +153,30 @@ namespace eosio { namespace chain { namespace webassembly {
...
       for (const auto& f: finalizers) {
+         EOS_ASSERT( f.description.size() <= config::max_finalizer_description, wasm_execution_error, "Finalizer description greater than 256" );
          f_weight_sum += f.fweight;
          unique_finalizer_keys.insert(f.public_key);
-         unique_finalizers.insert(f.description);
       }
 
-      EOS_ASSERT( finalizers.size() == unique_finalizers.size(), wasm_execution_error, "Duplicate finalizer description in finalizer set" );
       EOS_ASSERT( finalizers.size() == unique_finalizer_keys.size(), wasm_execution_error, "Duplicate finalizer bls key in finalizer set" );
       EOS_ASSERT( finset.fthreshold > f_weight_sum / 2, wasm_execution_error, "Finalizer set threshold cannot be met by finalizer weights" );
```
greg7mdp, line 192: why `/ 2`?

libraries/chain/webassembly/privileged.cpp
heifner: This not `std::span` but `using span = eosio::vm::span<T, Extent>;`

libraries/chain/finalizer_set.cpp
heifner: Oh you just meant `move constructor`, not `copy constructor`. Yep, agree.

libraries/chain/include/eosio/chain/config.hpp
heifner: Added `_size`

linh2931 (approved):

libraries/chain/webassembly/privileged.cpp
greg7mdp: Oh, missed that, thanks. Still `vector` should be std::vector, right?

libraries/chain/webassembly/privileged.cpp
greg7mdp: Well, the spec mentions 15, which is 2/3+1 of 21.  I wonder if we should not check that `threshold > (f_weight_sum / 3) * 2`?

libraries/chain/webassembly/privileged.cpp
heifner: yes, changed to `std::vector`

libraries/chain/webassembly/privileged.cpp
heifner: I agree, seems like it should be `threshold >= (f_weight_sum * 2 / 3) + 1`  or `threshold > (f_weight_sum * 2 / 3)` .

heifner: Conversation summary: The host function will enforce `f_weight_sum / 2`. It will be up to system contract to enforce a higher threshold if desired. 

> So if the chain would prefer to be more secure against finality violations at the cost of increased risk of losing liveness, it can increase the threshold above 1/2. Perhaps even all the way to 100% if it wanted.

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: CONSENSUS
summary: Modify `set_finalizer` host function structure.
Note:end

PR #1592, opened 2023-09-05
wrap `cfile`'s `fclose()` to avoid warning on latest glibc

spoonincode:
latest glibc adds a `__nonnull` attribute on `fclose()` which ends up causing a bunch of warnings,
```
libraries/libfc/include/fc/io/cfile.hpp:22:65: warning: ignoring attributes on template argument ‘int (*)(FILE*)’ [-Wignored-attributes]
   22 |    using unique_file = std::unique_ptr<FILE, decltype( &fclose )>;
```

Not really sure the best way of fixing this. Simply wrapping the call like this gets rid of the warning.

### DIFF
--- a/libraries/libfc/include/fc/io/cfile.hpp
+++ b/libraries/libfc/include/fc/io/cfile.hpp
@@ -19,7 +19,11 @@
 namespace fc {
 
 namespace detail {
-   using unique_file = std::unique_ptr<FILE, decltype( &fclose )>;
+   static void close_file(FILE* f) {
+      fclose(f);
+   }
+
+   using unique_file = std::unique_ptr<FILE, decltype( &close_file )>;
 }
 
 class cfile_datastream;
@@ -32,7 +36,7 @@ class cfile {
    friend class temp_cfile;
 public:
    cfile()
-     : _file(nullptr, &fclose)
+     : _file(nullptr, &detail::close_file)
    {}
 
    cfile(cfile&& other)
--- a/libraries/libfc/include/fc/io/cfile.hpp
+++ b/libraries/libfc/include/fc/io/cfile.hpp
@@ -19,7 +19,7 @@
 namespace fc {
 
 namespace detail {
-   static void close_file(FILE* f) {
+   inline void close_file(FILE* f) {
       fclose(f);
    }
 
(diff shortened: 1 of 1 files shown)


libraries/libfc/include/fc/io/cfile.hpp
```diff
@@ -19,7 +19,11 @@
 namespace fc {
 
 namespace detail {
-   using unique_file = std::unique_ptr<FILE, decltype( &fclose )>;
+   static void close_file(FILE* f) {
```
heifner, line 22: I believe `inline` would be better here than `static`.

greg7mdp:
Why not just replace the `unique_file` definition with:
`using unique_file = std::unique_ptr<FILE, decltype( [](FILE *f) noexcept { (void)fclose(f); } )>;`
and not provide `&fclose` when constructing a unique_file?

libraries/libfc/include/fc/io/cfile.hpp
spoonincode: ah yep

spoonincode:
> Why not just replace the `unique_file` definition with: `using unique_file = std::unique_ptr<FILE, decltype( [](FILE *f) noexcept { (void)fclose(f); } )>;` and not provide `&fclose` when constructing a unique_file?

That caused a different warning for me (that to be honest I didn't investigate)
```
libraries/libfc/include/fc/io/cfile.hpp:31:7: warning: ‘fc::cfile’ has a field ‘fc::detail::unique_file fc::cfile::_file’ whose type has internal linkage [-Wsubobject-linkage]
   31 | class cfile {
```

libraries/libfc/include/fc/io/cfile.hpp
```diff
@@ -19,7 +19,11 @@
 namespace fc {
 
 namespace detail {
-   using unique_file = std::unique_ptr<FILE, decltype( &fclose )>;
+   inline void close_file(FILE* f) {
```
greg7mdp, line 22:
```suggestion
   inline void close_file(FILE* f) noexcept {
```

greg7mdp (approved):

heifner (approved):

Issue #1558, opened 2023-08-24
IF: Add new snapshot version to hold consensus hotstuff state


heifner:
Depends on #1523 and #2080.

We will store the active finalizer set in the snapshot. Note this will be a non-backwards compatible snapshot version. Leap 5.0 can import older versions of snapshots that it currently supports, but it can only export the new version of the snapshot (v7).

Remove `if not activateIF` from `ship_streamer_test.py` and uncomment `nodeos_snapshot_diff_if_test` in `tests/CMakeLists.txt`.

linh2931:
Will get the wiring ready first.

PR #1598, opened 2023-09-05
IF: set_finalizers updates

heifner:
- Add `finalizer_set` to `block_header` extension
- Add `last_proposed_finalizer_set_generation` to `block_header_state`
- Add `set_finalzers` host function to `eos-vm-oc/intrinsic_mapping`
- Add to/from variant tests for `bls_public_key`
- Add activate protocol feature instant finality host function `set_finalizers` unittest.

Note: snapshot tests are failing because of `block_header_state` addition of `last_proposed_finalizer_set_generation`. I think the snapshot test should be fixed as part of https://github.com/AntelopeIO/leap/issues/1558. I've commented it out and added a warning for now.


libraries/chain/webassembly/privileged.cpp
```diff
@@ -156,7 +156,7 @@ namespace eosio { namespace chain { namespace webassembly {
    struct abi_finalizer_authority {
       std::string              description;
       uint64_t                 fweight = 0; // weight that this finalizer's vote has for meeting fthreshold
-      std::array<uint8_t, 144> public_key_g1_jacobian;
+      std::array<uint8_t, 96>  public_key_g1_affine_le;
```
greg7mdp, line 159: Shouldn't this also include the proof of possession?

greg7mdp (approved):

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -53,6 +54,7 @@ namespace detail {
       uint32_t                          dpos_proposed_irreversible_blocknum = 0;
       uint32_t                          dpos_irreversible_blocknum = 0;
       producer_authority_schedule       active_schedule;
+      uint32_t                          last_proposed_finalizer_set_generation = 0; // TODO: Add to snapshot_block_header_state_v3
```
linh2931, line 57: It would be easier to find this if a warning is added here.

libraries/chain/webassembly/privileged.cpp
```diff
@@ -183,7 +183,7 @@ namespace eosio { namespace chain { namespace webassembly {
          EOS_ASSERT( f.description.size() <= config::max_finalizer_description_size, wasm_execution_error,
                      "Finalizer description greater than ${s}", ("s", config::max_finalizer_description_size) );
          f_weight_sum += f.fweight;
-         std::optional<bls12_381::g1> pk = bls12_381::g1::fromJacobianBytesLE(f.public_key_g1_jacobian);
+         std::optional<bls12_381::g1> pk = bls12_381::g1::fromAffineBytesLE(f.public_key_g1_affine_le);
```
linh2931, line 186: The default `raw` parameter for `romAffineBytesLE` is false. You need to set it to true as we are using raw format.

heifner: What does `check` do?

libraries/chain/webassembly/privileged.cpp
heifner: Proof of possession will be done in contract. It will not be enforced in consensus.

ericpassmore:
Note:start
group: IF
category: CONSENSUS
summary: Updated block header state adding `last_proposed_finalizer_set_generation` and adding `finalizer_set` to block header extension. Created host function and protocol feature `set_finalizers`. 
Note:end

Issue #1277, opened 2023-06-13
Migrate mod_exp to BoringSSL, removing GMP requirement


bhazzard:
Details to be filled in by @spoonincode

bhazzard:
Unit tests could potentially be a separate PR.

BenjaminGormanPMP:
Issue moved from iteration to backlog. Per @spoonincode - this issue is dependent on https://github.com/AntelopeIO/leap/issues/644.

spoonincode:
> Wondering if [Boost multiprecision](https://www.boost.org/doc/libs/1_83_0/libs/multiprecision/doc/html/index.html) could be used?

https://www.boost.org/doc/libs/1_83_0/libs/multiprecision/doc/html/boost_multiprecision/perf/integer_performance.html
table 1.71 makes it look like the non-GMP impls Boost MP supports are slower than GMP too

greg7mdp:
Ha, that's right. Not quite as bad as your `Boring` results though. Not sure how critical this is.

PR #1233, opened 2023-06-05
Replace OpenSSL usage with BoringSSL

vladtr:
This PR replaces OpenSSL usage with BoringSSL in leap. This is based on the previous effort https://github.com/EOSIO/eos/pull/9026 and described in detail here - https://github.com/AntelopeIO/leap/issues/644 including all technical challenges of replacement.

Integration of BoringSSL in this PR following same approach as Chromium https://source.chromium.org/chromium/chromium/src/+/main:third_party/boringssl/ and based on a python script that generates all the build structure (generate_build_files.py). 

We had to patch build environment generation script to include missing dependencies for leap and detailed description of what needs to be done to upgrade BoringSSL in future is described in readme: https://github.com/AntelopeIO/leap/blob/5d1ce58cdd79eb96a2e83c98b8c27c1155273c80/libraries/libfc/third_party/boringssl/README.md

All tests seems to pass, but this is still in draft for now pending additional testing

Following, benchmarks of BoringSSL vs OpenSSL (2970WX CPU):

```
-BORINGSSL---------------------------------------------------------------------------

function                    runs        average       minimum       maximum

alt_bn_128:
alt_bn128_add               1,000      25,773 ns      25,257 ns      36,098 ns
alt_bn128_mul               1,000     123,121 ns     112,201 ns     241,455 ns
alt_bn128_pair (1 pair)     1,000   2,963,966 ns   2,941,270 ns   3,686,675 ns
alt_bn128_pair (10 pairs)   1,000  11,161,420 ns  11,028,624 ns  15,951,337 ns

blake2:
blake2                      1,000         368 ns         340 ns       6,513 ns

hash:
sha1 (64 bytes)             1,000         100 ns          90 ns       8,627 ns
sha1 (4096 bytes)           1,000       1,947 ns       1,913 ns       2,415 ns
sha256 (64 bytes)           1,000          98 ns          90 ns         701 ns
sha256 (4096 bytes)         1,000       2,077 ns       2,033 ns       6,693 ns
sha512 (64 bytes)           1,000         251 ns         240 ns       5,440 ns
sha512 (4096 bytes)         1,000       6,977 ns       6,923 ns      10,219 ns
ripemd160 (64 bytes)        1,000         298 ns         290 ns       4,809 ns
ripemd160 (4096 bytes)      1,000       8,299 ns       8,215 ns      11,873 ns
sha3-256 (64 bytes)         1,000         777 ns         741 ns       1,904 ns
[...]

PR #1599, opened 2023-09-06
Replace OpenSSL dependency with pinned BoringSSL

spoonincode:
Historically leap has used OpenSSL for TLS support and various crypto primitives (such as SHA256 and R1 key math). TLS support was removed a while ago leaving just the crypto primitives. Still, even with the rather limited usage, we'd like to avoid depending on system provided libraries for consensus bits, or at least "mathy" consensus bits, as these libraries may unexpectedly change some nuance out from under us resulting in consensus failure. A good example of such a change occurred in openssl 3.0.4 (that did _not_ affect Leap).

Because we no longer directly depend on a TLS library as part of our build, we can pin these crypto primitives to a specific point in time without fear of TLS security flaws in the future. Using BoringSSL as a pinned drop in replacement appears to be a great fit. BoringSSL is designed to be simply submoduled in and static linked against. Also, BoringSSL is a fork of OpenSSL so its behavior and API is likely to be the same as existing OpenSSL usage making it less risky than wholesale replacement and refactor with a different library.

This PR builds on #1233 with some changes, notably:

We now maintain our own fork of BoringSSL. While using upstream directly is nice, the trade off of that approach is that it's harder to audit the changes to the Python generation script. Since #1277 + #1255 will require modifying BoringSSL anyways, the fork seems inevitable. [Keep in mind that](https://boringssl.googlesource.com/boringssl/+/HEAD/INCORPORATING.md#which-branch-to-use),
> BoringSSL usage typically follows a “live at head” model. Projects pin to whatever the current latest of BoringSSL is at the time of update, and regularly update it to pick up new changes.
> 
> While the BoringSSL repository may contain project-specific branches, e.g. chromium-2214, those are _not_ supported release branches and must not as such.

So we are simply using yesterday's HEAD for the fork.

While BoringSSL really can simply be "dropped in" as a submodule that comes with a catch that a recent version of Go is required to build. We don't want that. Thus we end up with something I consider really ugly: committing post processed files in to our repository. We will simply have to tolerate this, [as it's just how BoringSSL works](https://boringssl.googlesource.com/boringssl/+/HEAD/INCORPORATING.md#build-support),

> Generally one checks in these generated files alongside the hand-written build files. Periodically an engineer updates the BoringSSL revision, regenerates these files and checks in the updated result. As an example, see how this is done in Chromium.

To make it worse, the generated files must be placed directly outside of the `src` submodule directory. I would have liked to place these files in to a `generated` directory to more cleanly delineate the generated files from non-generated files, but alas without hacking up the `generated_build_files.py` script more that doesn't seem possible.

I've tried to mitigate this ugliness a little by adding a small test in the workflow that regenerates the intermediate files and will fail if the files don't match what is committed. An example of failure can be found, https://github.com/AntelopeIO/leap/actions/runs/6091633748/job/16528533086

Also, as a reviewer note, the first commit in the PR contains the changes before running the `make regenerate_boringssl` target, and the third commit contains those generated files added to the repo. Ultimately the changes outside that third commit are quite minimal.


libraries/libfc/src/crypto/elliptic_common.cpp
```diff
@@ -94,8 +94,8 @@ namespace fc { namespace ecc {
             FC_ASSERT( EC_GROUP_get_order( group, order, ctx ) );
             BN_rshift1( order, order );
             private_key_secret bin;
-            FC_ASSERT( BN_num_bytes( order ) == static_cast<int>(bin.data_size()) );
-            FC_ASSERT( BN_bn2bin( order, (unsigned char*) bin.data() ) == static_cast<int>(bin.data_size()) );
+            FC_ASSERT( BN_num_bytes( order ) == bin.data_size() );
+            FC_ASSERT( BN_bn2bin( order, (unsigned char*) bin.data() ) == bin.data_size() );
```
spoonincode, line 98: These are a good example of while BoringSSL is a drop in replacement, it isn't without slight differences. A warning was introduced here because BoringSSL returns a `size_t` while OpenSSL returned an `int`.

libraries/libfc/src/crypto/elliptic_secp256k1.cpp
```diff
@@ -79,6 +81,14 @@ namespace fc { namespace ecc {
...
     }
 
+    private_key private_key::generate()
+    {
+       private_key ret;
+       do {
+         RAND_bytes((uint8_t*)ret.my->_key.data(), sizeof(ret.my->_key.data()));
+       } while(!secp256k1_ec_seckey_verify(detail::_get_context(), (const uint8_t*)ret.my->_key.data()));
+       return ret;
+    }
```
spoonincode, line 91: This change requires special attention during review. BoringSSL does not support the K1 curve. We've previously transitioned to using libsecp256k1 for all things K1... except for private key generation as performed here.

`_key` is a `private_key_secret`,
https://github.com/AntelopeIO/leap/blob/409eb5960c09b8b539984d2f098cf746be48b640/libraries/libfc/src/crypto/_elliptic_impl_priv.hpp#L22

and `private_key_secret` is a `fc::sha256`,
https://github.com/AntelopeIO/leap/blob/409eb5960c09b8b539984d2f098cf746be48b640/libraries/libfc/include/fc/crypto/elliptic.hpp#L23

The original code in this PR is probably incorrect and does not generate 32 random bytes, which would be a disaster.

libraries/CMakeLists.txt
```diff
@@ -7,6 +7,8 @@ set(BN256_INSTALL_COMPONENT "dev")
 
 set( Boost_USE_MULTITHREADED      ON )
 set( Boost_USE_STATIC_LIBS ON CACHE STRING "ON or OFF" )
+# don't include boost mysql library as it does a find_package(OpenSSL) thus finding the system openssl which could conflict with the bundled boringssl
+set( BOOST_EXCLUDE_LIBRARIES "mysql" )
```
spoonincode, line 11: This is something we'll need to be mindful of on boost bumps (or really any bundled dependency).

spoonincode:
That seems reasonable, let me play with that and see if there are any downsides.

spoonincode:
@greg7mdp I'm not sure how to best accomplish that. The primary problem here is that the `generate_build_files.py` is very inflexible with paths: the boringssl source directory (which is the submodule here) _must_ be at `src`, and the generated files _must_ be in the parent directory of `src`. That means placing the generated files in the boringssl submodule would require moving all the files around; that's not good.

Obviously we could change `generate_build_files.py` to be more flexible, but I like keeping the changes as minimal as possible (maybe they can even be upstreamed at some point).

One possibility is to utilize two submodules. We keep the close-to-upstream boringssl as it is now, but then have a new "`boringssl-generated`" repo that contains the generated files with the `src` directory pointing to the `boringssl` repo. Then leap only includes that "`boringssl-generated`" repo as a submodule. This certainly has some positives, but I'm unsure how others feel about that approach?

greg7mdp:
> One possibility is to utilize two submodules.

This is a little awkward, but I think this is preferable to adding all the generated files to leap itself, and I can't think of a better way to do it. I'm curious to hear what others think.

linh2931:
> One possibility is to utilize two submodules. We keep the close-to-upstream boringssl as it is now, but then have a new "`boringssl-generated`" repo that contains the generated files with the `src` directory pointing to the `boringssl` repo. Then leap only includes that "`boringssl-generated`" repo as a submodule.

I think main goals would be simple to upgrade in the future and fast to check out.  This approach seems to achieve those.

spoonincode:
Okay, I've migrated the generated build files to
https://github.com/AntelopeIO/boringssl-build/
There is a [workflow there](https://github.com/AntelopeIO/boringssl-build/blob/failure_example/.github/workflows/testgen.yaml) that checks the committed files match. Notice though that I changed the mechanism to `git ls-files`. A failure example was committed in the `failure_example` branch, https://github.com/AntelopeIO/boringssl-build/actions/runs/6269480194/job/17025961921

greg7mdp (approved):

linh2931 (approved):

greg7mdp (approved):

Issue #1596, opened 2023-09-05
Use raw Affine little-endian form of g1/g2 for encoding to BLS public key / signature


arhag:
Example for writing out bytes for a public key:
```
g1 pk = ...;
array<uint8_t, 96> msg = pk.toAffineBytesLE(true);
```

Something similar needed signatures.

linh2931:
Resolved by https://github.com/AntelopeIO/leap/pull/1600.

PR #1600, opened 2023-09-06
use raw Affine little-endian form of g1/g2 for encoding to BLS public key and signature

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/1596.

Use raw Affine little-endian format for BLS public key and signature.


greg7mdp (approved):

heifner (changes requested):
Need to update the from string as well (parse base64). Will also need to update tests.

linh2931:
> Need to update the from string as well (parse base64). Will also need to update tests.

Thanks! I forgot that. Will do.

heifner (approved):

PR #1606, opened 2023-09-07
IF: Check if input strings are invalid in conversion to public key or signature

linh2931:
When a string is converted to a public key or signature, check if the string is invalid.

### DIFF
--- a/libraries/libfc/src/crypto/bls_signature.cpp
+++ b/libraries/libfc/src/crypto/bls_signature.cpp
@@ -17,7 +17,7 @@ namespace fc::crypto::blslib {
 
          std::array<uint8_t, 192> bytes = fc::crypto::blslib::deserialize_base64<std::array<uint8_t, 192>>(data_str);
 
-         constexpr bool check = false; // default
+         constexpr bool check = true; // check if base64str is invalid
          constexpr bool raw = true;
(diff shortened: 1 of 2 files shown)


heifner (approved):
It would be nice to have a test that triggers this failure. I looked around some but couldn't find any examples. Maybe @mschoenebeck could provide an example.

libraries/libfc/src/crypto/bls_signature.cpp
```diff
@@ -17,7 +17,7 @@ namespace fc::crypto::blslib {
 
          std::array<uint8_t, 192> bytes = fc::crypto::blslib::deserialize_base64<std::array<uint8_t, 192>>(data_str);
 
-         constexpr bool check = false; // default
+         constexpr bool check = true; // check if base64str is invalid
          constexpr bool raw = true;
          std::optional<bls12_381::g2> g2 = bls12_381::g2::fromAffineBytesLE(bytes, check, raw);
```
greg7mdp, line 22: Not related to this PR, but I don't think the APIs from bls12_381 are well designed. These bool parameters could be replaced by enum template parameters which would be clearer and more efficient.

ericpassmore:
Note:start
group: IF
category: CRYPTO
summary: Add a check to make sure key or signature is valid when converted from a string.
Note:end

Issue #1257, opened 2023-06-08
Increase default read-only-threads


bhazzard:
Currently 8 threads.  Increase to 128 threads.

PR #1611, opened 2023-09-08
Increase maximum supported number of read-only threads to 128

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/1257

A number of PRs have been created to make it possible to increase the maximum allowed number of read-only threads to `128` (see tasks in https://github.com/eosnetworkfoundation/product/issues/149).

Specifically, 
* https://github.com/AntelopeIO/leap/pull/1488 reserves `10` pages (`11` slices) for each read-only thread while keeps `528` pages (`529` slices) for the main-thread memory. With maximum 128 read-only threads , virtual memory required by OC is `15TB` (OC's main thread uses `4TB` VM (by `529` slices) and the read-only threads use `11TB` (`128 * 11 * 8GB`)). It is about `11.7%` of total VM space in a 64-bit Linux machine (about `128TB`).

* https://github.com/AntelopeIO/leap/pull/1549 removes the need of each read-only threads maintaining a separate WASM instantiated module cache. All threads (including the main thread) shares the same cache. This ensures memory used by WASM instantiated module cache does not grow as the number of read-only threads increases.


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -617,9 +617,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    };
 
    uint32_t _ro_thread_pool_size{0};
-   // Due to uncertainty to get total virtual memory size on a 5-level paging system for eos-vm-oc and
-   // possible memory exhuastion for large number of contract usage for non-eos-vm-oc, set a hard limit
-   static constexpr uint32_t         _ro_max_threads_allowed{8};
+   static constexpr uint32_t         _ro_max_threads_allowed{128};
```
heifner, line 620: Can you add the info from the PR description as a comment here:
```
10 pages (11 slices) for each read-only thread keeps 528 pages (529 slices) for the main-thread memory. With maximum 128 read-only threads , virtual memory required by OC is 15TB (OC's main thread uses 4TB VM (by 529 slices) and the read-only threads use 11TB (128 * 11 * 8GB)). It is about 11.7% of total VM space in a 64-bit Linux machine (about 128TB).
```

heifner (changes requested):

heifner (approved):

spoonincode (approved):

Issue #1586, opened 2023-08-31
IF: Add block extension to include latest commit QC known to proposer


arhag:
Proposer would take the latest irreversible HotStuff proofs (the latest block proposal with the `final_on_qc` pointing to an irreversible block along with QC for it) known to it and include it in a new block extension for the block that it constructs. This would only be the _irreversible_ HotStuff proofs which means they would always be relevant for advancing LIB on the irreversible blockchain.

heifner:
Add to block extension `signed_block.block_extensions` the latest vector of `[hs_proposal_message, quorum_certificate_message]` for a committed hs_proposal. This irreversible hs proof will be used during syncing to advance LIB.

The accepted block signal can be processed by `qc_chain` to move LIB if they are relevant. They will not be relevant if you are not syncing as they will likely be behind current. (Do we need to optimize this or can `qc_chain` quickly tell?). Either way `qc_chain` will advance LIB as it normally would, which is still to be determined #1522. 

On finalize of pending block the `controller` will request all commitments greater than its current from `qc_chain`. `qc_chain` is required to store them until LIB. The `controller` will place them in the block extension and also in pending block state. They should be "pulled forward" by pending block state so any unwinding of blocks also unwinds the known and reported committed values. Note this is needed so that we do not need to repeat commitment values in block extension. It is the responsibility of `qc_chain` to provide us all relevant commitment pairs from current commitment pairs.

heifner:
1. The `block_header` extension is included in the digest that is the `block_id`. It needs to be information that can be deterministically verified. The QC info in the block extension is not deterministic. That is fine for a `block` extension, but not for a `block_header` extension.

2. You need the one at time of `finalizer_set` change as well as periodically.

greg7mdp:
> It needs to be information that can be deterministically verified.

I'm not quite understanding why the QC is not deterministic, while the `hs_finalizer_set_extension` that we include in `block_header` extensions is?

> You need the one at time of finalizer_set change as well as periodically.

Why? I thought that as long as we knew what `finalizer_set` was active when the QC was signed, that would be enough to verify the QC and then advance the LIB.

heifner:
> > It needs to be information that can be deterministically verified.
> 
> I'm not quite understanding why the QC is not deterministic, while the `hs_finalizer_set_extension` that we include in `block_header` extensions is?

The QC that any individual node has at any given time is according to network conditions; it is not deterministic on all nodes.

[...]

PR #1612, opened 2023-09-08
IF: Add hs_irreversible_blocknum to block_header_state

heifner:
- Add `hs_irreversible_blocknum` to `block_header_state` to track hotstuff irreversible block number.
- Update `fork_database` to use either `hs_irreversible_blocknum` or `dpos_irreversible_blocknum` for LIB.
- Add `mark_irreversible` to `fork_database` to be used by `chain_pacemaker` to advance LIB which will then be used by `controller` `log_irreversible` to advance LIB.
- Added `copyable_atomic` type for use by `block_header_state` `hs_irreversible_blocknum` since it will be modified after the `block_state` is signaled and potentially accessed from other threads.
- Added `fc::atomic_shared_ptr` and `copyable_atomic_shared_ptr` these may not actually be needed. But could be used if we wish to store active `finalizer_set` in `block_state`.

Need to add tests for the `fork_database::mark_irreversible` but getting this PR up so it can be used by #1586


heifner:
Closing. Going to explore not storing in block_header_state.

Issue #1548, opened 2023-08-23
IF: propagate hs_*_messages


heifner:
Need to support configuration where BPs are not directly connected. For example: `BP <-> relay-node <-> relay-node <-> BP`.
Also need to prevent infinite message loop. This is accomplished for trx and blocks by keep track of what messages have been received on each connection and not echoing back the message to that connection.

arhag:
Changes may instead need to be outside of net plugin, e.g. changes to qc_chain to support an qc tree analogous to fork database for reversible blocks. This would be needed to detect duplicates to drop them on the floor and avoid the infinite message loop problem.

fcecin:
I'm thinking that we want the rebroadcasting to be encapsulated inside of the IF engine, since it is already tasked with calling "broadcast hs message (x)" as a side-effect of message processing. So the IF engine will also make the decision to disseminate (route) the message to the network during the processing of each incoming message.

Here's my current thinking about what these decisions would look like, for each message type:

- New View: rebroadcast if a message triggers a local view change (is this right?);
- New Block: rebroadcast if local qc_chain didn't know about that block ID (is this right?);
- New Proposal: rebroadcast if qc_chain does not considerthe incoming proposal a repetition, irrelevant or obsolete (how exactly?)
- New Vote: rebroacast if... (??)

This is in addition to the broadcast-this-new-message requests that are already inside qc_chain. That would mean qc_chain is also a gossip router now.

A free optimization can be achieved by not ever rebroadcasting a message to the original sender. This can be achieved by passing an identifier of the sender to the IF engine together with the incoming message notification, which can then pass that as an optional argument to the send/broadcast requests (i.e. do broadcast, but exclude this given destination/connection/node).

A further optimization then is the creation of a "hotstuff" connection/channel/subscription type, similar to what is being done to block connections, transaction connections, and the new peer addressbook/discovery feature (last time I checked, month+ ago). All these broadcasts are limited to those connections only, which would allow e.g. hotstuff messages to be disseminated inside of the network core, eventually reaching all infrastructure nodes (and any external/downstream nodes interested in collecting this information and which have a "hotstuff" connection to a core network node).

I am assuming that all of the hotstuff messages are going to be what we are seeing in `libraries/chain/include/eosio/chain/hotstuff.hpp`, i.e. small messages.

[...]

PR #1613, opened 2023-09-09
IF: Infrastructure for hs message filtering

### LINKED ISSUE #1605: IF: HS network message filtering infrastructure (net plugin <-> IF engine)
Add code for:

- Net plugin notifies the connection_id of incoming HS messages to IF engine (this will be done by an additional parameter in the existing callback function signature(s)).

- IF engine can (optionally) supply a connection_id to exclude, when passing an outgoing (broadcast) HS message to Net plugin; this is done whenever the IF engine will (in the future) decide to propagate a message, instead of sending an original message to broadcast.

- IF engine calls the Net plugin whenever it detects that it has received a potentially malicious message (duplicates, irrelevant messages, messages with invalid signatures, i.e. an enum "reason"...) from a given connection_id direct peer, allowing the Net plugin to disconnect that direct peer that it decides is misbehaving (for example, by counting how frequently each type of messaging offense happen in a counter that would be added to the corresponding connection object -- which could then e.g. "cool down" periodically, or maybe not).

fcecin:
- connection ID of hs message sender (net plugin -> IF engine)
- connection ID to exclude when propagating (IF engine -> net plugin)
- added callback to warn of bad HS messages from a connection (IF engine -> net plugin)
- added sample use of the warning callback for discarded proposals and votes
- helpers on qc_chain to allow for easy message propagation (no propagation decisions added)
- no-ops for this feature added to test_pacemaker (which does not have multi-hop)

Closes #1605

Obsoletes PR #1583, which can be closed as a message propagation strategy (for closing #1548) will be decided and then implemented on top of the code in this PR (suggestion https://github.com/AntelopeIO/leap/pull/1583/files/e1cc627a75853e2929df25cbcc3ee4f76066ba37#r1316210590 can be implemented when we remove the qc_chain profiler).

### DIFF
--- a/libraries/chain/include/eosio/chain/hotstuff.hpp
+++ b/libraries/chain/include/eosio/chain/hotstuff.hpp
@@ -57,7 +57,7 @@ namespace eosio::chain {
 
    using hs_message = std::variant<hs_vote_message, hs_proposal_message, hs_new_block_message, hs_new_view_message>;
 
-   enum hs_message_warning : uint32_t {
+   enum class hs_message_warning {
       discarded,               // default code for dropped messages (irrelevant, redundant, ...)
       duplicate_signature,     // same message signature already seen
       invalid_signature,       // invalid message signature
--- a/libraries/chain/include/eosio/chain/hotstuff.hpp
+++ b/libraries/chain/include/eosio/chain/hotstuff.hpp
@@ -57,6 +57,13 @@ namespace eosio::chain {
 
    using hs_message = std::variant<hs_vote_message, hs_proposal_message, hs_new_block_message, hs_new_view_message>;
 
+   enum hs_message_warning : uint32_t {
+      discarded,               // default code for dropped messages (irrelevant, redundant, ...)
+      duplicate_signature,     // same message signature already seen
+      invalid_signature,       // invalid message signature
+      invalid                  // invalid message (other reason)
+   };
+
    struct finalizer_state {
       bool chained_mode = false;
       fc::sha256 b_leaf;
--- a/libraries/hotstuff/chain_pacemaker.cpp
+++ b/libraries/hotstuff/chain_pacemaker.cpp
@@ -120,13 +120,13 @@ namespace eosio { namespace hotstuff {
 
    void chain_pacemaker::register_bcast_function(std::function<void(const std::optional<uint32_t>&, const chain::hs_message&)> broadcast_hs_message) {
       FC_ASSERT(broadcast_hs_message, "on_hs_message must be provided");
-      std::lock_guard g( _hotstuff_global_mutex ); // not actually needed but doesn't hurt
+      // no need to std::lock_guard g( _hotstuff_global_mutex ); here since pre-comm init
       bcast_hs_message = std::move(broadcast_hs_message);
    }
 
    void chain_pacemaker::register_warn_function(std::function<void(const uint32_t, const chain::hs_message_warning&)> warning_hs_message) {
       FC_ASSERT(warning_hs_message, "must provide callback");
-      std::lock_guard g( _hotstuff_global_mutex ); // not actually needed but doesn't hurt
+      // no need to std::lock_guard g( _hotstuff_global_mutex ); here since pre-comm init
       warn_hs_message = std::move(warning_hs_message);
    }
 
--- a/libraries/hotstuff/chain_pacemaker.cpp
+++ b/libraries/hotstuff/chain_pacemaker.cpp
@@ -118,10 +118,16 @@ namespace eosio { namespace hotstuff {
       _head_block_state = chain->head_block_state();
    }
 
-   void chain_pacemaker::register_bcast_function(std::function<void(const chain::hs_message&)> broadcast_hs_message) {
+   void chain_pacemaker::register_bcast_function(std::function<void(const std::optional<uint32_t>&, const chain::hs_message&)> broadcast_hs_message) {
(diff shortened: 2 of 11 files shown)


libraries/chain/include/eosio/chain/hotstuff.hpp
```diff
@@ -57,6 +57,13 @@ namespace eosio::chain {
 
    using hs_message = std::variant<hs_vote_message, hs_proposal_message, hs_new_block_message, hs_new_view_message>;
 
+   enum hs_message_warning : uint32_t {
```
heifner, line 60: Make an `enum class`.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -142,10 +142,11 @@ namespace eosio::hotstuff {
...
 
-      void process_proposal(const hs_proposal_message& msg); //handles proposal
-      void process_vote(const hs_vote_message& msg); //handles vote
-      void process_new_view(const hs_new_view_message& msg); //handles new view
-      void process_new_block(const hs_new_block_message& msg); //handles new block
+      // is_loopback is used to check if the call is processing a self-receipt message (if so, never propagate it)
+      void process_proposal(const hs_proposal_message& msg, bool is_loopback = false); //handles proposal
+      void process_vote(const hs_vote_message& msg, bool is_loopback = false); //handles vote
+      void process_new_view(const hs_new_view_message& msg, bool is_loopback = false); //handles new view
+      void process_new_block(const hs_new_block_message& msg, bool is_loopback = false); //handles new block
```
heifner, line 149: I would prefer to be explicit and not have the default `= false`

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -159,10 +160,12 @@ namespace eosio::hotstuff {
...
       std::vector<hs_proposal_message> get_qc_chain(const fc::sha256& proposal_id); //get 3-phase proposal justification
 
-      void send_hs_proposal_msg(const hs_proposal_message& msg); //send vote msg
-      void send_hs_vote_msg(const hs_vote_message& msg); //send proposal msg
-      void send_hs_new_view_msg(const hs_new_view_message& msg); //send new view msg
-      void send_hs_new_block_msg(const hs_new_block_message& msg); //send new block msg
+      void send_hs_proposal_msg(const hs_proposal_message& msg, bool propagation = false); //send vote msg
+      void send_hs_vote_msg(const hs_vote_message& msg, bool propagation = false); //send proposal msg
+      void send_hs_new_view_msg(const hs_new_view_message& msg, bool propagation = false); //send new view msg
+      void send_hs_new_block_msg(const hs_new_block_message& msg, bool propagation = false); //send new block msg
```
heifner, line 166: I would prefer to be explicit and not have the default `= false`.
Also a couple of the comments are wrong and they are all completely useless as they do not tell you anything more than the method name.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -159,10 +160,12 @@ namespace eosio::hotstuff {
...
+
+      void send_hs_message_warning(const chain::hs_message_warning code = hs_message_warning::discarded); //use generic discard reason if none given
```
heifner, line 168: Remove default `= hs_message_warning::discarded`. Should be explicit about it.

libraries/hotstuff/qc_chain.cpp
```diff
@@ -630,6 +642,8 @@ namespace eosio::hotstuff {
    // Called from the main application thread
    void qc_chain::on_beat(){
 
+      _sender_connection_id = std::nullopt;
```
heifner, line 645: Storing the sender connection id seems like an anti-pattern to me. It provides a point of contention between the threads interacting with `qc_chain`. I know this is all protected by a large mutex, but if we want to remove that in the future seems much cleaner to pass along connection id via the call stack.

heifner (changes requested):

libraries/hotstuff/chain_pacemaker.cpp
```diff
@@ -118,12 +118,18 @@ namespace eosio { namespace hotstuff {
...
-   void chain_pacemaker::register_bcast_function(std::function<void(const chain::hs_message&)> broadcast_hs_message) {
+   void chain_pacemaker::register_bcast_function(std::function<void(const std::optional<uint32_t>&, const chain::hs_message&)> broadcast_hs_message) {
       FC_ASSERT(broadcast_hs_message, "on_hs_message must be provided");
       std::lock_guard g( _hotstuff_global_mutex ); // not actually needed but doesn't hurt
       bcast_hs_message = std::move(broadcast_hs_message);
    }
 
+   void chain_pacemaker::register_warn_function(std::function<void(const uint32_t, const chain::hs_message_warning&)> warning_hs_message) {
+      FC_ASSERT(warning_hs_message, "must provide callback");
+      std::lock_guard g( _hotstuff_global_mutex ); // not actually needed but doesn't hurt
```
linh2931, line 129: If it is not needed, remove it; but leave a comment explaining it is not required.

heifner (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: MESSAGES
summary: Filters hotstuff messages by accepting an optional list of peers to exclude from sending messages.
Note:end

Issue #1546, opened 2023-08-22
IF: net_plugin lib<->head catchup transistion change


heifner:
When lib is much closer to head, it will be very likely for a node to find its head is less than lib and think it needs to sync from a peer when in reality it is only a bit behind. Need a better heuristic for when to switch from in_sync<->head<->lib catchup modes. 

This might be as simple as a hard-coded delta from peer head before assuming it is syncing (or should sync), but will require some testing.

heifner:
- `sync_manager::recv_handshakerecv_handshake` - `head` < `msg.lib` is more likely to be true. The inverse is also more likely to be true. Both conditions sync 1 & sync 2 need to not trigger lib sync unless there is a distance between head<->lib of (latency + `sync_diff_range`). Where `sync_diff_range` could be hard-coded. But I believe setting it to half of `p2p-keepalive-interval-ms`  (default 10s) is a better option. `p2p-keepalive-interval-ms` makes sense as a value to tie it to because this corresponds to how long you wait before pinging a node if you have not heard from it. The default would be 5 seconds (10 blocks).

PR #1617, opened 2023-09-11
IF: prevent unnecessary lib<->head catchup transition changes

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/1546

Instant finality makes it likely peers think their lib and head are not in sync but in reality they are only within small difference. To avoid unnecessary catchups, a minimal distance between lib and head is introduced. `min_blocks_difference` must be reached before catchup starts.

`min_blocks_difference` is defined as the number of blocks produced during the half keep alive interval (how long a node waits before contacting its peer).

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -241,6 +241,12 @@ namespace eosio {
       std::atomic<stages> sync_state{in_sync};
       std::atomic<uint32_t> sync_ordinal{0};
 
+      // Instant finality makes it likely peers think their lib and head are
+      // not in sync but in reality they are only within small difference.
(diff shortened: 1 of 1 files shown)


heifner (changes requested):
I think we should target this to main. A private chain with a small number of block producers would hit similar situtation.

linh2931:
> I think we should target this to main. A private chain with a small number of block producers would hit similar situtation.

Rebasing to main brings in a large number of files and causes merge conflicts. I will create a fresh PR based on main directly and close this after that, so it is easier to see the changes.

linh2931:
Close this. Use https://github.com/AntelopeIO/leap/pull/1619 instead (which targets to `main`).

Issue #1525, opened 2023-08-17
IF: Generate BLS key pairs in leap-util


arhag:
Depends https://github.com/AntelopeIO/leap/issues/1584 and https://github.com/AntelopeIO/leap/issues/1596.

Ensure changes in https://github.com/AntelopeIO/bls12-381/issues/10 are brought into Leap as a submodule update.

There is no need for cleos to create a BLS key pair since BLS key pairs are only needed by some node operators (block producer candidates).

But the basic interface for creating key pairs for other key types (K1, R1) in cleos should be adopted in leap-util but specifically for creating a BLS key pair (there is no need for leap-util to generate K1 or R1 keys).

However, it should also generate a Proof of Possession (PoP) and include that signature in the output as well to act as a proof that the private key for the public key is known. This will be necessary for registering the public key as a finalizer key later in the system contracts. Use https://github.com/AntelopeIO/bls12-381/blob/d79a1836649d6f9b7494ccc3e9a1320a7470acaa/src/signatures.cpp#L630 to generate the PoP.

There should also be another sub-command to regenerate the PoP from a given private key. This will not only output the signature but also the public key corresponding to the provided private key.

linh2931:
Resolved by https://github.com/AntelopeIO/leap/pull/1594

arhag:
Reopening because https://github.com/AntelopeIO/leap/pull/1594 does not use `pop_prove` but should.

PR #1625, opened 2023-09-12
IF: Use bls12_381::pop_prove to generate POP and add get_pop_str method to libfc

linh2931:
Fixed https://github.com/AntelopeIO/leap/issues/1525 to use BLS library `pop_prove` to generate POP.
* Added `get_pop_str` to libfc `bls_private_key` using `pop_prove`;
* Modified leap-util to use `get_pop_str`;
* Updated the test.

### DIFF
--- a/libraries/libfc/include/fc/crypto/bls_private_key.hpp
+++ b/libraries/libfc/include/fc/crypto/bls_private_key.hpp
@@ -28,9 +28,7 @@ namespace fc::crypto::blslib {
 
          bls_public_key     get_public_key() const;
          bls_signature      sign( const std::vector<uint8_t>& message ) const;
-
-         // Returns proof of possession
-         bls_signature      pop_proof() const;
+         bls_signature      proof_of_possession() const;
 
          static bls_private_key generate();
 
--- a/libraries/libfc/include/fc/crypto/bls_private_key.hpp
+++ b/libraries/libfc/include/fc/crypto/bls_private_key.hpp
@@ -28,6 +28,7 @@ namespace fc::crypto::blslib {
 
          bls_public_key     get_public_key() const;
          bls_signature      sign( const std::vector<uint8_t>& message ) const;
+         std::string        get_pop_str() const;
 
          static bls_private_key generate();
 
--- a/libraries/libfc/include/fc/crypto/bls_private_key.hpp
+++ b/libraries/libfc/include/fc/crypto/bls_private_key.hpp
@@ -28,7 +28,9 @@ namespace fc::crypto::blslib {
 
          bls_public_key     get_public_key() const;
          bls_signature      sign( const std::vector<uint8_t>& message ) const;
-         std::string        get_pop_str() const;
+
+         // Returns proof of possession
+         bls_signature      pop_proof() const;
 
          static bls_private_key generate();
 
--- a/libraries/libfc/src/crypto/bls_private_key.cpp
+++ b/libraries/libfc/src/crypto/bls_private_key.cpp
@@ -13,7 +13,7 @@ namespace fc::crypto::blslib {
       return bls_public_key(pk);
    }
 
-   bls_signature bls_private_key::pop_proof() const
+   bls_signature bls_private_key::proof_of_possession() const
    {
       bls12_381::g2 proof = bls12_381::pop_prove(_sk);
       return bls_signature(proof);
--- a/libraries/libfc/src/crypto/bls_private_key.cpp
+++ b/libraries/libfc/src/crypto/bls_private_key.cpp
@@ -13,6 +13,14 @@ namespace fc::crypto::blslib {
       return bls_public_key(pk);
    }
 
+   std::string bls_private_key::get_pop_str() const
+   {
+      bls12_381::g2 proof = bls12_381::pop_prove(_sk);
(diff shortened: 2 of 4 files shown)


libraries/libfc/include/fc/crypto/bls_private_key.hpp
```diff
@@ -28,6 +28,7 @@ namespace fc::crypto::blslib {
 
          bls_public_key     get_public_key() const;
          bls_signature      sign( const std::vector<uint8_t>& message ) const;
+         std::string        get_pop_str() const;
```
heifner, line 31: Add comment or rename to `proof_of_possession_str()`. If adding comment also note this is in AffineLE format.

arhag, line 31: Instead of returning the base64 encoded string of the `pop_prove` g2, I think this should be called `pop_proof()` and it should return a `bls_signature`.

Then in leap-util you can convert it into the `SIG_BLS_` prefixed base64 encoded string with a checksum.

tests/leap_util_bls_test.py
```diff
@@ -100,13 +100,12 @@ def check_create_key_results(rslts):
...
     assert "PVT_BLS_" in results["Private key"]
     assert "PUB_BLS_" in results["Public key"]
-    assert "SIG_BLS_" in results["Proof of Possession"]
 
 def get_results(rslts):
     # sample output looks like
     # Private key: PVT_BLS_kRhJJ2MsM+/CddO...
     # Public key: PUB_BLS_lbUE8922wUfX0Iy5...
-    # Proof of Possession: SIG_BLS_olZfcFw...
+    # Proof of Possession: 3jwkVUUYahHgsnmnEA...
```
arhag, line 108: This would be `SIG_BLS_3jwkVUUYahHgsnmnEA...`.

arhag (changes requested):

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -13,6 +13,14 @@ namespace fc::crypto::blslib {
       return bls_public_key(pk);
    }
 
+   std::string bls_private_key::get_pop_str() const
+   {
+      bls12_381::g2 proof = bls12_381::pop_prove(_sk);
+      constexpr bool raw = true;
+      std::array<uint8_t, 192> bytes = proof.toAffineBytesLE(raw);
+      return fc::crypto::blslib::serialize_base64<std::array<uint8_t, 192>>(bytes);
```
dimas1185, line 21: I assume you need to add a `SIG_BLS_` prefix here

dimas1185:
```suggestion
      return fc::crypto::blslib::bls_signature(bls12_381::pop_prove(_sk)).to_string({});
```

linh2931: Thanks. Please see the latest commit.

dimas1185 (changes requested):
please include this update (should be in a main branch of bls):
https://github.com/AntelopeIO/bls12-381/pull/11

libraries/libfc/src/crypto/bls_private_key.cpp
```diff
@@ -13,6 +13,12 @@ namespace fc::crypto::blslib {
       return bls_public_key(pk);
    }
 
+   bls_signature bls_private_key::pop_proof() const
+   {
+      bls12_381::g2 proof = bls12_381::pop_prove(_sk);
```
arhag, line 18: As @dimas1185 mentioned, the bls12-381 submodule should be updated to use the following commit: https://github.com/AntelopeIO/bls12-381/commits/5b18ef4f8add17d17aa8fe9233aa363829aca90a. Then this call to `pop_prove` would generate the correct signature.

arhag (changes requested):

libraries/libfc/src/crypto/bls_private_key.cpp
arhag: Also, I think renaming it would be better to rename the function to either `proof_of_possession()` or, if we want to keep it short, `generate_pop()`. 

@heifner, it seems like you would prefer `proof_of_possession()`?

heifner: Yes, I think `pop` is not necessarily immediately clear.

dimas1185: `pop` says nothing to general public. I support using `proof_of_possession`

dimas1185: I also tested `leap-util` generated keys with requested bls commit and it works. I have changes that are ready for PR. Waiting for Lin to push this.

linh2931: I will change it to `proof_of_possession()` and update the submodule.

heifner (approved):

arhag (approved):

dimas1185 (approved):

ericpassmore:
Note:start
group: IF
category: MESSAGES
summary: Generate Proof of Possession Key with raw, affine, little-endian
Note:end

PR #1638, opened 2023-09-14
Refactor adding EosioChain INTERFACE library for use in  add_eosio_test_executable macro and by itself.

oschwaldp-oci:
This refactor creates an `INTERFACE` library called `EosioChain` which can be used externally to `leap` as a library for `target_link_libraries`.  In an repository that wants to depend on and link against `EosioChain`, it can now be done by simply:

```
find_package(eosio)
target_link_libraries( <myNewTarget> PRIVATE EosioChain ...)
```


heifner (approved):

spoonincode (changes requested):
How difficult would it be to refactor this cmake file to present maybe a `add_library(libtester STATIC IMPORTED)`. Then the existing `add_eosio_test_executable` would just link to the `IMPORTED libtester`, but when you're not making a test executable you would just `target_link_libraries(whatever libtester)`

Adding dependencies like this makes it difficult to know the union of all downstream dependencies when reviewing this file in the future.

oschwaldp-oci:
> How difficult would it be to refactor this cmake file to present maybe a `add_library(libtester STATIC IMPORTED)`. Then the existing `add_eosio_test_executable` would just link to the `IMPORTED libtester`, but when you're not making a test executable you would just `target_link_libraries(whatever libtester)`
> 
> Adding dependencies like this makes it difficult to know the union of all downstream dependencies when reviewing this file in the future.

@spoonincode -- How about going the direction of an `INTERFACE` library like you had also mentioned out of band earlier?  Proposal: [3e22bd1](https://github.com/AntelopeIO/leap/pull/1638/commits/3e22bd19882d61db43ed6e2c089526941c529f63)

spoonincode:
Yes true, `INTERFACE` is considerably easier than `IMPORTED` in this case since we have a handful of static libraries.

spoonincode (approved):

Issue #1556, opened 2023-08-24
DTrx: Remove deferred trxs configuration


bhazzard:
This is the solution for removing deferred transactions prior to activation of a protocol feature

remove the deferred transaction config option and hardcode the behavior as it was configured as:

```
max-scheduled-transaction-time-per-block-ms = 0
```

and block delayed transactions

PR #1643, opened 2023-09-15
Remove defer transaction configuration options

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/1556

* removed `--max-scheduled-transaction-time-per-block-ms`
* removed `--incoming-defer-ratio`
* simplified `process_scheduled_and_incoming_trxs`
* blocked delayed transactions
* removed member `delay` from `transaction_context` class
* added and updated tests
* modified documents
* other necessary changes


heifner (approved):
So happy to see this!

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2370,142 +2338,6 @@ bool producer_plugin_impl::process_unapplied_trxs(const fc::time_point& deadline
...
 
-void producer_plugin_impl::process_scheduled_and_incoming_trxs(const fc::time_point& deadline, unapplied_transaction_queue::iterator& itr) {
-   // scheduled transactions
-   int    num_applied         = 0;
-   int    num_failed          = 0;
-   int    num_processed       = 0;
-   bool   exhausted           = false;
-   double incoming_trx_weight = 0.0;
-
-   auto&              blacklist_by_id     = _blacklisted_transactions.get<by_id>();
```
arhag, line 2381: I think we should be able to also remove `_blacklisted_transactions` (and `remove_expired_blacklisted_trxs`) now.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2370,142 +2338,6 @@ bool producer_plugin_impl::process_unapplied_trxs(const fc::time_point& deadline
...
-      };
-
-      try {
-         auto             start        = fc::time_point::now();
-         auto             trx_tracker  = _time_tracker.start_trx(false, start); // delayed transaction cannot be transient
-         fc::microseconds max_trx_time = fc::milliseconds(_max_transaction_time_ms.load());
-         if (max_trx_time.count() < 0)
-            max_trx_time = fc::microseconds::maximum();
-
-         auto trace = chain.push_scheduled_transaction(trx_id, deadline, max_trx_time, 0, false);
```
arhag, line 2460: Unfortunately, we cannot completely get rid of calling `push_scheduled_transaction` yet. We still need to call it for expired transactions in `generated_transaction_multi_index` so that we can expire any ones still existing. We may even need to do that over several blocks if there is a large queue to work through.

But we can greatly simplify the logic in this function because we just need to go through the index `by_expiration` and pick out the ones that have expired to call `push_scheduled_transaction` with them. Additionally, we can provide a small hard-coded number of them to process at the beginning of the block before giving other transaction types their turn.

arhag (changes requested):

plugins/producer_plugin/producer_plugin.cpp
heifner: I would limit it to say a hardcoded 50ms deadline.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2370,71 +2302,21 @@ bool producer_plugin_impl::process_unapplied_trxs(const fc::time_point& deadline
...
-         if (pr.trx_exhausted) {
-            ++itr; // leave in incoming
-         } else {
-            itr = _unapplied_transactions.erase(itr);
-         }
-
-         if (exhausted)
-            break;
+   while (expired_itr != expired_idx.end()) {
+      if (expired_itr->expiration > pending_block_time) {
```
heifner, line 2318: Needs to be `>=`

heifner (changes requested):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2493,16 +2379,14 @@ void producer_plugin_impl::process_scheduled_and_incoming_trxs(const fc::time_po
...
-      if (sch_itr_next == sch_idx.end())
+      if (expired_itr_next == expired_idx.end())
          break;
-      sch_itr = sch_idx.lower_bound(boost::make_tuple(next_delay_until, next_id));
+      expired_itr = expired_idx.lower_bound(boost::make_tuple(next_expired, next_id));
    }
 
-   if (scheduled_trxs_size > 0) {
+   if (expired_size > 0) {
       fc_dlog(_log, "Processed ${m} of ${n} scheduled transactions, Applied ${applied}, Failed/Dropped ${failed}",
```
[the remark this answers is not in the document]
linh2931: I was as thinking about adding `expired`; then changed my mind on the ground that those were just scheduled transactions in general.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1995,14 +1957,9 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
...
-
-            auto scheduled_trx_deadline = preprocess_deadline;
-            if (_max_scheduled_transaction_time_per_block_ms >= 0) {
-               scheduled_trx_deadline = std::min<fc::time_point>(
-                  scheduled_trx_deadline, fc::time_point::now() + fc::milliseconds(_max_scheduled_transaction_time_per_block_ms));
-            }
-            // may exhaust scheduled_trx_deadline but not preprocess_deadline, exhausted preprocess_deadline checked below
-            process_scheduled_and_incoming_trxs(scheduled_trx_deadline, incoming_itr);
+            // Hard-code the deadline to retire expired deferred trxs to 50ms
+            auto deferred_trxs_deadline = std::min<fc::time_point>(preprocess_deadline, fc::time_point::now() + fc::milliseconds(50));
```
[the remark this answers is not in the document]
linh2931: Is the above `auto deferred_trxs_deadline = std::min<fc::time_point>(preprocess_deadline, fc::time_point::now() + fc::milliseconds(50));` not `min of preprocess_deadline and now() + 50ms`?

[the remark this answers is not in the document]
heifner: Weird, you have exactly what I would expect. I could swear that was `auto deferred_trxs_deadline = fc::time_point::now() + fc::milliseconds(50);`

plugins/producer_plugin/producer_plugin.cpp
heifner: See `controller::validate_expiration` which has `EOS_ASSERT( trx.expiration.to_time_point() >= pending_block_time()`

linh2931: Thanks!

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2370,71 +2302,21 @@ bool producer_plugin_impl::process_unapplied_trxs(const fc::time_point& deadline
...
-      }
-
-      const transaction_id_type trx_id         = sch_itr->trx_id; // make copy since reference could be invalidated
-      const auto                sch_expiration = sch_itr->expiration;
-      auto                      sch_itr_next   = sch_itr; // save off next since sch_itr may be invalidated by loop
-      ++sch_itr_next;
-      const auto next_delay_until = sch_itr_next != sch_idx.end() ? sch_itr_next->delay_until : sch_itr->delay_until;
-      const auto next_id          = sch_itr_next != sch_idx.end() ? sch_itr_next->id : sch_itr->id;
-
-      num_processed++;
```
arhag, line 2412: You lost this. Currently, `num_processed` will remain at 0.

arhag (changes requested):

plugins/producer_plugin/producer_plugin.cpp
linh2931: Thanks! I am double checking.

plugins/producer_plugin/producer_plugin.cpp
arhag: I wonder if 50ms too much? Expiring a deferred transaction is pretty quick. So an attacker would need to schedule a lot of deferred transactions per block to continually use up the full 50 ms capacity per block. This would cost them RAM but only for 1 block which they could reuse in the next block. So really it only costs them the CPU used to schedule the smallest deferred transactions they can as quickly as possible.

heifner: Sure 25ms or 10ms would work. I don't think we need it configurable.

linh2931: Oha, I missed this comment. Addressed it now. Make it to 10ms to be on safe side.

arhag (changes requested):
https://github.com/AntelopeIO/leap/issues/1556 says to also "block delayed transactions". This means the check in `transaction_context::init_for_input_trx` should be updated to disallow non-zero `delay_sec` for all transaction types (not just read-only and dry-run ones). It also means that `transaction_context::schedule_transaction` can be removed.

libraries/chain/transaction_context.cpp
```diff
@@ -330,10 +328,6 @@ namespace eosio { namespace chain {
       for( uint32_t i = 1; i <= num_original_actions_to_execute; ++i ) {
          execute_action( i, 0 );
       }
-
-      if( delay != fc::microseconds() ) {
```
arhag, line 334: You can also remove the `delay` member variable from `transaction_context`.

linh2931:
> #1556 says to also "block delayed transactions". This means the check in `transaction_context::init_for_input_trx` should be updated to disallow non-zero `delay_sec` for all transaction types (not just read-only and dry-run ones). It also means that `transaction_context::schedule_transaction` can be removed.

I will updated the tests.

libraries/chain/transaction_context.cpp
linh2931: Removing `delay` breaks a number of seemingly unrelated tests. Will look into it further.

linh2931: It was my mistake. Done.

unittests/protocol_feature_tests.cpp
```diff
@@ -526,182 +526,6 @@ BOOST_AUTO_TEST_CASE( replace_deferred_test ) try {
...
-         trace1 = t;
-      }
-   } );
-
-   BOOST_REQUIRE_EQUAL(0u, index.size());
-
-   c.push_action( config::system_account_name, "reqauth"_n, "alice"_n, fc::mutable_variant_object()
-      ("from", "alice"),
-      5, 2
-   );
```
arhag, line 590: I think most of this test case is still salvageable even if this action can no longer be pushed.

Maybe I'm not remembering the details of this test correctly, but I don't think pushing this delayed transaction really even gets us much in this test. Maybe it was just there to ensure that a deferred transaction generated prior to activation of the protocol feature would still be retired correctly.

So I think you should be able to remove it, adapt some of the `BOOST_REQUIRE_EQUAL`s below, and still keep the bulk of this test case which is checking the transaction extensions which are there to allow for unique deferred transaction IDs.

arhag (changes requested):

unittests/protocol_feature_tests.cpp
linh2931: Thanks. I have restored and adapted `protocol_feature_tests/no_duplicate_deferred_id_test` instead of removing it outright.

unittests/delay_tests.cpp
```diff
@@ -359,2049 +289,23 @@ BOOST_AUTO_TEST_CASE(delete_auth_test) { try {
...
-
-   liquid_balance = get_currency_balance(chain, "tester"_n);
-   BOOST_REQUIRE_EQUAL(asset::from_string("96.0000 CUR"), liquid_balance);
-   liquid_balance = get_currency_balance(chain, "tester2"_n);
-   BOOST_REQUIRE_EQUAL(asset::from_string("4.0000 CUR"), liquid_balance);
-
-} FC_LOG_AND_RETHROW() }/// schedule_test
-
-// test canceldelay action cancelling a delayed transaction
-BOOST_AUTO_TEST_CASE( canceldelay_test ) { try {
```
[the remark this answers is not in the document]
arhag: Never mind. It is not possible to use the native `canceldelay` to cancel a contract-generated deferred transaction. And I don't think it is worth providing a testing mode to controller to allow submitted delayed transactions just to have something to test `canceldelay` with.

So you can just remove the `cancedelay_test` as you originally did here.

libraries/chain/controller.cpp
```diff
@@ -3720,6 +3716,22 @@ void controller::code_block_num_last_used(const digest_type& code_hash, uint8_t
...
+   auto gto = my->db.find<generated_transaction_object, by_trx_id>(trx_id);
+   if (gto) {
+      my->db.modify<generated_transaction_object>(*gto, [&]( auto& gtx ) {
+         gtx.sender = account_name();
+
+         fc::uint128 _id(trx_id._hash[3], trx_id._hash[2]);
+         gtx.sender_id = (unsigned __int128)_id;
+      });
+   }
+}
```
heifner, line 3733: This should be moved to the test that calls it. That function could be made a `friend` of controller so that it has access to `mutable_db()`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -108,19 +108,6 @@ bool exception_is_exhausted(const fc::exception& e) {
...
-};
-
-struct by_id;
-struct by_expiry;
-
-using transaction_id_with_expiry_index = multi_index_container<
-   transaction_id_with_expiry,
-   indexed_by<hashed_unique<tag<by_id>, BOOST_MULTI_INDEX_MEMBER(transaction_id_with_expiry, transaction_id_type, trx_id)>,
-              ordered_non_unique<tag<by_expiry>, BOOST_MULTI_INDEX_MEMBER(transaction_id_with_expiry, fc::time_point, expiry)>>>;
-
```
heifner, line 123: `producer_plugin` no longer needs to include boost multi index or have the multi index aliases.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2374,78 +2306,36 @@ bool producer_plugin_impl::process_unapplied_trxs(const fc::time_point& deadline
...
       }
 
+      const transaction_id_type trx_id             = expired_itr->trx_id; // make copy since reference could be invalidated
+      auto                      expired_itr_next   = expired_itr; // save off next since expired_itr may be invalidated by loop
+      ++expired_itr_next;
+      const auto next_expired = expired_itr_next != expired_idx.end() ? expired_itr_next->expiration : expired_itr->expiration;
+      const auto next_id      = expired_itr_next != expired_idx.end() ? expired_itr_next->id : expired_itr->id;
+
+      num_processed++;
+
```
heifner, line 2338: `max_trx_time` below can be hard-coded to `fc::microseconds::maximum()` as it is not used in `push_scheduled_transaction` when trx is expired.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2497,16 +2385,14 @@ void producer_plugin_impl::process_scheduled_and_incoming_trxs(const fc::time_po
       }
       LOG_AND_DROP();
 
-      incoming_trx_weight += _incoming_defer_ratio;
-
-      if (sch_itr_next == sch_idx.end())
+      if (expired_itr_next == expired_idx.end())
          break;
-      sch_itr = sch_idx.lower_bound(boost::make_tuple(next_delay_until, next_id));
+      expired_itr = expired_idx.lower_bound(boost::make_tuple(next_expired, next_id));
```
heifner, line 2390: Don't actually need this `lower_bound` now as the `expired_itr_next` can just be used for `expired_itr`. I'm ok with not updating this and leaving as is if you prefer.

heifner (changes requested):

libraries/chain/controller.cpp
heifner: That seems preferable to controller "knowing" this test function.

linh2931: I think it more. Another advantage of making it as a friend is moving `modify_gto_for_canceldelay_test` out of controller can prevent it from being abused.  Will make changes. Thanks.

linh2931: Thanks @heifner for helping me to get the friend to work!

heifner (approved):

unittests/delay_tests.cpp
```diff
@@ -308,87 +419,75 @@ BOOST_AUTO_TEST_CASE(delete_auth_test) { try {
 
    BOOST_REQUIRE_EQUAL(transaction_receipt::executed, trace->receipt->status);
 
-   liquid_balance = get_currency_balance(chain, "eosio.token"_n);
-   BOOST_REQUIRE_EQUAL(asset::from_string("999900.0000 CUR"), liquid_balance);
+   chain.produce_blocks();
+
+   // Propose and approve an msig trx that transfers "quantity" tokens
```
arhag, line 424: Prior to to this point you needed to update the permission for the active permission to add a delay to it. That seems to have gotten lost within this test.

arhag (changes requested):

unittests/delay_tests.cpp
linh2931: Thanks for your thorough review!

arhag (approved):

PR #1645, opened 2023-09-15
Remove max-nonprivileged-inline-action-size option

### LINKED ISSUE #1614: Allow non privileged inline actions larger then 4KiB
Remove `max-non-privileged-inline-action-size` option and the associated checks in the code. Add a test.

There is a 4 KiB limit on inline action size if the sender is not a privileged account. There isn't a business need for this constraint, and it makes our EOS-EVM integration more difficult. We cannot easily change the limit because it is subjectively enforced and we need to convince the whole network to change that config option. Instead we should update the code to no longer enforce this option. 

```[tasklist]
### Tasks
```

heifner:
Remove the option `max-nonprivileged-inline-action-size`
- This PR reverts https://github.com/EOSIO/eos/pull/9262
- The subjective limit makes it difficult to coordinate increasing the value across all nodes.
  - The default subjective limit of 4k is much lower than the on-chain limit of 512K which limits useful use cases including EOS-EVM integration.
  - The objective on-chain limit `max_inline_action_size` provides protection against abuse.
  - The EOS mainnet objective on-chain limit is currently set to `524287`.  https://bloks.io/account/eosio?loadContract=true&tab=Tables&account=eosio&scope=eosio&limit=100&table=global
- Added new tests for the default 512K objective limit.

Resolves #1614


linh2931 (approved):
I like your PR description!

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -453,7 +450,7 @@ namespace eosio { namespace testing {
    class tester : public base_tester {
    public:
       tester(setup_policy policy = setup_policy::full, db_read_mode read_mode = db_read_mode::HEAD, std::optional<uint32_t> genesis_max_inline_action_size = std::optional<uint32_t>{}, std::optional<uint32_t> config_max_nonprivileged_inline_action_size = std::optional<uint32_t>{}) {
```
greg7mdp, line 452: should you remove the parameter as well, or maybe change the name to `unused`.
`std::optional<uint32_t> config_max_nonprivileged_inline_action_size = std::optional<uint32_t>{}`

greg7mdp:
Should remove comments in `libraries/chain/include/eosio/chain/config-dir/config.ini `

```
# maximum allowed size (in bytes) of an inline action for a nonprivileged account (eosio::chain_plugin)
# max-nonprivileged-inline-action-size = 4096
```

libraries/testing/include/eosio/testing/tester.hpp
heifner: Thanks, not sure how my search missed that.

greg7mdp (approved):

linh2931 (approved):

PR #1649, opened 2023-09-18
avoid possible memcpy with nullptr when action data size is 0

spoonincode:
When action data size is 0, it was possible for a `memcpy` call to contain a source pointer of `nullptr` and length of `0`. This is not defined behavior, but a careful analysis of leap's supported platforms and prior builds shows no evidence this could be a source of indeterminism so this change is not being treated as a security fix.

### DIFF
--- a/libraries/chain/webassembly/action.cpp
+++ b/libraries/chain/webassembly/action.cpp
@@ -5,10 +5,8 @@
 namespace eosio { namespace chain { namespace webassembly {
    int32_t interface::read_action_data(legacy_span<char> memory) const {
       auto s = context.get_action().data.size();
-      if( s == 0 ) return s;
-      if( memory.size() == 0 ) return s;
-
       auto copy_size = std::min( static_cast<size_t>(memory.size()), s );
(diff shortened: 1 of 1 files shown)


libraries/chain/webassembly/action.cpp
```diff
@@ -5,6 +5,7 @@
 namespace eosio { namespace chain { namespace webassembly {
    int32_t interface::read_action_data(legacy_span<char> memory) const {
       auto s = context.get_action().data.size();
+      if( s == 0 ) return s;
       if( memory.size() == 0 ) return s;
 
       auto copy_size = std::min( static_cast<size_t>(memory.size()), s );
```
greg7mdp, line 11:
```suggestion
      auto copy_size = std::min( static_cast<size_t>(memory.size()), s );
      if( copy_size == 0 ) return s;
```

greg7mdp (changes requested):

greg7mdp (approved):

heifner (approved):

PR #1659, opened 2023-09-19
Add multi-hop network to test_pacemaker

### LINKED ISSUE #1658: IF: add multi-hop networking to test_pacemaker
`test_pacemaker` currently assumes all nodes are directly connected to each other. Must change it so that a test may specify an arbitrary mesh of connections between nodes, which will allow us to write unit tests for #1548.

fcecin:
- test_pacemaker allows each test to define a specific connection mesh between nodes
- added a simple unit test that can be used to test a solution for #1548 (just needs to introduce a disconnect between nodes)
- added support to write multi-hop unit tests (with selective message propagation by type)

The hotstuff protocol has "phases", and between two adjacent phases, we are typically using different message types. This allows us to write tests where we loop propagating a specific message type across the network, for as many hops as necessary, until the `dispatch` method on the `test_pacemaker` returns an empty vector, signaling no more message propagation happened. And then, we can start propagating the message type that's expected for the next phase of the protocol (these were generated during propagation of the previous type, but were just retained).

All unit tests still use a full connection mesh, which is easily created by calling `test_pacemaker::connect(vector-with-all-node-names)`. The added test is meant to be changed by the inclusion of an explicit disconnect between the two rotating leaders; this can be used as the first test of a solution to message propagation.

Closes #1658

### DIFF
--- a/libraries/hotstuff/test/test_pacemaker.cpp
+++ b/libraries/hotstuff/test/test_pacemaker.cpp
@@ -15,7 +15,7 @@ namespace eosio::hotstuff {
       _next_leader = next_leader;
    };
 
-   void test_pacemaker::set_finalizers(std::vector<name> finalizers) {
+   void test_pacemaker::set_finalizers(const std::vector<name>& finalizers) {
       _finalizers = finalizers;
    };
 
@@ -27,7 +27,7 @@ namespace eosio::hotstuff {
       _quorum_threshold = threshold;
    }
 
-   void test_pacemaker::add_message_to_queue(hotstuff_message msg) {
+   void test_pacemaker::add_message_to_queue(const hotstuff_message& msg) {
       _pending_message_queue.push_back(msg);
    }
 
@@ -56,7 +56,7 @@ namespace eosio::hotstuff {
       return it->second.count(node2) > 0;
    }
 
-   void test_pacemaker::pipe(std::vector<test_pacemaker::hotstuff_message> messages) {
+   void test_pacemaker::pipe(const std::vector<test_pacemaker::hotstuff_message>& messages) {
       auto itr = messages.begin();
       while (itr != messages.end()) {
          _pending_message_queue.push_back(*itr);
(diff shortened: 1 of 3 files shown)


libraries/hotstuff/test/test_pacemaker.cpp
```diff
@@ -39,55 +64,62 @@ namespace eosio::hotstuff {
...
+               throw std::runtime_error("unknown message variant");
+            }
+
+            dispatched_messages.push_back(*msg_itr);
+         } else {
+            kept_messages.push_back(*msg_itr);
+         }
 
          ++msg_itr;
       }
```
greg7mdp, line 120: I suggest this more modern version:
```
      for (const auto& msg_pair : message_queue) {
         const auto& [sender_id, msg] = msg_pair;
         size_t v_index = msg.index();

         if (msg_type == hs_all_messages || msg_type == v_index) {

            if (v_index == hs_proposal) {
               ++proposals_count;
               on_hs_proposal_msg(std::get<hs_proposal_message>(msg), sender_id);
            } else if (v_index == hs_vote) {
               ++votes_count;
               on_hs_vote_msg(std::get<hs_vote_message>(msg), sender_id);
            } else if (v_index == hs_new_block) {
               ++new_blocks_count;
               on_hs_new_block_msg(std::get<hs_new_block_message>(msg), sender_id);
            } else if (v_index == hs_new_view) {
               ++new_views_count;
               on_hs_new_view_msg(std::get<hs_new_view_message>(msg), sender_id);
            } else {
               throw std::runtime_error("unknown message variant");
            }

            dispatched_messages.push_back(msg_pair);
         } else {
            kept_messages.push_back(msg_pair);
         }
      }
```

libraries/hotstuff/test/test_pacemaker.cpp
```diff
@@ -181,7 +213,7 @@ namespace eosio::hotstuff {
       while (qc_itr != _qcc_store.end()){
          const name                & qcc_name = qc_itr->first;
          std::shared_ptr<qc_chain> & qcc_ptr  = qc_itr->second;
-         if (qcc_ptr->get_id_i() != id && is_qc_chain_active(qcc_name) )
+         if (qcc_ptr->get_id_i() != id && is_qc_chain_active(qcc_name) && is_connected(id, qcc_ptr->get_id_i()))
```
greg7mdp, line 216: Please replace these old-style loops:
```
   void test_pacemaker::on_hs_proposal_msg(const hs_proposal_message& msg, name id) {
      auto qc_itr = _qcc_store.begin();
      while (qc_itr != _qcc_store.end()){
         const name                & qcc_name = qc_itr->first;
         std::shared_ptr<qc_chain> & qcc_ptr  = qc_itr->second;
         if (qcc_ptr->get_id_i() != id && is_qc_chain_active(qcc_name) && is_connected(id, qcc_ptr->get_id_i()))
            qcc_ptr->on_hs_proposal_msg(0, msg);
         qc_itr++;
      }
   }
```

with:

```
   void test_pacemaker::on_hs_proposal_msg(const hs_proposal_message& msg, name id) {
      for (const auto& [qcc_name, qcc_ptr] : _qcc_store) {
         if (qcc_ptr->get_id_i() != id && is_qc_chain_active(qcc_name) && is_connected(id, qcc_ptr->get_id_i()))
            qcc_ptr->on_hs_proposal_msg(0, msg);
      }
   }
```

greg7mdp (changes requested):

heifner (approved):

ericpassmore:
Note:start
group: IF
category: MESSAGES
summary: Tests message passing for specific network topology setup. 
Note:end

PR #1669, opened 2023-09-22
Update install_deps.sh

mschoenebeck:
Added `libstdc++-12-dev` which is a dependency of `llvm`.

### DIFF
--- a/scripts/install_deps.sh
+++ b/scripts/install_deps.sh
@@ -15,6 +15,7 @@ apt-get install -y \
     libgmp-dev \
     libncurses5 \
     libssl-dev \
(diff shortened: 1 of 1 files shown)


spoonincode:
Can you describe the problem you encountered and how this change fixed it? Need to understand if this ought to go back to a stable release branch as a bug fix.

`install_deps.sh` is really meant to be "install deps _for pinned builds_"; this differentiation has been a point of confusion in the past. But, that's why `install_deps.sh` does not install llvm (nor boost, for 3.x/4.0) -- those need to be built as part of the pinned build.

heifner:
No reply. Closing.

Issue #1641, opened 2023-09-14
reproducible binary builds


spoonincode:
Covers work originally documented as proposals in,

https://github.com/eosnetworkfoundation/product/blob/main/build-system/proposals/initial_tenets.md#single-linux-binary-build-to-rule-them-all-well-per-platform--arch
https://github.com/eosnetworkfoundation/product/blob/main/build-system/proposals/initial_tenets.md#fully-reproducible-binary-builds

PR #1670, opened 2023-09-22
bump libsecp256k1 to v0.4.0

spoonincode:
This bumps the submoduled libsecp256k1 library to v0.4.0 (at the time of the previous submodulization libsecp256k1 did not define version numbers -- current `master` was simply taken)

The primary motivator for this change is a comment from the v0.3.1 release notes,
> We strongly recommend updating to 0.3.1 if you use or plan to use Clang >=14 to compile libsecp256k1. [This release fixes] "constant-timeness" issue with Clang >=14 that could leave applications using libsecp256k1 vulnerable to a timing side-channel attack. The fix avoids secret-dependent control flow and secret-dependent memory accesses in conditional moves of memory objects when libsecp256k1 is compiled with Clang >=14.

As we bump our clang version for pinned/reproducible builds it seems prudent to take this fix since we'll be using the latest version of clang.

This library is, of course, deeply used in consensus. But if there is any third party library we use that aligns with our consensus demands it's going to be this one so I consider risk to be on the low side.

The only real change is the modification to how a few preprocessor macros get plumbed through. v0.3.0 release notes mentions,
> Removed the configuration header `src/libsecp256k1-config.h`. We recommend passing flags to `./configure` or `cmake` to set configuration options (see `./configure --help` or `cmake -LH`). If you cannot or do not want to use one of the supported build systems, pass configuration flags such as `-DSECP256K1_ENABLE_MODULE_SCHNORRSIG` manually to the compiler (see the file `configure.ac` for supported flags).

While the library does include a CMakeLists.txt now, it's fraught with issues similar to other third party libraries when we integrate their cmake rules: libtester grief. So, I've elected to just manually plumb through the previous defines that were once in `libsecp256k1-config.h`. I also added `SECP256K1_STATIC=1` since it's needed on Windows.

I noticed no change in recovery performance. I did see a very very slight slow down in signing performance (something like 24.7µs to 25.3µs). I don't consider that slow down material enough to be alarming.

The only other thing I thought was interesting in the change log was,
> The `secp256k1_nonce_function_rfc6979` nonce function, used by default by `secp256k1_ecdsa_sign`, now reduces the message hash modulo the group order to match the specification. This only affects improper use of ECDSA signing API.

This is interesting because "Antelope Canonical K1 Signatures" requires fiddling with the nonce function,
https://github.com/AntelopeIO/leap/blob/a44813512b082b420b1706c75f88681dac20c5d8/libraries/libfc/src/crypto/elliptic_impl_priv.cpp#L83-L89
I was unable to find any potential conflicts in this area though, as we don't care about the message hash the nonce function generates.

Work on #1641


libraries/libfc/src/crypto/elliptic_secp256k1.cpp
```diff
@@ -26,7 +26,7 @@ namespace fc { namespace ecc {
     namespace detail
     {
         const secp256k1_context* _get_context() {
-            static secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_VERIFY | SECP256K1_CONTEXT_SIGN);
+            static secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_NONE);
```
spoonincode, line 29: Per v0.2.0 release notes,
> Deprecated context flags `SECP256K1_CONTEXT_VERIFY` and `SECP256K1_CONTEXT_SIGN`. Use `SECP256K1_CONTEXT_NONE` instead.

greg7mdp (approved):

heifner (approved):

PR #1672, opened 2023-09-23
IF: Filter duplicate votes

fcecin:
- qc_chain does not aggregate a signature from a finalizer that has already voted on a proposal
- added test_pacemaker::duplicate()
- added hotstuff_8 unit test to test for duplicate votes

This fix is related to #1548 (lack of duplicate filtering only manifests with message propagation)

### DIFF
--- a/libraries/hotstuff/test/test_pacemaker.cpp
+++ b/libraries/hotstuff/test/test_pacemaker.cpp
@@ -79,7 +79,7 @@ namespace eosio::hotstuff {
          if (v_index == msg_type)
             dup.push_back(msg_pair);
       }
-      _pending_message_queue = dup;
+      _pending_message_queue = std::move(dup);
    }
 
    std::vector<test_pacemaker::hotstuff_message> test_pacemaker::dispatch(std::string memo, hotstuff_message_index msg_type) {
--- a/libraries/hotstuff/test/test_pacemaker.cpp
+++ b/libraries/hotstuff/test/test_pacemaker.cpp
@@ -70,6 +70,18 @@ namespace eosio::hotstuff {
       }
    }
 
+   void test_pacemaker::duplicate(hotstuff_message_index msg_type) {
+      std::vector<test_pacemaker::hotstuff_message> dup;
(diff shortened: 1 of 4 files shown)


libraries/hotstuff/test/test_pacemaker.cpp
```diff
@@ -70,6 +70,18 @@ namespace eosio::hotstuff {
...
+   void test_pacemaker::duplicate(hotstuff_message_index msg_type) {
+      std::vector<test_pacemaker::hotstuff_message> dup;
+      for (const auto& msg_pair : _pending_message_queue) {
+         const auto& [sender_id, msg] = msg_pair;
+         size_t v_index = msg.index();
+         dup.push_back(msg_pair);
+         if (v_index == msg_type)
+            dup.push_back(msg_pair);
+      }
+      _pending_message_queue = dup;
```
greg7mdp, line 82:
```suggestion
      _pending_message_queue = std::move(dup);
```

greg7mdp: Does it matter where the duplicates are added? Should they be together in the vector?

fcecin: Yes, they need to be together to trigger duplicate vote processing. If the message queue is `abcdefghijklmnopqrstuabcdefghijklmnopqrstu` votes, after processing the first `o` vote (15th vote) a quorum is met and `process_vote` correctly won't try to aggregate any more votes. But if you flush the duplicates as  `aabbccddeeffgghhiijjkkllmmnnooppqqrrssttuu`, then the second time `process_vote` sees `a`, it will try to aggregate it again (quorum not met yet) and then a problem occurs.

heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: MESSAGES
summary: Filter out duplicate messages to prevent infinite loops, and support configuration where BPs are not directly connected. BPs may be indirectly connected via a relay node. 
Note:end

PR #1674, opened 2023-09-24
fix some different signedness warnings in `custom_appbase_tests`

spoonincode:
Annoyingly, `std::count_if()` returns a signed type. So.. besides more refactoring and/or casting and/or new assignments.. this change was the best I came up with even though `num_expected` represents something unsigned. Feel free to suggest other approaches though

### DIFF
--- a/libraries/custom_appbase/tests/custom_appbase_tests.cpp
+++ b/libraries/custom_appbase/tests/custom_appbase_tests.cpp
@@ -290,7 +290,7 @@ BOOST_AUTO_TEST_CASE( execute_from_read_only_and_read_exclusive_queues ) {
    }
 
    auto app_thread = start_app_thread(app);
-   constexpr size_t num_expected = 13u; // 16 - 3 read_write
+   constexpr int num_expected = 13; // 16 - 3 read_write
 
    auto read_thread1 = start_read_thread(app);
    auto read_thread2 = start_read_thread(app);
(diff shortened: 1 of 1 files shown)


linh2931 (approved):
Using `int` is OK here, as we are defining a constexpr and we know the values (13 and 600). Both are valid integers (positive).

libraries/custom_appbase/tests/custom_appbase_tests.cpp
```diff
@@ -350,7 +350,7 @@ BOOST_AUTO_TEST_CASE( execute_many_from_read_only_and_read_exclusive_queues ) {
    app->executor().set_to_read_window([](){return false;});
 
    // post functions
-   constexpr size_t num_expected = 600u;
+   constexpr int num_expected = 600;
```
greg7mdp, line 353: Wouldn't this change cause a warning in `for (size_t i = 400; i < num_expected; i+=5)` below?
I think it would make sense to cast the result of `std::count_if` to `size_t`. The result will always be positive, and the only reason it returns a signed value is for compatibility with `std::count` I believe.

spoonincode: makes it clean on gcc 12.2 :woman_shrugging:  not sure what exactly the rules are: the warning is
> when a comparison between signed and unsigned values could produce an incorrect result when the signed value is converted to unsigned

So maybe there are additional rules here beyond just simply two types differentiating in signness which are required to be met before the warning is flagged?

greg7mdp (approved):

PR #1675, opened 2023-09-25
RFC: chainbase environment check tweaks; looser in some cases, stricter in others

spoonincode:
nodeos, via chainbase, constructs objects directly in a pool of memory with a dedicated allocator, and the lifetime of these objects extend indefinitely. This pool of memory is nominally a memory mapped file. Clearly, the layout of these objects must remain consistent across restarts of nodeos and, in an effort to prevent inconsistency when nodeos is rebuilt with different compilers or library versions, an environment check is performed: the environment at launch must match the environment when the memory pool was created. Failure of the environment check halts start up and requires the user to roll back nodeos or reset chain state (via genesis or snapshot).

The environment check is made up of five items:
* Compiler version
* Boost version
* `NDEBUG` set
* OS
* CPU architecture

The latter three are typically not a visible difference for users so I'm not too concerned with them.

The boost version check absolutely matters. Layout of boost objects placed in to the memory pool (such as `boost::mutex`) have changed over time. Thus, this check is required. However, as of Leap 5.0, the boost version will be pinned as part of an in-tree boost. So users rebuilding Leap 5.x will never have the boost version unexpectedly changed between rebuilds.

The compiler version check is more interesting. I am uncertain what this is protecting against; is it just paranoia? Structure layout is specified by the platform ABI (otherwise you would not be able to pass `struct`s between code compiled with different compilers), so different versions of the compiler should generate `struct`s with identical layout. Standard C++ library types provided via libstdc++ or libc++ that might be constructed in the memory pool, as far as I know, have ABI guarantees by the standard library otherwise it would not be possible to pass a `std::string` between shared objects compiled with different versions. To be clear, not all C++ libraries strive for this kind of ABI compatibility, but my understanding is that libstdc++ and libc++ do. (You can see this surface a little bit in how long it took the standard libraries to include some complex libraries, like `filesystem` or `format`, as part of the default library where you didn't have to pass a special `-lstdc++fs` or `-fexperimental-library`: the library authors were not willing to commit to the long term ABI compatibility for those libraries just yet.)

Thus, I am uncertain why we need to protect against the compiler version changing. But, there is one case that is dangerous: the layout of standard library types is not the same between different standard libraries. For example, a `std::string` from libstdc++ is not compatible in ABI with a `std::string` from libc++. And yet, this clearly breaking difference is a case that is _not_ detected by chainbase's current environment check!

This PR tweaks the environment check to be a little more lenient and checks only that the major version of the compiler and standard library matches. Primary motivation is to avoid user surprise on an environment check failure when Ubuntu rolls out an upgrade from gcc 9.2.0 to 9.3.0, and the user stumbles on this when upgrading from Leap 5.0.1 to 5.0.2, for example.


greg7mdp:
Possibly I am misunderstanding, but these seem contradictory:
`Standard C++ library types provided via libstdc++ or libc++ that might be constructed in the memory pool, as far as I know, have ABI guarantees by the standard library otherwise it would not be possible to pass a std::string between shared objects compiled with different versions.`

and

`the layout of standard library types is not the same between different standard libraries. For example, a std::string from libstdc++ is not compatible in ABI with a std::string from libc++.`

greg7mdp (approved):

spoonincode:
> I don't see the corresponding chainbase PR

I probably should have set it to draft; I didn't think to make a chainbase PR since I wast just initially trying to solicit comments and feedback. But I guess it's hard to leave feedback with no PR on the code that is changed, so I added AntelopeIO/chainbase#22


> Possibly I am misunderstanding, but these seem contradictory: `Standard C++ library types provided via libstdc++ or libc++ that might be constructed in the memory pool, as far as I know, have ABI guarantees by the standard library otherwise it would not be possible to pass a std::string between shared objects compiled with different versions.`
> 
> and
> 
> `the layout of standard library types is not the same between different standard libraries. For example, a std::string from libstdc++ is not compatible in ABI with a std::string from libc++.`

libstdc++ maintains public ABI guarantees for their `std::string`, and libc++ maintains public ABI guarantees for their `std::string`, but they are not compatible. So you can't mix and match.

But upon more reflection, "public ABI" is certainly different then the entire memory layout of some type. For example, a `std::string` may have a public ABI containing an opaque pointer type, and the contents of that opaque pointer type can change in different versions of libstdc++ without changing the public ABI, and not break two shared libraries passing a `std::string` around since it's the libstdc++ shared library that deals with the private opaque pointer data.

Thus it seems this change is probably not 100% safe if we construct stdlib types in the shared memory region, as it theory the private memory layout could change between even a patch level release.

I wonder if there is any reasonable way we can audit for usage of stdlib types in chainbase usage? For example, `shared_string` is a `shared_cow_string`; `shared_vector` is a `boost::interprocess::vector<T, allocator<T>>` (which technically could use a stdlib type internally). If we could be certain we aren't using stdlib types in any chainbase storage, this just wouldn't be a problem anyways.

spoonincode:
Without fully auditing objects that are constructed in chainbase, this seems too risky for now

PR #1685, opened 2023-09-27
PH: Add a longer running read-only trx for tests

heifner:
* Create a `logging-template.json` for tests instead of using the example one in `programs/nodeos` since the configured `gelf` logging was wasting a thread and cpu when no gelf logging is being used in tests.
* Add a new `doitslow` action to `payloadless` contract for use in tests and performance harness runs.

Example of using the new longer running read-only trx in a performance run:
https://github.com/AntelopeIO/leap/actions/runs/6318554443
```
./tests/PerformanceHarnessScenarioRunner.py findMax testApiOpMode --max-tps-to-test 1000 overrideBasicTestConfig -v --http-threads 4 --api-nodes-read-only-threads 8 --read-only-write-window-time-us 1000 --read-only-read-window-time-us 450000 --account-name payloadless --abi-file payloadless.abi --wasm-file payloadless.wasm --contract-dir unittests/test-contracts/payloadless --user-trx-data-file tests/PerformanceHarness/readOnlySlowTrxData.json
```
```
      "TPS": {
        "min": 995,
        "max": 1004,
        "avg": 1000.0,
        "sigma": 0.8301741920760902,
        "emptyBlocks": 0,
        "numBlocks": 297,
        "configTps": 1000,
        "configTestDuration": 150,
        "tpsPerGenerator": [
          1000
        ],
        "generatorCount": 1
      },
      "TrxCPU": {
        "min": 5655.0,
        "max": 10308.0,
        "avg": 6548.978346666667,
        "sigma": 1362.3947007864986,
        "samples": 150000
      },
```
Which means 1000*6549=6,549,000us worth of trx execution within a 1,000,000us window.


oschwaldp-oci (approved):

unittests/test-contracts/payloadless/payloadless.cpp
```diff
@@ -5,3 +5,46 @@ using namespace eosio;
...
+   return bool(s == 0);
+}
+
+
+void payloadless::doitslow() {
+   print("Im a payloadless slow action");
+
+   for (size_t p = 2; p <= cpu_prime_max; p += 1) {
+      if (is_prime(p) && is_mersenne_prime(p)) {
+         // We need to keep an eye on this to make sure it doesn't get optimized out. So far so good.
```
linh2931, line 45: How do we tell if it is not optimized out? Can it be checked automatically?

unittests/test-contracts/payloadless/payloadless.cpp
```diff
@@ -5,3 +5,46 @@ using namespace eosio;
...
+      return false;
+   }
+
+   bool prime = true;
+   const int to = sqrt(p);
+   int i;
+   for (i = 3; i <= to; i += 2) {
+      if (!((prime = bool(p)) % i)) break;
+   }
+   return prime;
```
linh2931, line 24: As `prime` is initialized to true, would the function always returns true? On the other hand, this function only concerns about `slowness`.

linh2931 (commented):
I like the example in the PR description, which shows the read-only threads are occupied most of the time.

unittests/test-contracts/payloadless/payloadless.cpp
heifner: Could check the elapsed time in the returned trace is greater than say 100us.

unittests/test-contracts/payloadless/payloadless.cpp
linh2931: Oh I missed the assignment `prime = bool(p)`.

linh2931 (approved):

unittests/test-contracts/payloadless/payloadless.cpp
```diff
@@ -5,3 +5,46 @@ using namespace eosio;
...
+      return true;
+   } else if (p <= 1 || p % 2 == 0) {
+      return false;
+   }
+
+   bool prime = true;
+   const int to = sqrt(p);
+   int i;
+   for (i = 3; i <= to; i += 2) {
+      if (!((prime = bool(p)) % i)) break;
```
oschwaldp-oci, line 22: How does it work?  I'm thinking the prime calc somewhere in here should be `p % i`... setting `prime = bool(p)` makes prime always true as long as `p > 0` but I don't see how you get the needed` p % i` out of this

unittests/test-contracts/payloadless/payloadless.cpp
heifner: I agree, this just always returns true. I'll fix it.

unittests/test-contracts/payloadless/payloadless.cpp
oschwaldp-oci: I believe it is supposed to be `prime = bool(p % i)`?

oschwaldp-oci (changes requested):

oschwaldp-oci (approved):

unittests/test-contracts/payloadless/payloadless.cpp
```diff
@@ -17,9 +17,11 @@ bool is_prime(int p) {
...
    bool prime = true;
    const int to = sqrt(p);
-   int i;
-   for (i = 3; i <= to; i += 2) {
-      if (!((prime = bool(p)) % i)) break;
+   for (int i = 3; i <= to; i += 2) {
+      if (p % i == 0) {
+         prime = false;
+         break;
+      }
```
linh2931, line 24: This is more clear.

linh2931 (approved):

PR #1695, opened 2023-09-29
Chainbase should balk on 3.1-4.0 chainbases

### LINKED ISSUE #1570: Chainbase should balk on 3.1-4.0 chainbases
## High Level Overview
The layout of chainbase in 5.0 is incompatible with prior versions. So we need to have detection and helpful error messages to handle gracefully.

## Original Content of Issue
Most likely 5.0 should balk on 3.1-4.0 chainbases, and vise versa. We can lean on the environment check for 99% of this, but no t 100%.

greg7mdp:
Resolves #1570 .

Updates chainbase to tip of `main` branch, where the version marker embedded in the `chainbase` db file was updated (in order for Leap 5.0 to not try loading chainbases dbs in the old format).

Here is the error displayed when this is attempted (Loading a Leap `3.1.1` db into `5.0`):

image

We could probably make the error display more user-friendly, i.e. state the likely cause of the problem and suggest a solution, However, I think this is a task that should be done holistically for all possible errors, and is deserving of its own issue.


heifner (approved):

spoonincode:
includes AntelopeIO/chainbase#20 & AntelopeIO/chainbase#23

spoonincode (approved):

Issue #1555, opened 2023-08-24
DTrx: Protocol Feature removing deferred transactions


bhazzard:
Disabling deferred transactions via protocol features will come in two stages.

`DISABLED_DEFERRED_TRXS_STAGE_1`:

Once this first protocol feature is activated, the behavior of the `send_deferred` and `cancel_deferred` host functions changes so that they become no-ops. In addition, any block that retires a deferred transaction with a status other than `expired` is invalid. Also, a deferred transaction can be retired as `expired` at any time regardless of whether its `delay_until` or `expiration` times have been reached.

Furthermore, the changes to the producer_plugin for the resolution of https://github.com/AntelopeIO/leap/issues/1556 should be enhanced to be aware of when this protocol feature is activated. When it is activated, the `retire_expired_deferred_trxs` function in the producer plugin will not break from the loop if `expired_itr->expiration >= pending_block_time`. It will expire deferred transactions (any order is okay) until it reaches the hard-coded wall-clock deadline for that block. 

`DISABLED_DEFERRED_TRXS_STAGE_2`:

This second protocol feature should depend on the first one (using the dependency mechanism in the protocol feature system to enforce this).

Its activation handler (`on_activation`) should iterate through any remaining deferred transactions in `generated_transaction_multi_index` to refund the RAM paid for the sender of that deferred transaction and to delete the deferred transaction from the state. It should clear out all deferred transactions within the activation handler.

Furthermore, the block validation rules should be changed again to not allow any deferred transactions to be retired in a block (not even as `expired`).

PR #1697, opened 2023-09-29
Implement DISABLE_DEFERRED_TRXS_STAGE_1 and DISABLE_DEFERRED_TRXS_STAGE_2 protocol features

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/1555

Once `DISABLE_DEFERRED_TRXS_STAGE_1` is activated, the behavior of the `send_deferred` and `cancel_deferred` host functions and `canceldelay` native action changes so that they become no-ops. In addition, a block will expire deferred transactions until it reaches the hard-coded wall-clock deadline for that block; any block that retires a deferred transaction with a status other than expired is invalid. Also, a deferred transaction can be retired as expired at any time regardless of whether its delay_until or expiration times have been reached.

`DISABLE_DEFERRED_TRXS_STAGE_2` depends on `DISABLE_DEFERRED_TRXS_STAGE_11. On activation of `DISABLE_DEFERRED_TRXS_STAGE_2`, its activation handler iterates through any remaining deferred transactions to refund the RAM paid for the sender of that deferred transaction and to delete the deferred transaction from the state. All deferred transactions are cleared out within the activation handler.


libraries/testing/tester.cpp
```diff
@@ -1227,6 +1227,17 @@ namespace eosio { namespace testing {
...
+         // Before deferred trxs feature is fully disabled, existing tests involving
+         // deferred trxs need to be exercised to make sure existing behaviors are
+         // maintained. Excluding DISABLE_DEFERRED_TRXS_STAGE_1 and DISABLE_DEFERRED_TRXS_STAGE_2
+         // from full protocol feature list such that existing tests can run.
+         // Remove the following exclusion once eferred trxs feature is fully disabled
+         // and tests are adapted.
+         std::string codename = f.second.codename;
+         if( codename.compare( "DISABLE_DEFERRED_TRXS_STAGE_1" ) == 0 || codename.compare( "DISABLE_DEFERRED_TRXS_STAGE_2" ) == 0 ) {
+            continue;
+         }
```
heifner, line 1239: Instead of this I think it would be better to create a new `setup_policy` that is everything but these two protocol features and use that for tests that need it. Then `preactivate_all_builtin_protocol_features()` means *all*.

Also why `compare` instead of just `==`?

linh2931: I was thinking about introducing a new specific policy too. But when we finally remove the deferred trxs, we need to update a number of places to change the policy back to `all`, instead of only changing one place.

Either way is fine with me. 

Yes, I will change to use `==` for more natural looking. Thanks.

linh2931: I have thought it about more. It is better to let `preactivate_all_builtin_protocol_features` mean really `all`. Will change the code. Thanks.

unittests/delay_tests.cpp
```diff
@@ -1439,7 +1439,7 @@ BOOST_AUTO_TEST_CASE( canceldelay_test ) { try {
 
 // test canceldelay action under different permission levels
 BOOST_AUTO_TEST_CASE( canceldelay_test2 ) { try {
-   validating_tester chain;
+   validating_tester_no_disable_deferrd_trx chain;
```
heifner, line 1442: Need a test that shows that deferred trxs are not allowed (do not validate) in a block after activation.

heifner (changes requested):

libraries/chain/protocol_feature_manager.cpp
```diff
@@ -272,6 +272,40 @@ Adds new cryptographic host functions
 */
             {}
          } )
+         (  builtin_protocol_feature_t::disable_deferred_trxs_stage_1, builtin_protocol_feature_spec{
+            "DISABLE_DEFERRED_TRXS_STAGE_1",
+            fc::variant("8c909017896bab2e294978e9e5cbcb804fa882a52587702ee65b05cf67d59222").as<digest_type>(),
```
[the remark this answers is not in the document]
linh2931: Which `sha` command did you use? I used Linux command `sha256sum` starting with `Builtin` and ends with the newline after `reached`.

libraries/chain/protocol_feature_manager.cpp
```diff
@@ -272,6 +272,40 @@ Adds new cryptographic host functions
...
+            // SHA256 hash of the raw message below within the comment delimiters (do not modify message below).
+/*
+Builtin protocol feature: DISABLE_DEFERRED_TRXS_STAGE_1
+
+Once this first disabling deferred transactions protocol feature is activated,
+the behavior of the send_deferred and cancel_deferred host functions and
+canceldelay native action changes so that they become no-ops. In addition,
+a block will expire deferred transactions until it reaches the hard-coded
+wall-clock deadline for that block; any block that retires a deferred
+transaction with a status other than expired is invalid. Also, a deferred
```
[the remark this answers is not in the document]
linh2931: Oh I misunderstood the requirements. Do you mean the controller should invalid a deferred transaction if it is not an expired one?

libraries/chain/protocol_feature_manager.cpp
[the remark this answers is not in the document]
heifner: https://emn178.github.io/online-tools/sha256.html

libraries/chain/protocol_feature_manager.cpp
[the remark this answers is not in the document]
heifner: Yes.

unittests/delay_tests.cpp
linh2931: A test `disable_deferred_trxs_stage_1_retire_test` was added.

libraries/chain/protocol_feature_manager.cpp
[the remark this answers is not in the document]
linh2931: With the on-line tool, I got the same result as Linux command. Will send you a screenshot offline.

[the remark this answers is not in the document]
heifner: Weird, I get that now also. Not sure what I did wrong the first time. Sorry about that.

libraries/chain/protocol_feature_manager.cpp
```diff
@@ -272,6 +272,40 @@ Adds new cryptographic host functions
...
+transaction can be retired as expired at any time regardless of whether its
+delay_until or expiration times have been reached.
+*/
+            {}
+         } )
+         (  builtin_protocol_feature_t::disable_deferred_trxs_stage_2, builtin_protocol_feature_spec{
+            "DISABLE_DEFERRED_TRXS_STAGE_2",
+            fc::variant("dd53e40a1832078eadec7817f300c2ad20e6d170ecd8103f88b02be944aa3d46").as<digest_type>(),
```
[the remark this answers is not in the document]
heifner: My mistake.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2306,8 +2312,10 @@ void producer_plugin_impl::retire_expired_deferred_trxs(const fc::time_point& de
    auto               expired_itr         = expired_idx.begin();
 
    while (expired_itr != expired_idx.end()) {
-      if (expired_itr->expiration >= pending_block_time) {
-         break; // not expired yet
+      // * Before disable_deferred_trxs_stage_1 is activated, retire only expired deferred trxs.
+      // * After disable_deferred_trxs_stage_1, retire any deferred trxs in any order
+      if (!chain.is_builtin_activated( builtin_protocol_feature_t::disable_deferred_trxs_stage_1) && expired_itr->expiration >= pending_block_time) { // before stage_1 and not expired yet
```
heifner, line 2317: Should cache the value of `chain.is_builtin_activated( builtin_protocol_feature_t::disable_deferred_trxs_stage_1)` outside the while loop.

heifner (changes requested):

heifner (approved):

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -512,6 +515,12 @@ namespace eosio { namespace testing {
       bool validate() { return true; }
    };
 
+   class tester_no_disable_deferrd_trx : public tester {
```
arhag, line 518: Typo in `deferrd`. Should be called `tester_no_disable_deferred_trx`.

libraries/testing/include/eosio/testing/tester.hpp
```diff
@@ -657,6 +666,12 @@ namespace eosio { namespace testing {
       bool                     skip_validate = false;
    };
 
+   class validating_tester_no_disable_deferrd_trx : public validating_tester {
```
arhag, line 669: Typo in `deferrd`. Should be called `validating_tester_no_disable_deferred_trx`.

unittests/deep_mind_tests.cpp
```diff
@@ -49,7 +49,7 @@ struct deep_mind_log_fixture
 
 struct deep_mind_tester : deep_mind_log_fixture, validating_tester
 {
-   deep_mind_tester() : validating_tester({}, &deep_mind_logger) {}
+   deep_mind_tester() : validating_tester({}, &deep_mind_logger, setup_policy::full_but_disable_deferrd_trx) {}
```
arhag, line 52: Why do the deep_mind_tests require deferred transactions to not be disabled?

unittests/currency_tests.cpp
```diff
@@ -18,7 +18,7 @@ using namespace eosio::chain;
 using namespace eosio::testing;
 using namespace fc;
 
-class currency_tester : public validating_tester {
+class currency_tester : public validating_tester_no_disable_deferrd_trx {
```
arhag, line 21: All the currency_tests except `test_deferred_failure` should work fine even with deferred transactions disabled.

I would rather see that one test case handled as a special case (maybe my making the `currency_tester` a template?) so that most of the test cases can run in the proper environment we expect the network to eventually move to which is disabling the deferred transactions.

arhag (changes requested):

unittests/currency_tests.cpp
linh2931: Thanks. Agreed. I was planning to look into further today to adapt those tests which had to use `validating_tester_no_disable_deferred_trx` such that both before and after work.

.cicd/defaults.json
```diff
@@ -4,6 +4,6 @@
        "prerelease":false
     },
     "eossystemcontracts":{
-       "ref":"release/3.1"
+       "ref":"main"
```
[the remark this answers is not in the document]
linh2931: Changed to `release/3.2`, so we won't forget to switch to it when doing Leap `release/5.0`.

libraries/chain/protocol_feature_manager.cpp
```diff
@@ -272,6 +272,40 @@ Adds new cryptographic host functions
...
+            // SHA256 hash of the raw message below within the comment delimiters (do not modify message below).
+/*
+Builtin protocol feature: DISABLE_DEFERRED_TRXS_STAGE_2
+Depends on: DISABLE_DEFERRED_TRXS_STAGE_1
+
+On activation of this second disabling deferred transactions protocol feature,
+its activation handler iterates through any remaining deferred transactions
+to refund the RAM paid for the sender of that deferred transaction and to delete
+the deferred transaction from the state. All deferred transactions are cleared
+out within the activation handler.
```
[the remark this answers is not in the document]
linh2931: Thanks. Your description is much cleaner and without alluding any implementation details.

libraries/testing/tester.cpp
```diff
@@ -1197,7 +1202,8 @@ namespace eosio { namespace testing {
       preactivate_protocol_features(features);
    }
 
-   void base_tester::preactivate_all_builtin_protocol_features() {
+
+   void base_tester::preactivate_all_builtin_protocol_features_common(const std::vector<builtin_protocol_feature_t>& ordered_builtins) {
```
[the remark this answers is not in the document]
linh2931: Thanks. Refactored the code. It is now cleaner, with one fewer function.

unittests/deep_mind_tests.cpp
linh2931: `producer_block_id` differs between different sets of activated protocol features, due to actions to activate the protocol features are different.

unittests/currency_tests.cpp
linh2931: Made the changes. One question: I tried to add a new test for test_proxy which sets delay to 0 and uses `full` policy. Somehow the transfer did not happen.

arhag: The proxy contract will use deferred transactions regardless of whether the delay is set to 0 or not. So it makes sense that the deferred transaction would fail to execute in this branch, regardless of whether the disable deferred transaction protocol features are activated or not.

linh2931: Thanks for your explanation!

arhag (approved):

Issue #1662, opened 2023-09-20
Investigate and fix throughput issues related to read only transactions and  HTTP


ericpassmore:
## Summary 
Initial performance testing of read only transactions indicated read-only transactions are not reaching peak throughput. This issues focuses on the C/C++ code for leap and enabling peak throughput. 

## Related Issues 
Performance testing code and the performance testing harness are covered by #1634

heifner:
Completed by #1702, #1699, #1692, but mostly by #1685

PR #1702, opened 2023-10-02
Reduce the number of thread hops for read-only trxs

heifner:
Optimize read-only trx execution by reducing the number of thread hops for execution.

This branch: https://github.com/AntelopeIO/leap/actions/runs/6365190135 16501 TPS
Compared with [GH-1662-close](https://github.com/AntelopeIO/leap/pull/1699): https://github.com/AntelopeIO/leap/actions/runs/6356928532 15001 TPS


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2659,11 +2657,8 @@ void producer_plugin::log_failed_transaction(const transaction_id_type&    trx_i
 
 // Called from only one read_only thread
 void producer_plugin_impl::switch_to_write_window() {
-   if (_log.is_enabled(fc::log_level::debug)) {
-      auto now = fc::time_point::now();
-      fc_dlog(_log, "Read-only threads ${n}, read window ${r}us, total all threads ${t}us",
-              ("n", _ro_thread_pool_size)("r", now - _ro_read_window_start_time)("t", _ro_all_threads_exec_time_us.load()));
-   }
+   fc_ilog(_log, "Read-only threads ${n}, read window ${r}us, total all threads ${t}us",
+           ("n", _ro_thread_pool_size)("r", fc::time_point::now() - _ro_read_window_start_time)("t", _ro_all_threads_exec_time_us.load()));
```
heifner, line 2661: This is debatable change. Happy to make this `debug` if it seems too chatty.

linh2931: Yes I think it would be too chatty. API operators probably are not interested in the details of window switching.

heifner: Changed to debug level.

linh2931 (approved):

libraries/custom_appbase/include/eosio/chain/application.hpp
```diff
@@ -30,14 +30,38 @@ class priority_queue_executor {
...
+
+   // not thread safe, only call at program startup from the main thread (thread that calls app().exec()
+   void init_main_thread_id() {
+      main_thread_id_ = std::this_thread::get_id();
+   }
+
+   // not thread safe, but as long as set_main_thread_id() only called at program startup from main thread before
+   // other threads are created then this will be safe to access.
+   std::thread::id get_main_thread_id() const {
+      assert(main_thread_id_ != std::thread::id());
```
greg7mdp, line 51: Why not initialize this in the member, so it is initialized when the executor is constructed?
`std::thread::id                    main_thread_id_ { std::thread::id()) };`

heifner: I was not thinking it would always be created on the main thread. But in practice I guess it would be.

heifner: Actually the tests create it on a diff thread. I can re-work those tests.

greg7mdp: I assumed the `Application` was always created on the main thread, but if that's not the case you can ignore my comment.

heifner: In the nomial use-case it is, it just isn't for our tests, but I should be able to fix the tests.

plugins/http_plugin/include/eosio/http_plugin/common.hpp
```diff
@@ -164,8 +164,8 @@ inline auto make_http_response_handler(http_plugin_state& plugin_state, detail::
       plugin_state.bytes_in_flight += payload_size;
 
       // post back to an HTTP thread to allow the response handler to be called from any thread
-      boost::asio::post(plugin_state.thread_pool.get_executor(),
-                        [&plugin_state, session_ptr, code, payload_size, response = std::move(response), content_type]() {
+      boost::asio::dispatch(plugin_state.thread_pool.get_executor(),
```
greg7mdp, line 167: I'm having a hard time figuring out why `dispatch` is appropriate here? Whay can't we just execute the lambda here?

greg7mdp (approved):

plugins/http_plugin/include/eosio/http_plugin/common.hpp
heifner: `dispatch` is the one that allows execution of the lambda here. `post` is the one that make sure it always executes outside this thread.

greg7mdp: Do we really need to call `dispatch` instead of just executing the lambda's code here?

heifner: This can be called from the application thread, in that case we want to do the json parsing on the http thread.

greg7mdp (approved):

PR #1706, opened 2023-10-03
P2P: Advance read pointer when dropping trx

### LINKED ISSUE #1705: Dropping trx when syncing not clearing input buffer
image

Issue introduced by: https://github.com/AntelopeIO/leap/pull/1215

heifner:
Fix issue introduced by #1215 where read buffer not updated when dropping a trx while syncing. Also change log to `debug` level as there is nothing wrong here, just a peer sent you a trx.

Resolves #1705

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3025,7 +3025,8 @@ namespace eosio {
          return true;
       }
       if (my_impl->sync_master->syncing_from_peer()) {
-         peer_wlog(this, "syncing, dropping trx");
+         peer_dlog(this, "syncing, dropping trx");
+         pending_message_buffer.advance_read_ptr( message_length );
          return true;
       }
 


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3025,7 +3025,8 @@ namespace eosio {
          return true;
       }
       if (my_impl->sync_master->syncing_from_peer()) {
-         peer_wlog(this, "syncing, dropping trx");
+         peer_dlog(this, "syncing, dropping trx");
+         pending_message_buffer.advance_read_ptr( message_length );
```
linh2931, line 3029: Should we have a test for this, now that wlog is changed to dlog? (would the wlog have helped identify the failure?)

heifner: Peers should not send you trx when you are syncing. However, they may not realize you are syncing especially when you first connect. Our integration tests will still log this as they run in full debug logging mode.

linh2931: I meant a test that shows the buffer does advance when this happens.

heifner: Unfortunately the net_plugin is not currently setup for unittests. Would be nice to refactor net_plugin to allow for these types of tests. As is there is no way to force net_plugin to send trxs when syncing.

linh2931 (approved):

greg7mdp (approved):

PR #1709, opened 2023-10-03
PH: Disable p2p trx execution on validation and api nodes

heifner:
The performance harness is evaluating performance of BP nodes or API nodes according to what they can individually handle. There is no need for the validation node or the API node to speculatively execute trxs. This removes load from the overall test rig allowing for better results. Also it better represents real-world API configuration.

### DIFF
--- a/tests/PerformanceHarness/performance_test_basic.py
+++ b/tests/PerformanceHarness/performance_test_basic.py
@@ -124,6 +124,7 @@ class PerformanceTestBasic:
 
             def configureValidationNodes():
                 validationNodeSpecificNodeosStr = ""
+                validationNodeSpecificNodeosStr += '--p2p-accept-transactions false '
                 if "v2" in self.nodeosVers:
                     validationNodeSpecificNodeosStr += '--plugin eosio::history_api_plugin --filter-on "*" '
                 else:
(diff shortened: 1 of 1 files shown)


tests/PerformanceHarness/performance_test_basic.py
```diff
@@ -124,6 +124,7 @@ def __post_init__(self):
 
             def configureValidationNodes():
                 validationNodeSpecificNodeosStr = ""
+                validationNodeSpecificNodeosStr += '--p2p-accept-transactions false '
```
linh2931, line 127: Why not remove the statement above and do `validationNodeSpecificNodeosStr = '--p2p-accept-transactions false ' directly?

Same comment for below.

linh2931: You don't have to make the changes. I am approving it now so that it can make to 5.0.

linh2931 (approved):

tests/PerformanceHarness/performance_test_basic.py
oschwaldp-oci: That would be a fine change.  Previously all additions to `validationNodeSpecificNodeosStr` were conditional so it was needed to initialize the variable.  Looks like down below `apiNodeSpecificNodeosStr` could've been initialized directly with the first option instead of to a blank string first.

PR #1718, opened 2023-10-04
Fix: segfault switching between read/write windows

### LINKED ISSUE #1716: Test failure: read-only-trx-parallel-test
Example of timeout: https://github.com/AntelopeIO/leap/actions/runs/6408329747/job/17397449786
Seg-faulting on my machine.

heifner:
* Always execute the lambda after `pop()`ing it from the queue. This is needed as the `read_write` queue is used to post the lambda that switches to the read window. The read window can then start and begin pushing future tasks to the `read_write` queue. This allows multiple threads to access the `read_write` at one time as the original call to switch to the read window is running unlocked, once it finishes it pops from the queue while other threads can access it.
* Modify `read_only_trx_test.py` so it can run in non-debug mode. In debug mode it generates an excessive amount of logging slowing down the test. Running in non-debug allowed the test to more easily trigger the segfault. Also enhanced the test to provide better error output if it does fail.

This fix needs to be back-ported to 4.0.

Resolves #1716


heifner:
Two clean runs:
https://github.com/AntelopeIO/leap/actions/runs/6410560475
https://github.com/AntelopeIO/leap/actions/runs/6410559733

ericpassmore (approved):
LGTM

linh2931 (approved):

greg7mdp (approved):

PR #1719, opened 2023-10-04
[4.0] Fix possible segfault switching between read/write windows

### LINKED ISSUE #1490: Tests: fix plugin_test test failure
https://github.com/AntelopeIO/leap/actions/runs/5857836866/job/15880851171

```
failed to read response from monitor process
    {}
    plugin_test  compile_monitor.cpp:326 get_connection_to_compile_monitor

appbase: exception thrown during plugin "eosio::chain_plugin" initialization.
Miscellaneous exception
warn  2023-08-14T16:05:10.608 plugin_te test_read_only_trx.cpp:124    operator()           ] 3100000 misc_exception: Miscellaneous exception
failed to read response from monitor process
    {}
    plugin_test  compile_monitor.cpp:326 get_connection_to_compile_monitor
rethrow
    {}
    plugin_test  chain_plugin.cpp:1110 plugin_initialize
```

heifner:
* Always execute the lambda after pop()ing it from the queue. This is needed as the read_write queue is used to post the lambda that switches to the read window. The read window can then start and begin pushing future tasks to the read_write queue. This allows multiple threads to access the read_write at one time as the original call to switch to the read window is running unlocked, once it finishes it pops from the queue while other threads can access it. Backport of #1718 
* Fix for producer shutdown to prevent segfault and deadlock on shutdown. Backport of 1368ecd0b72f1387f0545c6deef9bd494b1aa7fe

Backport of #1718

Resolves #1490

### DIFF
--- a/libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
+++ b/libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
@@ -13,6 +13,12 @@ class exec_pri_queue : public boost::asio::execution_context
 {
 public:
 
+   void stop() {
+      std::lock_guard g( mtx_ );
+      exiting_blocking_ = true;
+      cond_.notify_all();
+   }
+
    void enable_locking(uint32_t num_threads, std::function<bool()> should_exit) {
       assert(num_threads > 0 && num_waiting_ == 0);
       lock_enabled_ = true;
--- a/libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
+++ b/libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
@@ -53,17 +53,6 @@ public:
       handlers_ = prio_queue();
    }
 
-   // only call when no lock required
-   bool execute_highest()
-   {
-      if( !handlers_.empty() ) {
-         handlers_.top()->execute();
-         handlers_.pop();
-      }
-
-      return !handlers_.empty();
-   }
-
 private:
    // has to be defined before use, auto return type
    auto pop() {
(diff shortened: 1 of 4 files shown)


libraries/custom_appbase/include/eosio/chain/exec_pri_queue.hpp
```diff
@@ -68,6 +63,19 @@ class exec_pri_queue : public boost::asio::execution_context
...
+   {
+      if( !handlers_.empty() ) {
+         auto t = pop();
+         bool empty = handlers_.empty();
+         t->execute();
+         return !empty;
+      }
+
+      return false;
+   }
```
greg7mdp, line 77: No need to change, but just to chat I would have written it like that.

```suggestion
   bool execute_highest()
   {
      bool empty = handlers_.empty();
      if( !empty ) {
         auto t = pop();
         empty = handlers_.empty();
         t->execute();
      }

      return !empty;
   }
```

greg7mdp (approved):

linh2931 (approved):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1389,20 +1389,13 @@ void producer_plugin::plugin_startup()
...
-     chain_plugin::handle_bad_alloc();
-   } catch(const fc::exception& e) {
-      edump((e.to_detail_string()));
-   } catch(const std::exception& e) {
-      edump((fc::std_exception_wrapper::from_current_exception(e).to_detail_string()));
-   }
-
+   boost::system::error_code ec;
+   my->_timer.cancel(ec);
+   boost::system::error_code ro_ec;
```
spoonincode, line 1394: there doesn't appear to be a need for the second `ro_ec` here

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1389,20 +1389,13 @@ void producer_plugin::plugin_startup()
...
+   my->_ro_timer.cancel(ro_ec);
+   app().executor().stop();
```
spoonincode, line 1396: Can you explain this line some more? In what cases would we make it here and the appbase executor would still be running?

spoonincode (approved):

PR #1728, opened 2023-10-06
[5.0] Test: Retry set contract on failure

### LINKED ISSUE #1501: Test failure: nodeos_extra_packed_data_test failed
https://github.com/AntelopeIO/leap/actions/runs/5871403809/job/15921525923?pr=1498

2032 - nodeos_extra_packed_data_test (Failed)

2023-08-15T21:03:57.805143              cmd: /__w/leap/leap/build/bin/cleos --url http://localhost:8788  --wallet-url http://localhost:9899 --no-auto-keosd get transaction_trace 423f9bf05fc9b864d929eda83824354fa5ca368df9aec3086ed390635412bcbe
2023-08-15T21:03:57.827271               FAILURE - Exception during "get transaction_trace". Exception message: Error 3200005: http request fail
Error Details:
Error code 404
: {"code":404,"message":"Trace API: transaction id missing in the transaction id log files","error":{"code":0,"name":"","what":"","details":[]}}

.  stdout: .  cmd Duration=0.021 sec. Context: (transaction id=423f9bf05fc9b864d929eda83824354fa5ca368df9aec3086ed390635412bcbe)
2023-08-15T21:03:57.827657               ERROR: Exception during "get transaction_trace". Exception message: Error 3200005: http request fail
Error Details:
Error code 404
: {"code":404,"message":"Trace API: transaction id missing in the transaction id log files","error":{"code":0,"name":"","what":"","details":[]}}

.  stdout: .  cmd Duration=0.021 sec. Context: (transaction id=423f9bf05fc9b864d929eda83824354fa5ca368df9aec3086ed390635412bcbe)
2023-08-15T21:03:57.827937  Test failed.

heifner:
Occasionally `nodeos_extra_packed_data_test.py` test would fail because the setting of the `eosio.system` contract would take longer than the max trx cpu time. I think this test was more likely to fail in this way because it spawns more nodeos than other tests: `tests/nodeos_extra_packed_data_test.py -v -p 8`. There is no reason for this test to spawn so many producers.

Added a retry to `publishContract` to give the `set contract` a few tries before considering it a true failure.

Resolves #1501

### DIFF
--- a/tests/TestHarness/transactions.py
+++ b/tests/TestHarness/transactions.py
@@ -160,6 +160,7 @@ class Transactions(NodeosQueries):
         cmd += "" if abiFile is None else (" " + abiFile)
         if Utils.Debug: Utils.Print("cmd: %s" % (cmd))
         retries = 0
+        trans=None
         while retries < retryNum:
             trans=None
             if Utils.Debug and retries > 0:
--- a/tests/TestHarness/transactions.py
+++ b/tests/TestHarness/transactions.py
@@ -199,6 +199,8 @@ class Transactions(NodeosQueries):
                     return None
 
             NodeosQueries.validateTransaction(trans)
+            if not waitForTransBlock:
+                return trans
             transId=NodeosQueries.getTransId(trans)
             if self.waitForTransactionInBlock(transId, timeout=5):
                 break
--- a/tests/TestHarness/transactions.py
+++ b/tests/TestHarness/transactions.py
@@ -152,48 +152,58 @@ class Transactions(NodeosQueries):
         return popen, cmdArr
 
     # publish contract and return transaction as json object
-    def publishContract(self, account, contractDir, wasmFile, abiFile, waitForTransBlock=False, shouldFail=False, sign=False):
+    def publishContract(self, account, contractDir, wasmFile, abiFile, waitForTransBlock=True, shouldFail=False, sign=False, retryNum:int=5):
+        assert(isinstance(retryNum, int))
         signStr = NodeosQueries.sign_str(sign, [ account.activePublicKey ])
-        cmd="%s %s -v set contract -j %s %s %s" % (Utils.EosClientPath, self.eosClientArgs(), signStr, account.name, contractDir)
+        cmd=f"{Utils.EosClientPath} {self.eosClientArgs()} -v set contract -j -f {signStr} {account.name} {contractDir}"
         cmd += "" if wasmFile is None else (" "+ wasmFile)
         cmd += "" if abiFile is None else (" " + abiFile)
         if Utils.Debug: Utils.Print("cmd: %s" % (cmd))
         trans=None
-        start=time.perf_counter()
-        try:
-            trans=Utils.runCmdReturnJson(cmd, trace=False)
-            self.trackCmdTransaction(trans)
-            if Utils.Debug:
-                end=time.perf_counter()
-                Utils.Print("cmd Duration: %.3f sec" % (end-start))
-        except subprocess.CalledProcessError as ex:
-            if not shouldFail:
-                end=time.perf_counter()
-                out=ex.output.decode("utf-8")
-                msg=ex.stderr.decode("utf-8")
-                Utils.Print("ERROR: Exception during set contract. stderr: %s.  stdout: %s.  cmd Duration: %.3f sec." % (msg, out, end-start))
-                return None
-            else:
-                retMap={}
-                retMap["returncode"]=ex.returncode
-                retMap["cmd"]=ex.cmd
-                retMap["output"]=ex.output
-                retMap["stderr"]=ex.stderr
-                return retMap
+        retrys = 0
+        while retrys < retryNum:
+            if Utils.Debug and retrys > 0:
+                Utils.Print(f"Retrying: {cmd}")
+            retrys = retrys + 1
+            start=time.perf_counter()
+            try:
+                trans=Utils.runCmdReturnJson(cmd, trace=False)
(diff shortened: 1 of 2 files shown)


linh2931 (approved):

tests/TestHarness/transactions.py
```diff
@@ -152,48 +152,58 @@ def transferFundsAsync(self, source, destination, amountStr, memo="memo", force=
...
+                    retMap["output"]=bytes(str(trans),'utf-8')
+                    return retMap
+                else:
+                    Utils.Print("ERROR: The publish contract did not fail as expected.")
+                    return None
+
+            NodeosQueries.validateTransaction(trans)
+            transId=NodeosQueries.getTransId(trans)
+            if self.waitForTransactionInBlock(transId, timeout=5):
+                break
```
oschwaldp-oci, line 204: This no longer supports `waitForTransBlock=False` correctly.

oschwaldp-oci (changes requested):

tests/TestHarness/transactions.py
```diff
@@ -152,48 +152,60 @@ def transferFundsAsync(self, source, destination, amountStr, memo="memo", force=
         return popen, cmdArr
 
     # publish contract and return transaction as json object
-    def publishContract(self, account, contractDir, wasmFile, abiFile, waitForTransBlock=False, shouldFail=False, sign=False):
+    def publishContract(self, account, contractDir, wasmFile, abiFile, waitForTransBlock=True, shouldFail=False, sign=False, retryNum:int=5):
```
oschwaldp-oci, line 155: Changing the default `waitForTransBlock` value here could have repercussions in the tests where it was assumed that the default is `False` and could be relying on that.  I might recommend leaving the default as `False` but setting it to `True` in your test case you're trying to fix.  That way you don't have to audit every instance of the tests that use `publishContract`.

oschwaldp-oci (changes requested):

tests/TestHarness/transactions.py
heifner: I did audit all calls. I thought about removing the option actually because it should always be `True`, but decided the bigger diff was not worth it. There is only one case were we were not passing `True` and it should be true.

oschwaldp-oci (approved):

tests/TestHarness/transactions.py
```diff
@@ -152,48 +152,60 @@ def transferFundsAsync(self, source, destination, amountStr, memo="memo", force=
...
+        assert(isinstance(retryNum, int))
         signStr = NodeosQueries.sign_str(sign, [ account.activePublicKey ])
-        cmd="%s %s -v set contract -j %s %s %s" % (Utils.EosClientPath, self.eosClientArgs(), signStr, account.name, contractDir)
+        cmd=f"{Utils.EosClientPath} {self.eosClientArgs()} -v set contract -j -f {signStr} {account.name} {contractDir}"
         cmd += "" if wasmFile is None else (" "+ wasmFile)
         cmd += "" if abiFile is None else (" " + abiFile)
         if Utils.Debug: Utils.Print("cmd: %s" % (cmd))
         trans=None
```
greg7mdp, line 162: I can't tell if it is needed, but I think we should move `trans=None` in the while loop.

greg7mdp (approved):

tests/TestHarness/transactions.py
```diff
@@ -152,48 +152,60 @@ def transferFundsAsync(self, source, destination, amountStr, memo="memo", force=
...
-                retMap["output"]=bytes(str(trans),'utf-8')
-                return retMap
-            else:
-                Utils.Print("ERROR: The publish contract did not fail as expected.")
-                return None
-
-        NodeosQueries.validateTransaction(trans)
-        return self.waitForTransBlockIfNeeded(trans, waitForTransBlock, exitOnError=False)
+        retrys = 0
+        while retrys < retryNum:
```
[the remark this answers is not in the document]
heifner: Yes, I grew up in southern Illinois and can barely speak/read/write English.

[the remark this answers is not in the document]
heifner: Fixed.

tests/TestHarness/transactions.py
heifner: Not needed, but doesn't hurt. Moved into while.

tests/TestHarness/transactions.py
```diff
@@ -152,48 +152,60 @@ def transferFundsAsync(self, source, destination, amountStr, memo="memo", force=
...
-                return retMap
-            else:
-                Utils.Print("ERROR: The publish contract did not fail as expected.")
-                return None
-
-        NodeosQueries.validateTransaction(trans)
-        return self.waitForTransBlockIfNeeded(trans, waitForTransBlock, exitOnError=False)
+        retries = 0
+        while retries < retryNum:
+            trans=None
```
oschwaldp-oci, line 164: `trans=None` needs to be outside the `while` for the `return trans` at the end of the function line 208 otherwise you'll get an exception because `trans` has not been defined.

oschwaldp-oci (changes requested):

tests/TestHarness/transactions.py
oschwaldp-oci: It doesn't make too much sense, but in theory `retries < retryNum` could be false at the beginning and the while loop would never be exercised in the pathological case of `retryNum=0`.

oschwaldp-oci (changes requested):

oschwaldp-oci (approved):

PR #1732, opened 2023-10-06
[5.0] P2P: Modify log level for message to avoid confusion

### LINKED ISSUE #1507: The "bad block exception" logs as an error
The following warning happens in the syslog regularly. But logged as an "error".
Probably, It could be reduced to "warn" instead of "error".
image

heifner:
This PR conforms the `net_plugin` logging to meet the objectives set forth in: https://github.com/AntelopeIO/leap/blob/main/docs/01_nodeos/06_logging/01_logging-levels.md

Specifically many `error` level messages were changed to `warn`, `info`, or `debug`.

Resolves #1507

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1296,7 +1296,7 @@ namespace eosio {
       boost::system::error_code ec;
       socket->set_option( nodelay, ec );
       if( ec ) {
-         peer_elog( this, "connection failed (set_option): ${e1}", ( "e1", ec.message() ) );
+         peer_wlog( this, "connection failed (set_option): ${e1}", ( "e1", ec.message() ) );
          close();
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -2939,7 +2940,7 @@ namespace eosio {
          }
 
       } catch( const fc::exception& e ) {
-         peer_elog( this, "Exception in handling message: ${s}", ("s", e.to_detail_string()) );
+         peer_wlog( this, "Exception in handling message: ${s}", ("s", e.to_detail_string()) );
```
greg7mdp, line 2943: I think "in handling message" is a bit strange, but no big deal.

greg7mdp (approved):

linh2931 (approved):

Issue #1730, opened 2023-10-06
unlinkable blocks syncing with 5.0-rc1


heifner:
This needs to be restored

https://github.com/AntelopeIO/leap/pull/1643/files#diff-42816449499831004c7f5a1e66370fdd9fa0036312ad44da12cb963523e9c6a5R249-R250

It prevents delayed transactions from being processed. For example: 
https://bloks.io/block/320405089
[527bbf9029f604e7021be729c858fb520001c5c5413cbee2fc08dbcf735534bc](https://bloks.io/transaction/527bbf9029f604e7021be729c858fb520001c5c5413cbee2fc08dbcf735534bc)

PR #1734, opened 2023-10-06
[5.0] Fix unlinked blocks caused by deferred trx removal

linh2931:
Resolved https://github.com/AntelopeIO/leap/issues/1730.

Unlinked block error was encountered in a replay of mainnet. It was caused by blocking any deferred trxs by https://github.com/AntelopeIO/leap/pull/1643.

This PR
* restored deferred trxs handling in transaction_context
* updated related tests

I have run replay tests from the mainnet and have passed the failed block 320405089 and trx 527bbf9029f604e7021be729c858fb520001c5c5413cbee2fc08dbcf735534bc,
starting from snapshot-2023-07-12-16-eos-v6-0319983883. The previously failed block was at Jul-14-2023, 10:33:36 PM.

I will do more replay tests over the weekend.

### DIFF
--- a/libraries/chain/transaction_context.cpp
+++ b/libraries/chain/transaction_context.cpp
@@ -247,7 +247,9 @@ namespace eosio { namespace chain {
                                                  uint64_t packed_trx_prunable_size )
    {
       const transaction& trx = packed_trx.get_transaction();
-      EOS_ASSERT( trx.delay_sec.value == 0, transaction_exception, "transaction cannot be delayed" );
+      if ( is_transient() ) {
+         EOS_ASSERT( trx.delay_sec.value == 0, transaction_exception, "transaction cannot be delayed" );
+      }
       if( trx.transaction_extensions.size() > 0 ) {
          disallow_transaction_extensions( "no transaction extensions supported yet for input transactions" );
       }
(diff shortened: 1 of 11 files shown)


libraries/chain/transaction_context.cpp
```diff
@@ -247,7 +247,9 @@ namespace eosio { namespace chain {
                                                  uint64_t packed_trx_prunable_size )
    {
       const transaction& trx = packed_trx.get_transaction();
-      EOS_ASSERT( trx.delay_sec.value == 0, transaction_exception, "transaction cannot be delayed" );
+      if ( is_transient() ) {
+         EOS_ASSERT( trx.delay_sec.value == 0, transaction_exception, "transaction cannot be delayed" );
+      }
```
heifner, line 252: Once `DISABLE_DEFERRED_TRXS_STAGE_2` is activated blocks should not validate that contain a `delayed` trx. We should have a test for this as well.

heifner (changes requested):

linh2931:
I started another test from a January 2022 snapshot. I print out when a transaction's `delay_sec` is greater than 0. There were a number of them. The test has been running for 2 days and is going fine.

libraries/chain/transaction_context.cpp
heifner: I originally said stage 2, but I think we want this for stage 1. My understanding is stage 1 is stop deferred/delayed trxs and stage 2 is cleanup.

linh2931: You are right. After stage 1, no new deferred trxs are allowed; existing ones are starting to be retired as expired and
they should not reach block validation.
After stage 2, the whole system should not have any delayed trxs left.
It's better to assert after stage 1. Will change.

arhag (approved):
It seems like we can now undo the `modify_gto_for_canceldelay_test` hack if we wanted to since libtester allows for delayed transactions can be created prior to stage 1 activation.

But I am good with the changes in this PR as is.

linh2931:
> It seems like we can now undo the `modify_gto_for_canceldelay_test` hack if we wanted to since libtester allows for delayed transactions to be properly created prior to stage 1 activation.
> 
> But I am good with the changes in this PR as is.

Thanks. I will undo it on `main` later.

heifner (approved):

arhag (approved):

heifner (approved):

PR #1741, opened 2023-10-10
[5.0] Support throttling block syncing to peers

### LINKED ISSUE #1295: P2P: Ability to limit peer bandwidth consumption

jgiszczak:
Peer to peer listen ports may now apply a block sync throttle.  Each occurrence of `p2p-listen-endpoint` has an independent throttle specification.  Here's the updated help text:

```
--p2p-listen-endpoint arg (=0.0.0.0:9876:0)
                                        The actual host:port[:<rate-cap>] used
                                        to listen for incoming p2p connections.
                                        May be used multiple times.   The
                                        optional rate cap will limit block sync
                                        bandwidth to the specified rate.  A
                                        number alone will be interpreted as
                                        bytes per second.  The number may be
                                        suffixed with units.  Supported units
                                        are:   'B/s', 'KB/s', 'MB/s, 'GB/s',
                                        'TB/s', 'KiB/s', 'MiB/s', 'GiB/s',
                                        'TiB/s'.  Transactions and blocks
                                        outside of sync mode are not throttled.
                                          Examples:
                                            192.168.0.100:9876:1MiB/s
                                            node.eos.io:9876:1512KB/s
                                            node.eos.io:9876:0.5GB/s
                                            [2001:db8:85a3:8d3:1319:8a2e:370:7348]:9876:250KB/s
```
Parsing of the rate cap accepts fractional numbers expressed as decimals.  The throttle rate is in bytes per second.  Per the help text, shorthand suffixes are supported, and transactions and blocks being propagated across the peer to peer network between synchronized nodes are not throttled.  Only blocks transmitted to a syncing node are throttled.  A throttle of `0` or `0B/s` means unthrottled, and is the default.

The limit for specifying a rate cap is 30 characters, including the colon field separator.  This is sufficient characters to specify any value in the 64 bit range of `size_t` with room left for a suffix.  If the rate cap does not parse or the value multiplied by the suffix exceeds 18,446,744,073,709,551,615 an exception will be thrown and the node will not start.

IPv6 addresses are supported for listen endpoints.  They must be in square bracket format: `[<ipv6 address>]:port` optionally followed by `:<rate-cap>`.

> [!NOTE]
> Inbound connections from IPs which are configured `p2p-peer-address`es are exempt from throttle rate caps.  Care should be taken in NAT environments to avoid inadvertently exempting connections due to overlapping subnets.

Throttling is stable even at exceptionally low byterates, but of course on a busy network if the throttle is less than the average block size on the network, clients using that listen port will never catch up to the head block.  Such configurations are allowed but not recommended.

### Suggested network topology
Together with #1411 which allows multiple listen endpoints, an edge node for public peering might be configured as follows:
Leap Throttle Sample Network Diagram (truncated)

Resolves #1295.

Retargetting #1540 to `release/5.0`.


plugins/net_plugin/net_plugin.cpp
```diff
@@ -1667,14 +1739,24 @@ namespace eosio {
          sb = cc.fetch_block_by_number( num ); // thread-safe
       } FC_LOG_AND_DROP();
       if( sb ) {
-         enqueue_block( sb, true );
+         if( block_sync_rate_limit > 0 && peer_syncing_from_us ) {
```
dimas1185, line 1742: please add comment in a code of what is happening here. It took me a while to understand your approach so I think having a comment will be helpful

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,216 @@
...
+    cluster.launchUnstarted(2)
+
+    throttledNode = cluster.getNode(3)
+    while True:
+        try:
+            response = throttlingNode.processUrllibRequest('prometheus', 'metrics', returnType=ReturnType.raw, printReturnLimit=16).decode()
+        except urllib.error.URLError:
+            # catch ConnectionRefusedEror waiting for node to finish startup and respond
+            time.sleep(0.5)
+            continue
```
dimas1185, line 128: I don't like that potentially it can go into endless loop. I suggest to add a counter, assert or something

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,216 @@
...
+        else:
+            if len(response) < 100:
+                # tolerate HTTPError as well (method returns only the exception code)
+                continue
```
dimas1185, line 132: same here and for other endless loop `continue`

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,216 @@
...
+            connPorts = prometheusHostPortPattern.findall(response)
+            if len(connPorts) < 3:
+                # wait for node to be connected
+                time.sleep(0.5)
+                continue
```
dimas1185, line 137: potential endless loop

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,216 @@
...
+            if time.time() > clusterStart + 30: errorExit('Timed out')
+            break
+
+    while True:
+        try:
+            response = throttledNode.processUrllibRequest('prometheus', 'metrics', returnType=ReturnType.raw, printReturnLimit=16).decode()
+        except urllib.error.URLError:
+            # catch ConnectionRefusedError waiting for node to finish startup and respond
+            time.sleep(0.5)
+            continue
```
dimas1185, line 157: potential endless loop

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,216 @@
...
+        else:
+            if len(response) < 100:
+                # tolerate HTTPError as well (method returns only the exception code)
+                time.sleep(0.5)
+                continue
```
dimas1185, line 162: potential endless loop

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,216 @@
...
+            connPorts = prometheusHostPortPattern.findall(response)
+            if len(connPorts) < 2:
+                # wait for sending node to be connected
+                continue
```
dimas1185, line 166: potential endless loop

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,216 @@
...
+            Print('Throttling Node Start State')
+            throttlingNodePortMap = {port: id for id, port in connPorts}
+            startSyncThrottlingBytesSent = extractPrometheusMetric(throttlingNodePortMap['9879'],
+                                                                    'block_sync_bytes_sent',
+                                                                    response)
+            startSyncThrottlingState = extractPrometheusMetric(throttlingNodePortMap['9879'],
+                                                               'block_sync_throttling',
+                                                               response)
+            Print(f'Start sync throttling bytes sent: {startSyncThrottlingBytesSent}')
+            Print(f'Start sync throttling node throttling: {"True" if startSyncThrottlingState else "False"}')
```
dimas1185, line 147: do we want some asserts instead(or in addition) to those `Print` statements?

tests/p2p_sync_throttle_test.py
```diff
@@ -0,0 +1,216 @@
...
+    response = throttledNode.processUrllibRequest('prometheus', 'metrics', exitOnError=True, returnType=ReturnType.raw, printReturnLimit=16).decode()
+    Print('Throttled Node End State')
+    endSyncThrottledBytesReceived = extractPrometheusMetric(throttledNodePortMap['9878'],
+                                                            'block_sync_bytes_received',
+                                                            response)
+    Print(f'End sync throttled bytes received: {endSyncThrottledBytesReceived}')
+    throttlingElapsed = endThrottlingSync - clusterStart
+    throttledElapsed = endThrottledSync - clusterStart
+    Print(f'Unthrottled sync time: {throttlingElapsed} seconds')
+    Print(f'Throttled sync time: {throttledElapsed} seconds')
```
dimas1185, line 208: Can we safely check here that `throttlingElapsed` is greater than `throttledElapsed` plus some threshold?
I assume it can't be 1 sec and can't be 30 sec. So it can be like 5 or 10 seconds or so.

dimas1185 (approved):

tests/p2p_sync_throttle_test.py
jgiszczak: Revise with counter.

tests/p2p_sync_throttle_test.py
jgiszczak: Revise with counter.

tests/p2p_sync_throttle_test.py
jgiszczak: Revised with counter.

tests/p2p_sync_throttle_test.py
jgiszczak: Revised with counter.

tests/p2p_sync_throttle_test.py
jgiszczak: Revised with counter.

plugins/net_plugin/net_plugin.cpp
jgiszczak: Comment added.

tests/p2p_sync_throttle_test.py
jgiszczak: There is an assert.  It's a single check to be sure the node was throttled at some point.  Line 219.

tests/p2p_sync_throttle_test.py
jgiszczak: In an effort to keep the test duration reasonable, we're not attempting to measure a time difference.  The test is short enough that the variance in the timing of net connections being established and the Prometheus metrics being updated (both of which happen only once every three seconds) is enough that the test is unreliable when only a time difference is measured.  As hardware improves, that will only get worse.

Issue #1743, opened 2023-10-10
Add context to contract exceptions


YaroShkvorets:
## Problem

There is that annoying problem with errors coming from smart contracts.
When your smart contract makes inline action calls to external contracts and one of these actions fails, you just get an error with zero context, so you can’t tell which smart contract and action that error came from. You can probably recognize your own errors but not some random errors coming from other contracts.

I.e. you only get this:
```
Error 3070002: Runtime Error Processing WASM
Error Details:
access violation
pending console output:
Stack Trace:
executor.cpp:257 execute
apply_context.cpp:124 exec_one
```

Assertion "check" failures also have this problem:

```
Error 3050003: eosio_assert_message assertion failure
Error Details:
assertion failure with message: read
pending console output:
Stack Trace:
cf_system.cpp:22 eosio_assert_message
apply_context.cpp:124 exec_one
```

This makes debugging quite painful.

## Proposal

Add additional context to the error message:
- smart contract account
- action name
- receiver (to distinguish action calls from notification handlers)

PR #1744, opened 2023-10-10
Add action context to contract error message

YaroShkvorets:
This PR addresses #1743

I'm adding `contract::action @ receiver` in front of `pending console output` but let me know if there are other considerations.

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -131,7 +131,7 @@ void apply_context::exec_one()
                }
             }
          }
-      } FC_RETHROW_EXCEPTIONS( warn, "pending console output: ${console}", ("console", _pending_console_output) )
+      } FC_RETHROW_EXCEPTIONS( warn, "${account}::${action} @ ${receiver} pending console output: ${console}", ("console", _pending_console_output)("account", act->account)("action", act->name)("receiver", receiver) )
 
       if( control.is_builtin_activated( builtin_protocol_feature_t::action_return_value ) ) {
          act_digest =   generate_action_digest(


libraries/chain/apply_context.cpp
```diff
@@ -131,7 +131,7 @@ void apply_context::exec_one()
                }
             }
          }
-      } FC_RETHROW_EXCEPTIONS( warn, "pending console output: ${console}", ("console", _pending_console_output) )
+      } FC_RETHROW_EXCEPTIONS( warn, "${account}::${action} @ ${receiver} pending console output: ${console}", ("console", _pending_console_output)("account", act->account)("action", act->name)("receiver", receiver) )
```
heifner, line 134: `cleos` outputs this as `${receiver} <= ${account}::${action}`. I'm not sure it matters much but would be interested if anyone has a preference. @arhag

YaroShkvorets: I like that.

arhag: I don't have a strong preference how those three pieces of information should be presented. I agree it is a great developer experience improvement to include it though in whichever from. 

I'm wondering if we should also include the `action_ordinal` to disambiguate which one it comes from if there were multiple instances of the same action and receiver. But then again, even without this change, the transaction trace should already have enough information to determine which exact action within the trace the error came from.

YaroShkvorets: Changed to `${receiver} <= ${account}::${action}`.

I feel like the meaning of `action_ordinal` number would be hard to figure out for the developers, so would be more confusing than helpful. But maybe it's just me.

heifner:
Thank you @YaroShkvorets

ericpassmore:
Note:start
group: CONTRACTS
category: LOGGING
summary: Improve debugging for inline actions by adding additional context of the account that owns the action, the action that failed, and the receiver (to distinguish action calls from notification handlers)
Note:end

YaroShkvorets:
Curious @heifner is it too late to merge this into `release/5.0` branch?

heifner:
> Curious @heifner is it too late to merge this into `release/5.0` branch?

@bhazzard

YaroShkvorets:
Thanks. Opened: https://github.com/AntelopeIO/leap/pull/1902

PR #1752, opened 2023-10-11
[5.0] Fix compiler warning

heifner:
Fix compiler warning: `lambda capture 'return_failure_traces' is not used`


linh2931 (approved):
Sorry for this. Thanks for fixing it.

greg7mdp (approved):

PR #1768, opened 2023-10-12
Support arbitrary labels on p2p peers.

jgiszczak:
This PR creates a new configuration file and command line option and a new net API endpoint for labeling peers.  For the configuration file and command line options, a `p2p-peer-address` option must exist first.  It can then be labeled using `p2p-peer-label`.

```
p2p-peer-address=p2p.eos.io:9876
p2p-peer-label=p2p.eos.io:9876:Favorite peer
```
The new API endpoint is available at
`/v1/net/label`
and accepts the same format as the configuration option.

Peer labels are available as a new field with key `"label"` in the JSON output of  the `/v1/net/connections` API endpoint and in the Prometheus metrics for each peer.  Peer labels can accept an arbitrary string of arbitrary length.  They are not transmitted over the network to other peers.

For peer addresses which are restricted to only transactions or only blocks, the full address must be specified to match:
```
p2p-peer-address=p2p.trx.eos.io:9876:trx
p2p-peer-label=p2p.trx.eos.io:9876:trx:transaction only peer
```
The label text will include the restriction as prefix, e.g. `trx:transaction only peer`.

### DIFF
--- a/plugins/net_api_plugin/net_api_plugin.cpp
+++ b/plugins/net_api_plugin/net_api_plugin.cpp
@@ -57,6 +57,8 @@ void net_api_plugin::plugin_startup() {
    app().get_plugin<http_plugin>().add_async_api({
        CALL_WITH_400(net, net_rw, net_mgr, connect,
             INVOKE_R_R(net_mgr, connect, std::string), 201),
+       CALL_WITH_400(net, net_rw, net_mgr, label,
+            INVOKE_R_R(net_mgr, label, std::string), 201),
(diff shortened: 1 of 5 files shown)


bhazzard:
In the Antelope Node Operators Roundtable, we discussed that it might be worth refactoring the p2p configuration to improve readability and to avoid parsing complexity as we add more options to this configuration. Given that this won't be in 5.X, it makes sense to add scope.

ericpassmore:
significant changes in code, closing this PR.

Issue #1485, opened 2023-08-11
Tests: fix test_resmon_plugin test failure


heifner:
https://github.com/AntelopeIO/leap/actions/runs/5834112382/job/15823238447

spoonincode:
saw again on
https://github.com/AntelopeIO/leap/actions/runs/6180131909/job/16776270770

pretty weird looking

spoonincode:
https://github.com/AntelopeIO/leap/actions/runs/6395708177/job/17360355859

linh2931:
https://github.com/AntelopeIO/leap/actions/runs/6424323959/job/17444943478?pr=1726

PR #1774, opened 2023-10-13
[4.0] hardening resource monitor manager plugin shutdown handling

linh2931:
https://github.com/AntelopeIO/leap/issues/1485 reports occasional resource manager plugin test failures. The failure is shown as
`/__w/leap/leap/plugins/resource_monitor_plugin/test/test_resmon_plugin.cpp(147): fatal error: in "resmon_plugin_tests/startupNormal": unexpected exception thrown by plugin_startup({"/tmp"})`
without any other additional information.

Further investigation reveals the scheduled monitor timer task is not cancelled during plugin shutdown and currently monitor task might be still running before the thread exits. This PR hardens the plugin  shutdown process. 

Resolves https://github.com/AntelopeIO/leap/issues/1485


plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -6,6 +6,8 @@
 #include <eosio/chain/application.hpp>
 #include <eosio/chain/exceptions.hpp>
 
+#include <thread>
```
heifner, line 9: This all seems overly complicated. 
You should be able to simply add a mutex for the timer and call cancel on it in `stop()`. Or move the thread and ctx into file_space_handler and order it so it destructs appopriately.
Also please update to use `named_thread_pool`  of size 1.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -131,26 +133,60 @@ namespace eosio::resource_monitor {
...
+      space_monitor_task_running = false;
+   }
+
+   // called on main thread from plugin shutdown()
+   void stop() {
+      // signalling space_monitor_loop that plugin is being stopped
+      stopping = true;
+
+      // cancel outstanding scheduled space_monitor_loop task
+      timer.cancel();
```
heifner, line 176: This is not thread safe as-is. `timer.cancel()` is not thread safe. You need a mutex to here to access `timer`.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -131,26 +133,60 @@ namespace eosio::resource_monitor {
...
       update_warning_interval_counter();
 
       timer.expires_from_now( boost::posix_time::seconds( sleep_time_in_secs ));
-
+      if (stopping) {
+         space_monitor_task_running = false;
+         return;
+      }
       timer.async_wait([this](auto& ec) {
          if ( ec ) {
```
heifner, line 156: Should check if it is aborted and not log as aborted means it was canceled.

heifner (changes requested):

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
linh2931: Was trying to avoid locking. Will change.

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -5,19 +5,51 @@
...
+            for (bfs::directory_iterator itr(dir); itr != bfs::directory_iterator(); ++itr) {
+               if (bfs::is_directory(itr->path())) {
+                  add_file_system( itr->path() );
+               }
+            }
+         }
+
+         thread_pool.start(thread_pool_size,
+            []( const fc::exception& e ) {
+              elog("Exception in read-only thread pool, exiting: ${e}", ("e", e.to_detail_string()) );
```
heifner, line 37: s/read-only/resource monitor/

heifner (changes requested):

plugins/resource_monitor_plugin/include/eosio/resource_monitor_plugin/file_space_handler.hpp
```diff
@@ -5,19 +5,51 @@
...
+      }
+
+      // called on main thread from plugin shutdown()
+      void stop() {
+         {
+            std::lock_guard<std::mutex> g( timer_mtx );
+            boost::system::error_code ec;
+            timer.cancel(ec);
+         }
+         thread_pool.stop();
```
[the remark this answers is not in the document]
linh2931: Thanks for your insight. I updated the code.

spoonincode (approved):

heifner (approved):

PR #1791, opened 2023-10-17
[5.0] P2P: Fix: Throttling of last block of request caused lost block

### LINKED ISSUE #1776: Test Failure: p2p_sync_throttle_test
https://github.com/AntelopeIO/leap/actions/runs/6513109793/job/17692267461

heifner:
When throttling was enabled on a connection, the `net_plugin` would assume the last block of a `peer_requested` was sent and reset the `peer_requested` indicating no other sync blocks need to be sent to the peer.  This caused the block to not be sent or rescheduled to be sent later resulting in the peer not receiving the block. The peer would have to wait for a full timeout before requesting the required block from a different peer.

Resolves #1776

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -1732,10 +1732,6 @@ namespace eosio {
          peer_dlog( this, "enqueue sync block ${num}", ("num", peer_requested->last + 1) );
       }
       uint32_t num = peer_requested->last + 1;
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):
I think I'd get rid of `num`, and use `cc.fetch_block_by_number( peer_requested->last + 1 );` and `peer_requested->last` below, but it is a matter of preference.

linh2931 (approved):

PR #1795, opened 2023-10-18
automatically create `experimental-binaries` package on release

### LINKED ISSUE #1781: automatically create `experimental-binaries` package on release

spoonincode:
The `experimental-binaries` package is where we stash away some versioned packages that are not fully supported but we want indefinitely (they would eventually expire as a workflow artifact). Historically we've used this for `leap-dev.deb` which is used in CI, and ARM packages for DUNE -- both packages we don't fully support thus don't want them highly visible as a release asset.

Now that we're no longer creating ARM packages for DUNE, we can go ahead and automatically create the `experimental-binaries` package that solely stores `leap-dev.deb`. This is a rather obtuse procedure that has been done manually so automating it is a great improvement.

Clearly one of the problems with reviewing this is.. how to test it?! I don't have a good answer for that. I developed and tested it manually in a private repo. I think we'll just have to iterate on it if it doesn't work for future releases :confused: 

I made some changes to `asset-artifact-download-action` as part of this work -- AntelopeIO/asset-artifact-download-action#4. This got YOLOed merged but we can refine it some more if something comes out of the review. (we want the new `wait-for-exact-target` feature here so if someone releases before CI is complete at that ref, we wait around for CI to complete so we can upload

⚠️ 5.0 is the first version that has both a ubuntu20 & ubuntu22 `leap-dev.deb`; this might confuse existing workflows that just install `leap-dev*.deb` or such -- they might get the package for a different ubuntu version then they're running.

Resolves #1781

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,54 @@
+name: Release Actions
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  eb:
+    name: experimental-binaries
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      packages: write
+      actions: read
+    steps:
+      - name: Create Dockerfile
+        run: |
+          cat <<EOF > Dockerfile
+          FROM scratch
+          LABEL org.opencontainers.image.description="A collection of experimental Leap binary packages"
+          COPY *.deb /
+          EOF
+      - name: Get ubuntu20 leap-dev.deb
+        uses: AntelopeIO/asset-artifact-download-action@v3
+        with:
+          owner: ${{github.repository_owner}}
+          repo: ${{github.event.repository.name}}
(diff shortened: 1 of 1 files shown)


linh2931 (commented):
You don't want this target to `release\5.0` branch?

spoonincode:
For some reason I was under the impression that release triggers always run what's in `main` (default branch).. but I'm glad you asked since that's not true after looking/experimenting again! They do run with the workflow on the released commit.

So yes, this should be sent back to at least 4.0... maybe even potentially 3.2 as we still have unreleased commits on that branch?

linh2931:
Or if you want it to be safe, do `release/5.0` first. If it works as expected, then back port to 3.2 and 4.0.

spoonincode:
replacing with #1797 targeting 5.0, thanks @linh2931

PR #1800, opened 2023-10-19
[5.0] Replaced `cpu-effort-percent` with `produce-block-offset-ms`

### LINKED ISSUE #1784: Rename cpu-effort-percentage to round-cpu-effort-percentage and default to 95%
With the 5.0 RC2 release, the cpu-effort-percentage parameter functionality was dramatically changed from not controlling each individual block's cpu effort during signing, but now controls the ENTIRE round's cpu effort for ALL blocks cumulatively for the round.

While this is a great new feature to add, having it reuse the original parameter from previous versions that functioned differently I believe is a mistake and could easily clear up some confusion by simply renaming the new parameter vs.s reusing the original.

Also, on most public networks we operate on, the default of 90% round-cpu-effort-percentage is probably a bit too low as it kicks out the last block before it's scheduled time, which leads to some confusion because blocks are arriving early (and most networks don't have to send last block that aggressively).  We've found most networks tend to function properly with the last block coming out about 300ms early, which would be roughly 95% round-cpu-effort-percentage I believe (if I understand it correctly, that is :))

heifner:
Replaced `cpu-effort-percent` with `produce-block-offset-ms`.
- `produce-block-offset-ms` specifies how much time to leave at the end of the 12 block production round for the blocks (mainly the last block) to reach the next block producer. This value can be specified as larger than 500ms if needed. The default is 450ms.
- See https://github.com/AntelopeIO/leap/pull/867 for implementation of new behavior.

Updated documentation: https://docs.eosnetwork.com/manuals/leap/latest/nodeos/plugins/producer_plugin/block-producing-explained/ 
from: https://github.com/AntelopeIO/leap/blob/39327d80665dd275680b58587e89fac3a3b90891/docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md

```
For simplicity of the explanation let's consider the following notations:

* `r` = `producer_repetitions = 12` (hard-coded value)
* `m` = `max_block_cpu_usage` (on-chain consensus value)
* `u` = `max_block_net_usage` (on-chain consensus value)
* `t` = `block-time`
* `e` = `produce-block-offset-ms` (nodeos configuration)
* `w` = `block-time-interval = 500ms` (hard-coded value)
* `a` = `produce-block-early-amount = w - (w - (e / r)) = e / r ms` (how much to release each block of round early by)
* `l` = `produce-block-time = t - a`
* `p` = `produce block time window = w - a` (amount of wall clock time to produce a block)
* `c` = `billed_cpu_in_block = minimum(m, w - a)`
* `n` = `network tcp/ip latency`
* `h` = `block header validation time ms`

Peer validation for similar hardware/version/config will be <= `m`

**Let's consider the example of the following two BPs and their network topology as depicted in the below diagram**

```
         +------+     +------+       +------+     +------+
      -->| BP-A |---->| BP-A |------>| BP-B |---->| BP-B |
         +------+     | Peer |       | Peer |     +------+
                      +------+       +------+
```

`BP-A` will send block at `l` and, `BP-B` needs block at time `t` or otherwise will drop it.

If `BP-A`is producing 12 blocks as follows `b(lock) at t(ime) 1`, `bt 1.5`, `bt 2`, `bt 2.5`, `bt 3`, `bt 3.5`, `bt 4`, `bt 4.5`, `bt 5`, `bt 5.5`, `bt 6`, `bt 6.5` then `BP-B` needs `bt 6.5` by time `6.5` so it has `.5` to produce `bt 7`.

Please notice that the time of `bt 7` minus `.5` equals the time of `bt 6.5` therefore time `t` is the last block time of `BP-A` and when `BP-B` needs to start its first block.

A block is produced and sent when either it reaches `m` or `u` or `p`. 

Starting in Leap 4.0, blocks are propagated after block header validation. This means instead of `BP-A Peer` & `BP-B Peer` taking `m` time to validate and forward a block it only takes a small number of milliseconds to verify the block header and then forward the block.

Starting in Leap 5.0, blocks in a round are started immediately after the completion of the previous block. Before 5.0, blocks were always started on `w` intervals and a node would "sleep" between blocks if needed. In 5.0, the "sleeps" are all moved to the end of the block production round. 

## Example 1: block arrives 110ms early
* Assuming zero network latency between all nodes.
* Assuming blocks do not reach `m` and therefore take `w - a` time to produce.
* Assume block completion including signing takes zero time.
* `BP-A` has e = 120, n = 0ms, h = 5ms, a = 10ms
* `BP-A` sends b1 at `t1-10ms` => `BP-A-Peer` processes `h=5ms`, sends at `t-5ms` => `BP-B-Peer` processes `h=5ms`, sends at `t-0ms` => arrives at `BP-B` at `t`.
* `BP-A` starts b2 at `t1-10ms`, sends b2 at `t2-20ms` => `BP-A-Peer` processes `h=5ms`, sends at `t2-15ms` => `BP-B-Peer` processes `h=5ms`, sends at `t2-10ms` => arrives at `BP-B` at `t2-10ms`.
* `BP-A` starts b3 at `t2-20ms`, ...
* `BP-A` starts b12 at `t11-110ms`, sends b12 at `t12-120ms` => `BP-A-Peer` processes `h=5ms`, sends at `t12-115ms` => `BP-B-Peer` processes `h=5ms`, sends at `t12-110ms` => arrives at `BP-B` at `t12-110ms`

## Example 2: block arrives 80ms early
* Assuming zero network latency between `BP-A` & `BP-A Peer` and between `BP-B Peer` & `BP-B`.
* Assuming 150ms network latency between `BP-A Peer` & `BP-B Peer`.
* Assuming blocks do not reach `m` and therefore take `w - a` time to produce.
* Assume block completion including signing takes zero time.
* `BP-A` has e = 240, n = 0ms/150ms, h = 5ms, a = 20ms
* `BP-A` sends b1 at `t1-20ms` => `BP-A-Peer` processes `h=5ms`, sends at `t-15ms` =(150ms)> `BP-B-Peer` processes `h=5ms`, sends at `t+140ms` => arrives at `BP-B` at `t+140ms`.
* `BP-A` starts b2 at `t1-20ms`, sends b2 at `t2-40ms` => `BP-A-Peer` processes `h=5ms`, sends at `t2-35ms` =(150ms)> `BP-B-Peer` processes `h=5ms`, sends at `t2+120ms` => arrives at `BP-B` at `t2+120ms`.
* `BP-A` starts b3 at `t2-40ms`, ...
* `BP-A` starts b12 at `t11-220ms`, sends b12 at `t12-240ms` => `BP-A-Peer` processes `h=5ms`, sends at `t12-235ms` =(150ms)> `BP-B-Peer` processes `h=5ms`, sends at `t12-80ms` => arrives at `BP-B` at `t12-80ms`

## Example 3: block arrives 16ms late and is dropped
* Assuming zero network latency between `BP-A` & `BP-A Peer` and between `BP-B Peer` & `BP-B`.
* Assuming 200ms network latency between `BP-A Peer` & `BP-B Peer`.
* Assuming blocks do not reach `m` and therefore take `w - a` time to produce.
* Assume block completion including signing takes zero time.
* `BP-A` has e = 204, n = 0ms/200ms, h = 10ms, a = 17ms
* `BP-A` sends b1 at `t1-17ms` => `BP-A-Peer` processes `h=10ms`, sends at `t-7ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t+203ms` => arrives at `BP-B` at `t+203ms`.
* `BP-A` starts b2 at `t1-17ms`, sends b2 at `t2-34ms` => `BP-A-Peer` processes `h=10ms`, sends at `t2-24ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t2+186ms` => arrives at `BP-B` at `t2+186ms`.
* `BP-A` starts b3 at `t2-34ms`, ...
* `BP-A` starts b12 at `t11-187ms`, sends b12 at `t12-204ms` => `BP-A-Peer` processes `h=10ms`, sends at `t12-194ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t12+16ms` => arrives at `BP-B` at `t12+16ms`

## Example 4: full blocks are produced early
* Assuming zero network latency between `BP-A` & `BP-A Peer` and between `BP-B Peer` & `BP-B`.
* Assuming 200ms network latency between `BP-A Peer` & `BP-B Peer`.
* Assume all blocks are full as there are enough queued up unapplied transactions ready to fill all blocks.
* Assume a block can be produced with 200ms worth of transactions in 225ms worth of time. There is overhead for producing the block.
* `BP-A` has e = 120, m = 200ms, n = 0ms/200ms, h = 10ms, a = 10ms
* `BP-A` sends b1 at `t1-275s` => `BP-A-Peer` processes `h=10ms`, sends at `t-265ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t-55ms` => arrives at `BP-B` at `t-55ms`.
* `BP-A` starts b2 at `t1-275ms`, sends b2 at `t2-550ms (t1-50ms)` => `BP-A-Peer` processes `h=10ms`, sends at `t2-540ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t2-330ms` => arrives at `BP-B` at `t2-330ms`.
* `BP-A` starts b3 at `t2-550ms`, ...
* `BP-A` starts b12 at `t11-3025ms`, sends b12 at `t12-3300ms` => `BP-A-Peer` processes `h=10ms`, sends at `t12-3290ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t12-3080ms` => arrives at `BP-B` at `t12-3080ms`


Running wasm-runtime=eos-vm-jit eos-vm-oc-enable on relay node will reduce the validation time.
```
Resolves #1784


docs/01_nodeos/03_plugins/producer_plugin/index.md
```diff
@@ -72,20 +72,9 @@ Config Options for eosio::producer_plugin:
...
-  --cpu-effort-percent arg (=80)        Percentage of cpu block production time
-                                        used to produce block. Whole number
-                                        percentages, e.g. 80 for 80%
-  --last-block-cpu-effort-percent arg (=80)
-                                        Percentage of cpu block production time
-                                        used to produce last block. Whole
-                                        number percentages, e.g. 80 for 80%
+  --produce-block-offset-ms arg (=450)  The number of milliseconds early the
+                                        last block of a production round should
+                                        be produced.
```
greg7mdp, line 77: What setting for `--produce-block-offset-ms` would be equivalent to the default values we had previously?

heifner: Default `cpu-effort-percent` in `5.0.0-rc2` was `90%` which was equivalent to `produce-block-offset-ms = 600`

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
```diff
@@ -4,87 +4,61 @@ content_title: Block Production Explained
...
-(t-375ms) <-> (+125ms) <-> (+125ms) <-> arrive at `BP-B` with 125ms to spare.
+Starting in Leap 5.0, blocks in a round are started immediately after the completion of the previous block. Before 5.0, blocks were always started on `w` intervals and a node would "sleep" between blocks if needed. In 5.0, the "sleeps" are all moved to the end of the block production round. 
 
-## Example 5
-`BP-A` has 10% e and m = 200ms, c = 50ms, n = 0ms, a = 450ms:
-(t-450ms) <-> (+50ms) <-> (+50ms) <-> arrive at `BP-B` with 350ms to spare.
+## Example 1: block arrives 110ms early (zero network latency between BP and immediate peer)
+* `BP-A` has e = 120, n = 0ms, h = 5ms, a = 10ms
+* `BP-A` sends b1 at `t1-10ms` => `BP-A-Peer` processes `h=5ms`, sends at `t-5ms` => `BP-B-Peer` processes `h=5ms`, sends at `t-0ms` => arrives at `BP-B` at `t`.
+* `BP-A` starts b2 at `t1-10ms`, sends b2 at `t2-20ms` => `BP-A-Peer` processes `h=5ms`, sends at `t2-15ms` => `BP-B-Peer` processes `h=5ms`, sends at `t2-10ms` => arrives at `BP-B` at `t2-10ms`.
```
greg7mdp, line 46: I thought it started `b2` as soon as `b1` was completed, and that it didn't wait for `t1-10ms`

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
```diff
@@ -4,87 +4,61 @@ content_title: Block Production Explained
...
 
-## Example 3
-`BP-A` has 30% e and m = 200ms, c = 150ms, n = 0ms, a = 350ms:
-(t-350ms) <-> (+150ms) <-> (+150ms) <-> arrive at `BP-B` with 50ms to spare.
+Starting in Leap 4.0, blocks are propagated after block header validation. This means instead of `BP-A Peer` & `BP-B Peer` taking `m` time to validate and forward a block it only takes a small number of milliseconds to verify the block header and then forward the block.
 
-## Example 4
-`BP-A` has 25% e and m = 200ms, c = 125ms, n = 0ms, a = 375ms:
-(t-375ms) <-> (+125ms) <-> (+125ms) <-> arrive at `BP-B` with 125ms to spare.
+Starting in Leap 5.0, blocks in a round are started immediately after the completion of the previous block. Before 5.0, blocks were always started on `w` intervals and a node would "sleep" between blocks if needed. In 5.0, the "sleeps" are all moved to the end of the block production round.
```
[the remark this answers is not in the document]
heifner: Anytime a block is ready it is sent immediately. That has been true since I think 2.x, not sure which x without digging into it. What is diff in 5.0 is that we start the next block immediately. In previous versions the block would be sent out, but we would wait until the next block boundary to start the next block.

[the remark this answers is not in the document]
heifner: Yes, that is true.

Lets say you have enough queued up trxs to fill a 200ms block in 200ms. That can only be true if you have 200ms of trxs ready to go. Lets also say you have enough to fill 12 200ms blocks. So you have 2.4s of trxs ready to go. In previous versions those would be but into a blocks like: [200ms] (sleep for 300ms) [200ms] (sleep for 300ms) ...
Now they would be put into blocks like: [200ms][200ms][200ms](sleep 3.6secs). This actually provides better latency for the trxs that make it into blocks, but no worse latency for those trxs that do not make it into blocks as they would not have made it into blocks either way.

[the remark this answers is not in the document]
greg7mdp: That is *not* what the example describes:

* `BP-A` sends b1 at `t1-10ms` => ...

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
heifner: It does start `b2` as soon as `b1` is complete. `b1` is complete at `t1-10ms` which is when `b2` starts.

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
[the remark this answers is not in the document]
heifner: The current examples assume non-full blocks. I'll add an example with full blocks.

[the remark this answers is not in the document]
greg7mdp: true, good point.

But what happen if we don't have enough transactions, do we produce and send the block as soon as we run out of transactions.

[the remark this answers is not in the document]
heifner: If we don't have enough transactions we wait for the configured `w - a` amount, adding trxs to a block as they come in.

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
greg7mdp: Why would b1 be complete at `t1-10ms`? Couldn't it reach the `m` limit before that?

heifner: All the existing examples assume non-full blocks, none of them reach `m` limit. I'll add an example where `m` is reached.

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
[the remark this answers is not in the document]
greg7mdp: Oh, I missed that, thanks, it makes sense!

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
```diff
@@ -4,87 +4,61 @@ content_title: Block Production Explained
...
+* `a` = `produce-block-early-amount = w - (w - (e / r)) = e / r ms` (how much to release each block of round early by)
+* `l` = `produce-block-time = t - a`
+* `p` = `produce block time window = w - a` (amount of wall clock time to produce a block)
+* `c` = `billed_cpu_in_block = minimum(m, w - a)`
+* `n` = `network tcp/ip latency`
+* `h` = `block header validation time ms`
+
+Peer validation for similar hardware/version/config will be <= `m`
+
+**Let's consider for exemplification the following two BPs and their network topology as depicted in the below diagram**
```
heifner, line 22: Why `exemplification` is used here instead of just `for example` is beyond me. Seems like we are trying to sound smart.

greg7mdp: I think I'd write:
`Let's consider the exemple of the following two BPs and their network topology as depicted in the below diagram`

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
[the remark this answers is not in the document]
greg7mdp: That's great, so there are really no drawbacks to the new mode, only benefits!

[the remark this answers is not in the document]
heifner: The only drawback I see is that if you have non-full blocks, trxs arriving in the `e`  interval for block 12 would have to wait for the next producer to place it in there first block instead of making it into block 12. But in pre-5.0, block 12 has to be set very small to provide time for it to reach the next BP. This is a very small gain in latency for trxs in block 12 time while losing latency on trxs when blocks are full. Seems like a clear win.

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
```diff
@@ -4,87 +4,81 @@ content_title: Block Production Explained
...
+* `BP-A` sends b1 at `t1-17ms` => `BP-A-Peer` processes `h=10ms`, sends at `t-7ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t+203ms` => arrives at `BP-B` at `t+203ms`.
+* `BP-A` starts b2 at `t1-17ms`, sends b2 at `t2-34ms` => `BP-A-Peer` processes `h=10ms`, sends at `t2-24ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t2+186ms` => arrives at `BP-B` at `t2+186ms`.
+* `BP-A` starts b3 at `t2-34ms`, ...
+* `BP-A` starts b12 at `t11-187ms`, sends b12 at `t12-204ms` => `BP-A-Peer` processes `h=10ms`, sends at `t12-194ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t12+16ms` => arrives at `BP-B` at `t12-16ms`
+
+## Example 4: full blocks are produced early
+* Assuming zero network latency between BP and immediate BP peer.
+* Assume all blocks are full as there are enough queued up unapplied transactions ready to fill all blocks.
+* Assume a block can be produced with 200ms worth of transactions in 225ms worth of time. There is overhead for producing the block.
+* `BP-A` has e = 120, n = 200ms, h = 10ms, a = 10ms
```
greg7mdp, line 77: should it be `m = 200ms` instead of `n`?

greg7mdp:
This document `10_block-producing-explained.md` is excellent... for sure it should be kept and maintained.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -613,6 +612,17 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
 
+   void set_produce_block_offset(uint32_t produce_block_offset_ms) {
+      EOS_ASSERT(produce_block_offset_ms < (config::producer_repetitions * config::block_interval_ms), plugin_config_exception,
+                 "produce-block-offset-ms ${p} must be [0 - ${max})", ("p", produce_block_offset_ms)("max", config::producer_repetitions * config::block_interval_ms));
+      _produce_block_cpu_effort_us = fc::milliseconds( config::block_interval_ms - produce_block_offset_ms / config::producer_repetitions );
+   }
+
+   fc::microseconds get_produce_block_offset() const {
+      return fc::milliseconds( (config::block_interval_ms * config::producer_repetitions) -
+                               ((_produce_block_cpu_effort_us.count()/1000) * config::producer_repetitions) );
```
greg7mdp, line 623: Why not? 
```suggestion
      return fc::milliseconds(produce_block_offset_ms);
```

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
heifner: Updated

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
```diff
@@ -4,87 +4,81 @@ content_title: Block Production Explained
...
+* `BP-A` sends b1 at `t1-17ms` => `BP-A-Peer` processes `h=10ms`, sends at `t-7ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t+203ms` => arrives at `BP-B` at `t+203ms`.
+* `BP-A` starts b2 at `t1-17ms`, sends b2 at `t2-34ms` => `BP-A-Peer` processes `h=10ms`, sends at `t2-24ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t2+186ms` => arrives at `BP-B` at `t2+186ms`.
+* `BP-A` starts b3 at `t2-34ms`, ...
+* `BP-A` starts b12 at `t11-187ms`, sends b12 at `t12-204ms` => `BP-A-Peer` processes `h=10ms`, sends at `t12-194ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t12+16ms` => arrives at `BP-B` at `t12-16ms`
+
+## Example 4: full blocks are produced early
+* Assuming zero network latency between BP and immediate BP peer.
+* Assume all blocks are full as there are enough queued up unapplied transactions ready to fill all blocks.
+* Assume a block can be produced with 200ms worth of transactions in 225ms worth of time. There is overhead for producing the block.
+* `BP-A` has e = 120, m = 200ms, n = 200ms, h = 10ms, a = 10ms
```
greg7mdp, line 77: `n = 200ms`? isn't `n` network latency now?

heifner: Yes, I meant `n` to mean network latency all along.

plugins/producer_plugin/producer_plugin.cpp
heifner: Because `_produce_block_cpu_effort_us` is the offset for each individual block.

greg7mdp: if you look above, `_produce_block_cpu_effort_us = fc::milliseconds( config::block_interval_ms - produce_block_offset_ms / config::producer_repetitions )`. Just replace in `get_produce_block_offset` and I believe what I suggested is correct.

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
greg7mdp: This seems contradictory 
image

heifner: That is 0 latency between `BP-A` and `BP-A Peer` and between `BP-B Peer` and `BP-B`. There is 200ms between `BP-A Peer` and `BP-B Peer`. Maybe I should spell that out.

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
```diff
@@ -4,87 +4,81 @@ content_title: Block Production Explained
...
+
+## Example 3: block arrives 16ms late and is dropped
+* Assuming zero network latency between BP and immediate BP peer.
+* Assuming blocks do not reach `m` and therefore take `w - a` time to produce.
+* Assume block completion including signing takes zero time.
+* `BP-A` has e = 204, n = 200ms, h = 10ms, a = 17ms
+* `BP-A` sends b1 at `t1-17ms` => `BP-A-Peer` processes `h=10ms`, sends at `t-7ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t+203ms` => arrives at `BP-B` at `t+203ms`.
+* `BP-A` starts b2 at `t1-17ms`, sends b2 at `t2-34ms` => `BP-A-Peer` processes `h=10ms`, sends at `t2-24ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t2+186ms` => arrives at `BP-B` at `t2+186ms`.
+* `BP-A` starts b3 at `t2-34ms`, ...
+* `BP-A` starts b12 at `t11-187ms`, sends b12 at `t12-204ms` => `BP-A-Peer` processes `h=10ms`, sends at `t12-194ms` =(200ms)> `BP-B-Peer` processes `h=10ms`, sends at `t12+16ms` => arrives at `BP-B` at `t12-16ms`
```
linh2931, line 71: Shouldn't it be arriving at B at `t12+16ms`? (the `+`).

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: In the existing code, you multiply by `config::producer_repetitions`, so it is *not* for each individual block.

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
greg7mdp: Yes, I don't think it is clear as is.

plugins/producer_plugin/producer_plugin.cpp
heifner: Sorry, I'm not following. There is no variable `produce_block_offset_ms`. We could save it for this purpose, but it can be easily calculated. The user setting is for offset of the 12 block round. The internal variable is how much window of time to give an individual block.

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
heifner: Yes. Thanks.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: Oh, my bad, I thought it was a member. Of course you recalculate it here. 
I think it may be more straightforward to save it as a member variable instead of recalculating it.

docs/01_nodeos/03_plugins/producer_plugin/10_block-producing-explained.md
heifner: Updated.

plugins/producer_plugin/producer_plugin.cpp
heifner: If we keep it separate I'm afraid they may drift. Seems better to have one value of truth.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1033,8 +1043,8 @@ void producer_plugin::set_program_options(
           "account that can not access to extended CPU/NET virtual resources")
          ("greylist-limit", boost::program_options::value<uint32_t>()->default_value(1000),
           "Limit (between 1 and 1000) on the multiple that CPU/NET virtual resources can extend during low usage (only enforced subjectively; use 1000 to not enforce any limit)")
-         ("cpu-effort-percent", bpo::value<uint32_t>()->default_value(config::default_block_cpu_effort_pct / config::percent_1),
-          "Percentage of cpu block production time used to produce block. Whole number percentages, e.g. 80 for 80%")
+         ("produce-block-offset-ms", bpo::value<uint32_t>()->default_value(config::default_produce_block_offset_ms),
+          "The number of milliseconds early the last block of a production round should be produced.")
```
greg7mdp, line 1047: really it defines the latest time in the round the last block will be produced. It could be produced much earlier than that.

heifner: `The number of milliseconds early the last block of a production round can be produced.` ?
`The minimum time to allow for the last block of a production round to make it to next block producer.`
`The minimum time to produce the last block of a production round.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: It seems odd to store not the one provided by the user, but the one calculated, but I guess the calculated one is the one we use. So it is fine.

plugins/producer_plugin/producer_plugin.cpp
heifner: `The minimum time to reserve at the end of a production round for blocks to propagate to the next block producer.`
`The minimum offset from the end of a block production round to reserve for block propagation.`

heifner: `Set to a value large enough for your blocks to have time to reach the next block producer.`

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -42,28 +42,28 @@ namespace block_timing_util {
...
       uint32_t production_round_start_block_slot =
             (block_slot / chain::config::producer_repetitions) * chain::config::producer_repetitions;
       uint32_t production_round_index = block_slot % chain::config::producer_repetitions;
       return chain::block_timestamp_type(production_round_start_block_slot - 1).to_time_point() +
-             fc::microseconds(cpu_effort_us * production_round_index);
+             fc::microseconds(cpu_effort_us.count() * production_round_index);
    }
 
-   inline fc::time_point calculate_producing_block_deadline(uint32_t cpu_effort_us, chain::block_timestamp_type block_time) {
+   inline fc::time_point calculate_producing_block_deadline(fc::microseconds cpu_effort_us, chain::block_timestamp_type block_time) {
```
greg7mdp, line 57: maybe we can rename the variable `cpu_effort` now that the type enforces microseconds?

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -42,28 +42,28 @@ namespace block_timing_util {
    // In the past, a producer would always start a block `config::block_interval_us` ahead of its block time. However,
    // it causes the last block in a block production round being released too late for the next producer to have
    // received it and start producing on schedule. To mitigate the problem, we leave no time gap in block producing. For
-   // example, given block_interval=500 ms and cpu effort=400 ms, assuming the our round start at time point 0; in the
+   // example, given block_interval=500 ms and cpu effort=400 ms, assuming our round starts at time point 0; in the
    // past, the block start time points would be at time point -500, 0, 500, 1000, 1500, 2000 ....  With this new
    // approach, the block time points would become -500, -100, 300, 700, 1100 ...
-   inline fc::time_point production_round_block_start_time(uint32_t cpu_effort_us, chain::block_timestamp_type block_time) {
+   inline fc::time_point production_round_block_start_time(fc::microseconds cpu_effort_us, chain::block_timestamp_type block_time) {
```
greg7mdp, line 48: maybe we can rename the variable `cpu_effort` now that the type enforces microseconds?

plugins/producer_plugin/include/eosio/producer_plugin/block_timing_util.hpp
```diff
@@ -118,7 +118,7 @@ namespace block_timing_util {
    // Return the *next* block start time according to its block time slot.
    // Returns empty optional if no producers are in the active_schedule.
    // block_num is only used for watermark minimum offset.
-   inline std::optional<fc::time_point> calculate_producer_wake_up_time(uint32_t cpu_effort_us, uint32_t block_num,
+   inline std::optional<fc::time_point> calculate_producer_wake_up_time(fc::microseconds cpu_effort_us, uint32_t block_num,
```
greg7mdp, line 121: maybe we can rename the variable `cpu_effort` now that the type enforces microseconds?

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: This variable defines the point in time `produce-block-offset-ms` before the end of the 12 block round, and nodeos will ensure that the last block of the round is produced by this time at the latest.

docs/01_nodeos/03_plugins/producer_plugin/index.md
```diff
@@ -72,20 +72,9 @@ Config Options for eosio::producer_plugin:
...
-                                        microseconds. Valid range 0 ..
-                                        -block_time_interval.
-  --cpu-effort-percent arg (=80)        Percentage of cpu block production time
-                                        used to produce block. Whole number
-                                        percentages, e.g. 80 for 80%
-  --last-block-cpu-effort-percent arg (=80)
-                                        Percentage of cpu block production time
-                                        used to produce last block. Whole
-                                        number percentages, e.g. 80 for 80%
+  --produce-block-offset-ms arg (=450)  The number of milliseconds early the
```
linh2931, line 75: As this is a new option, we have freedom to pick an intuitive name. The `produce` indicates it is a producer_plugin option, so it is better be `producer`; `produce` is a verb; it does not sound right when you read `produce block offset` in whole. This option is for a whole round; `block` is misleading in the name. `offset` is ambiguous; what does it refers to?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -501,14 +500,14 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    std::atomic<int32_t>                              _max_transaction_time_ms; // modified by app thread, read by net_plugin thread pool
    std::atomic<uint32_t>                             _received_block{0};       // modified by net_plugin thread pool
    fc::microseconds                                  _max_irreversible_block_age_us;
-   int32_t                                           _cpu_effort_us = 0;
+   // produce-block-offset is in terms of the complete round, internally use calculated value for each block of round
+   fc::microseconds                                  _produce_block_cpu_effort_us;
```
linh2931, line 504: `produce` is not needed in this name.

docs/01_nodeos/03_plugins/producer_plugin/index.md
linh2931: Maybe default should be 480, as it is divisible by 12.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -613,6 +612,17 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    void repost_exhausted_transactions(const fc::time_point& deadline);
    bool push_read_only_transaction(transaction_metadata_ptr trx, next_function<transaction_trace_ptr> next);
 
+   void set_produce_block_offset(uint32_t produce_block_offset_ms) {
+      EOS_ASSERT(produce_block_offset_ms < (config::producer_repetitions * config::block_interval_ms), plugin_config_exception,
+                 "produce-block-offset-ms ${p} must be [0 - ${max})", ("p", produce_block_offset_ms)("max", config::producer_repetitions * config::block_interval_ms));
+      _produce_block_cpu_effort_us = fc::milliseconds( config::block_interval_ms - produce_block_offset_ms / config::producer_repetitions );
```
linh2931, line 618: I always have to look further, when I see one side is `_us` and the other side is `fc::milliseconds`.

plugins/producer_plugin/producer_plugin.cpp
heifner: Reading them all again, I think: `The minimum time to reserve at the end of a production round for blocks to propagate to the next block producer.`

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -613,6 +612,17 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+   }
+
+   fc::microseconds get_produce_block_offset() const {
```
linh2931, line 621: At the first glance, I think this returns the user supplied value.

linh2931 (commented):
* I don't see in the code where you send out a block right away after it is produced.
* Should some new tests be added for this change?

plugins/producer_plugin/include/eosio/producer_plugin/producer_plugin.hpp
```diff
@@ -17,7 +17,7 @@ class producer_plugin : public appbase::plugin<producer_plugin> {
    struct runtime_options {
       std::optional<int32_t>   max_transaction_time;
       std::optional<int32_t>   max_irreversible_block_age;
-      std::optional<int32_t>   cpu_effort_us;
+      std::optional<int32_t>   produce_block_offset_ms;
```
greg7mdp, line 20: I'd add the comment here as well 
```suggestion
      std::optional<int32_t>   produce_block_offset_ms; // minimum time to reserve at the end of a production round for blocks to propagate to the next block producer.
```

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: excellent!

heifner:
> * I don't see in the code where you send out a block right away after it is produced.

See `producer_plugin` use of `block_is_exhausted()`. 
https://github.com/AntelopeIO/leap/blob/54e42bc1980d959b2348fc65d127ad9031514af5/plugins/producer_plugin/producer_plugin.cpp#L2527-L2527
https://github.com/AntelopeIO/leap/blob/54e42bc1980d959b2348fc65d127ad9031514af5/plugins/producer_plugin/producer_plugin.cpp#L849-L849
https://github.com/AntelopeIO/leap/blob/54e42bc1980d959b2348fc65d127ad9031514af5/plugins/producer_plugin/producer_plugin.cpp#L2494-L2494

> * Should some new tests be added for this change?

This is not new behavior, it has shipped a block off early since v2.0.4. See https://github.com/EOSIO/eos/pull/8651 
We should consider refactoring `producer_plugin` so it is easier to test. Or it might be possible to use `mock_time` like the `test_trx_retry_db` does. However, I believe that is out of scope for this PR. Please create a GitHub issue to work that later.

docs/01_nodeos/03_plugins/producer_plugin/index.md
heifner: The name was suggested by community. Do you have an alternative name you would suggest?
No reason for it to be evenly divided by 12 ,but maybe we should round up in the division by 12 in `set_produce_block_offset` so 450 becomes 456 instead of 444.

plugins/producer_plugin/producer_plugin.cpp
heifner: I added `produce` here to emphasize that it controls the block production block effort not subjective block production effort. I can drop `produce` if you think it is clearer.

plugins/producer_plugin/producer_plugin.cpp
heifner: Renamed to drop the `_us`

plugins/producer_plugin/producer_plugin.cpp
heifner: It does. We do not have a `milliseconds` type only a `microseconds` type.

greg7mdp (approved):

linh2931:
> See `producer_plugin` use of `block_is_exhausted()`.
Thanks! I was thinking about where change was for keeping production without waiting, but asking a wrong question.

linh2931:
> This is not new behavior, it has shipped a block off early since v2.0.4. See [EOSIO/eos#8651](https://github.com/EOSIO/eos/pull/8651)
> We should consider refactoring `producer_plugin` so it is easier to test. Or it might be possible to use `mock_time` like the `test_trx_retry_db` does. However, I believe that is out of scope for this PR. Please create a GitHub issue to work that later.

https://github.com/AntelopeIO/leap/issues/1804

heifner:
> Thanks! I was thinking about where change was for keeping production without waiting, but asking a wrong question.

See https://github.com/AntelopeIO/leap/pull/867 which is the change to calculations on when to start a block.

docs/01_nodeos/03_plugins/producer_plugin/index.md
heifner: Updated to use `std::ceil` on `produce-block-offset-ms / 12.0`

plugins/producer_plugin/producer_plugin.cpp
linh2931: OK, keep produce here.

docs/01_nodeos/03_plugins/producer_plugin/index.md
linh2931: OK, just keep the name suggested by the community. After reading the code, I got used to it.

linh2931 (approved):

docs/01_nodeos/03_plugins/producer_plugin/index.md
heifner: Just realized there was no reason to round and lose precision. Updated to calculate in microseconds.

linh2931 (approved):

greg7mdp (approved):

PR #1807, opened 2023-10-23
Avoid including `eos-vm-oc.hpp` which includes `types.hpp`, so the c++17 compilation doesn't propagate throughout

greg7mdp:


heifner (approved):

spoonincode:
This is fine but it comes at the cost of repeating in two different files,
https://github.com/AntelopeIO/leap/blob/6dbb6c24f6379e5db6bd8fa9dad8da97d260ecb8/libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/memory.hpp#L14
https://github.com/AntelopeIO/leap/blob/6dbb6c24f6379e5db6bd8fa9dad8da97d260ecb8/libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/eos-vm-oc.hpp#L21

I wonder instead if you could add this to `eos-vm-oc.h` -- there is already `#ifdef __cplusplus` in there -- so that it would only be defined once.

greg7mdp:
> I wonder instead if you could add this to `eos-vm-oc.h` -- there is already `#ifdef __cplusplus` in there -- so that it would only be defined once.

Done, thanks for the suggestion.

heifner (approved):

spoonincode (approved):

ericpassmore:
Note:start
group: BUG
category: INTERNALS
summary: To prevent compiling code as c++ 17, avoid using the eos-vm-oc.hpp which included types.hpp
Note:end

Issue #1690, opened 2023-09-27
Performance: Validate Expected Throughput


ericpassmore:
Using the performance harness demonstrate a performance run of 30,000 TPS or greater. This is consistent with previous performance runs. Target for this work is post Leap-5 release.

bhazzard:
Not actually release dependent. Marked for Leap 6 milestone just to track it.

PR #1814, opened 2023-10-24
[5.0] PH: Reliability improvements

heifner:
- Performance harness tests were failing because the last few trxs from the `trx_generator` were being dropped. Modified the `trx_generator` to wait for all writes to finish before terminating.
- Modify the performance harness to wait for the complete test time for the trxs in blocks.
- Update the performance harness to `--print-missing-transactions` by default as this is very useful for debugging why the test failed.
- Updated Cluster.py to support `--produce-block-offset-ms 0` for older versions of `nodeos`.

Still need to run a complete test run on a quiescent machine.

Issue #1690

### DIFF
--- a/tests/PerformanceHarness/log_reader.py
+++ b/tests/PerformanceHarness/log_reader.py
@@ -39,7 +39,7 @@ class TpsTestConfig:
     numTrxGensUsed: int = 0
     targetTpsPerGenList: List[int] = field(default_factory=list)
     quiet: bool = False
-    printMissingTransactions: bool=False
+    printMissingTransactions: bool=True
 
 @dataclass
 class stats():
--- a/tests/PerformanceHarness/performance_test_basic.py
+++ b/tests/PerformanceHarness/performance_test_basic.py
@@ -733,7 +733,7 @@ class PtbArgumentsHandler(object):
         ptbBaseParserGroup.add_argument("--save-state", help=argparse.SUPPRESS if suppressHelp else "Whether to save node state. (Warning: large disk usage)", action='store_true')
         ptbBaseParserGroup.add_argument("--quiet", help=argparse.SUPPRESS if suppressHelp else "Whether to quiet printing intermediate results and reports to stdout", action='store_true')
         ptbBaseParserGroup.add_argument("--prods-enable-trace-api", help=argparse.SUPPRESS if suppressHelp else "Determines whether producer nodes should have eosio::trace_api_plugin enabled", action='store_true')
-        ptbBaseParserGroup.add_argument("--print-missing-transactions", type=bool, help=argparse.SUPPRESS if suppressHelp else "Toggles if missing transactions are be printed upon test completion.", default=True)
+        ptbBaseParserGroup.add_argument("--print-missing-transactions", type=bool, help=argparse.SUPPRESS if suppressHelp else "Print missing transactions upon test completion.", default=True)
         ptbBaseParserGroup.add_argument("--account-name", type=str, help=argparse.SUPPRESS if suppressHelp else "Name of the account to create and assign a contract to", default="eosio")
         ptbBaseParserGroup.add_argument("--contract-dir", type=str, help=argparse.SUPPRESS if suppressHelp else "Path to contract dir", default="unittests/contracts/eosio.system")
         ptbBaseParserGroup.add_argument("--wasm-file", type=str, help=argparse.SUPPRESS if suppressHelp else "WASM file name for contract", default="eosio.system.wasm")
--- a/tests/PerformanceHarness/performance_test_basic.py
+++ b/tests/PerformanceHarness/performance_test_basic.py
@@ -171,7 +171,7 @@ class PerformanceTestBasic:
         quiet: bool=False
         delPerfLogs: bool=False
         expectedTransactionsSent: int = field(default_factory=int, init=False)
-        printMissingTransactions: bool=False
+        printMissingTransactions: bool=True
         userTrxDataFile: Path=None
         endpointMode: str="p2p"
         apiEndpoint: str=None
@@ -495,7 +495,7 @@ class PerformanceTestBasic:
         scrapeTrxGenTrxSentDataLogs(trxSent, self.trxGenLogDirPath, self.ptbConfig.quiet)
         if len(trxSent) != self.ptbConfig.expectedTransactionsSent:
             print(f"ERROR: Transactions generated: {len(trxSent)} does not match the expected number of transactions: {self.ptbConfig.expectedTransactionsSent}")
-        blocksToWait = 2 * self.ptbConfig.testTrxGenDurationSec + 10
+        blocksToWait = self.data.startBlock + (2 * self.ptbConfig.testTrxGenDurationSec) + 10
         trxNotFound = self.validationNode.waitForTransactionsInBlockRange(trxSent, self.data.startBlock, blocksToWait)
         self.data.ceaseBlock = self.validationNode.getHeadBlockNum()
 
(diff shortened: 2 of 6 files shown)


tests/PerformanceHarness/performance_test_basic.py
```diff
@@ -495,7 +495,7 @@ def configureConnections():
         scrapeTrxGenTrxSentDataLogs(trxSent, self.trxGenLogDirPath, self.ptbConfig.quiet)
         if len(trxSent) != self.ptbConfig.expectedTransactionsSent:
             print(f"ERROR: Transactions generated: {len(trxSent)} does not match the expected number of transactions: {self.ptbConfig.expectedTransactionsSent}")
-        blocksToWait = 2 * self.ptbConfig.testTrxGenDurationSec + 10
+        blocksToWait = self.data.startBlock + (2 * self.ptbConfig.testTrxGenDurationSec) + 10
```
oschwaldp-oci, line 498: `blocksToWait` is meant to be the number of blocks to wait after a given block. so the `startBlock` was not included on purpose.  You should see the `startBlock` being added into the call to `waitForTrasactionsInBlockRange` in `performance_test_basic.py` and then in `Node.py` `waitForTransactionsInBlockRange` -> `overallEndBlock = startBlock + maxFutureBlocks`

tests/PerformanceHarness/log_reader.py
```diff
@@ -39,7 +39,7 @@ class TpsTestConfig:
     numTrxGensUsed: int = 0
     targetTpsPerGenList: List[int] = field(default_factory=list)
     quiet: bool = False
-    printMissingTransactions: bool=False
+    printMissingTransactions: bool=True
```
oschwaldp-oci, line 42: The reason this was not defaulted to `True` initially was that if the transaction generators initially fail to connect completely, then the log file is riddled with potentially tens of thousands of log messages for dropped transactions.

tests/PerformanceHarness/performance_test_basic.py
heifner: Ok. I looked at that wrong. It must have been failing because it took  more than 10 blocks for the trx generators to fire up as it was consistently failing on not getting the trxs from the blocks when they were there. Since the function is called `waitForTransactionsInBlockRange, I think I'll change it to take a range instead of start and offset.

tests/TestHarness/Cluster.py
```diff
@@ -461,6 +461,11 @@ def connectGroup(group, producerNodes, bridgeNodes) :
                 if "--plugin eosio::history_api_plugin" in args:
                     argsArr.append("--is-nodeos-v2")
                     break
+
+        # Handle common case of specifying no block offset for older versions
+        if "v2" in self.nodeosVers or "v3" in self.nodeosVers or "v4" in self.nodeosVers:
```
greg7mdp, line 466: I like this a little better, no need to change though.
```suggestion
        if any(s in self.nodeosVers for s in ["v2", "v3", "v4"]):
```

tests/trx_generator/trx_provider.cpp
```diff
@@ -73,15 +73,29 @@ namespace eosio::testing {
...
 
    void p2p_connection::disconnect() {
-      ilog("Closing socket.");
-      _p2p_socket.close();
-      ilog("Socket closed.");
+      int max    = 30;
+      int waited = 0;
+      while (_sent.load() != _sent_callback_num.load() && waited < max) {
+         ilog("disconnect waiting on ack - sent ${s} | acked ${a} | waited ${w}",
+              ("s", _sent.load())("a", _sent_callback_num.load())("w", waited));
```
greg7mdp, line 80: The ilog may not be accurate as we reload the `atomic`.  Maybe do something like this:
```suggestion
      for (int sent = _sent.load(), sent_callback_num = _sent_callback_num.load(); sent != sent_callback_num && waited < max; ) {
         ilog("disconnect waiting on ack - sent ${s} | acked ${a} | waited ${w}",
              ("s", sent)("a", sent_callback_num)("w", waited));
```

tests/trx_generator/trx_provider.cpp
```diff
@@ -73,15 +73,29 @@ namespace eosio::testing {
...
    }
 
    void p2p_connection::send_transaction(const chain::packed_transaction& trx) {
       send_buffer_type msg = create_send_buffer(trx);
-      _p2p_socket.send(boost::asio::buffer(*msg));
-      trx_acknowledged(trx.id(), fc::time_point::min()); //using min to identify ack time as not applicable for p2p
+
+      ++_sent;
+      _strand.post( [this, msg{std::move(msg)}, id{trx.id()}]() {
+         boost::asio::write(_p2p_socket, boost::asio::buffer(*msg));
```
greg7mdp, line 95:
```suggestion
         boost::asio::write(_p2p_socket, boost::asio::buffer(std::move(*msg)));
```

tests/trx_generator/trx_provider.cpp
```diff
@@ -73,15 +73,29 @@ namespace eosio::testing {
...
+      }
    }
 
    void p2p_connection::send_transaction(const chain::packed_transaction& trx) {
       send_buffer_type msg = create_send_buffer(trx);
-      _p2p_socket.send(boost::asio::buffer(*msg));
-      trx_acknowledged(trx.id(), fc::time_point::min()); //using min to identify ack time as not applicable for p2p
+
+      ++_sent;
+      _strand.post( [this, msg{std::move(msg)}, id{trx.id()}]() {
```
greg7mdp, line 94: I'm not quite getting the benefit of sending the transaction in a `strand`. Can you explain briefly?

greg7mdp (changes requested):

tests/trx_generator/trx_provider.cpp
heifner: boost asio sockets are not thread safe. The strand makes sure it is only accessed by thread pool thread.

tests/trx_generator/trx_provider.cpp
heifner: `boost::asio::buffer` never has ownership of data, it is simply a span of it.

greg7mdp: Yes, right, I missed that, it is not very obvious.

greg7mdp (approved):

tests/PerformanceHarness/performance_test_basic.py
```diff
@@ -733,7 +733,7 @@ def _createBaseArgumentParser(defEndpointApiDef: str, defProdNodeCnt: int, defVa
         ptbBaseParserGroup.add_argument("--save-state", help=argparse.SUPPRESS if suppressHelp else "Whether to save node state. (Warning: large disk usage)", action='store_true')
         ptbBaseParserGroup.add_argument("--quiet", help=argparse.SUPPRESS if suppressHelp else "Whether to quiet printing intermediate results and reports to stdout", action='store_true')
         ptbBaseParserGroup.add_argument("--prods-enable-trace-api", help=argparse.SUPPRESS if suppressHelp else "Determines whether producer nodes should have eosio::trace_api_plugin enabled", action='store_true')
-        ptbBaseParserGroup.add_argument("--print-missing-transactions", help=argparse.SUPPRESS if suppressHelp else "Toggles if missing transactions are be printed upon test completion.", action='store_true')
+        ptbBaseParserGroup.add_argument("--print-missing-transactions", type=bool, help=argparse.SUPPRESS if suppressHelp else "Toggles if missing transactions are be printed upon test completion.", default=True)
```
linh2931, line 736: Should `are be` be `are`?

tests/trx_generator/trx_provider.cpp
```diff
@@ -73,15 +73,31 @@ namespace eosio::testing {
    }
 
    void p2p_connection::disconnect() {
-      ilog("Closing socket.");
-      _p2p_socket.close();
```
linh2931, line 77: Now that `_p2p_socket.close();` is removed, where is disconnect triggered?

tests/trx_generator/trx_provider.hpp
```diff
@@ -2,9 +2,13 @@
 
 #include<eosio/chain/transaction.hpp>
 #include<eosio/chain/block.hpp>
-#include<boost/asio/ip/tcp.hpp>
-#include<fc/network/message_buffer.hpp>
 #include<eosio/chain/thread_utils.hpp>
+
+#include<fc/network/message_buffer.hpp>
```
linh2931, line 7: Is `message_buffer.hpp` required in this header file?

tests/trx_generator/trx_provider.cpp
heifner: At destruction of `p2p_connection`.

tests/trx_generator/trx_provider.hpp
heifner: No.

tests/PerformanceHarness/performance_test_basic.py
heifner: Rewrote it.

linh2931 (approved):

Issue #1794, opened 2023-10-18
Test Failure: failed to read response from monitor process in snapshot_scheduler_test


linh2931:
Seen this again. A core dump somewhere?

https://github.com/AntelopeIO/leap/actions/runs/6555274088/job/17803647521?pr=1792#step:5:25942

spoonincode:
https://github.com/AntelopeIO/leap/actions/runs/6568816849/job/17843972230

heifner:
https://github.com/AntelopeIO/leap/actions/runs/6579506381/job/17875830636

heifner:
https://github.com/AntelopeIO/leap/actions/runs/6587504928/job/17898239176

BTW, I ran the plugin_test `tests/plugin_test --catch_system_errors=no` all night locally in a loop and it never failed once.

heifner:
https://github.com/AntelopeIO/leap/actions/runs/6591432570/job/17910300584

spoonincode:
https://github.com/AntelopeIO/leap/actions/runs/6628104758/job/18004907600

bhazzard:
Kevin mentioned on 10/24/23 that he ran it all night without a single failure.

bhazzard:
Kevin also mentioned a theory that maybe something else was writing to the same directory, though most tests use a tmp directory.

bhazzard:
move to 5.0.1 milestone

linh2931:
The problem is shifted to read_only_trxs:
https://github.com/AntelopeIO/leap/actions/runs/6631880102/job/18016715919?pr=1819

Looks like some interferences in tests within plugin_test or in parallel test.

spoonincode:
imo until we learn otherwise, let's consider this issue strictly for failures on `snapshot_scheduler_test` which maybe hopefully was fixed with #1821 (since there was clearly threading violations in it, crashes/weirdness isn't a stretch), so let's close it. If we should open another issue for the read only trx tests failing, let's do that.

linh2931:
https://github.com/AntelopeIO/leap/issues/1823 opened for read only trx tests.

linh2931:
Happened in `snapshot_scheduler_test` in this CICD run: https://github.com/AntelopeIO/leap/actions/runs/6643256190/job/18050335448?pr=1819

Issue #1736, opened 2023-10-08
Test Failure: eos-vm-oc compile_monitor failed to read response from monitor process


linh2931:
Got this failure while trying to reproduce resource monitor plugin test failure by manually repeating CI:

https://github.com/AntelopeIO/leap/actions/runs/6444613909/job/17505200114

PR #1821, opened 2023-10-25
[5.0] refactor threading of `snapshot_scheduler_test`

spoonincode:
In the process of investigating #1794 I got my build in to a state (via various compile options & compilers I guess) where `snapshot_scheduler_test` would consistently fail via
```
test_snapshot_scheduler.cpp(115): fatal error: in "producer_snapshot_scheduler_tests/snapshot_scheduler_test": critical check validate_snapshot_request(4, 12, 10, true) has failed
```
Upon closer look, some aspects of threading look invalid and racy in this test. After this refactor my troublesome build always passes. I'd like to consider this resolving #1794 until we know otherwise (we should know shortly :slightly_smiling_face:)

In particular,
There is a race between the `app->exec()` and the `chain_plug->chain().block_start.connect()` -- blocks could be created before that connection is made and thus missed by the connection. I resolved this by moving the connection to occur before `app->exec()`. I believe this is where my local failure stumbled.

`pp->schedule_snapshot()` and `pp->get_snapshot_requests()` touch variables that are not guarded by any mutex. These should only be called by the "main thread" (in this case the `app_thread`) so wrap them in a `.post()` to the app.

I also replaced the effectively `sleep(10)` for a wait on block 20 (however long it takes).

### DIFF
--- a/tests/test_snapshot_scheduler.cpp
+++ b/tests/test_snapshot_scheduler.cpp
@@ -60,6 +60,9 @@ BOOST_AUTO_TEST_CASE(snapshot_scheduler_test) {
          std::promise<std::tuple<producer_plugin*, chain_plugin*>> plugin_promise;
          std::future<std::tuple<producer_plugin*, chain_plugin*>> plugin_fut = plugin_promise.get_future();
 
+         std::promise<void> at_block_20_promise;
+         std::future<void> at_block_20_fut = at_block_20_promise.get_future();
+
          std::thread app_thread([&]() {
             try {
                fc::logger::get(DEFAULT_LOGGER).set_log_level(fc::log_level::debug);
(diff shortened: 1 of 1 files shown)


linh2931 (approved):
Great fix!

The `failed to read response from monitor process` failure now appears in another test: https://github.com/AntelopeIO/leap/actions/runs/6634407607/job/18023991204. I added some some debug logging in `compile_monitor.cpp` in `oc_compile_monitor_debug` branch, run CICD manually hoping to see some logs, but the logs I added did not show up. I thought the monitor process might have already exited before `get_connection_to_compile_monitor` was called.

I could not reproduce it locally.

spoonincode:
Not sure... I've been working under the premise that the error message was actually indicative of a crash since that's what we were thinking with #1736/#1766. But these tests we're seeing the error message on are of the few that reuse appbase in this manner... so maybe it's something with the way plugins are getting destroyed/recreated that's messing something up in OC's external process glue.

fwiw I've been manually hitting retry on this PR's jobs some this evening (slow going), and have a handful of runs that passed (no errors so far),
https://github.com/AntelopeIO/leap/actions/runs/6634069566/job/18023135214
https://github.com/AntelopeIO/leap/actions/runs/6634069566/job/18023135923
https://github.com/AntelopeIO/leap/actions/runs/6634069566/job/18023136456
https://github.com/AntelopeIO/leap/actions/runs/6634069566/job/18023136593
https://github.com/AntelopeIO/leap/actions/runs/6634069566/job/18023637759
https://github.com/AntelopeIO/leap/actions/runs/6634069566/job/18023755275
https://github.com/AntelopeIO/leap/actions/runs/6634069566/job/18023846713
https://github.com/AntelopeIO/leap/actions/runs/6634069566/job/18024167831
https://github.com/AntelopeIO/leap/actions/runs/6634069566/job/18024243576
https://github.com/AntelopeIO/leap/actions/runs/6634069566/job/18024819879

heifner (approved):

PR #1827, opened 2023-10-26
[5.0] don't run EOS VM OC's monitor compile task callback when socket being dtored

### LINKED ISSUE #1823: Test Failure: "failed to read response from monitor process" in read_only_trxs/with_3_read_only_threads
Previously we saw `failed to read response from monitor process` in `snapshot_scheduler_test`. PR #1821 fixed that.

Now we see it happening in `read_onlu_trxs`:
https://github.com/AntelopeIO/leap/actions/runs/6641360943/job/18044099007

spoonincode:
This may resolve #1823 & #1794 based on core dump analysis of an `oc-monitor` crash.

A single OC monitor process is created at launch of a parent process which plans to use OC (nodeos, a unit test application, etc). Critically, a single io_context is used in this single monitor process. As `code_cache`s are created on the parent process, `compile_monitor_session`s are created in the monitor process for each of those `code_cache` instances.

When a `code_cache` needs to compile a contract, it sends a message to its out of process `compile_monitor_session`. This will then call off to _another_ process -- the OC trampoline -- but the most important actions here to note is that the `compile_monitor_session` notes in its `current_compiles` a socket to receive a reply on from the trampoline,
https://github.com/AntelopeIO/leap/blob/81a9d5cfe948b8da849798f87e51268ce1e38c75/libraries/chain/webassembly/runtimes/eos-vm-oc/compile_monitor.cpp#L110
and then goes off to do an `async_wait()` on that socket,
https://github.com/AntelopeIO/leap/blob/81a9d5cfe948b8da849798f87e51268ce1e38c75/libraries/chain/webassembly/runtimes/eos-vm-oc/compile_monitor.cpp#L116

Consider that `current_compiles` holds some outstanding sockets because compiles are ongoing. Now consider that `code_cache` is destroyed. When that happens the `read_message_from_nodeos()` (which is a misnomer: it's really read_message_from_codecache) will indicate a signal,
https://github.com/AntelopeIO/leap/blob/81a9d5cfe948b8da849798f87e51268ce1e38c75/libraries/chain/webassembly/runtimes/eos-vm-oc/compile_monitor.cpp#L57-L61
this signal is an indication that the `compile_monitor_session` can be destroyed (since its code_cache pairing has been destroyed). So.. it's destroyed,
https://github.com/AntelopeIO/leap/blob/81a9d5cfe948b8da849798f87e51268ce1e38c75/libraries/chain/webassembly/runtimes/eos-vm-oc/compile_monitor.cpp#L211-L215

As `compile_monitor_session` is destroyed its `current_compiles` is first to be dtored. As part of this destruction the `local::datagram_protocol::socket` will be destroyed ergo it will be `cancel()`ed egro it will make a callback to the `async_wait()` with `boost::asio::error::operation_aborted`. That's not strictly a problem, but things fall apart quickly here because the `async_wait()` handler then accesses an iterator to the currently-being-destroyed `current_compiles`, along with potentially using the currently-being-destroyed socket.

So, bail doing anything in `async_wait()`'s callback if there is an error.

Unfortunately while all the above makes sense, I can't really explain why the failure doesn't happen more often with just nodeos. My hunch is that a crashing oc-monitor on nodeos shutdown is simply never noticed nor has any negative side effects. It's only in these new tests that expect oc-monitor to (as designed) be long lived where something gets angry that oc-monitor has disappeared (because it crashed)

### DIFF
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/compile_monitor.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/compile_monitor.cpp
@@ -114,7 +114,13 @@ struct compile_monitor_session {
    void read_message_from_compile_task(std::list<std::tuple<code_tuple, local::datagram_protocol::socket>>::iterator current_compile_it) {
       auto& [code, socket] = *current_compile_it;
       socket.async_wait(local::datagram_protocol::socket::wait_read, [this, current_compile_it](auto ec) {
-         //at this point we only expect 1 of 2 things to happen: we either get a reply (success), or we get no reply (failure)
+         //at this point we generally expect 1 of 2 things to happen: we either get a reply (success), or we get an error reading from the
+         // socket (failure). But there is also a third possibility that this compile_monitor_session is being destroyed and thus the
+         // socket is being destroyed by way of current_compiles being destroyed. Since this is an async_wait() and not an async_read(),
+         // for now just consider any error as being due to cancellation at dtor time and completely bail out (there aren't many other
+         // potential errors for an asnyc_wait)
+         if(ec)
+            return;
          auto& [code, socket] = *current_compile_it;
          auto [success, message, fds] = read_message_with_fds(socket);
          


heifner (approved):

libraries/chain/webassembly/runtimes/eos-vm-oc/compile_monitor.cpp
```diff
@@ -114,7 +114,13 @@ struct compile_monitor_session {
...
       auto& [code, socket] = *current_compile_it;
       socket.async_wait(local::datagram_protocol::socket::wait_read, [this, current_compile_it](auto ec) {
-         //at this point we only expect 1 of 2 things to happen: we either get a reply (success), or we get no reply (failure)
+         //at this point we generally expect 1 of 2 things to happen: we either get a reply (success), or we get an error reading from the
+         // socket (failure). But there is also a third possibility that this compile_monitor_session is being destroyed and thus the
+         // socket is being destroyed by way of current_compiles being destroyed. Since this is an async_wait() and not an async_read(),
+         // for now just consider any error as being due to cancellation at dtor time and completely bail out (there aren't many other
+         // potential errors for an asnyc_wait)
+         if(ec)
+            return;
```
linh2931, line 123: Do you want to log this for potential debugging or just too much noises?

linh2931 (approved):

libraries/chain/webassembly/runtimes/eos-vm-oc/compile_monitor.cpp
heifner: No higher than `debug` if you do log it; don't think we need it.

spoonincode: Logging config won't be active in the monitor process since it's forked off before `main()` -- so debug won't be visible, and any error/warn/info messages won't honor the user's logging config either. So probably best to not log anything here.

Issue #1804, opened 2023-10-20
Add tests for produce-block-offset-ms


linh2931:
This comes from review discussion from https://github.com/AntelopeIO/leap/pull/1800 for future work.

This is not new behavior, it has shipped a block off early since v2.0.4. See https://github.com/EOSIO/eos/pull/8651
We should consider refactoring producer_plugin so it is easier to test. Or it might be possible to use mock_time like the test_trx_retry_db does. However, I believe that is out of scope for this PR. Please create a GitHub issue to work that later.

bhazzard:
Added a task as mentioned in the text above.

PR #1840, opened 2023-10-27
[5.0] Normalize speculative & producer block intervals

### LINKED ISSUE #1837: 5.0: lots of "not producing block" messages on the backup BP
With 5.0 (latest), we get a lot of "not producing block" messages on the backup BP.  The backup BP has the wrong key on purpose.

In previous version we just get 1 of these per block.

```
$ grep 'Block #105321099 eosnationftw' /var/log/syslog | wc -l
763
```

image

heifner:
#1481 changed speculative block creation so that speculative blocks would always have a current block time. However, it didn't correctly handle the case where the configured producer was the scheduled producer but could not produce (e.g. incorrect configured private key). This PR simplifies the implementation to just use the producer schedule as configured by `produce-block-offset-ms` regardless if the block is speculative or a producer block. This normalizes block production time for speculative and producer blocks. I think this normalization is better than maintaining different block intervals for speculative vs producer nodes.

Resolves #1837

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -396,7 +396,7 @@ public:
                                   account_name                                first_auth,
                                   int64_t                                     sub_bill,
                                   uint32_t                                    prev_billed_cpu_time_us);
-   
+
    void        log_trx_results(const transaction_metadata_ptr& trx, const transaction_trace_ptr& trace, const fc::time_point& start);
    void        log_trx_results(const transaction_metadata_ptr& trx, const fc::exception_ptr& except_ptr);
    void        log_trx_results(const packed_transaction_ptr& trx,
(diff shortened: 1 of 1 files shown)


linh2931 (commented):
Is it possible to add a test where the scheduled producer could not produce?

heifner:
> Is it possible to add a test where the scheduled producer could not produce?

I think that would be under https://github.com/AntelopeIO/leap/issues/1804

There are tests in `test_block_timing_util.cpp` that test the timings, so I feel the functionality is well tested; just not the actual usage here.

linh2931:
> I think that would be under #1804
> 
> There are tests in `test_block_timing_util.cpp` that test the timings, so I feel the functionality is well tested; just not the actual usage here.

I meant the situation reported by #1837. No more lots of "not producing" messages after this fix. But if it is too much effort, there is no need to do that, as the fix makes things simpler. I am approving the PR.

linh2931 (approved):

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1830,7 +1830,7 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
    }
 
    // create speculative blocks at regular intervals, so we create blocks with "current" block time
-   _pending_block_deadline = now + fc::microseconds(config::block_interval_us);
+   _pending_block_deadline = block_timing_util::calculate_producing_block_deadline(_produce_block_cpu_effort, block_time);
```
greg7mdp, line 1833: I believe that `calculate_producing_block_deadline` will return `std::nullopt` if `producers.empty()`.
Is it OK to not have a deadline in that case?

heifner: This is calling `calculate_producing_block_deadline` not `calculate_producer_wake_up_time`

greg7mdp: Oh, my bad, thanks!

greg7mdp (approved):

Issue #1573, opened 2023-08-27
Tests: disable EOS VM OC's subjective limits during unit tests


spoonincode:
When nodeos uses EOS VM OC it (..except for when a special compilation flag is used..) uses OC in a tier up manner: contract execution is dispatched to the baseline runtime (EOS VM JIT) while OC is in the process of compiling the contract _or potentially forever_ should OC decide to blacklist the contract due to it failing subjective compilation limits.

When EOS VM OC is used in unit tests it's used in a different manner: contract actions are not dispatched until EOS VM OC has finished compilation, and only EOS VM OC is used -- there is no baseline. I tend to refer to this as synchronous mode. This is an important mode of operation because generally all actions in a test need to be dispatched to OC to ensure complete test coverage with OC.

However, this means if a unit test runs afoul of EOS VM OC's subjective complication limits the unit test will fail: that unit test cannot dispatch the expected action because OC has blacklisted the contract and there is no baseline that can be used. This is most notably problematic for EOS EVM because EOS EVM's exhaustive tests run afoul of OC's subjective limits, but we'd really like to run those EVM exhaustive tests with OC to make sure everything is operating correctly with 5.0's default `eos-vm-oc-enable=auto`

OC's subjective limits also prevent ASAN from being used while OC is enabled, because ASAN's 16TB of virtual memory usage is flagged by OC's subjective limits.

OC's "synchronous mode" used during unit tests should disable its subjective limits. Bonus points for making the subjective limits knobs that a unit test can control as that may make it easier to ensure that the subjective limits are working properly.

PR #1843, opened 2023-10-31
[5.0] Disable EOS VM OC's subjective compilation limits in unit tests

linh2931:
When EOS VM OC is used in unit tests, contract actions are not dispatched until EOS VM OC has finished compilation. If a unit test violates EOS VM OC's subjective compilation limits, the unit test will fail. This is most notably problematic for EOS EVM because EOS EVM's exhaustive tests run afoul of OC's subjective limits. Another problem with OC's subjective limits  are they prevent ASAN from being used, because ASAN's 16TB of virtual memory usage is flagged by OC's subjective limits.

This PR disables OC subjective limits: `cpu limits`, `vm limits`, `stack size limit`, and `generated code size limit` in unit tests. It also adds unit tests to make sure limits to work properly.

Resolves https://github.com/AntelopeIO/leap/issues/1573


unittests/eosvmoc_limits_tests.cpp
```diff
@@ -0,0 +1,139 @@
...
+BOOST_AUTO_TEST_CASE( stack_limit ) { try {
+   eosvmoc::config eosvmoc_config;
+
+   // disable non-tested limits
+   eosvmoc_config.cpu_limits.rlim_cur       = RLIM_INFINITY;
+   eosvmoc_config.vm_limits.rlim_cur        = RLIM_INFINITY;
+   eosvmoc_config.generated_code_size_limit = std::numeric_limits<size_t>::max();
+
+   // The larget stack size in the test is 104. 
+   // Set stack_size_limit to highest value to verify wasm_execution_error is raised
```
heifner, line 108: I find this comment confusing. Maybe `// set stack_size_limit less than stack size of test`

unittests/eosvmoc_limits_tests.cpp
```diff
@@ -0,0 +1,139 @@
...
+   eosvmoc_config.cpu_limits.rlim_cur       = RLIM_INFINITY;
+   eosvmoc_config.vm_limits.rlim_cur        = RLIM_INFINITY;
+   eosvmoc_config.generated_code_size_limit = std::numeric_limits<size_t>::max();
+
+   // The larget stack size in the test is 104. 
+   // Set stack_size_limit to highest value to verify wasm_execution_error is raised 
+   eosvmoc_config.stack_size_limit = 103;
+   limit_violated_test(eosvmoc_config);
+
+   // set stack_size_limit to lowest value to verify wasm_execution_error is not raised
```
heifner, line 112: I find this comment confusing as well. Maybe `// set stack_size_limit one above the actual needed stack size`

unittests/eosvmoc_limits_tests.cpp
```diff
@@ -0,0 +1,139 @@
...
+   eosvmoc::config eosvmoc_config;
+
+   // disable non-tested limits
+   eosvmoc_config.cpu_limits.rlim_cur = RLIM_INFINITY;
+   eosvmoc_config.vm_limits.rlim_cur  = RLIM_INFINITY;
+   eosvmoc_config.stack_size_limit    = std::numeric_limits<uint64_t>::max();
+
+   // The sgenerated code size in the test is 36856. 
+   // Set generated_code_size_limit to highest value to verify wasm_execution_error
+   // is raised
```
heifner, line 128: Same as above this comment could be clearer I think.

unittests/eosvmoc_limits_tests.cpp
```diff
@@ -0,0 +1,139 @@
...
+   eosvmoc_config.vm_limits.rlim_cur  = RLIM_INFINITY;
+   eosvmoc_config.stack_size_limit    = std::numeric_limits<uint64_t>::max();
+
+   // The sgenerated code size in the test is 36856. 
+   // Set generated_code_size_limit to highest value to verify wasm_execution_error
+   // is raised 
+   eosvmoc_config.generated_code_size_limit = 36856;
+   limit_violated_test(eosvmoc_config);
+
+   // set generated_code_size_limit to lowest value to verify wasm_execution_error is not raised
```
heifner, line 132: This comment as well I think needs to be clearer.

heifner (changes requested):

heifner (approved):

libraries/testing/tester.cpp
```diff
@@ -317,6 +317,17 @@ namespace eosio { namespace testing {
...
+      // libtester does not enforce EOSVMOC limits for all tests except those in
+      // eosvmoc_limits_test.cpp, where one or more of the limits are set to
+      // max to indicate the limits are enforced.
+      if (cfg.eosvmoc_config.cpu_limits.rlim_cur != RLIM_INFINITY && cfg.eosvmoc_config.vm_limits.rlim_cur != RLIM_INFINITY && cfg.eosvmoc_config.stack_size_limit != std::numeric_limits<uint64_t>::max() && cfg.eosvmoc_config.generated_code_size_limit != std::numeric_limits<size_t>::max()) {
+         // set to inifinity or max such that they are not enforced.
+         cfg.eosvmoc_config.cpu_limits.rlim_cur       = RLIM_INFINITY;
+         cfg.eosvmoc_config.vm_limits.rlim_cur        = RLIM_INFINITY;
+         cfg.eosvmoc_config.stack_size_limit          = std::numeric_limits<uint64_t>::max();
+         cfg.eosvmoc_config.generated_code_size_limit = std::numeric_limits<size_t>::max();
+      }
```
spoonincode, line 329: I can't help but feel `std::optional` would be a better choice for these items. `default_config()` could be modified to simply `.reset()` each of them,
https://github.com/AntelopeIO/leap/blob/e4df534c675e8133b257a783cde4326c76618c93/libraries/testing/include/eosio/testing/tester.hpp#L400-L410
and since that occurs before validating_tester's `Lambda conf_edit`, eosvmoc_limits_tests can still override them back to however it wants them.

Regardless of `optional` usage, doing this part in `default_config()` feels like right spot since that's where other libtester-specific configuration items are set.

An optional should be able to be plumbed through easily, and something like
https://github.com/AntelopeIO/leap/blob/e4df534c675e8133b257a783cde4326c76618c93/libraries/chain/webassembly/runtimes/eos-vm-oc/compile_trampoline.cpp#L172-L174
becomes
```c++
if(conf.cpu_limits) {
   struct rlimit cpu_limit = {*conf.cpu_limit, *conf.cpu_limit};
   setrlimit(RLIMIT_CPU, &cpu_limit);
}
```

spoonincode (changes requested):
The only feedback I'd consider important is moving the libtester configuration in to `default_config()` if we can, which ought to remove that large `if( .. &&` I believe.

Using an optional and getting rid of soft/hard as part of the config isn't a blocker.

linh2931:
> The only feedback I'd consider important is moving the libtester configuration in to `default_config()` if we can, which ought to remove that large `if( .. &&` I believe.
> 
> Using an optional and getting rid of soft/hard as part of the config isn't a blocker.

Thank you Matt @spoonincode for you suggestions! They make code much simpler. I have incorporated them.

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/config.hpp
```diff
@@ -7,11 +7,25 @@
...
+   // in the tests in unittests/eosvmoc_limits_tests.cpp.
+   std::optional<rlim_t>   cpu_limit {20u};
+   std::optional<rlim_t>   vm_limit  {512u*1024u*1024u};
+   std::optional<uint64_t> stack_size_limit {16u*1024u};
+   std::optional<size_t>   generated_code_size_limit {16u*1024u*1024u};
 };
 
 }}}
+
+FC_REFLECT(rlimit, (rlim_cur)(rlim_max))
```
spoonincode, line 30: I think you can remove this `FC_REFLECT` now

spoonincode (approved):

libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/config.hpp
linh2931: Thanks. I missed this.

PR #1845, opened 2023-10-31
[5.0] Minimize abort/start block for speculative blocks

heifner:
Simplify and fix issues with #1481 block deadline calculation. Remove the concept of a hard deadline as producers want to produce a block even if the deadline has passed for the block. This uses a conservative calculation for producer wake up time, but does verify it is in the future.

Resolves #1831

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -772,6 +772,7 @@ public:
    }
 
    void restart_speculative_block() {
+      // log message is used by Node.py verifyStartingBlockMessages in distributed-transactions-test.py test
       fc_dlog(_log, "Restarting exhausted speculative block #${n}", ("n", chain_plug->chain().head_block_num() + 1));
       // abort the pending block
       abort_block();
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -1844,7 +1844,7 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
 
    // create speculative blocks at regular intervals, so we create blocks with "current" block time
    _pending_block_deadline = block_timing_util::calculate_producing_block_deadline(_produce_block_cpu_effort, block_time);
-   if (in_speculating_mode()) { // if we are producing, then produce block even if deadline has pasted
+   if (in_speculating_mode()) { // if we are producing, then produce block even if deadline has passed
       // speculative block, no reason to start a block that will immediately be re-started, set deadline in the future
       // a block should come in during this time, if not then just keep creating the block every produce_block_cpu_effort
       if (now + fc::milliseconds(config::block_interval_ms/10) > _pending_block_deadline) {
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -1829,8 +1829,6 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
       last_start_block_time = now;
    }
 
-   // create speculative blocks at regular intervals, so we create blocks with "current" block time
-   _pending_block_deadline = block_timing_util::calculate_producing_block_deadline(_produce_block_cpu_effort, block_time);
    if (in_producing_mode()) {
       uint32_t production_round_index = block_timestamp_type(block_time).slot % chain::config::producer_repetitions;
       if (production_round_index == 0) {
@@ -1844,6 +1842,15 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
       }
    }
 
+   // create speculative blocks at regular intervals, so we create blocks with "current" block time
+   _pending_block_deadline = block_timing_util::calculate_producing_block_deadline(_produce_block_cpu_effort, block_time);
+   if (in_speculating_mode()) { // if we are producing, then produce block even if deadline has pasted
+      // speculative block, no reason to start a block that will immediately be re-started, set deadline in the future
+      // a block should come in during this time, if not then just keep creating the block every produce_block_cpu_effort
+      if (now + fc::milliseconds(config::block_interval_ms/10) > _pending_block_deadline) {
+         _pending_block_deadline = now + _produce_block_cpu_effort;
+      }
+   }
    const auto& preprocess_deadline = _pending_block_deadline;
 
    fc_dlog(_log, "Starting block #${n} at ${time} producer ${p}", ("n", pending_block_num)("time", now)("p", scheduled_producer.producer_name));
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -772,9 +772,9 @@ public:
    }
 
    void restart_speculative_block() {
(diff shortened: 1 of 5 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1844,6 +1842,15 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
       }
    }
 
+   // create speculative blocks at regular intervals, so we create blocks with "current" block time
+   _pending_block_deadline = block_timing_util::calculate_producing_block_deadline(_produce_block_cpu_effort, block_time);
+   if (in_speculating_mode()) { // if we are producing, then produce block even if deadline has pasted
```
greg7mdp, line 1847:
```suggestion
   if (in_speculating_mode()) { // if we are producing, then produce block even if deadline has passed
```

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -2488,9 +2495,13 @@ void producer_plugin_impl::schedule_production_loop() {
       chain::controller& chain = chain_plug->chain();
       fc_dlog(_log, "Speculative Block Created; Scheduling Speculative/Production Change");
       EOS_ASSERT(chain.is_building_block(), missing_pending_block_state, "speculating without pending_block_state");
-      auto wake_time = block_timing_util::calculate_producer_wake_up_time(_produce_block_cpu_effort, chain.pending_block_num(), chain.pending_block_timestamp(),
+      auto wake_time = block_timing_util::calculate_producer_wake_up_time(fc::microseconds{config::block_interval_us}, chain.pending_block_num(), chain.pending_block_timestamp(),
                                                                           _producers, chain.head_block_state()->active_schedule.producers,
                                                                           _producer_watermarks);
+      if (wake_time && fc::time_point::now() > *wake_time) {
+         // if wake time has already passed then use the block deadline instead
+         wake_time = _pending_block_deadline;
```
greg7mdp, line 2503: should we check that `_pending_block_deadline` > `fc::time_point::now()`?

greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: I mean, if `_pending_block_deadline` < `fc::time_point::now()`, we would still be setting a `wake_time ` that is in the past.

heifner: Right, setting it in the past will mean it immediately will wake up and start a new block. If the pending deadline has passed then that is what we want I believe.

greg7mdp: OK, perfect, thanks!

tests/TestHarness/Node.py
```diff
@@ -550,6 +550,38 @@ def findInLog(self, searchStr):
...
+            duplicatesFound = False
+            lastRestartBlockNum = 0
+
+            with open(f, 'r') as file:
+                for line in file:
+                    match = re.match(r".*Restarting exhausted speculative block #(\d+)", line)
+                    if match:
+                        lastRestartBlockNum = match.group(1)
+                    if re.match(r".*unlinkable_block_exception", line):
+                        lastRestartBlockNum = blockNumber
```
linh2931, line 569: `blockNumber` is first defined in line 572. Are we sure it has already been defined when `if re.match(r".*unlinkable_block_exception", line):` holds?

tests/TestHarness/Node.py
```diff
@@ -550,6 +550,38 @@ def findInLog(self, searchStr):
...
+            blockNumbers = set()
+            duplicateBlockNumbers = set()
+            duplicatesFound = False
+            lastRestartBlockNum = 0
+
+            with open(f, 'r') as file:
+                for line in file:
+                    match = re.match(r".*Restarting exhausted speculative block #(\d+)", line)
+                    if match:
+                        lastRestartBlockNum = match.group(1)
```
linh2931, line 567: Can "continue" be put here inside the match? The rest of matches will not be true.

tests/TestHarness/Node.py
```diff
@@ -550,6 +550,38 @@ def findInLog(self, searchStr):
...
+                    if match:
+                        blockNumber = match.group(1)
+                        if blockNumber != lastRestartBlockNum and blockNumber in duplicateBlockNumbers:
+                            print(f"Duplicate Staring block found: {blockNumber} in {f}")
+                            duplicatesFound = True
+                        if blockNumber != lastRestartBlockNum and blockNumber in blockNumbers:
+                            duplicateBlockNumbers.add(blockNumber)
+                        blockNumbers.add(blockNumber)
+
+            if duplicatesFound:
```
linh2931, line 580: What happens if the same Start Block is spread across two stderr files?

tests/TestHarness/Node.py
```diff
@@ -550,6 +550,38 @@ def findInLog(self, searchStr):
...
         return False
 
+    # verify only one or two 'Starting block' per block number unless block is restarted
+    def verifyStartingBlockMessages(self):
+        dataDir=Utils.getNodeDataDir(self.nodeId)
+        files=Node.findStderrFiles(dataDir)
+        for f in files:
+            blockNumbers = set()
+            duplicateBlockNumbers = set()
+            duplicatesFound = False
```
linh2931, line 560: should "duplicate" be renamed as only 3 or more is a failure.

plugins/producer_plugin/test/test_block_timing_util.cpp
```diff
@@ -76,21 +77,49 @@ BOOST_AUTO_TEST_CASE(test_calculate_block_deadline) {
       fc::mock_time_traits::set_now(seventh_block_time.to_time_point() - fc::milliseconds(500));
 
       BOOST_CHECK_EQUAL(calculate_producing_block_deadline(cpu_effort, seventh_block_time),
-                        seventh_block_time.to_time_point() - fc::milliseconds(100));
+                        seventh_block_time.to_time_point() - fc::milliseconds(700));
```
linh2931, line 80: Can a comment be added why 700, 800, 900?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -772,9 +772,9 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
    }
 
    void restart_speculative_block() {
+      fc_dlog(_log, "Restarting exhausted speculative block #${n}", ("n", chain_plug->chain().head_block_num() + 1));
```
linh2931, line 775: A comment can be added here to indicate this dlog is for integration test, so it won't be removed accidentally in the furture.

tests/TestHarness/Node.py
heifner: This `if` is inside the loop over the files. The first file that is found with a duplicate the verify stops. Might be better to scan all files for reporting. I'll make that change.

tests/TestHarness/Node.py
heifner: renamed

plugins/producer_plugin/producer_plugin.cpp
heifner: Adde comment

linh2931 (approved):

PR #1846, opened 2023-11-01
[5.0] Recover keys in chain thread pool

heifner:
Using the chain thread pool improves TPS (eosio.token transfer per second) significantly. From ~25K to ~31K. Since this is a rather simple change targeting it for 5.0. With this PR the sequence is net-thread => chain-thread => producer-thread => main-thread.

An improved version of this code that avoids a thread hop will target 6.0; net-thread => chain-thread => main-thread.

Removing 2 thread hops is also possible by recovering the keys on the net thread, however that overwhelms the main thread; net-thread => main-thread. That improvement would require throttling the net threads so they do not overwhelm the main thread. The current mechanism of a limit on the number of active trxs works as a protection but is not ideal for performance measurement as it drops trxs instead of just delaying them.

Issue #1690

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -802,7 +802,7 @@ public:
       fc::microseconds   max_trx_cpu_usage = max_trx_time_ms < 0 ? fc::microseconds::maximum() : fc::milliseconds(max_trx_time_ms);
 
       auto future = transaction_metadata::start_recover_keys(trx,
-                                                             _thread_pool.get_executor(),
+                                                             chain.get_thread_pool(),
                                                              chain.get_chain_id(),
                                                              fc::microseconds(max_trx_cpu_usage),
                                                              trx_type,


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -802,7 +802,7 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
       fc::microseconds   max_trx_cpu_usage = max_trx_time_ms < 0 ? fc::microseconds::maximum() : fc::milliseconds(max_trx_time_ms);
 
       auto future = transaction_metadata::start_recover_keys(trx,
-                                                             _thread_pool.get_executor(),
+                                                             chain.get_thread_pool(),
```
linh2931, line 805: Will this change have any side effects on controller's thread pool operation?

What in release/5.0 cause the slowdown? In release/3.2 the same producer_plugin's thread pool was used and 31k TPS was achieved.

greg7mdp, line 805: Why is using the chain `thread_pool` faster than using the producer_plugin `thread_pool`?

heifner: I think the issue is with the way it is using the same thread pool to recover the keys and then wait for the recover of the keys.

linh2931 (approved):

plugins/producer_plugin/producer_plugin.cpp
heifner: For additional information. `post` is simply the `boost::post()` line. `exec` is complete execution of the contract including loading WASM and everything. Clearly these `post()` calls are killing performance.
```
 post calls: 10000, total: 162933us, avg: 16us
 exec calls: 10000, total: 167621us, avg: 16us
```

greg7mdp (approved):

PR #1849, opened 2023-11-01
[5.0] globally (effectively) `--catch_system_errors=no` all boost-test based unit tests

spoonincode:
Some of our tests require `--catch_system_errors=no` to properly operate, but there is another compelling reason to blanket this option on _all_ tests: gathering core dumps. Default behavior (`--catch_system_errors=yes`) will block core dump creation for those tests should they fail. Instead of needing to remember to sprinkle `--catch_system_errors=no` in all the various `add_test()` declarations (prone to missing some), we have some other options,
* set `BOOST_TEST_CATCH_SYSTEM_ERRORS` environment variable in CI. This allows CI to gobble up core dumps from any test, but introduces a difference in CI vs non-CI behavior
* plumb the option in boost test's C++ initialization (but this requires manual plumbing for every test)
* set the `BOOST_TEST_DEFAULTS_TO_CORE_DUMP` preprocessor macro
  * Secondary option: plumb through a new `INTERFACE` library that depends on `Boost::unit_test_framework` to add  `BOOST_TEST_DEFAULTS_TO_CORE_DUMP`; but this still requires remembering in every place we use boost test we need to link to `leap_boost_unit_test` instead of `Boost::unit_test_framework`

As much as I detest "global" cmake compile definitions, the `BOOST_TEST_DEFAULTS_TO_CORE_DUMP` preprocessor macro seems the best option in terms of being a single location with consistent blanket behavior. Certainly open to opinions though.

Work on #640

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -175,6 +175,10 @@ endif()
 # do not attempt to use an external OpenSSL in any manner
 set(CMAKE_DISABLE_FIND_PACKAGE_OpenSSL On)
 
+# many tests require handling of signals themselves and even when they don't we'd like for them to generate a core dump, this
+# is effectively --catch_system_errors=no broadly across all tests
+add_compile_definitions(BOOST_TEST_DEFAULTS_TO_CORE_DUMP)
+
 add_subdirectory( libraries )
 add_subdirectory( plugins )
 add_subdirectory( programs )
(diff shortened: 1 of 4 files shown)


heifner (approved):
I've always considered default behavior of `--catch_system_errors=yes` to be an anti-feature.

greg7mdp (approved):

PR #1850, opened 2023-11-01
upload core dumps from failed tests in CI

### LINKED ISSUE #640: CICD: Capture core dump files from CI runs
This is a carry over of eosnetworkfoundation/mandel#139, but I think we're pretty good now about capturing everything but core dump files? So I'm eliding the title to just reference core dumps.

Original issue:
> b1's CI was diligent in [archiving all logs, data files, and even core dump files](https://github.com/EOSIO/eos/blob/62b8c9964cdf9a3b6afc5ae7cd562468920ef700/.cicd/test.sh#L25-L48) from the tests it ran. This makes it easier to troubleshoot why a test failed compared to just seeing the stdout of `ctest`. Any refactor of Mandel's CI should strive to archive all these artifacts too.

spoonincode:
Core dumps will now be uploaded from crashing tests in CI. Resolves #640. Example of failure and uploaded files can be found here,
https://github.com/AntelopeIO/leap/actions/runs/6724910321
(I trimmed the run down to just Tests & NP Tests on a single platform)

While I didn't change anything on the ENF runners to accomplish this, it's worth noting that this probably only works with ENF runners because of their behavior to put core dumps in to `/var/lib/systemd/coredump`. Something like Github's runners don't do that. While I like to keep the workflows runner-agnostic so that forks could potentially work in the future, afaik this won't break the workflow on other runners; you'll just miss out on the core dumps.

Coming along for the ride in some separate commits were a couple modifications to use node20 instead of node16. In some of our repos starting to notice warnings about using node16, for example: https://github.com/AntelopeIO/bls12-381/actions/runs/6633721660 (I'm not sure why we haven't seen warnings in leap; possibly a slow rollout by Github?). So,
* upgrade to actions/checkout@v4, and
* upgrade parallel-ctest-containers to use node20

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -138,7 +138,7 @@ jobs:
     runs-on: ["self-hosted", "enf-x86-hightier"]
     container:
       image: ${{fromJSON(needs.platform-cache.outputs.platforms)[matrix.cfg.base].image}}
-      options: --security-opt seccomp=unconfined
+      options: --security-opt seccomp=unconfined --mount type=bind,source=/var/lib/systemd/coredump,target=/cores
     steps:
       - uses: actions/checkout@v4
       - name: Download builddir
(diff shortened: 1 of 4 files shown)


heifner (approved):

linh2931 (approved):

.github/workflows/build.yaml
```diff
@@ -181,12 +188,17 @@ jobs:
           log-tarball-prefix: ${{matrix.cfg.name}}
           tests-label: nonparallelizable_tests
           test-timeout: 420
+      - name: Export core dumps
+        run: docker run --mount type=bind,source=/var/lib/systemd/coredump,target=/cores alpine sh -c 'tar -C /cores/ -c .' | tar x
```
spoonincode, line 192: I know this looks weird, and to add more confusion it's different than the parallel tests job because NP/LR tests don't run the job in a container until `parallel-ctest-containers` step.

On ENF runners the user is _not_ `root` but rather a user named `enf` that does not have sudo privileges (this is unlike GitHub runners). But `/var/lib/systemd/coredump` files are only readable by `root`. Since docker is not creating a userns in this case, we can use docker as a cheat to get access to the files without `sudo` and without addtional modificatons on the ENF runners.

PR #1851, opened 2023-11-01
Remove per-instance allocator and empty allocations from chainbase containers. Add `cow` vector.

### LINKED ISSUE #1652: Remove per-instance allocator and empty allocations from chainbase containers
There are a few improvements to chainbase that can reduce wasted space in the shared memory file that exist in the `chainbase-container-allocator-poc` branch

## Per-instance allocator storage

each `shared_vector` and `shared_cow_string` stores its own copy of the chainbase allocator. This is always the same pointer even when it has been cast into a different sized allocator (eg `node` vs `char` allocator). These types need access to the allocator for their mutating actions as well as destruction.  As it is always the same so the POC implements storage as a global static member of the container class with assertions that the assumption its the same holds. Some method of removing all the duplicated allocators suffices. In order to achieve this for `shared_vector` a new `shared_cow_vector` was created. this has the additional benefit of creating fewer spurious allocations in the shared memory file on mutations which may lead to a less fragmented file.

**note: removing allocators from `shared_cow_string` exposed an implementation bug where `sizeof(shared_cow_string)` was used in place of `sizeof(shared_cow_string::impl)` in the deallocation.  Somehow this was not creating issues?

## empty `shared_cow_string` allocations

When a `shared_cow_string` is initialized at runtime (eg, when a new account is created and its ABI is not set) it does not allocate a `shared_cow_string::impl` structure. The `nullptr` held in the `shared_cow_string` is equivalent to an empty string.  However, when a `shared_cow_string` is read from a Datastream or variant, the behavior is different. In that case `resize_and_fill` or an equivalent is used and this *always allocates* even if the size passed in is 0.  This results in extra bloat when loading from a snapshot that is not present when the state is derived from replay. The POC prevents allocation for empty structures. 

## Results

When testing the POC at block 310859547 from snapshot these changes reduced the initial state size from 23,263,420,992 bytes to 21,785,865,104 bytes (or around 6% savings)

greg7mdp:
Resolves #1652.

In addition to the chainbase changes detailed [here](https://github.com/AntelopeIO/chainbase/pull/27), this PR defines the chain's `shared_vector` as chainbase's `shared_cow_vector` instead of `boost::interprocess::vector<T, allocator<T>>`, and updates the classes using `shared` objects as necessary.

This PR reduces the chainbase memory usage as reported by the db_size api by **6.33%** (from `21.64 GiB` to `20.27 Gib`, measured after loading snapshot `snapshot-2023-05-15-16-eos-v6-0309964229.bin`).

In addition, we have also eliminated unnecessary buffer copies, and when temporary `shared` strings or vectors are created on the stack, they use the regular heap (via std::allocator) instead of the chainbase memory mapped segment.


libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -426,6 +424,31 @@ namespace chainbase {
...
    }
+
+   // chainbase::shared_cow_vector
+   template<typename DataStream, typename T> inline DataStream& operator<<( DataStream& s, const chainbase::shared_cow_vector<T>& v )  {
+      FC_ASSERT( v.size() <= MAX_SIZE_OF_BYTE_ARRAYS );
+      fc::raw::pack( s, fc::unsigned_int((uint32_t)v.size()));
+      const auto* data = v.data();
+      auto size = v.size();
+      for (std::size_t i = 0; i < size; ++i) {
+         fc::raw::pack(s, &data[i]);
```
heifner, line 435: Shouldn't this be `fc::raw::pack(s, data[i]);`

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -426,6 +424,31 @@ namespace chainbase {
...
+   }
+
+   template<typename DataStream, typename T> inline DataStream& operator>>( DataStream& s, chainbase::shared_cow_vector<T>& v )  {
+      fc::unsigned_int size;
+      fc::raw::unpack( s, size );
+      FC_ASSERT( size.value <= MAX_SIZE_OF_BYTE_ARRAYS );
+      FC_ASSERT( v.size() == 0 );
+      v.resize_and_fill(size.value, [&](auto* data, std::size_t sz) { 
+         for (std::size_t i = 0; i < sz; ++i) {
+            fc::raw::unpack(s, &data[i]);
```
heifner, line 447: Shouldn't this be `fc::raw::unpack(s, data[i]);`

heifner (changes requested):

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -100,23 +100,21 @@ namespace eosio::chain {
...
-         :shared_string(f,l,a)
+         shared_blob(InputIterator f, InputIterator l)
+         :shared_string(f,l)
          {}
 
-         shared_blob(const allocator_type& a)
-         :shared_string(a)
+         shared_blob()
+         :shared_string()
          {}
```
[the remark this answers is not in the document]
greg7mdp: No, because the `shared_blob` class doesn't have a constructor taking std::string (even though its base class does). I thought of adding one, but I thought this was the least intrusive change.

libraries/chain/include/eosio/chain/database_utils.hpp
```diff
@@ -183,7 +171,8 @@ namespace fc {
    inline
    void from_variant( const variant& v, eosio::chain::shared_blob& b ) {
       std::string _s = base64_decode(v.as_string());
-      b = eosio::chain::shared_blob(_s.begin(), _s.end(), b.get_allocator());
+      std::destroy_at(std::addressof(b));
+      new (std::addressof(b)) eosio::chain::shared_blob(_s.begin(), _s.end());
```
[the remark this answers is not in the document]
greg7mdp: No, because the `shared_blob` class doesn't have a constructor taking std::string (even though its base class does). I thought of adding one, but I thought this was the least intrusive change.

libraries/chain/include/eosio/chain/authority.hpp
```diff
@@ -116,20 +116,17 @@ struct shared_key_weight {
...
-            wa_ss.resize_and_fill( psz, [&wa]( char* data, std::size_t sz ) {
-               fc::datastream<char*> ds(data, sz);
-               fc::raw::pack(ds, wa);
-            });
-
-            return shared_key_weight(std::move(wa_ss), k.weight);
+            std::string wa_ss(psz, 'a');
+            fc::datastream<char*> ds(wa_ss.data(), psz);
+            fc::raw::pack(ds, wa);
+            key.pubkey.emplace<shared_string>(wa_ss);
```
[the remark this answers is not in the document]
greg7mdp: Yes! We didn't do it before but indeed we can. Done.

libraries/chain/include/eosio/chain/database_utils.hpp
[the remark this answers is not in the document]
heifner: I think we should add the constructor as this is rather ugly to do here.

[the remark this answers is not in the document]
greg7mdp: But I just added a `std::string_view` constructor to `shared_blob` so it works now.

libraries/chain/include/eosio/chain/types.hpp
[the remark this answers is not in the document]
greg7mdp: Yes, updated.

libraries/chain/include/eosio/chain/authority.hpp
```diff
@@ -221,19 +220,22 @@ struct authority {
...
-      keys.reserve(a.keys.size());
-      for(const key_weight& k : a.keys) {
-         keys.emplace_back(shared_key_weight::convert(keys.get_allocator(), k));
-      }
-      accounts = decltype(accounts)(a.accounts.begin(), a.accounts.end(), accounts.get_allocator());
-      waits = decltype(waits)(a.waits.begin(), a.waits.end(), waits.get_allocator());
-      return *this;
+   shared_authority() = default;
+
+   shared_authority(const authority& auth) :
```
heifner, line 225: `explicit`

libraries/chain/include/eosio/chain/producer_schedule.hpp
```diff
@@ -46,24 +46,28 @@ namespace eosio { namespace chain {
       };
    }
 
+   struct block_signing_authority_v0;
+
    struct shared_block_signing_authority_v0 {
-      shared_block_signing_authority_v0() = delete;
+      shared_block_signing_authority_v0() = default;
+      shared_block_signing_authority_v0( const block_signing_authority_v0& );
```
heifner, line 53: `explicit`

libraries/chain/include/eosio/chain/authority.hpp
```diff
@@ -221,19 +220,22 @@ struct authority {
...
+      threshold(auth.threshold),
+      keys(auth.keys),
+      accounts(auth.accounts),
+      waits(auth.waits)
+   {
+   }
+
+   shared_authority(authority&& auth) :
```
heifner, line 233: `explicit`

libraries/chain/include/eosio/chain/authority.hpp
```diff
@@ -116,20 +116,19 @@ struct shared_key_weight {
       return key_weight{key.to_public_key(), weight};
    }
 
-   static shared_key_weight convert(chainbase::allocator<char> allocator, const key_weight& k) {
-      return std::visit(overloaded {
-         [&](const auto& k1r1) {
-            return shared_key_weight(k1r1, k.weight);
+   shared_key_weight(const key_weight& k) : key(fc::ecc::public_key_shim()), weight(k.weight) {
```
heifner, line 119: `explicit`

libraries/chain/include/eosio/chain/producer_schedule.hpp
```diff
@@ -78,11 +82,12 @@ namespace eosio { namespace chain {
...
 
+   struct producer_authority_schedule;
+
    struct shared_producer_authority_schedule {
-      shared_producer_authority_schedule() = delete;
+      shared_producer_authority_schedule() = default;
 
-      explicit shared_producer_authority_schedule( chainbase::allocator<char> alloc )
-      :producers(alloc){}
+      shared_producer_authority_schedule( const producer_authority_schedule& );
```
heifner, line 90: `explicit`

libraries/chain/include/eosio/chain/database_utils.hpp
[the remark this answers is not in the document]
greg7mdp: I did!

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -99,25 +99,19 @@ namespace eosio::chain {
...
-
+   public:
+      shared_blob() = default;
 
-         shared_blob& operator=(const shared_blob& s) = default;
+      shared_blob(shared_blob&&) = default;
+      shared_blob(const shared_blob& s) = default;
 
-         shared_blob& operator=(shared_blob&& ) = default;
+      shared_blob(std::string_view s) : shared_string(s) {}
```
heifner, line 108: `explicit`

heifner (changes requested):

libraries/chain/include/eosio/chain/types.hpp
greg7mdp: Yes, good catch, thanks. There was also the issue that pack/unpack were not defined for shared?string/shared_vector. I just fixed this.

libraries/chain/include/eosio/chain/types.hpp
greg7mdp: Yes, good catch, thanks. There was also the issue that pack/unpack were not defined for shared?string/shared_vector. I just fixed this.

libraries/chain/include/eosio/chain/authority.hpp
greg7mdp: This makes some assignments a little more verbose, but OK.

libraries/chain/include/eosio/chain/authority.hpp
greg7mdp: This makes some assignments a little more verbose, but OK.

libraries/chain/include/eosio/chain/types.hpp
greg7mdp: done. I used placement new for the construction to make sure there isn't a copy.

libraries/chain/include/eosio/chain/producer_schedule.hpp
greg7mdp: Leaving as-is because the type of the constructor parameter leave little chance for unexpected conversions.

libraries/chain/include/eosio/chain/authority.hpp
greg7mdp: Leaving as-is because the type of the constructor parameter leave little chance for unexpected conversions.

libraries/chain/include/eosio/chain/authority.hpp
greg7mdp: Leaving as-is because the type of the constructor parameter leave little chance for unexpected conversions.

libraries/chain/include/eosio/chain/producer_schedule.hpp
greg7mdp: Thanks, made the constructor at line 52 explicit. Other one is very specific.

heifner: `explicit` on a default constructor has no affect, right?

greg7mdp: You are right, what was I thinking! Will remove the explicit.

libraries/chain/include/eosio/chain/protocol_state_object.hpp
```diff
@@ -16,9 +16,14 @@ namespace eosio { namespace chain {
     */
    class protocol_state_object : public chainbase::object<protocol_state_object_type, protocol_state_object>
    {
-      OBJECT_CTOR(protocol_state_object, (activated_protocol_features)(preactivated_protocol_features)(whitelisted_intrinsics))
-
    public:
+      template<typename Constructor>
+      protocol_state_object(Constructor&& c, chainbase::constructor_tag) :
+         id(0),
+         whitelisted_intrinsics(*activated_protocol_features.get_allocator(this)) {
```
heifner, line 23: I wonder if we should create a `shared_flat_multimap` that derives from `boost::interprocess::flat_multimap` that provides a default constructor that passes in `pinnable_mapped_file::get_allocator<char>(obj)`. Make every `shared_*` type have this functionality like `shared_cow_string` and `shared_cow_vector`.

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -86,10 +84,13 @@ namespace eosio::chain {
 
    using chainbase::allocator;
    using shared_string = chainbase::shared_string;
+   
    template<typename T>
-   using shared_vector = boost::interprocess::vector<T, allocator<T>>;
+   using shared_vector = chainbase::shared_vector<T>;
+   
    template<typename T>
    using shared_set = boost::interprocess::set<T, std::less<T>, allocator<T>>;
```
heifner, line 92: Not used, remove.

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -410,20 +409,78 @@ namespace eosio::chain {
...
+      fc::unsigned_int size;
+      fc::raw::unpack( s, size );
+      FC_ASSERT(size.value <= MAX_SIZE_OF_BYTE_ARRAYS);
+      FC_ASSERT(v.size() == 0);
+      v.clear_and_construct(size.value, 0, [&](auto* dest, std::size_t i) { 
+         new (dest) T(); // unpack expects a constructed variable
+         fc::raw::unpack(s, *dest);
+      });
+   }
+}
```
heifner, line 456: I think all of this should be moved to `database_utils.hpp`.

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -410,20 +409,78 @@ namespace eosio::chain {
...
-      fc::unsigned_int size; fc::raw::unpack( s, size );
-      FC_ASSERT( size.value <= MAX_SIZE_OF_BYTE_ARRAYS );
-      FC_ASSERT( v.size() == 0 );
-      v.resize_and_fill(size.value, [&s](char* buf, std::size_t sz) { s.read(buf, sz); });
+   template<typename Stream, typename T>
+   inline Stream& operator>>(Stream& s, chainbase::shared_cow_vector<T>& v)  {
+      fc::raw::unpack(s, v);
       return s;
    }
 }
```
heifner, line 486: I think all of this should be moved to `database_utils.hpp`

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -99,25 +100,23 @@ namespace eosio::chain {
...
-         shared_blob(const allocator_type& a)
-         :shared_string(a)
-         {}
+      shared_blob& operator=(const shared_blob& s) = default;
+      shared_blob& operator=(shared_blob&& ) = default;
+      shared_blob& operator=(std::string_view sv) {
+         static_cast<shared_string&>(*this) = sv;
+         return *this;
+      }
    };
```
heifner, line 120: Seems like `shared_blob` should be move to `database_utils.hpp`. Maybe all the `shared_*` types in here should be moved to `database_utils.hpp`

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -410,20 +409,78 @@ namespace eosio::chain {
...
+   template<typename Stream>
+   inline void pack(Stream& s, const chainbase::shared_string& v) {		
+      FC_ASSERT(v.size() <= MAX_SIZE_OF_BYTE_ARRAYS);		
+      fc::raw::pack(s, fc::unsigned_int((uint32_t)v.size()));		
+      if (v.size())
+         s.write(v.data(), v.size());		
+   }		
+
+   template<typename Stream>
+   inline void unpack(Stream& s, chainbase::shared_string& v) {
```
heifner, line 425: I realize that unpack of `std::string` does not check `MAX_SIZE_OF_BYTE_ARRAYS` but I wonder if that is a mistake. Also why are we enforcing these limits at all?

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -410,20 +409,78 @@ namespace eosio::chain {
...
+            s.read(buf, sz);            
+         });
+      }
+   }
+   
+   // pack/unpack chainbase::shared_cow_vector
+   // ----------------------------------------
+   template<typename Stream, typename T>
+   inline void pack(Stream& s, const chainbase::shared_cow_vector<T>& v)  {
+      FC_ASSERT(v.size() <= MAX_SIZE_OF_BYTE_ARRAYS);
```
heifner, line 439: The `MAX_SIZE_OF_BYTE_ARRAYS` is for bytes where used elsewhere. This should be specialized for `char` type if we want to enforce this.

libraries/chain/include/eosio/chain/types.hpp
```diff
@@ -410,20 +409,78 @@ namespace eosio::chain {
...
+      fc::raw::pack( s, fc::unsigned_int((uint32_t)v.size()));
+      for (const auto& el : v)
+         fc::raw::pack(s, el);
+   }
+
+   template<typename Stream, typename T>
+   inline void unpack(Stream& s, chainbase::shared_cow_vector<T>& v)  {
```
heifner, line 446: The `MAX_SIZE_OF_BYTE_ARRAYS` is for bytes where used elsewhere. This should be specialized for `char` type if we want to enforce this.

plugins/chain_plugin/test/test_chainbase_types.cpp
```diff
@@ -0,0 +1,123 @@
...
+      }
+   }
+
+   temp_directory& operator=(const temp_directory&) = delete;
+   temp_directory& operator=(temp_directory&& other) noexcept {
+      tmp_path.swap(other.tmp_path);
+      return *this;
+   }
+   const std::filesystem::path& path() const { return tmp_path; }
+};
```
heifner, line 37: Why not use `fc::temp_directory`?

plugins/chain_plugin/test/CMakeLists.txt
```diff
@@ -2,6 +2,7 @@ add_executable( test_chain_plugin
         test_account_query_db.cpp
         test_trx_retry_db.cpp
         test_trx_finality_status_processing.cpp
+        test_chainbase_types.cpp
```
heifner, line 5: I think this test should be in `unittests` not in `chain_plugin` tests.

libraries/chain/controller.cpp
```diff
@@ -1793,11 +1793,12 @@ struct controller_impl {
...
 
-               ps.activated_protocol_features.reserve( ps.activated_protocol_features.size()
-                                                         + new_protocol_feature_activations.size() );
-               for( const auto& feature_digest : new_protocol_feature_activations ) {
-                  ps.activated_protocol_features.emplace_back( feature_digest, pbhs.block_num );
-               }
+               std::size_t current_size = ps.activated_protocol_features.size();
+               std::size_t new_size     = current_size + new_protocol_feature_activations.size();
+               
+               ps.activated_protocol_features.clear_and_construct(new_size, current_size, [&](auto* dest, std::size_t idx) {
```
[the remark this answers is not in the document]
greg7mdp: I added `emplace_back` a while back, but I can't have an `append` of multiple items, because we need each value to be constructed in-place so the correct allocator is used.

libraries/chain/include/eosio/chain/protocol_state_object.hpp
greg7mdp: I'm not understanding the suggestion?

libraries/chain/controller.cpp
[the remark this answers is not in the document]
heifner: I was thinking the append could take a `const std::vector<T>&` and do what you have here.

libraries/chain/include/eosio/chain/types.hpp
greg7mdp: done.

libraries/chain/include/eosio/chain/protocol_state_object.hpp
heifner: What you have here is fine. I was thinking alternatively we could have:

```
template <typename K, typename V>
class shared_flat_multimap : public boost::interprocess::flat_multimap< K, V, std::less<K>, allocator< std::pair<K,V>>> {
public:
   shared_flat_multimap() : boost::interprocess::flat_multimap( pinnable_mapped_file::get_allocator<char>(this) ) {}
...
```

libraries/chain/include/eosio/chain/types.hpp
greg7mdp: done

libraries/chain/include/eosio/chain/types.hpp
greg7mdp: done

libraries/chain/include/eosio/chain/types.hpp
greg7mdp: I don't mind moving them, but I think because these are type definitions, them being in `types.hpp` makes sense.

heifner: I think having the typedefs in `types.hpp` is fine.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: I'd have to construct a vector of `activated_protocol_feature` just to pass it to `append`, now we construct the `activated_protocol_feature` in-place. 
I could also `emplace_back` the new protocol features. I assume that in most cases there is only one, right?

libraries/chain/include/eosio/chain/protocol_state_object.hpp
greg7mdp: As an alternative to what? `protocol_state_object`?

plugins/chain_plugin/test/test_chainbase_types.cpp
greg7mdp: Will change, thanks!

libraries/chain/include/eosio/chain/protocol_state_object.hpp
heifner: As an alternative to: 
```
   template<typename K, typename V>
   using shared_flat_multimap = boost::interprocess::flat_multimap< K, V, std::less<K>, allocator< std::pair<K,V> > >;
```

libraries/chain/controller.cpp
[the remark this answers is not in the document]
heifner: I was missing that `new_protocol_feature_activations` is not the same type as `preactivated_protocol_features`; was missing the `pbhs.block_num`. 

I do think a for loop of `emplace` would be much cleaner. Would be nice if `clear_and_construct` was private.

plugins/chain_plugin/test/test_chainbase_types.cpp
greg7mdp: done.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: I'll switch to a ` for loop of emplace`, but that would be very inefficient for many items.

libraries/chain/include/eosio/chain/protocol_state_object.hpp
greg7mdp: Oh, thanks, I finally understand your suggestion. Let me give that a try!

greg7mdp: I can't do that unfortunately. `pinnable_mapped_file::get_allocator<char>(this)` returns a `std::optional`, and when empty it means we should use the heap (std::allocator()). But `boost::interprocess::flat_multimap` has no easy way to support two different allocators. 

I guess I could do this by creating a `proxy` allocator which would use one or the other. I'm happy to do that, it may be overkill for just this one instance of using shared_flat_multimap, but if we ever used more it would be cleaner.

heifner: I think it is fine as is.

plugins/chain_plugin/test/CMakeLists.txt
greg7mdp: moved.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: done.

libraries/chain/include/eosio/chain/types.hpp
greg7mdp: Is [this](https://github.com/AntelopeIO/leap/pull/1851/commits/4b011f3aa088fcd8e4e48bfa704c3a38f8b2a0e4) what you meant?

libraries/chain/include/eosio/chain/types.hpp
greg7mdp: Is [this](https://github.com/AntelopeIO/leap/pull/1851/commits/4b011f3aa088fcd8e4e48bfa704c3a38f8b2a0e4) what you meant?

libraries/chain/include/eosio/chain/types.hpp
heifner: No, I mean you need this impl: https://github.com/AntelopeIO/leap/blob/dd4217f111619f758a13e5d06a95a84f179058f8/libraries/libfc/include/fc/container/flat.hpp#L13-L13

And the specializations for `char`

greg7mdp: Is this really needed? As far as I can tell, these are just optimizations, and we don't serialize any shared_vector of fundamental types.

greg7mdp: But probably I should use `MAX_NUM_ARRAY_ELEMENTS` instead of `MAX_SIZE_OF_BYTE_ARRAYS` for `shared_vector`.

greg7mdp: I have updated the serialization... please let me know what you think.

heifner (approved):

libraries/chain/include/eosio/chain/database_utils.hpp
```diff
@@ -212,6 +267,34 @@ namespace chainbase {
...
+   inline Stream& operator<<(Stream& s, const chainbase::shared_cow_vector<T>& v)  {
+      fc::raw::pack(s, v);
+      return s;
+   }
+
+   template<typename Stream, typename T>
+   inline Stream& operator>>(Stream& s, chainbase::shared_cow_vector<T>& v)  {
+      fc::raw::unpack(s, v);
+      return s;
+   }
```
spoonincode, line 297: `fc:raw::un/pack()` on a non-reflected class will call `operator<<`/`>>`,
https://github.com/AntelopeIO/leap/blob/f4b7ace8f753a4e2795e342178aed6c0d3114667/libraries/libfc/include/fc/io/raw.hpp#L660-L663
https://github.com/AntelopeIO/leap/blob/f4b7ace8f753a4e2795e342178aed6c0d3114667/libraries/libfc/include/fc/io/raw.hpp#L414-L419
https://github.com/AntelopeIO/leap/blob/f4b7ace8f753a4e2795e342178aed6c0d3114667/libraries/libfc/include/fc/io/raw.hpp#L369-L372

I wonder if we don't need both the un/pack() overrides above and operator<</>> (i.e. the implementations of un/pack() can be moved in to operator<</>>)

libraries/chain/include/eosio/chain/database_utils.hpp
```diff
@@ -195,7 +183,74 @@ namespace fc {
...
+      fc::raw::pack(s, fc::unsigned_int((uint32_t)v.size()));		
+      if (v.size())
+         s.write(v.data(), v.size());		
+   }		
+
+   template<typename Stream>
+   inline void unpack(Stream& s, chainbase::shared_string& v) {
+      fc::unsigned_int sz;
+      fc::raw::unpack(s, sz);
+      FC_ASSERT(sz.value <= MAX_SIZE_OF_BYTE_ARRAYS);
```
spoonincode, line 205: This is more strict than what existed previously, but so far I can't see any reason it'd be a problem..

libraries/chain/include/eosio/chain/database_utils.hpp
greg7mdp: Updated.

spoonincode (approved):

PR #1859, opened 2023-11-02
Remove thread hop to producer thread for trx signature recovery

heifner:
Remove the use of a `std::future` and `wait` in producer thread pool. Now after signature recovery is complete the chain thread posts directly to the main application thread. This improves the eosio token transfer per second (TPS) tests from ~31K TPS to ~36K TPS.

Transaction now transitions from net-thread => chain-thread => main-thread.

Removing the thread hop to chain-thread and performing key recovery on the net thread is possible (net-thread => main-thread), but testing shows it easily overwhelms the main-thread. Instead I think we need to create a memory pool of trxs that the main thread can pull from instead of pushing low priority tasks for trx execution to the main thread. That would greatly reduce the contention on the app post call and on the main thread priority queue. See https://github.com/AntelopeIO/leap/issues/1860

This builds off the improvement of #1846.

Removes `--producer-threads` option as there is now no use for the producer thread pool.

Issue #1690

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -397,13 +397,12 @@ public:
                                   int64_t                                     sub_bill,
                                   uint32_t                                    prev_billed_cpu_time_us);
 
-   void        log_trx_results(const transaction_metadata_ptr& trx, const transaction_trace_ptr& trace, const fc::time_point& start);
+   void        log_trx_results(const transaction_metadata_ptr& trx, const transaction_trace_ptr& trace);
    void        log_trx_results(const transaction_metadata_ptr& trx, const fc::exception_ptr& except_ptr);
    void        log_trx_results(const packed_transaction_ptr& trx,
                                const transaction_trace_ptr&  trace,
                                const fc::exception_ptr&      except_ptr,
                                uint32_t                      billed_cpu_us,
-                               const fc::time_point&         start,
                                bool                          is_transient);
    
    void        add_greylist_accounts(const producer_plugin::greylist_params& params) {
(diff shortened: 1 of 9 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -825,38 +815,57 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
+                                                           if (this->in_producing_mode()) {
+                                                              this->schedule_maybe_produce_block(true);
+                                                           } else {
+                                                              this->restart_speculative_block();
+                                                           }
+                                                        }
+                                                     }
+                                                     CATCH_AND_CALL(exception_handler);
+                                                  });
+                         } );
```
greg7mdp, line 868: I think it would be better to remove all occurences of this-> in this code.

greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
heifner: Removed

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -825,38 +815,57 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
-                                          } else {
-                                             self->restart_speculative_block();
-                                          }
-                                       }
-                                    }
-                                    CATCH_AND_CALL(exception_handler);
-                                 });
-                           }
-                        });
+      boost::asio::post( chain.get_thread_pool(),
```
linh2931, line 818: This section of code is deep nested and becomes complicated. Can more comments be added and/or it be refactored a bit?

heifner: Cleaned it up some.

greg7mdp (approved):

linh2931 (approved):

PR #1863, opened 2023-11-06
[5.0] PH: Post on resolver strand to ensure execution on strand

### LINKED ISSUE #1858: `performance_test_basic_read_only_trxs` failure
https://github.com/AntelopeIO/leap/actions/runs/6737103548/job/18314197007

heifner:
Unable to get anything meaningful from the core dumps of the failed test of #1858. Before this change the `async_resolve` would run on the main thread even though the resolver was given a different thread to execute on. This change makes sure the resolver is only touched by the resolver thread.

Resolves #1858

### DIFF
--- a/tests/trx_generator/http_client_async.hpp
+++ b/tests/trx_generator/http_client_async.hpp
@@ -80,8 +80,10 @@ class session : public std::enable_shared_from_this<session> {
       req_.prepare_payload();
 
       // Look up the domain name
-      resolver_.async_resolve(
-          host, std::to_string(port), [self = this->shared_from_this()](beast::error_code ec, auto res) { self->on_resolve(ec, res); });
+      boost::asio::post( resolver_.get_executor(), [host, port, self=this->shared_from_this()] {
+         self->resolver_.async_resolve(
+                 host, std::to_string(port), [self](beast::error_code ec, auto res) { self->on_resolve(ec, res); });
+      });
    }
 
    void on_resolve(beast::error_code ec, tcp::resolver::results_type results) {


greg7mdp:
I don't see the benefit of using a different strand for `resolver_` and `stream_`?

heifner:
> I don't see the benefit of using a different strand for `resolver_` and `stream_`?

No benefit I see either. I'll fix it.

tests/trx_generator/http_client_async.hpp
```diff
@@ -80,8 +80,10 @@ class session : public std::enable_shared_from_this<session> {
       req_.prepare_payload();
 
       // Look up the domain name
-      resolver_.async_resolve(
-          host, std::to_string(port), [self = this->shared_from_this()](beast::error_code ec, auto res) { self->on_resolve(ec, res); });
+      boost::asio::post( resolver_.get_executor(), [host, port, self=this->shared_from_this()] {
+         self->resolver_.async_resolve(
+                 host, std::to_string(port), [self](beast::error_code ec, auto res) { self->on_resolve(ec, res); });
+      });
```
greg7mdp, line 86: Does this change make a difference? I would have thought that, since we provide an executor when building the resolver, the async functions would already be posted on this executor?

heifner: This code follows the example of `libs/beast/example/http/client/async/http_client_async.cpp` however I think it would be better to follow https://www.boost.org/doc/libs/1_83_0/libs/beast/example/http/client/crawl/http_crawl.cpp as that is a multi-threaded example. I'll close this PR and create a new one with those changes.

PR #1866, opened 2023-11-06
[5.0] PH: Improve error handling and use one strand

### LINKED ISSUE #1858: `performance_test_basic_read_only_trxs` failure

heifner:
Update async http client to use one strand like the multi-threaded boost example.
Improve error reporting so an un-caught exception is not generated.

The core file of #1858 did not reveal anything useful. Do a bit of cleanup and hope for the best.

Maybe Resolves #1858

### DIFF
--- a/tests/trx_generator/http_client_async.hpp
+++ b/tests/trx_generator/http_client_async.hpp
@@ -4,6 +4,9 @@
 // libs/beast/example/http/client/async/http_client_async.cpp
 // with minimum modification and yet reusable.
 //
+// Updated to use strand like in the boost example
+// libs/beast/example/http/client/crawl/http_crawl.cpp
+//
 //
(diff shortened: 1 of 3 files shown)


greg7mdp:
Overall lgtm. I'm wondering about the `sleep(1)` calls when disconnecting. it seems that it would prevent the threads from doing anything else, but maybe it is not an issue for PH?

heifner:
> Overall lgtm. I'm wondering about the `sleep(1)` calls when disconnecting. it seems that it would prevent the threads from doing anything else, but maybe it is not an issue for PH?

Yeah, the trx generator is not doing anything at that point except for waiting on async operations to finish.

linh2931 (approved):

Issue #1854, opened 2023-11-02
Benchmark for leap BLS Host Functions


bhazzard:
```[tasklist]
### Tasks
- [x] Add benchmarking of BLS host functions into Leap Benchmark suites
- [x] Adapt BLS benchmarking to cover new BLS host functions
```

linh2931:
Implemented benchmarking to compare Jacobian and Affine forms.

Jacobian, on commit `9c42d4248e4dccae0f2edc6f9214a62794afa0f2`

```
function                                runs        average       minimum       maximum

bls:
bls_g1_add                              5,000       3,750 ns       3,572 ns      12,580 ns
bls_g1_add_mont                         5,000       3,420 ns       3,253 ns      12,142 ns
bls_g2_add                              5,000      10,512 ns       9,811 ns      21,856 ns
bls_g2_add_mont                         5,000       9,707 ns       9,229 ns      21,503 ns
bls_pairing 1 pair                      5,000   1,789,295 ns   1,680,824 ns   7,485,899 ns
bls_pairing_mont 1 pair                 5,000   1,731,573 ns   1,684,301 ns   3,213,257 ns
bls_pairing 3 pairs                     5,000   2,724,449 ns   2,657,346 ns   4,076,564 ns
bls_pairing_mont 3 pairs                5,000   2,725,833 ns   2,655,432 ns   6,151,175 ns
bls_g1_weighted_sum 1 point             5,000     177,830 ns     175,772 ns     204,178 ns
bls_g1_weighted_sum_mont 1 point        5,000     183,024 ns     180,422 ns   1,554,149 ns
bls_g1_weighted_sum 3 points            5,000     615,286 ns     605,572 ns   1,985,376 ns
bls_g1_weighted_sum_mont 3 points       5,000     595,925 ns     586,211 ns   1,955,829 ns
bls_g1_weighted_sum 5 points            5,000     724,105 ns     714,152 ns   2,096,116 ns
bls_g1_weighted_sum_mont 5 points       5,000     728,274 ns     718,108 ns   2,090,126 ns
bls_g2_weighted_sum 1 point             5,000     735,463 ns     712,924 ns   2,095,999 ns
bls_g2_weighted_sum_mont 1 point        5,000     731,287 ns     708,690 ns   2,084,301 ns
bls_g2_weighted_sum 3 points            5,000   2,295,894 ns   2,237,104 ns   3,633,355 ns
[...]

PR #1884, opened 2023-11-09
[5.0] Benchmark BLS host functions

linh2931:
Add benchmarking of BLS host functions into `Leap` Benchmark suites. Those host functions are directly benchmarked without using CDT wrappers.

Use `benchmark/benchmark -f bls` to benchmark. A sample result looks like

```
function                    runs        average       minimum       maximum

bls:
bls_g1_add                  1,000       2,736 ns       2,607 ns      10,048 ns
bls_g2_add                  1,000       6,766 ns       6,461 ns      13,983 ns
bls_g1_mul                  1,000     193,676 ns     191,533 ns     204,904 ns
bls_g2_mul                  1,000     486,744 ns     480,722 ns     516,755 ns
bls_pairing 1 pair          1,000   1,284,147 ns   1,263,355 ns   1,521,806 ns
bls_pairing 3 pairs         1,000   2,095,239 ns   2,060,983 ns   2,448,659 ns
bls_g1_exp 1 point          1,000     491,239 ns     485,616 ns     626,591 ns
bls_g1_exp 3 points         1,000     698,879 ns     692,786 ns     741,049 ns
bls_g2_exp 1 point          1,000   1,239,919 ns   1,229,475 ns   1,559,680 ns
bls_g2_exp 3 points         1,000   1,841,482 ns   1,821,692 ns   2,167,092 ns
bls_g1_map                  1,000     297,041 ns     291,710 ns     310,379 ns
bls_g2_map                  1,000     380,135 ns     374,088 ns     438,751 ns
bls_fp_mod                  1,000         764 ns         668 ns       7,571 ns
```

Resolves https://github.com/AntelopeIO/leap/issues/1854

### DIFF
--- a/benchmark/CMakeLists.txt
+++ b/benchmark/CMakeLists.txt
@@ -1,7 +1,8 @@
 file(GLOB BENCHMARK "*.cpp")
 add_executable( benchmark ${BENCHMARK} )
 
-target_link_libraries( benchmark fc Boost::program_options bn256)
+target_link_libraries( benchmark eosio_testing fc Boost::program_options bn256)
 target_include_directories( benchmark PUBLIC
                             "${CMAKE_CURRENT_SOURCE_DIR}"
+                            "${CMAKE_CURRENT_BINARY_DIR}/../unittests/include"
                           )
--- a/benchmark/bls.cpp
+++ b/benchmark/bls.cpp
@@ -1,14 +1,8 @@
 #include <benchmark.hpp>
-#include <eosio/testing/tester.hpp>
-#include <eosio/chain/abi_serializer.hpp>
-#include <eosio/chain/platform_timer.hpp>
-#include <eosio/chain/controller.hpp>
 #include <eosio/chain/apply_context.hpp>
 #include <eosio/chain/webassembly/interface.hpp>
-#include <fc/variant_object.hpp>
-#include <fc/io/json.hpp>
+#include <eosio/testing/tester.hpp>
 #include <test_contracts.hpp>
-#include <fc/log/logger.hpp>
 #include <bls12-381/bls12-381.hpp>
 #include <random>
 
@@ -17,14 +11,18 @@ using namespace eosio::chain;
 using namespace eosio::testing;
 using namespace bls12_381;
 
+// benchmark BLS host functions directly without relying on CDT wrappers.
+// to run a benchmarking session, in the build directory, type
+// benchmark/benchmark -f bls
+
 namespace eosio::chain {
 // placed in eosio::chain name space so that interface_in_benchmark
 // can access transaction_context's private member max_transaction_time_subjective.
 // interface_in_benchmark is a friend of transaction_context.
 
-// To benchmark host functions directly without going through CDT
-// wrappers, we need to contruct a eosio::chain::webassembly::interface
-// object, as host functions are implemented in eosio::chain::webassembly::interface
+// To benchmark host functions directly without CDT wrappers,
+// we need to contruct a eosio::chain::webassembly::interface object,
+// as host functions are implemented in eosio::chain::webassembly::interface
 struct interface_in_benchmark {
    interface_in_benchmark() {
       // prevent logging from messing up benchmark results display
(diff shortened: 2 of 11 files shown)


benchmark/bls.cpp
```diff
@@ -0,0 +1,417 @@
...
+
+      // build apply context from the control and transaction context
+      apply_ctx = std::make_unique<apply_context>(*control, *trx_ctx, 1);
+
+      // finally build the interface
+      interface = std::make_unique<webassembly::interface>(*apply_ctx);
+   }
+
+   std::unique_ptr<tester>                      chain;
+   controller*                                  control;
```
heifner, line 90: For the little use of `control` seems like `chain->control` would be better than having this member.

benchmark/bls.cpp
```diff
@@ -0,0 +1,417 @@
...
+using namespace bls12_381;
+
+// benchmark BLS host functions directly without relying on CDT wrappers.
+// to run a benchmarking session, in the build directory, type
+// benchmark/benchmark -f bls
+
+namespace eosio::chain {
+// placed in eosio::chain name space so that interface_in_benchmark
+// can access transaction_context's private member max_transaction_time_subjective.
+// interface_in_benchmark is a friend of transaction_context.
```
heifner, line 21: Seems like you could make `eosio::test::interface_in_benchmark` a friend. Just need to forward declare.
Or `eosio::benchmark` either one.

```
namespace eosio::test {
   struct interface_in_benchmark; // for benchmark testing
}

friend struct test::interface_in_benchmark;
```

benchmark/bls.cpp
```diff
@@ -0,0 +1,417 @@
...
+      // construct a packed transaction
+      ptrx = std::make_unique<packed_transaction>(*trx, eosio::chain::packed_transaction::compression_type::zlib);
+
+      // build transaction context from the packed transaction
+      timer = std::make_unique<platform_timer>();
+      trx_timer = std::make_unique<transaction_checktime_timer>(*timer);
+      trx_ctx = std::make_unique<transaction_context>(*control, *ptrx, ptrx->id(), std::move(*trx_timer));
+      trx_ctx->max_transaction_time_subjective = fc::microseconds::maximum();
+      trx_ctx->init_for_input_trx( ptrx->get_unprunable_size(), ptrx->get_prunable_size() );
+      trx_ctx->exec();
```
heifner, line 80: I don't think the `exec()` is needed, and probably cleaner not to call it. All you want is a setup apply context.

benchmark/bls.cpp
```diff
@@ -0,0 +1,417 @@
...
+   std::unique_ptr<apply_context>               apply_ctx;
+   std::unique_ptr<webassembly::interface>      interface;
+};
+} // namespace eosio::chain
+
+
+namespace benchmark {
+
+// utilility to create a random scalar
+std::array<uint64_t, 4> random_scalar()
```
heifner, line 105: I wonder if this should use `bls12_381::secret_key`. Are the new enforcement of valid data going to be tripped by this?

benchmark/bls.cpp
```diff
@@ -0,0 +1,417 @@
...
+      fc::logger::get(DEFAULT_LOGGER).set_log_level(fc::log_level::off);
+
+      // create a chain
+      fc::temp_directory tempdir;
+      auto conf_genesis = tester::default_config( tempdir );
+      auto& cfg = conf_genesis.second.initial_configuration;
+      // configure large cpu usgaes so expensive BLS functions like pairing
+      // can run a reasonable number of times within a trasaction time
+      cfg.max_block_cpu_usage        = 500'000;
+      cfg.max_transaction_cpu_usage  = 480'000;
```
heifner, line 38: Seems like these could be set much much larger so you don't have to limit benchmarking to 480ms.

benchmark/bls.cpp
```diff
@@ -0,0 +1,417 @@
...
+      dis(gen) % bls12_381::fp::Q[0],
+      dis(gen) % bls12_381::fp::Q[1],
+      dis(gen) % bls12_381::fp::Q[2],
+      dis(gen) % bls12_381::fp::Q[3]
+   };
+}
+
+// some functions like pairing/exp are expensive. set a hard limit to
+// the number of runs such that the transanction does not reach deadline
+constexpr uint32_t num_runs_limit = 200;
```
heifner, line 121: Is this needed if you just set the block/trx times much larger?

benchmark/bls.cpp
```diff
@@ -0,0 +1,417 @@
...
+{
+   std::array<uint64_t, 4> k = random_scalar();
+   return bls12_381::g1::one().mulScalar(k);
+}
+
+// utilility to create a random g2
+bls12_381::g2 random_g2()
+{
+   std::array<uint64_t, 4> k = random_scalar();
+   return bls12_381::g2::one().mulScalar(k);
```
heifner, line 134: Does using `one()` provide a good distribution here? Would it be better to have random for both? Same question for g1.

heifner (changes requested):

benchmark/bls.cpp
linh2931: `exec()` is needed to generate action traces. `apply_context`'s constructor calls `transaction_context::get_action_trace()`. Its assert
```
EOS_ASSERT( 0 < action_ordinal && action_ordinal <= trace->action_traces.size()
```
will fail if `exec()` is not called.

benchmark/bls.cpp
linh2931: Thanks. This make BLS benchmark code belongs to a single `eosio::benchmark` name space.

benchmark/bls.cpp
linh2931: I was thinking we should be close to real block and transaction cpu limits. We don't need that. After I changed to very large values, all expensive functions can be benchmarked without any issues. Thanks!

benchmark/bls.cpp
linh2931: I copied this code from the BLS library's unittest.cpp. As `k` is randomized, it might be random enough for benchmark/tests.

benchmark/bls.cpp
linh2931: I think the way `g1` and `g2` are generated this benchmarking (multiple by `g1::one()` or `g2::one()`) makes the resulting `g1` and `g2` are on the curve. Current `bls_pairing`, `bls_g1_map`, and `bls_g2_map` already verify input, and I tried g1/g2 add with validation enabled. They all work.

heifner (approved):

benchmark/CMakeLists.txt
```diff
@@ -1,7 +1,8 @@
 file(GLOB BENCHMARK "*.cpp")
 add_executable( benchmark ${BENCHMARK} )
 
-target_link_libraries( benchmark fc Boost::program_options bn256)
+target_link_libraries( benchmark eosio_testing fc Boost::program_options bn256)
 target_include_directories( benchmark PUBLIC
                             "${CMAKE_CURRENT_SOURCE_DIR}"
+                            "${CMAKE_CURRENT_BINARY_DIR}/../unittests/include"
```
spoonincode, line 7: We don't need to fix this now (or even when merging to main) but we should try to avoid these sort of declarations. We probably want a target similar to `eosio_testing_contracts`,
https://github.com/AntelopeIO/leap/blob/581cd2b605ea0f8c2e21493a2647d02e31047dfc/libraries/testing/CMakeLists.txt#L39-L41

arhag (approved):

benchmark/CMakeLists.txt
linh2931: Thanks @spoonincode. I was doing the cleos set code issue yesterday and not getting around with this. As the PR has already merged into 5.0, for cleaner tracking, I will merge the PR to main as is, and use your suggested way later.

PR #1887, opened 2023-11-10
[5.0] Add additional tests for mapped_private

heifner:
Add `mapped_private` to `db_modes_test.sh`.

### DIFF
--- a/tests/db_modes_test.sh
+++ b/tests/db_modes_test.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
-# This test is intended to verify that switching between DB modes "just works". Addtionally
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

spoonincode (changes requested):
`mapped_private` only works on some platforms so need to skip it on unsupported platforms

greg7mdp:
> `mapped_private` only works on some platforms so need to skip it on unsupported platforms

There are two features "mapped_private" depends on, the `soft-dirty` feature, and the `copy_on_write` (i.e. `MAP_PRIVATE`) file mapping option. The `mapped_private` mode would still work if `soft-dirty` is not available, but  `copy_on_write` is.

spoonincode (approved):

PR #1889, opened 2023-11-13
[3.2] Do not require trailing `=` for base64 encoded strings

heifner:
All previous versions of nodeos required a trailing `=` for base64 encoded strings. `fc::variant` appends an extra `=` for all of its base64 encoded strings. However, non-fc base64 encoders do not add an extra `=` when one is not needed. This PR modifies `fc::variant` to support non-fc base64 encoded strings that do not have the extra `=`. 

In 5.0, `fc::variant` was updated to not add the extra `=` to base64 encoded strings. See #1888. This PR allows `nodeos` to support the new 5.0 version of base64 encoded strings.

Issue #1461

### DIFF
--- a/libraries/libfc/src/variant.cpp
+++ b/libraries/libfc/src/variant.cpp
@@ -525,10 +525,14 @@ blob variant::as_blob()const
       {
          const string& str = get_string();
          if( str.size() == 0 ) return blob();
-         if( str.back() == '=' )
-         {
-            std::string b64 = base64_decode( get_string() );
+         try {
+            // variant adds `=` to end of base64 encoded string (see as_string() above) which produces invalid base64
+            // variant in 5.0 no longer appends the '=' character to conform to valid base64 encoding
+            // fc version of base64_decode allows for extra `=` at the end of the string
+            std::string b64 = base64_decode( str );
             return blob( { std::vector<char>( b64.begin(), b64.end() ) } );
+         } catch(const std::exception&) {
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):
So I assume that if there is an extra `=` character at the end, `base64_decode` ignores it, right?

heifner:
> So I assume that if there is an extra `=` character at the end, `base64_decode` ignores it, right?

Yes. And there is an existing test that verifies that: https://github.com/AntelopeIO/leap/blob/bce1b0b19d29140f212fcc75904473a164cf68a9/libraries/libfc/test/test_base64.cpp#L40-L45

libraries/libfc/src/variant.cpp
```diff
@@ -525,10 +525,14 @@ blob variant::as_blob()const
...
-         if( str.back() == '=' )
-         {
-            std::string b64 = base64_decode( get_string() );
+         try {
+            // variant adds `=` to end of base64 encoded string (see as_string() above) which produces invalid base64
+            // variant in 5.0 no longer appends the '=' character to conform to valid base64 encoding
+            // fc version of base64_decode allows for extra `=` at the end of the string
+            std::string b64 = base64_decode( str );
             return blob( { std::vector<char>( b64.begin(), b64.end() ) } );
+         } catch(const std::exception&) {
```
spoonincode, line 534: `fc::assert_exception`?

spoonincode (approved):

PR #1893, opened 2023-11-13
[5.0 -> main] Fix base64 encoding - take 2

### LINKED ISSUE #1461: Invalid Base64 padding returned by APIs

heifner:
#1482 fixed base64 encoding by removing the unneeded extra `=` that variant added to base64 encoded data. However, it also updated the base64 library with a stricter one. This PR leaves the old base64 encoding/decoding library but does remove the extra `=` from the base64 encoded data so non-fc base64 decoders will not claim the data is invalid.

This change still breaks backward compatibility with 3.2 & 4.0 `cleos/nodeos`. PRs to 3.2 (#1889) and 4.0 will be created that will allow them to work with this new base64 encoding (they will no longer expect the invalid `=` character).

PR #1886 reverts #1482. This PR provided an alternative fix for #1461.

Included in this PR is an optimization to the existing `base64_decode` to avoid a copy for our use-cases by returning a `vector<char>` instead of a `string`.
Included in this PR is a change to the `nodeos_run_test.py` to keep the `keosd` running with passed `--leaving-running` which is useful for local testing. This was used to manually test using 3.2 `cleos` with 5.0.

Merges `release/5.0` into `main` including #1888 

Resolves #1461

### DIFF
--- a/libraries/libfc/test/test_base64.cpp
+++ b/libraries/libfc/test/test_base64.cpp
@@ -26,21 +26,27 @@ BOOST_AUTO_TEST_CASE(base64dec) try {
    auto input = "YWJjMTIzJCYoKSc/tPUB+n5h"s;
    auto expected_output = "abc123$&()'?\xb4\xf5\x01\xfa~a"s;
 
-   BOOST_CHECK_EQUAL(expected_output, base64_decode(input));
+   std::vector<char> b64 = base64_decode(input);
+   std::string b64_str(b64.begin(), b64.end());
+   BOOST_CHECK_EQUAL(expected_output, b64_str);
 } FC_LOG_AND_RETHROW();
 
 BOOST_AUTO_TEST_CASE(base64urldec) try {
    auto input = "YWJjMTIzJCYoKSc_tPUB-n5h"s;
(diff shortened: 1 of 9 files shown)


spoonincode (approved):

libraries/libfc/test/test_base64.cpp
```diff
@@ -26,21 +26,27 @@ BOOST_AUTO_TEST_CASE(base64dec) try {
    auto input = "YWJjMTIzJCYoKSc/tPUB+n5h"s;
    auto expected_output = "abc123$&()'?\xb4\xf5\x01\xfa~a"s;
 
-   BOOST_CHECK_EQUAL(expected_output, base64_decode(input));
+   std::vector<char> b64 = base64_decode(input);
+   std::string b64_str(b64.begin(), b64.end());
+   BOOST_CHECK_EQUAL(expected_output, b64_str);
```
greg7mdp, line 31: Again, not a required change, just a suggestion for here and the few next instances of the same pattern.
```suggestion
   BOOST_CHECK_EQUAL(std::string_view(expected_output), std::string_view(b64.begin(), b64.end()));
```

greg7mdp (approved):

greg7mdp (approved):

Issue #1868, opened 2023-11-07
`cleos set code`/`cleos set abi` don't appear to work


spoonincode:
`cleos set code` and `cleos set abi` don't seem to work on main, and I'm suspicious they haven't worked in a long time.

The problem is `contractPath` is empty here when doing `set code` or `set abi` and, at least on main, this causes `std::filesystem::canonical()` to throw
https://github.com/AntelopeIO/leap/blob/161265052b10265da3ad3455606053b3e238268a/programs/cleos/main.cpp#L3474

This looks to be a straightforward fix; adding a test will take longer than the fix itself.

bhazzard:
We need to verify that this was broken prior to 5.0.

linh2931:
I am able to reproduce this problem in `3.2`. Will fix  it `in 3.2` and merge the fix to `4.0`, `5.0`, and `main`.

Will create an integration test in `5.0` and `main`.

PR #1897, opened 2023-11-14
[3.2] Fix non-working cleos set code and set abi commands

linh2931:
Cleos `set code` and `set abi` have not worked since `release/3.2`. The root cause is the common code `set_code_callback` for `set contract` and `set code` assumes `contract-dir` argument which is only applicable to `set contract`; and the common code `set_code_callback` for `set abi` and `set code` assumes `contract-dir` argument too.

This PR fixes the problem.

The fix has been tested manually. CI tests will be added to `release/5.0` to minimize changes to 3.2.

Resolves https://github.com/AntelopeIO/leap/issues/1868.

### DIFF
--- a/programs/cleos/main.cpp
+++ b/programs/cleos/main.cpp
@@ -3429,7 +3429,7 @@ int main( int argc, char** argv ) {
    contractSubcommand->add_flag( "--suppress-duplicate-check", suppress_duplicate_check, localized("Don't check for duplicate"));
 
    std::vector<chain::action> actions;
-   auto set_code_callback = [&]() {
+   auto set_code_callback = [&](bool is_set_contract) {
 
       std::vector<char> old_wasm;
       bool duplicate = false;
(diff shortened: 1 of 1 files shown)


programs/cleos/main.cpp
```diff
@@ -3447,12 +3447,17 @@ int main( int argc, char** argv ) {
...
         std::string wasm;
-        fc::path cpath = fc::canonical(fc::path(contractPath));
 
-        if( wasmPath.empty() ) {
-           wasmPath = (cpath / (cpath.filename().generic_string()+".wasm")).generic_string();
-        } else if ( boost::filesystem::path(wasmPath).is_relative() ) {
-           wasmPath = (cpath / wasmPath).generic_string();
+        // contractPath (set by contract-dir argument) is only applicable
+        // to "set contract" command
+        if(is_set_contract) {
```
spoonincode, line 3453: you might be able to use `!contractPath.empty()` instead of a new variable?

spoonincode (approved):

programs/cleos/main.cpp
linh2931: Thanks. My solution emphasizes the intention of `contractPath` only applicable to `set contract`. Your solution is simpler and covers emptiness check for any conditions. Changed to use yours.

heifner (approved):

PR #1909, opened 2023-11-16
[4.0 -> 5.0] Fix non-working cleos set code and set abi commands, and add tests

linh2931:
Cleos `set code` and `set abi` have not worked since `release/3.2`. The root cause is the common code `set_code_callback` for `set contract` and `set code` assumes `contract-dir` argument which is only applicable to `set contract`; and the common code `set_code_callback` for `set abi` and `set code` assumes `contract-dir` argument too.

Merges fix from `release/4.0` into `release/5.0` including [#1897](https://github.com/AntelopeIO/leap/pull/1897).

In addition to the fix, tests are added in this PR for `release/5.0` and `main`.

Resolves https://github.com/AntelopeIO/leap/issues/1868.


tests/cleos_set_code_test.py
```diff
@@ -0,0 +1,74 @@
...
+
+import random
+from TestHarness import Account, Cluster, TestHelper, Utils, WalletMgr
+
+###############################################################
+# cleos_set_code_test
+#
+# Test Cleos "set code" and "set abi" commands
+#
+###############################################################
```
heifner, line 11: This test is fine, but I think you could have just added it to `nodeos_run_test.py` to save on test run times.

tests/cleos_set_code_test.py
```diff
@@ -0,0 +1,74 @@
...
+
+Print=Utils.Print
+errorExit=Utils.errorExit
+
+args=TestHelper.parse_args({"--seed"
```
heifner, line 16: Why the `seed` argument?

tests/cleos_set_code_test.py
```diff
@@ -0,0 +1,74 @@
...
+                            ,"--dump-error-details","-v","--leave-running"
+                            ,"--keep-logs","--unshared"})
+
+pnodes=1
+topo="mesh"
```
heifner, line 21: The default is `mesh`

heifner (approved):

tests/cleos_set_code_test.py
linh2931: Thanks! I will change to use `nodeos_run_test.py` to save test setup time.

spoonincode (approved):

heifner (approved):

PR #1917, opened 2023-11-17
Fix a couple perf issues in `fc::raw::pack` and add tuple support.

greg7mdp:
- pack parameters were passed by value causing unneeded copies.
- packing a `char*` was unnecessarily allocating a temp `std::string`.
- added `std::tuple` support for `pack`, so that we can provide lists longer than pairs, so we don't have to do [this](https://github.com/AntelopeIO/leap/blob/4b34cf572886dda4641c650d36a142bd2aeea3bc/libraries/chain/include/eosio/chain/hotstuff.hpp#L17-L18).

### DIFF
--- a/libraries/libfc/include/fc/io/raw.hpp
+++ b/libraries/libfc/include/fc/io/raw.hpp
@@ -449,7 +449,12 @@ namespace fc {
 
     template<typename Stream, typename... Ts >
     inline void pack( Stream& s, const std::tuple<Ts...>& tup ) {
-       auto l = [&s](const auto& v) { fc::raw::pack( s, v ); };
+       auto l = [&s](const auto&... v) { fc::raw::pack( s, v... ); };
+       std::apply(l, tup);
+    }
+    template<typename Stream, typename... Ts >
+    inline void unpack( Stream& s, std::tuple<Ts...>& tup ) {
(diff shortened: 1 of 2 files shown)


libraries/libfc/include/fc/io/raw.hpp
```diff
@@ -442,6 +447,11 @@ namespace fc {
       }
     }
 
+    template<typename Stream, typename... Ts >
+    inline void pack( Stream& s, const std::tuple<Ts...>& tup ) {
+       auto l = [&s](const auto& v) { fc::raw::pack( s, v ); };
+       std::apply(l, tup);
+    }
```
heifner, line 454: What about `unpack` ?

heifner (changes requested):

libraries/libfc/include/fc/io/raw.hpp
greg7mdp: Thanks, adding it.

greg7mdp: Done. I also added a test for this.

heifner (approved):

spoonincode (approved):

greg7mdp:
FYI, I just added a test after both of you approved.

heifner (approved):

PR #1918, opened 2023-11-18
Support serialization of fixed-size arrays in abi_serializer.

### LINKED ISSUE #1931: Support serialization of fixed-size arrays in abi_serializer.
It is not clear we want to do this. It would require support in `abios` as well, in addition to `abi_serializer`, and verify that it works with cdt and other tools.

greg7mdp:
[partially] resolves #1931.

- add support for serialization of fixed size array types such as `uint8[5]`.
- [perf] add missing `reserve()` when deserializing a vector (or now an array), fix unnecessary copy in `vector<fc::variant> vars = var.get_array();`
- add tests (thanks to Kevin) for serialization of fixed size array types.

### DIFF
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -400,16 +400,15 @@ namespace eosio { namespace chain {
       auto h = ctx.enter_scope();
       auto rtype = resolve_type(type);
       auto ftype = fundamental_type(rtype);
-      auto btype = built_in_types.find(ftype );
-      bool is_var_array = is_szarray(rtype);
-      if( !is_var_array && btype != built_in_types.end() ) {
+      bool var_is_array = is_szarray(rtype);
+      if( auto btype = built_in_types.find(ftype ); !var_is_array && btype != built_in_types.end() ) {
          try {
             return btype->second.first(stream, is_array(rtype), is_optional(rtype), ctx.get_yield_function());
          } EOS_RETHROW_EXCEPTIONS( unpack_exception, "Unable to unpack ${class} type '${type}' while processing '${p}'",
                                    ("class", is_array(rtype) ? "array of built-in" : is_optional(rtype) ? "optional of built-in" : "built-in")
                                    ("type", impl::limit_size(ftype))("p", ctx.get_path_string()) )
       }
-      if ( is_array(rtype) || is_var_array ) {
+      if ( is_array(rtype) || var_is_array ) {
          ctx.hint_array_type_if_in_array();
          fc::unsigned_int size;
          try {
@@ -497,7 +496,7 @@ namespace eosio { namespace chain {
       auto v_itr = variants.end();
       auto s_itr = structs.end();
 
-      bool var_is_array = var.get_type() == fc::variant::array_type;
+      bool var_is_array = is_szarray(rtype);
       
       if( auto btype = built_in_types.find(fundamental_type(rtype)); !var_is_array && btype != built_in_types.end() ) {
          btype->second.second(var, ds, is_array(rtype), is_optional(rtype), ctx.get_yield_function());
--- a/libraries/chain/abi_serializer.cpp
+++ b/libraries/chain/abi_serializer.cpp
@@ -218,17 +218,23 @@ namespace eosio { namespace chain {
       return ends_with(type, "[]");
    }
 
-   bool abi_serializer::is_szarray(const string_view& type)const {
+   std::optional<fc::unsigned_int> abi_serializer::is_szarray(const string_view& type)const {
       auto pos1 = type.find_last_of('[');
       auto pos2 = type.find_last_of(']');
-      if(pos1 == string_view::npos || pos2 == string_view::npos) return false;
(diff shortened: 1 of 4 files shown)


greg7mdp:
I think we should check that the length of the array passed to serialize matches the size in the `rtype`.

libraries/chain/abi_serializer.cpp
```diff
@@ -400,15 +400,15 @@ namespace eosio { namespace chain {
       auto h = ctx.enter_scope();
       auto rtype = resolve_type(type);
       auto ftype = fundamental_type(rtype);
-      auto btype = built_in_types.find(ftype );
-      if( btype != built_in_types.end() ) {
+      bool var_is_array = is_szarray(rtype);
```
heifner, line 403: I think this can be:
`bool var_is_array = is_array(rtype) || is_szarray(rtype);`
and then just `var_is_array` below.

libraries/chain/abi_serializer.cpp
```diff
@@ -496,10 +496,11 @@ namespace eosio { namespace chain {
       auto v_itr = variants.end();
       auto s_itr = structs.end();
 
-      auto btype = built_in_types.find(fundamental_type(rtype));
-      if( btype != built_in_types.end() ) {
+      bool var_is_array = is_szarray(rtype);
```
heifner, line 499: I think this can be:
`bool var_is_array = is_array(rtype) || is_szarray(rtype);`
and then just `var_is_array` below.

libraries/chain/abi_serializer.cpp
```diff
@@ -400,15 +400,15 @@ namespace eosio { namespace chain {
       auto h = ctx.enter_scope();
       auto rtype = resolve_type(type);
       auto ftype = fundamental_type(rtype);
-      auto btype = built_in_types.find(ftype );
-      if( btype != built_in_types.end() ) {
+      bool var_is_array = is_szarray(rtype);
+      if( auto btype = built_in_types.find(ftype ); !var_is_array && btype != built_in_types.end() ) {
          try {
             return btype->second.first(stream, is_array(rtype), is_optional(rtype), ctx.get_yield_function());
```
heifner, line 406: I think you can also hard-code `false` for `is_array(rtype)`

libraries/chain/abi_serializer.cpp
```diff
@@ -496,10 +496,11 @@ namespace eosio { namespace chain {
       auto v_itr = variants.end();
       auto s_itr = structs.end();
 
-      auto btype = built_in_types.find(fundamental_type(rtype));
-      if( btype != built_in_types.end() ) {
+      bool var_is_array = is_szarray(rtype);
+      
+      if( auto btype = built_in_types.find(fundamental_type(rtype)); !var_is_array && btype != built_in_types.end() ) {
          btype->second.second(var, ds, is_array(rtype), is_optional(rtype), ctx.get_yield_function());
```
heifner, line 502: I think you can also hard-code `false` for `is_array(rtype)`

unittests/abi_tests.cpp
```diff
@@ -530,6 +530,44 @@ fc::variant verify_type_round_trip_conversion( const abi_serializer& abis, const
 }
 )=====";
 
+BOOST_AUTO_TEST_CASE(std_array_types)
+{ try {
+
+   const char* currency_abi = R"=====(
```
heifner, line 536: Rename to `test_abi`

heifner (changes requested):

libraries/chain/abi_serializer.cpp
greg7mdp: I see your point, but maybe we want a different serialization for fixed-size arrays without the size, as Matt's mentioned in telegram. Let le try that.

unittests/abi_tests.cpp
greg7mdp: I don't really mind, but this doesn't seem consistent with the other tests in the same file.

heifner: I mean `currency_abi` => `test_abi`

libraries/chain/abi_serializer.cpp
```diff
@@ -218,17 +218,23 @@ namespace eosio { namespace chain {
       return ends_with(type, "[]");
    }
 
-   bool abi_serializer::is_szarray(const string_view& type)const {
+   std::optional<fc::unsigned_int> abi_serializer::is_szarray(const string_view& type)const {
```
heifner, line 221: I think we need a test that demonstrates unmatched `std::array` size does not work and generates a meaningful error.

heifner (changes requested):

unittests/abi_tests.cpp
greg7mdp: Got it, makes sense, thanks!

libraries/chain/abi_serializer.cpp
greg7mdp: done.

heifner (approved):

greg7mdp:
Replaced by https://github.com/AntelopeIO/spring/pull/1614

Issue #1916, opened 2023-11-17
IF: Add tests for set_finalizers and hotstuff activation


heifner:
Add infrastructure to tester for `set_finalizers` similar to `set_producers`.
Test `set_finalizers` host function.
Test hotstuff activation and changing of producer schedule

arhag:
Related to #1523.

PR #1921, opened 2023-11-20
IF: Update bios contract to have `setfinset` which calls host `set_finalizers`

heifner:
First part of #1916 

- Update bios contract to have a `setfinset` that calls host function `set_finalizers`
- Add support to `tester` for `set_finalizers`
- Add simple test for host function serialization


libraries/testing/tester.cpp
```diff
@@ -1168,6 +1182,31 @@ namespace eosio { namespace testing {
 
    }
 
+   transaction_trace_ptr base_tester::set_finalizers(const vector<account_name>& finalizer_names) {
+      uint64_t fthreshold = finalizer_names.size() * 2 / 3;
```
linh2931, line 1186: Should this be upper bounded in case `2/3` is not divisible?

unittests/api_tests.cpp
```diff
@@ -3855,4 +3856,33 @@ BOOST_AUTO_TEST_CASE(get_code_hash_tests) { try {
...
+   // activate hotstuff
+   t.set_finalizers(producers);
+   auto block = t.produce_block();
+
+   std::optional<block_header_extension> ext = block->extract_header_extension(hs_finalizer_set_extension::extension_id());
+
+   BOOST_TEST(!!ext);
+   BOOST_TEST(std::get<hs_finalizer_set_extension>(*ext).finalizers.size() == producers.size());
+   BOOST_TEST(std::get<hs_finalizer_set_extension>(*ext).generation == 1);
+   BOOST_TEST(std::get<hs_finalizer_set_extension>(*ext).fthreshold == producers.size() / 3 * 2);
```
linh2931, line 3884: Can you check `hs_dpos_irreversible_blocknum` is `uint32_t max()` and `hs_block_confirmed` is `uint16_t max()` after hotstuff activation?

heifner: Good idea.

libraries/testing/tester.cpp
heifner: Doesn't really matter what this is set to. It is just a default value.

unittests/api_tests.cpp
heifner: This points out there is more plumbing needed to the tester as it currently does not have a chain_pacemaker wired up.

libraries/testing/tester.cpp
```diff
@@ -1168,6 +1182,31 @@ namespace eosio { namespace testing {
...
+
+      fc::variants finalizer_auths;
+      for (const auto& n: finalizer_names) {
+         crypto::blslib::bls_public_key pk = get_bls_public_key( n );
+         std::array<uint8_t, 96> a = pk._pkey.toAffineBytesLE();
+         std::vector<char> v(96);
+         memcpy(&v[0], a.data(), v.size());
```
greg7mdp, line 1193:
```suggestion
         std::vector<char> v(96);
         pk._pkey.toAffineBytesLE(std::span<uint8_t, 96>((uint8_t *)v.data(), 96));
```

libraries/testing/tester.cpp
```diff
@@ -1168,6 +1182,31 @@ namespace eosio { namespace testing {
...
+
+         finalizer_auths.emplace_back(
+            fc::mutable_variant_object()
+               ("description", n.to_string() + " description")
+               ("fweight", (uint64_t)1)
+               ("public_key_g1_affine_le", v) );
```
greg7mdp, line 1199:
```suggestion
               ("public_key_g1_affine_le", std::move(v)) );
```

libraries/testing/tester.cpp
```diff
@@ -1168,6 +1182,31 @@ namespace eosio { namespace testing {
...
+      }
+
+
+      fc::mutable_variant_object fin_set_variant;
+      fin_set_variant("fthreshold", fthreshold);
+      fin_set_variant("finalizers", std::move(finalizer_auths));
+
+      return push_action( config::system_account_name, "setfinset"_n, config::system_account_name,
+                          fc::mutable_variant_object()("fin_set", fin_set_variant));
```
greg7mdp, line 1208:
```suggestion
                          fc::mutable_variant_object()("fin_set", std::move(fin_set_variant)));
```

greg7mdp (approved):

unittests/api_tests.cpp
heifner: The `qc_chain`, `chain_pacemaker` needs a bit of refactoring so it can be used in `tester`. How about doing that in a diff PR?

linh2931 (approved):

Issue #1912, opened 2023-11-16
IF: Unification: Modify qc_chain to support core of new algorithm


arhag:
The proposal needs to reference a block it builds off of. This will be be referenced by the `previous` field in the block header. Other essential information to include in the proposal:
* `last_qc_block_height`: The block height of the most recent ancestor block that has a QC.
* `is_last_qc_strong`: This is a boolean that determines if the QC justification for the block referenced by last QC block height is strong or weak.

For now, we will not have all necessary information since the block header will not have a header extension (that is covered by https://github.com/AntelopeIO/leap/issues/1911). So to make progress with this issue, use the following values:
* `last_qc_block_height`: For now, assume this is the height of the previous block. No need yet for validated that you have a QC for it since there is no QC in the block extension (that is covered by a later issue).
* `is_last_qc_strong`: For now, always assume it is strong.

Modify qc_chain to take this new proposal generated from the `signed_block` message. Modify its tracked state to track the core information that we will eventually be tracking as part of a new `block_header_state`. That core information is represented by a `block_header_state_core` data structure which is associated with each block:

* `last_final_block_height`: This is the block height of the last irreversible (final) block.
* `final_on_strong_qc_block_height`: This is the block height of the block that would become irreversible if the associated block header was to achieve a strong QC.
* `last_qc_block_height`: This is the same as the last QC block height extract from the associated block.

Implement a state transition for the `block_header_state_core`. The transition will require as input `last_qc_block_height` and `is_last_qc_strong` and works as follows:
1. If input `last_qc_block_height` is the same as the old `last_qc_block_height` then the new `block_header_state_core` should be the same as the old one, and there is no need to consider the items below.
2. New `last_qc_block_height` is always the input `last_qc_block_height`.
3. If `is_last_qc_strong == true`, then the new `final_on_strong_qc_block_height` is the old `last_qc_block_height` assuming it was present. Otherwise, the new `final_on_strong_qc_block_height` should not be present.
4. If `is_last_qc_strong == true`, then the new `last_final_block_height` is the old `final_on_strong_qc_block_height` assuming it was present. Otherwise, the new `last_final_block_height` is the same as the old `last_final_block_height`.

PR #1923, opened 2023-11-21
IF: implement block_header_state_core and its state transition

linh2931:
Part 1 of https://github.com/AntelopeIO/leap/issues/1912. It implements `block_header_state_core` (to hold hotstuff core information associated with each block) and its state transitions.

Extensive unit tests are added.

Part 2 for qc_chain to track the state will be in a separate PR.

### DIFF
--- a/libraries/chain/block_header_state.cpp
+++ b/libraries/chain/block_header_state.cpp
@@ -35,8 +35,7 @@ namespace eosio { namespace chain {
                                                           bool     is_last_qc_strong) {
       // no state change if last_qc_block_height is the same
       if( last_qc_block_height == this->last_qc_block_height ) {
-         block_header_state_core result{*this};
-         return result;
+         return {*this};
       }
 
       EOS_ASSERT( last_qc_block_height > this->last_qc_block_height, block_validate_exception, "new last_qc_block_height must be greater than old last_qc_block_height" );
--- a/libraries/chain/block_header_state.cpp
+++ b/libraries/chain/block_header_state.cpp
@@ -24,17 +24,19 @@ namespace eosio { namespace chain {
    }
 
    block_header_state_core::block_header_state_core( uint32_t last_final_blk_ht,
-                                       std::optional<uint32_t> final_on_strong_qc_blk_ht,
-                                       std::optional<uint32_t> last_qc_blk_ht ) :
+                                      std::optional<uint32_t> final_on_strong_qc_blk_ht,
+                                      std::optional<uint32_t> last_qc_blk_ht )
+      :
       last_final_block_height(last_final_blk_ht),
       final_on_strong_qc_block_height(final_on_strong_qc_blk_ht),
       last_qc_block_height(last_qc_blk_ht) {}
 
    block_header_state_core block_header_state_core::next( uint32_t last_qc_block_height,
-                                                          bool is_last_qc_strong) {
+                                                          bool     is_last_qc_strong) {
       // no state change if last_qc_block_height is the same
       if( last_qc_block_height == this->last_qc_block_height ) {
-         return std::move(*this);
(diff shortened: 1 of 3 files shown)


heifner (changes requested):

libraries/chain/block_header_state.cpp
```diff
@@ -23,6 +23,52 @@ namespace eosio { namespace chain {
...
+                                       std::optional<uint32_t> last_qc_blk_ht ) :
+      last_final_block_height(last_final_blk_ht),
+      final_on_strong_qc_block_height(final_on_strong_qc_blk_ht),
+      last_qc_block_height(last_qc_blk_ht) {}
+
+   block_header_state_core block_header_state_core::next( uint32_t last_qc_block_height,
+                                                          bool is_last_qc_strong) {
+      // no state change if last_qc_block_height is the same
+      if( last_qc_block_height == this->last_qc_block_height ) {
+         return std::move(*this);
```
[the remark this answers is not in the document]
linh2931: Thanks! To be consistent with `next` of `block_header_state`, I have changed not doing `std::move`.

heifner (approved):

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -120,6 +120,32 @@ struct pending_block_header_state : public detail::block_header_state_common {
...
+ */
+struct block_header_state_core {
+   uint32_t last_final_block_height = 0; // the block height of the last irreversible
+                                         // (final) block.
+   std::optional<uint32_t> final_on_strong_qc_block_height; // the block height of
+                                                            // the block that would
+                                                            // become irreversible
+                                                            // if the associated block
+                                                            // header was to achieve a
+                                                            // strong QC.
```
greg7mdp, line 136: I think you could use longer lines (similar to line 119 above) so the comment would not be spread out on so many lines.

libraries/chain/block_header_state.cpp
```diff
@@ -23,6 +23,54 @@ namespace eosio { namespace chain {
       }
    }
 
+   block_header_state_core::block_header_state_core( uint32_t last_final_blk_ht,
+                                      std::optional<uint32_t> final_on_strong_qc_blk_ht,
+                                      std::optional<uint32_t> last_qc_blk_ht )
```
greg7mdp, line 28: maybe spell out `block_height` everywhere (instead of `blk_ht`)so it is consistent throughout (like on line 34).

libraries/chain/block_header_state.cpp
```diff
@@ -23,6 +23,54 @@ namespace eosio { namespace chain {
...
+      last_final_block_height(last_final_blk_ht),
+      final_on_strong_qc_block_height(final_on_strong_qc_blk_ht),
+      last_qc_block_height(last_qc_blk_ht) {}
+
+   block_header_state_core block_header_state_core::next( uint32_t last_qc_block_height,
+                                                          bool     is_last_qc_strong) {
+      // no state change if last_qc_block_height is the same
+      if( last_qc_block_height == this->last_qc_block_height ) {
+         block_header_state_core result{*this};
+         return result;
```
greg7mdp, line 39:
```suggestion
         return {*this};
```

greg7mdp (approved):

libraries/chain/include/eosio/chain/block_header_state.hpp
linh2931: Thanks. The shorter lines were ugly. Changed.

libraries/chain/block_header_state.cpp
linh2931: Thanks. Simpler.

greg7mdp (approved):

PR #1926, opened 2023-11-22
IF: Refactor chain_pacemaker ownership

### LINKED ISSUE #1916: IF: Add tests for set_finalizers and hotstuff activation

heifner:
- `chain_pacemaker` now owned by `controller` instead of `chain_plugin`
  - This simplifies test integration
  - Prepares for merger of `fork_database` in the future.
- Moved `hotstuff.hpp` into `hotstuff` library
- Added `hotstuff` library to `libtester`
- Added test for producer changes under hotstuff.
  - These tests can't be completed yet until we have the new hs lib represented in the block header state.

Resolves #1916

### DIFF
--- a/libraries/chain/include/eosio/chain/controller.hpp
+++ b/libraries/chain/include/eosio/chain/controller.hpp
@@ -302,14 +302,14 @@ namespace eosio { namespace chain {
 
          int64_t set_proposed_producers( vector<producer_authority> producers );
 
+         void create_pacemaker(std::set<chain::account_name> my_producers, std::map<std::string,std::string> finalizer_keys, fc::logger& hotstuff_logger);
+         void register_pacemaker_bcast_function(std::function<void(const std::optional<uint32_t>&, const hotstuff::hs_message&)> bcast_hs_message);
+         void register_pacemaker_warn_function(std::function<void(uint32_t, hotstuff::hs_message_warning)> warn_hs_message);
          // called by host function set_finalizers
          void set_proposed_finalizers( const finalizer_set& fin_set );
          void get_finalizer_state( hotstuff::finalizer_state& fs ) const;
          // called from net threads
          void notify_hs_message( const uint32_t connection_id, const hotstuff::hs_message& msg );
-         void create_pacemaker(std::set<chain::account_name> my_producers, std::map<std::string,std::string> finalizer_keys, fc::logger& hotstuff_logger);
-         void register_pacemaker_bcast_function(std::function<void(const std::optional<uint32_t>&, const hotstuff::hs_message&)> bcast_hs_message);
-         void register_pacemaker_warn_function(std::function<void(uint32_t, hotstuff::hs_message_warning)> warn_hs_message);
 
          bool light_validation_allowed() const;
          bool skip_auth_check()const;
--- a/libraries/chain/include/eosio/chain/controller.hpp
+++ b/libraries/chain/include/eosio/chain/controller.hpp
@@ -25,6 +25,7 @@ namespace eosio::hotstuff {
    struct hs_message;
    struct finalizer_state;
    enum class hs_message_warning;
+   using bls_pub_priv_key_map_t = std::map<std::string, std::string>;
 }
 
 namespace eosio { namespace chain {
@@ -302,7 +303,7 @@ namespace eosio { namespace chain {
 
          int64_t set_proposed_producers( vector<producer_authority> producers );
 
-         void create_pacemaker(std::set<chain::account_name> my_producers, std::map<std::string,std::string> finalizer_keys, fc::logger& hotstuff_logger);
+         void create_pacemaker(std::set<chain::account_name> my_producers, hotstuff::bls_pub_priv_key_map_t finalizer_keys, fc::logger& hotstuff_logger);
          void register_pacemaker_bcast_function(std::function<void(const std::optional<uint32_t>&, const hotstuff::hs_message&)> bcast_hs_message);
          void register_pacemaker_warn_function(std::function<void(uint32_t, hotstuff::hs_message_warning)> warn_hs_message);
          // called by host function set_finalizers
--- a/libraries/chain/include/eosio/chain/controller.hpp
+++ b/libraries/chain/include/eosio/chain/controller.hpp
@@ -6,7 +6,6 @@
 #include <eosio/chain/snapshot.hpp>
 #include <eosio/chain/protocol_feature_manager.hpp>
 #include <eosio/chain/webassembly/eos-vm-oc/config.hpp>
-#include <eosio/hotstuff/hotstuff.hpp>
 
 #include <chainbase/pinnable_mapped_file.hpp>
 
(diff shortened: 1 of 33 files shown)


libraries/hotstuff/include/eosio/hotstuff/hotstuff.hpp
```diff
@@ -4,18 +4,17 @@
 #include <fc/crypto/bls_private_key.hpp>
 #include <fc/crypto/bls_public_key.hpp>
 #include <fc/crypto/bls_signature.hpp>
-#include <fc/crypto/bls_utils.hpp>
 
 #include <boost/dynamic_bitset.hpp>
 
-namespace eosio::chain {
+namespace eosio::hotstuff {
```
greg7mdp, line 10: Why do we even have a `eosio::hotstuff` namespace? I think it would be better to have it all live under `eosio::chain`.  Many types already start with `hs_` so there is no ambiguity. And things like `quorum_certificate` are not ambiguous either. I even wonder why we have a separate directory and  library for hotstuff.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -154,7 +153,7 @@ namespace eosio::hotstuff {
 
       qc_chain(std::string id, base_pacemaker* pacemaker,
                std::set<name> my_producers,
-               chain::bls_key_map_t finalizer_keys,
+               std::map<std::string,std::string> finalizer_keys,
```
greg7mdp, line 156: I think it would be better to still have a named type for this.

libraries/hotstuff/include/eosio/hotstuff/test_pacemaker.hpp
```diff
@@ -64,20 +64,20 @@ namespace eosio { namespace hotstuff {
...
 
-      void send_hs_proposal_msg(const hs_proposal_message & msg, const std::string& id, const std::optional<uint32_t>& exclude_peer);
-      void send_hs_vote_msg(const hs_vote_message & msg, const std::string& id, const std::optional<uint32_t>& exclude_peer);
-      void send_hs_new_view_msg(const hs_new_view_message & msg, const std::string& id, const std::optional<uint32_t>& exclude_peer);
+      void send_hs_proposal_msg(const hs_proposal_message & msg, const std::string& id, const std::optional<uint32_t>& exclude_peer) override;
+      void send_hs_vote_msg(const hs_vote_message & msg, const std::string& id, const std::optional<uint32_t>& exclude_peer) override;
+      void send_hs_new_view_msg(const hs_new_view_message & msg, const std::string& id, const std::optional<uint32_t>& exclude_peer) override;
 
-      void send_hs_message_warning(const uint32_t sender_peer, const chain::hs_message_warning code);
+      void send_hs_message_warning(uint32_t sender_peer, const hs_message_warning code) override;
```
greg7mdp, line 80: aren't these final?

heifner: They are final now, but this is test code and there might be a need to override again for a specific test and performance is not important here.

libraries/hotstuff/include/eosio/hotstuff/hotstuff.hpp
heifner: The `hotstuff` namespace exists and I do think having this namespace and directory is helpful. Since it does exist, I believe these types belong in the namespace.

linh2931: I agree we probably don't need a separate directory for hotstuff, especially now that pacemaker is owned by controller. But we can do those move around later after prototyping.

libraries/hotstuff/CMakeLists.txt
```diff
@@ -15,3 +15,10 @@ target_include_directories( hotstuff
                           )
 
 add_subdirectory( test )
+
+install(TARGETS hotstuff
```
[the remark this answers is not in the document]
heifner: A directory inside `library/chain` makes sense to me. Do you want that done as part of this PR?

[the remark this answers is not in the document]
linh2931: Is this too much ask for you for now? Maybe it is best to do now, as it would make things simpler and avoid big changes later.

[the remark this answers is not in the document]
heifner: Before we do this, I think we should get @arhag take on it.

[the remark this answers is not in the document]
greg7mdp: I agree that this would be better, and probably better to do this in this PR because of the CMakeLists.txt changes. I think we can still have separate classes for `qc_chain` and `pacemaker`, and have these insulated by not accessing the controller directly, but as they are an integral part of the chain behavior, I really feel they should be in `libraries/chain`. 
I don't think they need a separate directory, I don't see the benefit.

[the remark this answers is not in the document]
heifner: The benefit of a diff directory to me is it keeps the hotstuff consensus logic separate.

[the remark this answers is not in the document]
greg7mdp: My personal thinking is that hotstuff It is just in two classes, and I don't think there will be ambiguity where it is implemented. I feel that the separate directory is unnecessary. Why not have separate dirs for the fork database, the abi_serializer, etc...?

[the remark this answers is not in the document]
linh2931: Yes, we should get Areg's opinion. You can have your current PR in as is.

libraries/chain/webassembly/privileged.cpp
```diff
@@ -196,7 +196,7 @@ namespace eosio { namespace chain { namespace webassembly {
 
       // system contract should perform a duplicate check and fthreshold check before calling
       EOS_ASSERT( finalizers.size() == unique_finalizer_keys.size(), wasm_execution_error, "Duplicate finalizer bls key in finalizer set" );
-      EOS_ASSERT( finset.fthreshold > f_weight_sum / 2, wasm_execution_error, "Finalizer set threshold cannot be met by finalizer weights" );
+      EOS_ASSERT( finset.fthreshold >= f_weight_sum / 2, wasm_execution_error, "Finalizer set threshold cannot be met by finalizer weights" );
```
heifner, line 199: @arhag any issue with changing this? Test case of two finalizers with weight of 1 each does not work unless this is set to `>=`.

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -19,6 +21,13 @@ namespace boost { namespace asio {
 
 namespace eosio { namespace vm { class wasm_allocator; }}
 
+namespace eosio::hotstuff {
+   struct hs_message;
+   struct finalizer_state;
+   enum class hs_message_warning;
+   using bls_pub_priv_key_map_t = std::map<std::string, std::string>;
+}
```
greg7mdp, line 29: I don't think there would be any ambiguity if these were defined in `eosio::chain`

greg7mdp (approved):

PR #1927, opened 2023-11-22
[IF] Modify quorum_certification to support strong and weak

### LINKED ISSUE #1881: IF: Unification: Modify quorum_certification to support strong and weak QCs
Votes for a proposal can either be strong or weak. A weak vote signs a different digest than the one used for a strong vote (the digest for a weak vote should be derived from the digest for a strong vote). The details of how these digests are computed or why a finalizer would choose to vote strong or weak is left for another issue. However, a finalizer can only choose between a weak or strong vote for any given proposal not both.

A `pending_quorum_certificate` is a data structure that contains the data needed to collect further finalizer votes to ultimately construct a valid QC that meets the necessary threshold for either a weak or strong QC (based on the reference finalizer policy):
* Strong bitset: 1 bit for each finalizer. If set, it means that the corresponding finalizer has contributed a strong vote to the strong aggregated signature.
* Strong aggregated signature: Aggregates all of the signatures corresponding to strong votes for the corresponding proposal of this QC.
* Weak bitset: 1 bit for each finalizer. If set, it means that the corresponding finalizer has contributed a weak vote to the weak aggregated signature. It is disallowed for a finalizer to have both a weak and strong vote present for any given QC.
* Weak aggregated signature: Aggregates all of the signatures corresponding to weak votes for the corresponding proposal of this QC.
* Note that we do not explicitly reference the proposal that this QC corresponds to. It is implied by the context in which the QC is included. The implied proposal also implies the finalizer policy to be used to validate the aggregate signatures.

There should be no intersection between the strong and weak bitsets.

A `pending_quorum_certificate` can be in one of the following states:
* `unrestricted`: Starting state.
* `restricted`: Enough votes to know it is impossible to reach the `strong` state.
* `weak_achieved`: Enough votes to have a valid weak QC but enough votes are pending that it is possible to still reach the `strong` state.
* `weak_final`: Enough votes to have a valid weak QC and also enough to know that it is impossible to reach the `strong` state.
* `strong`: Enough votes to have a valid strong QC.

If the accumulated weight of strong votes for a given proposal is greater than or equal to the  threshold for the referenced finalizer policy, the `pending_quorum_certificate` is in the `strong` state. Otherwise, if the accumulated weight of strong votes and weak votes together is greater than or equal to the threshold for the referenced finalizer policy, the `pending_quorum_certificate` is in either the `weak_achieved` or `weak_final` state.

A `valid_quorum_certificate` is a data structure that contains just the necessary data for someone with access to the corresponding finalizer policy to validate the QC as either a weak or strong QC:
* Optional strong bitset: Same as for `pending_quorum_certificate` except the votes are contributed to the single aggregated signature in this data structure. Should not be present if there are no weak votes.
* Optional weak bitset: Same as for `pending_quorum_certificate` except the votes are contributed to the single aggregated signature in this data structure. Should not be present if there are no weak votes.
* Aggregated signature: Aggregates all of the signatures corresponding to either strong or weak votes for the corresponding proposal of this QC.

So when there are sufficient votes, we can construct a `valid_quorum_certificate` from a `pending_quorum_certificate`. 

If there are sufficient strong votes for a strong QC (state: `strong`), then:
   + The strong bitset is copied over to the new data structure.
   + The weak bitset is omitted.
   + The aggregated signature is set to the strong aggregated signature.

If there are not sufficient strong votes for a strong QC but there are enough strong + weak votes for a weak QC (states: `weak_achieved` or `weak_final`), then:
   + The strong bitset is copied over to the new data structure.
   + The weak bitset is copied over to the new data structure.
   + The aggregated signature is set to the sum of the strong aggregated signature and weak aggregated signature.

Note (mostly relevant for an optimistically responsiveness implementation of IF): We wish to delay creating a `valid_quorum_certificate` from a `pending_quorum_certificate` if it has the status of `weak_achieved` because of a hope that it becomes `strong`. However, based on timeouts, we may wish to do the conversion rather than waiting longer and missing an opportunity to propagate a valid QC to finalizers. There is no value (other than maybe trying to increase representation of lagging finalizers in the QC) to wait to create a `valid_quorum_certificate` from a `pending_quorum_certificate` that has a `weak_final` status because it will never become `strong`.

Modify vote message to capture that the vote could either be strong or weak. For now, just keep them strong votes and use some finalizer digest (could just be block ID for now) as the hash to sign. Receiver of votes should aggregate the votes properly into `pending_quorum_certificate`. Anytime a vote is aggregated into `pending_quorum_certificate` its state should be re-evaluated.

greg7mdp:
Resolves #1881.


libraries/chain/include/eosio/chain/hotstuff.hpp
```diff
@@ -112,7 +115,7 @@ namespace eosio::chain {
 
 
 FC_REFLECT(eosio::chain::view_number, (bheight)(pcounter));
-FC_REFLECT(eosio::chain::quorum_certificate_message, (proposal_id)(active_finalizers)(active_agg_sig));
+FC_REFLECT(eosio::chain::quorum_certificate_message, (proposal_id)(strong_votes)(active_agg_sig));
```
fcecin, line 118: `weak_votes` is not sent? (just making sure)

fcecin (approved):

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -77,71 +77,261 @@ namespace eosio::hotstuff {
...
-              , active_agg_sig(msg.active_agg_sig) {
-               active_finalizers.resize(finalizer_count);
+      explicit pending_quorum_certificate(const fc::sha256& proposal_id,
+                                          const digest_type& proposal_digest,
+                                          size_t num_finalizers,
+                                          size_t quorum) :
+         pending_quorum_certificate(num_finalizers, quorum) {         
+         _proposal_id = proposal_id;
+         _proposal_digest.assign(proposal_digest.data(), proposal_digest.data() + 32);
+         _state = state_t::unrestricted;
```
linh2931, line 148: `_state` already is defaulted to unrestricted.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -77,71 +77,261 @@ namespace eosio::hotstuff {
...
       }
 
+      size_t num_weak()   const { return _weak_votes.count(); }
+      size_t num_strong() const { return _strong_votes.count(); }
+
+      bool   valid() const { return _state >= state_t::weak_achieved; }
```
linh2931, line 154: It is safer and clearer to spell out what states are valid, in case the order and/or elements in the enum change in the future.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -77,71 +77,261 @@ namespace eosio::hotstuff {
...
+         _num_finalizers = num_finalizers;
+         _state = state_t::unrestricted;
       }
 
-      void reset(const fc::sha256& proposal, size_t finalizer_size) {
-         proposal_id = proposal;
-         active_finalizers = hs_bitset{finalizer_size};
-         active_agg_sig = fc::crypto::blslib::bls_signature();
-         quorum_met = false;
+      bool add_strong_vote(const std::vector<uint8_t>& proposal_digest,
```
linh2931, line 185: Why this and other implementation not in .cpp file?

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -77,71 +77,261 @@ namespace eosio::hotstuff {
...
+               _state = state_t::weak_final;
+            break;
+            
+         case state_t::weak_final:
+         case state_t::strong:
+            // getting another weak vote... nothing to do
+            break;
+         }
+         return true;
       }
```
heifner, line 252: Please add unittests for `pending_quorum_certificate` state transitions via votes.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -77,71 +77,261 @@ namespace eosio::hotstuff {
...
+                         const bls_signature& sig) {
+         assert(index < _num_finalizers);
+         if (!_weak_votes.add_vote(proposal_digest, index, pubkey, sig))
+            return false;
+         size_t weak   = num_weak();
+         size_t strong = num_strong();
+         
+         switch(_state) {
+         case state_t::unrestricted:
+         case state_t::restricted:
```
linh2931, line 229: `restricted` should be moved out of here and moved down.

libraries/hotstuff/qc_chain.cpp
```diff
@@ -388,81 +350,74 @@ namespace eosio::hotstuff {
...
-               _b_leaf = proposal_candidate.proposal_id;
-
-               //todo : asynchronous?
-               //write_state(_liveness_state_file , _liveness_state);
-
-               send_hs_proposal_msg( std::nullopt, proposal_candidate );
-
-               fc_tlog(_logger, " === ${id} _b_leaf updated (process_vote): ${proposal_id}", ("proposal_id", proposal_candidate.proposal_id)("id", _id));
+            if (finalizers[i].public_key == vote.finalizer_key) {
+               digest_type digest = p->get_proposal_digest();
```
heifner, line 362: This should be called once outside of the loop. Also I think it should be renamed to `calculate_proposal_digest()` as it is not a getter. We may want to consider only creating the proposal digest once and caching it in the `hs_proposal_message` or somewhere else.

heifner (changes requested):

libraries/hotstuff/include/eosio/hotstuff/hotstuff.hpp
```diff
@@ -114,10 +116,10 @@ namespace eosio::hotstuff {
 
 
 FC_REFLECT(eosio::hotstuff::view_number, (bheight)(pcounter));
-FC_REFLECT(eosio::hotstuff::quorum_certificate_message, (proposal_id)(active_finalizers)(active_agg_sig));
+FC_REFLECT(eosio::hotstuff::quorum_certificate_message, (proposal_id)(strong_votes)(active_agg_sig));
```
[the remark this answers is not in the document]
greg7mdp: fixed.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -77,71 +77,261 @@ namespace eosio::hotstuff {
...
    using namespace eosio::chain;
-
-   class quorum_certificate {
+   
+   using bls_public_key  = fc::crypto::blslib::bls_public_key;
+   using bls_signature   = fc::crypto::blslib::bls_signature;
+   using bls_private_key = fc::crypto::blslib::bls_private_key;
+
+   static inline std::string bitset_to_string(const hs_bitset& bs) { std::string r; boost::to_string(bs, r); return r; }
+   static inline hs_bitset   vector_to_bitset(const std::vector<unsigned_int>& v) { return { v.cbegin(), v.cend() }; }
```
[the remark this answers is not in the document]
greg7mdp: changed

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
greg7mdp: I'm not sure I agree. We could also add another `valid` state after `weak_achieved` and forget to update `valid()`, but changed anyways.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
greg7mdp: Why would it be much better in the .cpp file? 
In any case like this it is easier to modify for now, and if we want later we can move the implementations in the .cpp.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -77,71 +77,261 @@ namespace eosio::hotstuff {
...
+                           const bls_signature& sig) {
+         assert(index < _num_finalizers);
+         if (!_strong_votes.add_vote(proposal_digest, index, pubkey, sig))
+            return false;
+         size_t weak   = num_weak();
+         size_t strong = num_strong();
+         
+         switch(_state) {
+         case state_t::unrestricted:
+         case state_t::restricted:
```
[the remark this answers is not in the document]
greg7mdp: Not true, restricted means we can't achieve strong, but we can still achieve weak_final.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
greg7mdp: Hum, I don't think so.

libraries/hotstuff/qc_chain.cpp
```diff
@@ -796,7 +750,14 @@ namespace eosio::hotstuff {
       EOS_ASSERT( b_lock != nullptr || _safety_state.get_b_lock().empty() , chain_exception, "expected hs_proposal ${id} not found", ("id", _safety_state.get_b_lock()) );
 
       //fc_tlog(_logger, " === update_high_qc : proposal.justify ===");
-      update_high_qc(quorum_certificate{proposal.justify, 21}); // TODO: use active schedule size
+      const hs_proposal_message *justify = get_proposal(proposal.justify.proposal_id);
```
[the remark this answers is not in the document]
greg7mdp: done.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
greg7mdp: removed

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -77,71 +77,261 @@ namespace eosio::hotstuff {
...
+
+   class valid_quorum_certificate {
+   public:
+      valid_quorum_certificate(const pending_quorum_certificate& qc) :
+         _proposal_id(qc._proposal_id),
+         _proposal_digest(qc._proposal_digest) {
+         if (qc._state == pending_quorum_certificate::state_t::strong) {
+            _strong_votes = qc._strong_votes._bitset;
+            _sig = qc._strong_votes._sig;
+         } if (qc._state > pending_quorum_certificate::state_t::weak_achieved) {
```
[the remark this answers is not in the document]
greg7mdp: updated.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -77,71 +77,261 @@ namespace eosio::hotstuff {
...
-         active_agg_sig = fc::crypto::blslib::bls_signature();
-         quorum_met = false;
+      bool add_strong_vote(const std::vector<uint8_t>& proposal_digest,
+                           size_t index,
+                           const bls_public_key& pubkey,
+                           const bls_signature& sig) {
+         assert(index < _num_finalizers);
+         if (!_strong_votes.add_vote(proposal_digest, index, pubkey, sig))
+            return false;
+         size_t weak   = num_weak();
```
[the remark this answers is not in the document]
greg7mdp: These are local variables of small functions, so I don't think it requires much effort to memorize the variables meaning, and I don't like reusing the name of member functions.

libraries/hotstuff/include/eosio/hotstuff/hotstuff.hpp
```diff
@@ -47,12 +47,14 @@ namespace eosio::hotstuff {
...
       fc::sha256                          proposal_id;
-      std::vector<chain::unsigned_int>    active_finalizers; //bitset encoding, following canonical order
+      std::vector<chain::unsigned_int>    strong_votes; //bitset encoding, following canonical order
+      std::vector<chain::unsigned_int>    weak_votes;   //bitset encoding, following canonical order
       fc::crypto::blslib::bls_signature   active_agg_sig;
    };
 
    struct hs_vote_message {
       fc::sha256                          proposal_id; //vote on proposal
+      bool                                strong;
```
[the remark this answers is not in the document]
greg7mdp: done

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
```diff
@@ -77,71 +77,261 @@ namespace eosio::hotstuff {
...
+   using bls_private_key = fc::crypto::blslib::bls_private_key;
+
+   static inline std::string bitset_to_string(const hs_bitset& bs) { std::string r; boost::to_string(bs, r); return r; }
+   static inline hs_bitset   vector_to_bitset(const std::vector<unsigned_int>& v) { return { v.cbegin(), v.cend() }; }
+   static inline std::vector<unsigned_int> bitset_to_vector(const hs_bitset& bs) { 
+      std::vector<unsigned_int> r;
+      r.resize(bs.num_blocks());
+      boost::to_block_range(bs, r.begin());
+      return r;
+   }
```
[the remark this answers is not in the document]
greg7mdp: done

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
greg7mdp: Will do in a separate PR.

libraries/hotstuff/qc_chain.cpp
greg7mdp: done.

libraries/hotstuff/qc_chain.cpp
```diff
@@ -388,81 +350,74 @@ namespace eosio::hotstuff {
...
+
+                        //todo : asynchronous?
+                        //write_state(_liveness_state_file , _liveness_state);
+
+                        send_hs_proposal_msg( std::nullopt, proposal_candidate );
+
+                        fc_tlog(_logger, " === ${id} _b_leaf updated (process_vote): ${proposal_id}", ("proposal_id", proposal_candidate.proposal_id)("id", _id));
+                     }
+                  }
+               }
```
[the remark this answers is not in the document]
greg7mdp: done.

heifner (approved):

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
linh2931: Some times new values are just added to the end of existing ones, as enum order is not significant, if the values aer not explicitly specified. This prevents that from happening.

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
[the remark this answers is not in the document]
linh2931: Not a big deal. It is just when reading `strong >= _quorum` and `strong + week`, it sounds strange.

linh2931 (approved):

libraries/hotstuff/include/eosio/hotstuff/qc_chain.hpp
greg7mdp: The test was ready, so I just added it to this PR. Caugh one small issue too :-).

libraries/hotstuff/test/hotstuff_tools.cpp
```diff
@@ -49,8 +54,189 @@ BOOST_AUTO_TEST_CASE(view_number_tests) try {
...
+         pending_quorum_certificate qc2(qc);
+
+         // add a strong vote
+         // -----------------
+         strong_vote(qc2, digest, 2);
+         BOOST_CHECK_EQUAL(qc2._state, state_t::weak_final);
+         BOOST_CHECK(qc2.is_quorum_met());
+      }
+   }
+
```
linh2931, line 241: Those tests are great. You can add one for invalid index.

linh2931 (approved):

libraries/hotstuff/test/hotstuff_tools.cpp
greg7mdp: Hum, I have an `assert()` there, not a `EOS_ASSERT`, so I don't think I can add a test for it.  
When calling `qc.add_vote()`, we already did a lookup in our finalizer list and matched the public key, so if the index is out of range in `qc.add_vote()`, we have a serious logic problem (cannot be a user or config error), and I think that an `assert` is warranted.

Issue #1910, opened 2023-11-16
IF: Unification: Deprecate proposal and new_block messages


arhag:
We do not need the new_block message anymore. Just get rid of it.

~~Remove proposal network message and instead generate a proposal to incorporate into qc_chain by generating one from the `signed_block` network message. We need to extract the relevant proposal information from the `signed_block`. Right now all that is needed is proposal it builds off of. This can be identified using the `previous` field of the `signed_block`. We can have the proposal ID be the block ID now to make this identification easy.~~

EDIT: We don't need to remove the proposal network message as part of this issue since we will make more significant changes soon that makes that work obsolete.

fcecin:
Part 1/2 of this issue (remove new block message) is addressed by #1919

For part 2/2 (remove proposal message), my current idea is:

- Remove message propagation from qc_chain::process_proposal, since it is no longer a network message
- Change test_pacemaker to not require hs_proposal_message multi-hop propagation (do what was being done for the now defunct hs_new_block_message)

This guarantees that the unit tests keep working with minimal modifications (since the unit tests don't have `signed_block`, so they will keep using hs_proposal_message internally to emulate it, at least for now).

Then:

- Produce a hs_proposal_message structure anywhere in e.g. net_plugin and pass it to the chain_pacemaker (e.g. when a `signed_block` is received, or at any time when a new proposal is supposed to be modeled inside of the qc_chain)
- Remove qc_chain::on_beat() (no longer needed; instead the production of a `signed_block` locally is no different than receiving a `signed_block` from a remote producer)

Does this look like a good idea?

arhag:
https://github.com/AntelopeIO/leap/pull/1919 is sufficient to close this issue. https://github.com/AntelopeIO/leap/pull/1929 is no longer needed.

PR #1929, opened 2023-11-23
IF: Remove proposal msg

fcecin:
__NOTE__: This is being submitted a *Draft* PR -- this is not intended for merging at this time; this is for discussions, and if someone wants to inspect the code and critique it.

This PR will implement part 2 of 2 of #1910 , completely removing proposal network messages. `hs_proposal_message` has been renamed to `hs_proposal` as @linh2931 had suggested.

As part of this PR, the `beat()` entrypoint into the IF engine has been completely removed: this has been done because the side-effect of `beat()` was to send out a proposal message to the network -- which does not exist anymore. So, instead of notifying the IF engine that "a block has been produced" through `void beat(void)`, we will need to create two other pacemaker methods (pseudo-code):

* `qc signed_block_produced(block_id)`, which is called by Leap when it needs a QC to put into a `signed_block` that is being produced and is going out, and
* `void signed_block_received(qc, block_id)`, which is called by Leap when it receives a `signed_block` from the network that has some block_id and a QC in it.

Both these methods will create and/or update proposal entries in the IF engine's proposal store.

The first effectively replaces the `beat` pattern, and the second replaces the `on-incoming-proposal-message(msg)` pattern.

Once these methods are implemented, ensuring the parameters and return values are modeled correctly and are sufficient information, we then need to find where in Leap to call them and do so.


fcecin:
I don't know how a signed_block QC is converted to an hs_proposal and vice-versa, so commit `69af81a` is as far as I can get on this issue for now. What is left to do in this PR is to fill in the body of the two new entry points, and then connect them to the rest of Leap.

libraries/hotstuff/chain_pacemaker.cpp
```diff
@@ -131,6 +131,25 @@ namespace eosio { namespace hotstuff {
       warn_hs_message = std::move(warning_hs_message);
    }
 
+   // signed_block is being produced and needs a QC to be put in it before it is broadcast
+   quorum_certificate_message chain_pacemaker::on_block_proposal(const chain::block_id_type& block_id) {
```
linh2931, line 135: The name `on_block_proposal` implies on receiving a block proposal. Can it be renamed?

linh2931: `on_producing_block`?

PR #1932, opened 2023-11-27
IF: Add a incremental merkle tree that does not do left/right bit flip

### LINKED ISSUE #1913: IF:  New incremental merkle tree
Add new incremental Merkle tree class (or add flag to existing one or make original into template to support old and new) that does not use the MSB to indicate whether it is a left or right child. This will replace the existing one within the new IF algorithm. Add unit tests.

Left and right orientation of nodes (inner or leaf) can be determined by their location within the tree (depth and ordinal). So the MSB was redundant.

Add helper function to get the coordinates (depth and ordinal) for a node reached from an existing node (left child, right child, or parent).

heifner:
- Renamed function `merkle` to `canonical_merkle` to make it clear it uses the bit flip implementation.
- Renamed `incremental_merkle` to `incremental_canonical_merkle` to make it clear it uses the bit flip implementation.
- Added `incremental_merkle_tree` as an incremental merkle tree implementation that does not use left/right bit flip.
- Added tests.

Resolves #1913

### DIFF
--- a/libraries/chain/include/eosio/chain/incremental_merkle.hpp
+++ b/libraries/chain/include/eosio/chain/incremental_merkle.hpp
@@ -90,14 +90,13 @@ inline void move_nodes(Container& to, Container&& from) {
  * change.  This allows proofs based on this merkle to be very stable
  * after some time has past only needing to update or add a single
  * value to maintain validity.
+ *
+ * @param canonical if true use the merkle make_canonical_pair which sets the left/right bits of the hash
  */
-template<typename DigestType, template<typename ...> class Container = vector, typename ...Args>
+template<typename DigestType, bool canonical = false, template<typename ...> class Container = vector, typename ...Args>
 class incremental_merkle_impl {
    public:
-      incremental_merkle_impl()
-      :_node_count(0)
-      {}
-
+      incremental_merkle_impl() = default;
       incremental_merkle_impl( const incremental_merkle_impl& ) = default;
       incremental_merkle_impl( incremental_merkle_impl&& ) = default;
       incremental_merkle_impl& operator= (const incremental_merkle_impl& ) = default;
@@ -188,7 +187,11 @@ class incremental_merkle_impl {
 
                // calculate the partially realized node value by implying the "right" value is identical
                // to the "left" value
-               top = DigestType::hash(make_canonical_pair(top, top));
+               if constexpr (canonical) {
+                  top = DigestType::hash(make_canonical_pair(top, top));
+               } else {
+                  top = DigestType::hash(std::make_pair(std::cref(top), std::cref(top)));
+               }
                partial = true;
             } else {
                // we are collapsing from a "right" value and an fully-realized "left"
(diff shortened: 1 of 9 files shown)


libraries/chain/include/eosio/chain/incremental_merkle.hpp
```diff
@@ -238,14 +243,17 @@ class incremental_merkle_impl {
...
+      friend struct fc::reflector<incremental_merkle_impl>;
+      uint64_t                         _node_count = 0;
       Container<DigestType, Args...>   _active_nodes;
 };
 
-typedef incremental_merkle_impl<digest_type>               incremental_merkle;
-typedef incremental_merkle_impl<digest_type,shared_vector> shared_incremental_merkle;
+typedef incremental_merkle_impl<digest_type, true>                incremental_canonical_merkle;
+typedef incremental_merkle_impl<digest_type, true, shared_vector> shared_incremental_canonical_merkle;
+typedef incremental_merkle_impl<digest_type>                      incremental_merkle_tree;
```
greg7mdp, line 254: why `_tree` at the end of this one... seems inconsistent to me.

unittests/merkle_tree_tests.cpp
```diff
@@ -0,0 +1,263 @@
...
+   auto node6 = fc::sha256::hash("Node6");
+   auto node7 = fc::sha256::hash("Node7");
+   auto node8 = fc::sha256::hash("Node8");
+   auto node9 = fc::sha256::hash("Node9");
+
+   tree.append(node1);
+   BOOST_CHECK_EQUAL(tree.get_root(), node1);
+
+   tree.append(node2);
+   BOOST_CHECK_EQUAL(tree.get_root().str(), fc::sha256::hash(make_canonical_pair(node1, node2)).str());
```
greg7mdp, line 34: why `.str()` here and below, and not on the previous one?

greg7mdp (approved):

unittests/merkle_tree_tests.cpp
heifner: The str() gives a readable error message.

greg7mdp: So the one at line 31 should also have a `str()`, right?

libraries/chain/include/eosio/chain/incremental_merkle.hpp
greg7mdp: Ah, names are hard. Canonical is not very meaningful to me, I think I'd prefer the following, but totally fine if you leave as is. 
My remark was that I thought having the suffix `_tree` in the last one seemed inconsistent, but again no big deal.

`merkle` -> `bitflip_merkle` or `bf_merkle`
`incremental_merkle` -> `incremental_bitflip_merkle` or `incremental_bf_merkle` 
new one `incremental_noop_merkle` or `incremental_passthru_merkle`

heifner: Renamed `incremental_canonical_merkle` to `incremental_canonical_merkle_tree`

greg7mdp (approved):

libraries/chain/include/eosio/chain/incremental_merkle.hpp
greg7mdp: Thanks! Please update the comment at the top of the PR as well.

unittests/merkle_tree_tests.cpp
```diff
@@ -0,0 +1,263 @@
...
+            fc::sha256::hash(std::make_pair(
+               fc::sha256::hash(std::make_pair(node9, node9)),
+               fc::sha256::hash(std::make_pair(node9, node9))
+            ))
+         )
+      )   ));
+   BOOST_CHECK_EQUAL(tree.get_root().str(), calculated_root.str());
+}
+
+BOOST_AUTO_TEST_SUITE_END()
```
linh2931, line 263: Like the new comprehensive unit tests.

linh2931 (approved):

Issue #1911, opened 2023-11-16
IF: Unification: Add block header extension


arhag:
Depends on https://github.com/AntelopeIO/leap/issues/1881 and https://github.com/AntelopeIO/leap/issues/1941.

The block header extension should carry the new information needed for a proposal:
* `last_qc_block_num`: The block height of the most recent ancestor block that has a QC.
* `is_last_qc_strong`: This is a boolean that determines if the QC justification for the block referenced by last QC block height is strong or weak.
* `new_finalizer_policy`: The new finalizer policy proposed in the block (may be nullopt).
* `new_proposer_policy`: The new proposer policy proposed in the block (may be nullopt).

Example:
```cpp
 struct instant_finality_extension {
    uint32_t last_qc_block_num {0};  ///< The block height of the most recent ancestor block that has a QC justification
    bool     is_last_qc_strong {false}; ///< whether the QC for the block referenced by last_qc_block_height is strong or weak.
    std::optional<finalizer_policy> new_finalizer_policy;
    std::optional<proposer_policy> new_proposer_policy;
 };
```

This should replace the extensions `finalizer_policy_extension` and `proposal_info_extension`.

Use the status of the `pending_quorum_certificate` to determine whether you have a valid QC and if so whether it is strong or weak. That determines the values of `last_qc_block_num` and `is_last_qc_strong`.

Use the most recent ancestor block that has a valid QC (whether the status is `strong`, `weak_final`, or `weak_achieved`) as the block referenced by `last_qc_block_num`.

The actual QC does not need to be included yet in a block extension. That will be handled in a separate issue.

arhag:
Partially done, but to complete we must first wait until at least https://github.com/AntelopeIO/leap/issues/2034 is complete and it would probably be best to wait until https://github.com/AntelopeIO/leap/issues/2046 is also complete.

greg7mdp:
I have finished this issue in my branch gh_2034.

greg7mdp:
@linh2931 @arhag I believe we can close this issue as completed by #2077

Issue #1941, opened 2023-11-30
IF: Unification: Create new block_header_state and block_state


linh2931:
Add new `block_header_state` and `block_state`. Use the `block_state` in a new fork database.

The old code in `block_header_state.{hpp,cpp}` and `block_state.{hpp,cpp}` should have been moved to `block_header_state_legacy.{hpp,cpp}` and `block_state_legacy.{hpp,cpp}`, respectively, in https://github.com/AntelopeIO/leap/issues/1940. For this issue, put the new code into a new `block_header_state.{hpp,cpp}` and `block_state.{hpp,cpp}` as appropriate.

greg7mdp:
Here is what we came up with in the meeting:

```c++
struct building_block_input {
   block_id_type                   previous;
   block_timestamp_type            timestamp;
   account_name                    producer;
   vector<digest_type>             new_protocol_feature_activations;
};
      
// this struct can be extracted from a building block
struct assembled_block_input : public building_block_input {
   digest_type                     transaction_mroot;
   digest_type                     action_mroot;
   std::optional<proposer_policy>  new_proposer_policy;
   std::optional<finalizer_policy> new_finalizer_policy;
   std::optional<quorum_certificate> qc;                 // assert(qc.block_height <= num_from_id(previous));
};

struct block_header_state {
   block_header                        header;
   protocol_feature_activation_set_ptr activated_protocol_features;

   block_header_state_core             core;
   incremental_merkle_tree             proposal_mtree;
   incremental_merkle_tree             finality_mtree;

   finalizer_policy_ptr                finalizer_policy; // finalizer set + threshold + generation, supports `digest()`
   proposer_policy_ptr                 proposer_policy;  // producer authority schedule, supports `digest()`
   
   flat_map<uint32_t, proposer_policy_ptr> proposer_policies;
   flat_map<uint32_t, finalizer_policy_ptr> finalizer_policies;

   digest_type          compute_finalizer_digest() const;
   block_timestamp_type timestamp() const;
   account_name         producer() const;
   block_id_type        previous() const;
   uint32_t             block_num() const { return num_from_id(previous()) + 1; }
   
   block_header_state   next(const assembled_block_input& data) const;
   
   // block descending from this need the provided qc in the block extension
   bool                 is_needed(const quorum_certificate& qc) const {
      return !core.last_qc_block_height || qc.block_height > *core.last_qc_block_height;
   }
};

struct block_state  {
   block_header_state         bhs; // provides parent link
[...]

PR #1943, opened 2023-11-30
IF: Unification: Add block header extension

greg7mdp:
Partially resolves https://github.com/AntelopeIO/leap/issues/1911.

The last part "Use the status of the pending_quorum_certificate to determine whether you have a valid QC and if so whether it is strong or weak. That determines the values of last_qc_block_height and is_last_qc_strong. Use the most recent ancestor block that has a valid QC (whether the status is strong, weak_final, or weak_achieved) as the block referenced by last_qc_block_height." will be done in #1941


heifner (approved):

libraries/chain/include/eosio/chain/hs_proposal_info.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
```
linh2931, line 1: Should this file be located in hotstuff directory? If so, no need `hs_` in the name.

greg7mdp: I can move it. I thought that because it is a block header extension, it should be in `chain`, but I have no strong opinion on this.

greg7mdp: @heifner what do you think, should I move `hs_proposal_info.hpp` to the `hotstuff` subdirectory?

heifner: Makes sense to move it to me.

greg7mdp: Done. 

By that rationale, `finalizer_set.hpp` should be in `hotstuff` as well. Should I move it in this PR?

greg7mdp: Please approve this PR, there are changes all over and I'll do it in a separate PR.

Issue #1940, opened 2023-11-30
Rename existing block_header_state and block_state to block_header_state_legacy and block_state_legacy


linh2931:
* Rename existing structs `block_header_state` to `block_header_state_legacy`, `block_state` to `block_state_lagacy`; 
* Change `block_header_state.{hpp,cpp}` to `block_header_state_legacy.{hpp,cpp}` and change `block_state.{hpp,cpp}` to `block_state_legacy.{hpp,cpp}`.

There should be no functional changes to the behavior of Leap. Snapshots should still work without any version bump required.

This should target `main`. This is needed in preparation for the Instant Finality work.

heifner:
The `_legacy`  versions should match what is in 5.0 not `hotstuff_integration` branch.

greg7mdp:
```c++
struct block_header_state {
   block_header                        header;
   protocol_feature_activation_set_ptr activated_protocol_features;

   block_header_state_core             core;
   uint32_t                            last_proposed_finalizer_set_generation = 0; // TODO: Add to snapshot_block_header_state_v3
   incremental_merkle_tree             proposal_mtree;
   incremental_merkle_tree             finality_mtree;

   finalizer_policy_ptr                finalizer_policy; // finalizer set + threshold + generation, supports `digest()`
   proposer_policy_ptr                 proposer_policy;  // producer authority schedule, supports `digest()`
   
   flat_map<uint32_t, proposer_policy_ptr> proposer_policies;
   flat_map<uint32_t, finalizer_policy_ptr> finalizer_policies;

   digest_type          compute_finalizer_digest() const;
   block_timestamp_type timestamp() const;
   account_name         producer() const;
   block_id_type        previous() const;
   uint32_t             block_num() const { return num_from_id(previous()) + 1; }
   
};

struct block_state  {
   block_header_state         bhs; // provides parent link
   block_id_type              id;
   digest_type                finalizer_digest;
   pending_quorum_certificate pending_qc;
};
```

linh2931:
Reopen to rename `pending_block_header_state` to `pending_block_header_state_legacy`

PR #1947, opened 2023-12-01
IF: Rename existing block_header_state and block_state to block_header_state_legacy and block_state_legacy

linh2931:
* Renamed structs `block_header_state` to `block_header_state_legacy`, `block_state` to `block_state_lagacy`; 
* Changed `block_header_state.{hpp,cpp}` to `block_header_state_legacy.{hpp,cpp}` and change `block_state.{hpp,cpp}` to `block_state_legacy.{hpp,cpp}`;
* Updated snapshot handling;
* Made the _legacy versions match what is in 5.0

Resolved https://github.com/AntelopeIO/leap/issues/1940


linh2931:
> @linh2931 , it looks loke you removed the current hotstuff_integration changes, so for example `block_state.hpp` was removed, and `block_header_state.hpp` is almost emptied. I think you should have left the existing code living in the old classes, even if it is unused for now. We will need both versions, the one with the old code, and the new hotstuff-based ones.
> 
> Looking at the changes, I think even more that this change should be done in the `main` branch, and merged into hotstuff_integration. @arhag what do you think?

My branch started yesterday. That's why it did not have today's hotstuff_integration changes. After I resolve the conflicts, it will have today's changes. I think we want to have a clean `block_state/block_header_state {hpp/cpp}` which only contain the code we are writing for hotstuff. That's why `block_header_state{hpp/cpp}` only have the new `block_header_state_core` class. @arhag.

greg7mdp:
Upon reflection, I don't think this renaming is the way to go (if it is done in `hotstuff_integration` and not `main`).

I think we should leave the old names alone, and create the new classes as `block_header_state_v2` and `block_state_v2`, which is what I have started to do in my branch, and move the hotstuff code to the new classes.

After merging `hotstuff_integration` to main, we can always do a renaming  `old classes -> legacy` and removing the `_v2` suffix from the new classes  if we like.

linh2931:
> Upon reflection, I don't think this renaming is the way to go (if it is done in `hotstuff_integration` and not `main`).
> 
> I think we should leave the old names alone, and create the new classes as `block_header_state_v2` and `block_state_v2`, which is what I have started to do in my branch, and move the hotstuff code to the new classes.
> 
> After merging `hotstuff_integration` to main, we can always do a renaming `old classes -> legacy` and removing the `_v2` suffix from the new classes if we like.

We should discuss this at this afternoon's meeting. If we were to do the renaming on main eventually, why not doing it now as we are developing the new classes? The merging seems not that bad as shown when merging from hotsuff_integration to this renaming branch.

PR #1950, opened 2023-12-05
IF: Merge main 12-05-2023 - affine non-montgomery bls public key and signature

heifner:
Merge `main` as of 12-05-2023 into `hotstuff_integration` includes bls12-381 update for proof of possession in affine non-montgomery form.
Use affine non-montgomery base64 encoding for `bls_public_key` and `bls_signature`.
Update host function `set_finalizers` to expect `bls_public_key` in affine non-montgomery form.


greg7mdp (approved):

linh2931 (approved):

SebastianCanonaco:
I hope that private key you are using is not being used on your environments....

heifner:
> I hope that private key you are using is not being used on your environments....

It is not. It is just a throw away key created for testing.

PR #1955, opened 2023-12-06
IF: merkle calculation for transaction_mroot and action_mroot

### LINKED ISSUE #1942: IF: Use merkle calculation for transaction_mroot and action_mroot that does not use bitflip implementation
- Switch to new implementation on hotstuff activation not on instant finality protocol feature activation.

heifner:
Use merkle calculation for `transaction_mroot` and `action_mroot` that does not use bitflip implementation after hotstuff activation.

Resolves #1942

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1877,20 +1877,31 @@ struct controller_impl {
 
       try {
 
+      uint32_t hs_lib = hs_irreversible_block_num.load();
+      const bool hs_active = hs_lib > 0;
+
       auto& pbhs = pending->get_pending_block_header_state();
 
       auto& bb = std::get<building_block>(pending->_block_stage);
 
       auto action_merkle_fut = post_async_task( thread_pool.get_executor(),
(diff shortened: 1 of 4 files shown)


libraries/chain/controller.cpp
```diff
@@ -1877,20 +1877,31 @@ struct controller_impl {
...
 
       auto action_merkle_fut = post_async_task( thread_pool.get_executor(),
-                                                [ids{std::move( bb._action_receipt_digests )}]() mutable {
-                                                   return canonical_merkle( std::move( ids ) );
+                                                [ids{std::move( bb._action_receipt_digests )}, hs_active]() mutable {
+                                                   if (hs_active) {
+                                                      return calculate_merkle( std::move( ids ) );
+                                                   } else {
+                                                      return canonical_merkle( std::move( ids ) );
+                                                   }
```
linh2931, line 1893: Can you make above code block line 1888 to 1893 as a method taking `hs_active` (or if_active) as a parameter? It can be reused in several places.

linh2931 (approved):

greg7mdp (approved):

PR #1958, opened 2023-12-06
Added pseudo code

systemzax:
Updated pseudo code : libraries/chain/hotstuff/hs_pseudo


greg7mdp (approved):

libraries/chain/hotstuff/hs_pseudo
```diff
@@ -0,0 +1,345 @@
...
+	incremental_block_mtree		finality_mtree;
+	uint32_t					policy_generation;
+}
+
+struct block_handle {
+	block_header_state_ptr 						bhs;
+	finalizer_policy_ptr 						fp;	
+	std::optional<pending_quorum_certificate>	pending_qc; 
+	std::optional<valid_quorum_certificate>		valid_qc; 
+	uint64_t id() const {return bhs->id;}
```
heifner, line 90: `uint64_t` -> `block_id_type` which is `sha256`

libraries/chain/hotstuff/hs_pseudo
```diff
@@ -0,0 +1,345 @@
...
+	sha256          proposal_digest; //proposal digest
+	bls_public_key  finalizer_key;
+	bls_signature   sig;
+	bool 			weak; //indicate if vote is weak, strong otherwise
+};
+
+
+//added as a block_header extension before signing
+struct hotstuff_header_extension {
+	uint32_t					last_qc_block_height;
```
heifner, line 151: Not sure why this is only the height. Doesn't it need to be block id or block digest?

libraries/chain/hotstuff/hs_pseudo
```diff
@@ -0,0 +1,345 @@
...
+}
+
+//when a node receives a vote on a proposal
+void on_vote_received(hs_vote_message v){
+	//check for duplicate or invalid vote, return in either case
+	//abstracted [...]
+	block_handle& bc = fork_db.get_block_by_id(v.block_id);
+	am_i_leader = [...] //abstracted, must return true if I am the leader, false otherwise
+	if(!am_i_leader) return;
+	//only leader need to take action on votes
```
heifner, line 318: Need to relay votes even if not a leader.

libraries/chain/hotstuff/hs_pseudo
systemzax: I think for the header extension we only store the last_qc_block_height and is_last_qc_strong.

Assuming the block is valid, we should always be able to locate an ancestor block at last_qc_block_height on that branch and get its id. 

See : 
https://github.com/AntelopeIO/leap/issues/1911

libraries/chain/hotstuff/hs_pseudo
systemzax: I added a comment to mention this. Also added a similar comment on_proposal_received for consistency.

ericpassmore:
Note:start
group: IF
category: DESIGN
summary: Proposed data-structures and terminology for new faster finality. 
Note:end

PR #1981, opened 2023-12-12
Small console optimization

heifner:
Use C++20 `std::stringstream::view()` instead of `std::stringstream::str()` to avoid string copy.


heifner:
Apparently not available in GNU 10.5 libstdc++

greg7mdp:
> Apparently not available in GNU 10.5 libstdc++

I assume that `rdbuf()->view()` is not available either, right?

heifner:
> > Apparently not available in GNU 10.5 libstdc++
> 
> I assume that `rdbuf()->view()` is not available either, right?

I would assume it is not according to cppreference.

PR #1990, opened 2023-12-14
IF: Add IF activation to integration test framework

heifner:
- Adds ability to activate instant finality to the integration test framework, specifically `Cluster.py launch()`.
  - Unique BLS keys and proof-of-possession is created for each test run similar to proposer keys.
  - Added example in `tests/distributed-transactions-test.py` & `tests/nodeos_run_test.py`
  - If we are happy with this approach, it can be added to all integration tests and new cmake tests can be added that use it.
    - Currently `LIB` is not progressing after activation.
- Copied over `eosio.bios` contract from https://github.com/AntelopeIO/reference-contracts/pull/33 which has updated `setfinalizer`.
- Updated `tutorials/bios-boot-tutorial/bios-boot-tutorial.py` to include `INSTANT_FINALITY` activation.


greg7mdp (approved):

heifner:
Replaced by https://github.com/AntelopeIO/leap/pull/1998

Issue #1957, opened 2023-12-06
Signal `signed_block_ptr` and `block_id` instead of `block_state_ptr`


heifner:
See https://github.com/AntelopeIO/leap/pull/1951#discussion_r1417647036

Currently the controller signals `block_state_ptr` via signals `accepted_block_header`, `accepted_block`, and `irreversible_block`. I think abstraction would be better served to signal `signed_block_ptr, block_id` for these instead.

PR #1996, opened 2023-12-17
Clean up controller signals to improve abstraction

linh2931:
Currently `controller` signals an all-purpose `block_state_legacy_ptr` for `accepted_block_header`, `accepted_block`, `and irreversible_block`. `block_state_legacy_ptr` is passed across functions, hiding actual parameters a function needs. Over 40 files depend on `block_state_legacy.hpp`. 

This PR
* Changed to signal types which are only required.
* Changed function interface to use exact parameters instead of one-for-all `block_state_legacy_ptr` and updated implementation
* Updated tests
* Removed unused signals `pre_accepted_block`, `accepted_transaction`, and `bad_alloc`.

Those changes 
* help make Instant Finality coding simpler by removing the need to signal variant types,
* reduce abstraction leakage,
* reduce the overall codebase dependency on to-be-obsolete `block_state_legacy_ptr` by removing the majority uses of it.

Resolved https://github.com/AntelopeIO/leap/issues/1957

### DIFF
--- a/libraries/chain/include/eosio/chain/controller.hpp
+++ b/libraries/chain/include/eosio/chain/controller.hpp
@@ -44,6 +44,8 @@ namespace eosio { namespace chain {
    // lookup transaction_metadata via supplied function to avoid re-creation
    using trx_meta_cache_lookup = std::function<transaction_metadata_ptr( const transaction_id_type&)>;
 
+   using block_signal_params = std::tuple<const signed_block_ptr&, const block_id_type&>;
+
    class fork_database;
 
    enum class db_read_mode {
@@ -326,10 +328,10 @@ namespace eosio { namespace chain {
 
          static std::optional<uint64_t> convert_exception_to_error_code( const fc::exception& e );
 
-         signal<void(uint32_t)>                        block_start;
-         signal<void(std::tuple<const signed_block_ptr&, const block_id_type&, const signed_block_header&, uint32_t>)> accepted_block_header;
-         signal<void(std::tuple<const signed_block_ptr&, const block_id_type&, const signed_block_header&, uint32_t>)> accepted_block;
-         signal<void(std::tuple<const signed_block_ptr&, const block_id_type&, uint32_t>)> irreversible_block;
+         signal<void(uint32_t)>             block_start;
+         signal<void(block_signal_params)>  accepted_block_header;
+         signal<void(block_signal_params)>  accepted_block;
+         signal<void(block_signal_params)>  irreversible_block;
          signal<void(const transaction_metadata_ptr&)> accepted_transaction;
          signal<void(std::tuple<const transaction_trace_ptr&, const packed_transaction_ptr&>)> applied_transaction;
 
--- a/libraries/chain/include/eosio/chain/controller.hpp
+++ b/libraries/chain/include/eosio/chain/controller.hpp
@@ -327,7 +327,7 @@ namespace eosio { namespace chain {
          static std::optional<uint64_t> convert_exception_to_error_code( const fc::exception& e );
 
          signal<void(uint32_t)>                        block_start;
-         signal<void(std::tuple<const signed_block_ptr&, const block_id_type&, const account_name&>)> accepted_block_header;
+         signal<void(std::tuple<const signed_block_ptr&, const block_id_type&, const signed_block_header&, uint32_t>)> accepted_block_header;
          signal<void(std::tuple<const signed_block_ptr&, const block_id_type&, const signed_block_header&, uint32_t>)> accepted_block;
          signal<void(std::tuple<const signed_block_ptr&, const block_id_type&, const signed_block_header&, uint32_t>)> irreversible_block;
          signal<void(const transaction_metadata_ptr&)> accepted_transaction;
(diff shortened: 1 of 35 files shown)


libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -326,24 +326,12 @@ namespace eosio { namespace chain {
 
          static std::optional<uint64_t> convert_exception_to_error_code( const fc::exception& e );
 
-         signal<void(uint32_t)>                        block_start; // block_num
-         signal<void(const signed_block_ptr&)>         pre_accepted_block;
-         signal<void(const block_state_legacy_ptr&)>   accepted_block_header;
-         signal<void(const block_state_legacy_ptr&)>   accepted_block;
-         signal<void(const block_state_legacy_ptr&)>   irreversible_block;
```
greg7mdp, line 333: Why the tuples instead of passing a list of arguments?

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -326,24 +326,12 @@ namespace eosio { namespace chain {
...
+         signal<void(uint32_t)>                        block_start;
+         signal<void(std::tuple<const signed_block_ptr&, const block_id_type&, const signed_block_header&, uint32_t>)> accepted_block_header;
+         signal<void(std::tuple<const signed_block_ptr&, const block_id_type&, const signed_block_header&, uint32_t>)> accepted_block;
+         signal<void(std::tuple<const signed_block_ptr&, const block_id_type&, uint32_t>)> irreversible_block;
```
greg7mdp, line 332: Why the tuples instead of passing a list of arguments?

greg7mdp: Never mind, got it, looks like it is a restriction from the `channels` of `plugin_interface`.

greg7mdp, line 332: I would create a type:
`using block_signal_params_t = std::tuple<const signed_block_ptr&, const block_id_type&, const signed_block_header&, uint32_t>;`

And use the same type for all 3 signals.

linh2931: Thanks. Will change.

heifner: Why `signed_block_header`. When is `signed_block_header != static_cast<signed_block_header>(*signed_block_ptr)` ?

heifner: I see no reason for `uint32_t` block_num. The user can just use `signed_block_ptr->block_num()`. The id is included because `calculate_id()` is expensive, but `block_num()` I think is efficient enough we don't need to include it in the signal.

linh2931: I decided to use `signed_block_header` and `block_num` explicitly for calling out parameters required by signal clients and for easier review to make sure nothing is missed during the changes. But we need to strike some balances between the number of parameters to signal and abstraction leakage.

linh2931: https://github.com/AntelopeIO/leap/pull/1996/commits/7c4ca8a5df4d7b5438c03663ddbbb1ad9951b8d0

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -35,15 +34,19 @@ struct session_base {
...
+   const chain::signed_block_ptr block;
+   const chain::block_id_type id;
+   uint32_t block_num;
 public:
-   send_update_send_queue_entry(std::shared_ptr<session_base> s, chain::block_state_legacy_ptr block_state)
+   send_update_send_queue_entry(std::shared_ptr<session_base> s, chain::signed_block_ptr block, chain::block_id_type id, uint32_t block_num)
          : session(std::move(s))
-         , block_state(std::move(block_state)){}
+         , block(std::move(block))
+         , id(std::move(id))
```
heifner, line 44: id doesn't move, should pass by reference.

heifner (changes requested):

greg7mdp (approved):

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
linh2931: Thanks. Fixed.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1329,10 +1329,18 @@ void producer_plugin_impl::plugin_startup() {
...
-         _accepted_block_header_connection.emplace(chain.accepted_block_header.connect([this](const auto& bsp) { on_block_header(bsp); }));
-         _irreversible_block_connection.emplace(
-            chain.irreversible_block.connect([this](const auto& bsp) { on_irreversible_block(bsp->block); }));
+         _accepted_block_connection.emplace(chain.accepted_block.connect([this](const block_signal_params& t) {
+            const auto& [ block, id ] = t;
+            on_block(block);
+          }));
+         _accepted_block_header_connection.emplace(chain.accepted_block_header.connect([this](const block_signal_params& t) {
+            const auto& [ block, id ] = t;
+            on_block_header(static_cast<signed_block_header>(*block).producer, block->block_num(), block->timestamp);
```
heifner, line 1338: This cast is not needed.

heifner (approved):

PR #1998, opened 2023-12-18
IF: Add IF activation to integration test framework

heifner:
- Adds ability to activate instant finality to the integration test framework, specifically `Cluster.py launch()`.
  - Unique BLS keys and proof-of-possession is created for each test run similar to proposer keys.
  - Added example in `tests/distributed-transactions-test.py` & `tests/nodeos_run_test.py`
  - If we are happy with this approach, it can be added to all integration tests and new cmake tests can be added that use it.
    - Currently `LIB` is not progressing after activation.
- Updated `tutorials/bios-boot-tutorial/bios-boot-tutorial.py` to include `INSTANT_FINALITY` activation.
- Updated `Cluster.py` to use `eosio.boot` contract for protocol feature activation instead of old bios contract.
- Updated `Cluster.py` to use new `setprods` of the bios contract.
- Updated `prod_preactivation_test.py` to work as before without needing old bios contracts to be setup by `Cluster.py`.


linh2931 (approved):
It looks good. We should see actual IF activation when other parts of code are in.

greg7mdp (approved):
Nice!

Issue #1865, opened 2023-11-06
Test Failure: plugin_http_api_test 5.0


heifner:
https://github.com/AntelopeIO/leap/actions/runs/6771687724/job/18403013839

heifner:
https://github.com/AntelopeIO/leap/actions/runs/6804586080/job/18531623820

heifner:
https://github.com/AntelopeIO/leap/actions/runs/6869950320/job/18684154078

Issue #1970, opened 2023-12-07
`larger_lib_test` test failure


spoonincode:
Don't recall ever seeing an issue with this test before so a bit surprising.

https://github.com/AntelopeIO/leap/actions/runs/7132387412/job/19423267255

bhazzard:
Let's do a quick look on this to determine if there is a more serious issue at play that needs to be fixed prior to 5.0.0 stable.

spoonincode:
https://github.com/AntelopeIO/leap/actions/runs/7213305978/job/19653335121

linh2931:
https://github.com/AntelopeIO/leap/actions/runs/7477546239/job/20350978144?pr=2067#step:4:592

arhag:
This test should just be removed. That is covered by https://github.com/AntelopeIO/leap/issues/2232.

PR #2000, opened 2023-12-19
[5.0] Test: Fix trx_generator handling of connection lost

heifner:
Add an `on_except` so `named_thread_pool` does not terminate when connection lost to client. This should fix the core dumps found in ci/cd runs such as https://github.com/AntelopeIO/leap/issues/1865
This is not a fix for #1865, as something caused nodeos to die for the `trx_generator` to lose connection and core dump. This PR however should remove the red herring of the trx_generator core files in ci/cd.

### DIFF
--- a/tests/trx_generator/trx_provider.cpp
+++ b/tests/trx_generator/trx_provider.cpp
@@ -35,7 +35,10 @@ namespace eosio::testing {
    }
 
    void provider_connection::init_and_connect() {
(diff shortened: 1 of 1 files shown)


spoonincode:
:+1: I believe seeing the same problem in #1970 failures

spoonincode (approved):

linh2931 (approved):

PR #2007, opened 2023-12-20
[5.0] Fix unlinkable block exceptions when block-log-retain-blocks = 0

### LINKED ISSUE #2006: When setting block-log-retain-blocks = 0 get lots of unlinkable blocks errors
In 5.0.0-rc3, when setting `block-log-retain-blocks = 0` get lots of unlinkable blocks errors.  This worked fine in 3.2.  Not tested in 4.0

Easy to reproduce on a heavy blockchain like wax, when starting from a snapshot. Seen on other lighter chains as well, even when starting from up-to-date.

heifner:
When running with `block-log-retain-blocks = 0`, many `unlinkable block` exceptions were generated.

Do not attempt to apply blocks that are before LIB. 

Blocks are received before LIB because we always sync from LIB to make sure we see all possible forks. However, between the request and the blocks coming in, LIB can move. When running with no block log, these blocks are not known to have already been processed. However, there is never any reason to attempt to apply blocks that are <= LIB, so just drop them instead of attempting to apply them.

Resolves #2006

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -3062,7 +3062,7 @@ namespace eosio {
                  ("h", my_impl->get_chain_head_num()));
       if( !my_impl->sync_master->syncing_from_peer() ) { // guard against peer thinking it needs to send us old blocks
          uint32_t lib_num = my_impl->get_chain_lib_num();
-         if( blk_num < lib_num ) {
+         if( blk_num <= lib_num ) {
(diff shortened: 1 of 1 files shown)


linh2931 (approved):

spoonincode (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: BUG
summary: Account for LIB movement between request and receiving blocks by dropping blocks before LIB.
Note:end

PR #2008, opened 2023-12-20
[5.0] Test: Check for unlinkable blocks while syncing

### LINKED ISSUE #2006: When setting block-log-retain-blocks = 0 get lots of unlinkable blocks errors

heifner:
Update `nodeos_startup_catchup.py` test to sync with `--block-log-retain-blocks=0` and verify no `unlinkable` blocks.
Test fails before #2007 fix.

Resolves #2006

### DIFF
--- a/tests/nodeos_startup_catchup.py
+++ b/tests/nodeos_startup_catchup.py
@@ -170,7 +170,10 @@ try:
             waitForBlock(node0, catchupHead+5, timeout=twoRoundsTimeout*2, blockType=BlockType.lib)
 
         Print("Restart catchup node")
-        catchupNode.relaunch()
+        addSwapFlags = None
+        if catchup_num % 3 == 0:
+            addSwapFlags = {"--block-log-retain-blocks": "0", "--delete-all": ""}
+        catchupNode.relaunch(skipGenesis=False, addSwapFlags=addSwapFlags)
         waitForNodeStarted(catchupNode)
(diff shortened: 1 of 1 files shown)


tests/nodeos_startup_catchup.py
```diff
@@ -189,6 +192,12 @@ def waitForNodeStarted(node):
         catchupNode.interruptAndVerifyExitStatus(60)
         catchupNode.popenProc=None
 
+        logFile = Utils.getNodeDataDir(catchupNodeNum) + "/stderr.txt"
+        f = open(logFile)
+        contents = f.read()
+        if contents.count("unlinkable_block_exception") > 3: # a few are fine
```
linh2931, line 198: Maybe you can even make it to 10 to prevent test flakeness.

linh2931 (approved):

tests/nodeos_startup_catchup.py
heifner: Lets see how it goes. If we see flakiness then we can adjust up.

greg7mdp (approved):

PR #2029, opened 2024-01-03
[5.0] P2P: Pause net_plugin during snapshot write

### LINKED ISSUE #1128: P2P: stop nodeos from syncing blocks from the network while snapshot in progress
If you run nodeos with `read-mode = irreversible` the snapshot process can start immediately when requested. (when running in `read-mode = head` it needs to wait for LIB first).

Once nodeos is at LIB it can start generating the snapshot. However it keeps syncing from the network. All this syncing from the network uses up RAM. When snapshots take > 40 minutes to generate (eg WAX), then the amount of memory used can be significant. On servers where there is insufficient RAM, this means swapping which means the snapshot takes longer, which means more RAM, ...

This is very obvious if you start from 1 day behind live blocks. Memory keeps piling up while snapshot is being generated when blocks can be synced quickly. Example chart. Every bump in the chart is a new snapshot being generated (every 3 hours of blocks time).

image

Further, this extra memory doesn't get released ever.

Compare BP node with snapshot node:

image

### LINKED ISSUE #2021: When taking a snapshot RAM usage increases significantly on 5.0.0-rc3 WAX
When taking a snapshot RAM usage increases significantly on 5.0.0-rc3.  Screenshot is from WAX, starting at ~23:10.  snapshot is started from producer plugin when the blockchain is up-to-date.  Previous snapshots (earlier in the screenshot) were taken while the chain was not up-to-date and we didn't see the same problem.

image

Seems 5.0.0 behaviour is different than earlier versions: https://github.com/AntelopeIO/leap/issues/1128

Note also high CPU usage.

heifner:
Pause `net_plugin` while generating snapshot. With the threads of the `net_plugin` continuing to run they queue up many blocks to be processed consuming large amounts of RAM.

Verification testing in progress.

Resolves #2021
Resolves #1128

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -251,6 +251,7 @@ struct controller_impl {
    named_thread_pool<chain>        thread_pool;
    deep_mind_handler*              deep_mind_logger = nullptr;
    bool                            okay_to_print_integrity_hash_on_stop = false;
+   std::atomic<bool>               writing_snapshot = false;
 
    thread_local static platform_timer timer; // a copy for main thread and each read-only thread
 #if defined(EOSIO_EOS_VM_RUNTIME_ENABLED) || defined(EOSIO_EOS_VM_JIT_RUNTIME_ENABLED)
@@ -3246,7 +3247,18 @@ fc::sha256 controller::calculate_integrity_hash() { try {
 
 void controller::write_snapshot( const snapshot_writer_ptr& snapshot ) {
    EOS_ASSERT( !my->pending, block_validate_exception, "cannot take a consistent snapshot with a pending block" );
-   return my->add_to_snapshot(snapshot);
+   my->writing_snapshot = true;
+   try {
+      my->add_to_snapshot(snapshot);
+      my->writing_snapshot = false;
+   } catch(...) {
+      my->writing_snapshot = false;
+      throw;
+   }
+}
+
+bool controller::is_writing_snapshot() const {
+   return my->writing_snapshot;
 }
 
 int64_t controller::set_proposed_producers( vector<producer_authority> producers ) {
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3247,18 +3247,15 @@ fc::sha256 controller::calculate_integrity_hash() { try {
 
 void controller::write_snapshot( const snapshot_writer_ptr& snapshot ) {
    EOS_ASSERT( !my->pending, block_validate_exception, "cannot take a consistent snapshot with a pending block" );
-   my->writing_snapshot = true;
-   try {
-      my->add_to_snapshot(snapshot);
-      my->writing_snapshot = false;
-   } catch(...) {
-      my->writing_snapshot = false;
-      throw;
-   }
+   my->writing_snapshot.store(true, std::memory_order_release);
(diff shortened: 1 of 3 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -537,6 +542,28 @@ namespace eosio {
...
+            return;
+
+         std::unique_lock l(pause_mtx);
+         paused = true;
+         while (!pause_cv.wait_for(l, std::chrono::milliseconds(10), [&]{ return !paused; } )) {
+            if (!cc.is_writing_snapshot()) {
+               paused = false;
+            }
+         }
+      }
```
heifner, line 557: It occurred to me that this could just be:
```cpp
      void wait_if_paused() {
         controller& cc = chain_plug->chain();
         if (cc.is_writing_snapshot()) {
            std::this_thread::sleep_for( 10ms );
         }
      }
```
And not use a `condition_variable` at all. Let me know if you prefer the simpler implementation.

spoonincode: Yes I would vote for the simple sleep loop in this case since it's already doing something effectively the same via `wait_for()` anyways. If we're going to add any complexity beyond a simple sleep loop it might be best as a signal from controller when a snapshot is complete so we don't need to loop (not suggesting we do that).

libraries/chain/controller.cpp
```diff
@@ -3246,7 +3247,18 @@ fc::sha256 controller::calculate_integrity_hash() { try {
...
+      my->add_to_snapshot(snapshot);
+      my->writing_snapshot = false;
+   } catch(...) {
+      my->writing_snapshot = false;
+      throw;
+   }
+}
+
+bool controller::is_writing_snapshot() const {
+   return my->writing_snapshot;
```
spoonincode, line 3261: maybe consider release/acquire on access of this atomic bool?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -2897,6 +2924,8 @@ namespace eosio {
             return;
          }
 
+         my_impl->wait_if_paused();
```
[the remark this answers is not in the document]
heifner: I think better after the max_write_queue_size check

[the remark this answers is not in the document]
heifner: Correct.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -537,6 +537,14 @@ namespace eosio {
       void start_expire_timer();
       void start_monitors();
 
+      // we currently pause on snapshot generation
+      void wait_if_paused() {
```
spoonincode, line 541: can be const now

spoonincode (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: PERFORMANCE
category: INTERNALS
summary: Pause syncing with peers during snapshots. Eliminates the need to queue blocks, which memory consumption and reduces snapshot time by over 10X. 
Note:end

plugins/net_plugin/net_plugin.cpp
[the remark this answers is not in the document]
heifner: Turns out, that disconnect is less than ideal. See https://github.com/AntelopeIO/spring/issues/592

PR #1074, opened 2023-04-24
refactor prometheus plugin

huangminghuang:
This PR

- refactor prometheus plugin to avoid recreating prometheus objects on every update,
- replace
  *  num_peers -> p2p_connections {direction="out"}
  * num_client -> p2p_connections {direction="in"}

- replace the following metrics with the sum of observed values so far to fully reflect the states in nodeos regardless of scrap rate
  * dropped_trxs -> dropped_connections_total
  * unapplied_transactions -> unapplied_transactions_total
  * blacklisted_transactions -> blacklisted_transactions_total
  * subjective_bill_account_size -> subjective_bill_account_size_total
  * scheduled_trxs -> scheduled_trxs_total
  * blocks_produced -> blocks_produced_total
  * trxs_produced -> trxs_produced_total
- add new metrics
  * blocks_incoming
  * trxs_incoming_total
  * cpu_usage_us
  * net_usage_us
- provides the explanations to the meanings of all metrics

PR #2038, opened 2024-01-04
[5.0] only register prometheus handlers when `prometheus_plugin` enabled; fixing memory leak

### LINKED ISSUE #2030: Investigate recent surge in growth rate of system RAM usage
We've seen a recent increase in the growth rate of system RAM usage that seemed to correspond with upgrading to Leap v5.0.0.

This shows a 90 day period for a public api server that was upgraded to Leap v5.0.0:
90 day public api v5.0.5

This shows the same 90 day period for a public api server that is still running Leap v4.0.4:
90 day public api v4.0.4

Note the difference in the rate of increase of system RAM usage between 5.0 and 4.0.

A different node operator provided this data:
Matthew
_("eos-bp24" is running 4.0, "eos-peer72" is running 5.0.)_

Note that the rate of increase in system RAM usage is not significantly different between 5.0 and 4.0.

So, we have 2 contrasting data points. Both show that the rate of growth of system RAM usage has recently increased. One suggests that it was caused by 5.0. The other suggests that it was NOT caused by 5.0.

spoonincode:
appbase plugins are constructed no matter if they are enabled or not. This meant no matter whether the `prometheus_plugin` was enabled it would install its handlers in to `http_plugin`, `net_plugin`, etc. These handlers always `post()` to `prometheus_plugin`'s strand. But when `prometheus_plugin` isn't enabled its strand is never serviced leading to an endlessly growing queue that is never drained. Each HTTP request would add on to this queue, for example.

Move the handler registration to `plugin_initialize()` so handlers will only be registered when the plugin is enabled ergo when the strand will be pumped. (this is actually how it was prior to #1074, which makes me wonder if there was a need for this change, but I am not immediately seeing one)

Resolves #2030

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -4597,7 +4597,6 @@ namespace eosio {
    // called from any thread
    void connections_manager::start_conn_timers() {
       start_conn_timer(connector_period, {}, timer_type::check); // this locks mutex
-      start_conn_timer(connector_period, {}, timer_type::stats); // this locks mutex
       if (update_p2p_connection_metrics) {
          start_conn_timer(connector_period + connector_period / 2, {}, timer_type::stats); // this locks mutex
       }
--- a/plugins/prometheus_plugin/prometheus_plugin.cpp
+++ b/plugins/prometheus_plugin/prometheus_plugin.cpp
@@ -18,13 +18,9 @@ namespace eosio {
    struct prometheus_plugin_impl {
 
       eosio::chain::named_thread_pool<struct prom> _prometheus_thread_pool;
-      boost::asio::io_context::strand              _prometheus_strand;
+      boost::asio::io_context::strand              _prometheus_strand{_prometheus_thread_pool.get_executor()};
       metrics::catalog_type                        _catalog;
       fc::microseconds                             _max_response_time_us;
-
-      prometheus_plugin_impl(): _prometheus_strand(_prometheus_thread_pool.get_executor()){ 
-         _catalog.register_update_handlers(_prometheus_strand);
-      }
    };
 
    prometheus_plugin::prometheus_plugin()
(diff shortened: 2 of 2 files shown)


linh2931 (approved):

greg7mdp (approved):

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4597,7 +4597,6 @@ namespace eosio {
    // called from any thread
    void connections_manager::start_conn_timers() {
       start_conn_timer(connector_period, {}, timer_type::check); // this locks mutex
-      start_conn_timer(connector_period, {}, timer_type::stats); // this locks mutex
```
spoonincode, line 4600: I wonder if this was mistakenly left in during some sort of refactor? `connection_statistics_monitor()` is clearly only valid when `update_p2p_connection_metrics` is a valid function (it even `assert()`s on it). I thought maybe `connection_statistics_monitor()` drives some other non-prometheus statistics when prometheus isn't enabled, but it doesn't seem so.

greg7mdp, line 4600: If this line is removed, won't it prevent the prometheus `net_plugin` stats from being updated, even if the prometheus plugin is enabled? 
Unless I missed some other piece of code calling `connections_manager::connection_statistics_monitor`, but I didn't find any.
Maybe we should just have that function return immediately if `update_p2p_connection_metrics` is empty?

heifner: See the next line below.

greg7mdp: Oh, I missed that, thanks Kevin.
I think it would be safer to check for `!!update_p2p_connection_metrics` inside `connection_statistics_monitor` rather than at the calling site, but I guess it works as is.

PR #2039, opened 2024-01-04
[5.0 -> main] only register prometheus handlers when `prometheus_plugin` enabled; fixing memory leak

### LINKED ISSUE #2030: Investigate recent surge in growth rate of system RAM usage

spoonincode:
main merge of #2038 to fix leak when prometheus not enabled; resolves #2030

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -4599,7 +4599,6 @@ namespace eosio {
    // called from any thread
    void connections_manager::start_conn_timers() {
       start_conn_timer(connector_period, {}, timer_type::check); // this locks mutex
-      start_conn_timer(connector_period, {}, timer_type::stats); // this locks mutex
       if (update_p2p_connection_metrics) {
          start_conn_timer(connector_period + connector_period / 2, {}, timer_type::stats); // this locks mutex
       }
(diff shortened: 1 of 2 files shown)


heifner (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Fix memory leak by moving initialization of http handler associated with app plugins. Prevents creation http requests that never get drained. This issue was only in leap v5-rc releases, never surfaced in a stable release.
Note:end

Issue #2022, opened 2023-12-29
5.0.0-rc3 WAX: Error on shutdown: corrupted size vs. prev_size


matthewdarwin:
This issue still exists: `corrupted size vs. prev_size`. Has been seen in many previous versions.  It is triggered on a normal clean shutdown. I don't know how often it happens because the state is written correctly and nodeos starts up.  So it's just a log message, but nothing bad happens....

```
Dec 29 00:13:41 wax-snap71c nodeos[2575092]: info  2023-12-29T00:13:41.514 nodeos    http_plugin.cpp:515           plugin_shutdown      ] exit shutdown
Dec 29 00:13:41 wax-snap71c nodeos[2575092]: info  2023-12-29T00:13:41.515 nodeos    main.cpp:155                  operator()           ] nodeos version v5.0.0wax01-rc3 v5.0.0wax01-rc3-10c04d06485c8a1fdb52f6036dfa90b93813cb3a-dirty
Dec 29 00:13:41 wax-snap71c nodeos[2575092]: info  2023-12-29T00:13:41.515 nodeos    main.cpp:62                   log_non_default_opti ] Non-default options:  [.........]
Dec 29 00:13:41 wax-snap71c nodeos[2575092]: corrupted size vs. prev_size
Dec 29 00:13:45 wax-snap71c nodeos[2593285]: nodeos still running
Dec 29 00:13:46 wax-snap71c snapshot[2554292]: get_info failed
Dec 29 00:13:46 wax-snap71c systemd[1]: nodeos.service: Main process exited, code=killed, status=6/ABRT
Dec 29 00:13:49 wax-snap71c snapshot[2554292]: get_info failed
Dec 29 00:13:50 wax-snap71c nodeos[2593285]: nodeos stopped
Dec 29 00:13:51 wax-snap71c systemd[1]: nodeos.service: Failed with result 'signal'.
Dec 29 00:13:51 wax-snap71c systemd[1]: Stopped nodeos.service - EOS Nation Nodeos.
Dec 29 00:13:51 wax-snap71c systemd[1]: nodeos.service: Consumed 8h 6min 43.460s CPU time.
```

matthewdarwin:
Today I got

```
nodeos[999210]: corrupted size vs. prev_size while consolidating
```

spoonincode:
Can you provide the core dump for these crashes?

matthewdarwin:
There is no core dump.. Core dumps are turned off in production systems.

[...]

PR #2042, opened 2024-01-05
[5.0] extend life of `http_plugin_state` to ensure no invalid memory accesses during nodeos shutdown

spoonincode:
During the destruction of `beast_http_session`,
https://github.com/AntelopeIO/leap/blob/04774eb7726ae95a6cb795b493fcf0f25021bc5f/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp#L270-L281
there are accesses to `plugin_state_` which is simply a reference to an object that lives in `http_plugin_impl`,
https://github.com/AntelopeIO/leap/blob/04774eb7726ae95a6cb795b493fcf0f25021bc5f/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp#L86
Additionally, notice that `beast_http_session`'s `socket_` has been constructed with the io_context from `http_plugin_impl::plugin_state`'s `thread_pool`. So destruction of the `socket_` will also require valid access to the `plugin_state` object.

But it is possible for a `beast_http_session` to outlive `http_plugin` which causes these accesses during destruction to be invalid. Consider a situation where appbase's executor (`priority_queue_executor`) has a queued callback holding the sole reference count of a `shared_ptr<beast_http_session>` and appbase is commanded to `quit()`. appbase's `exec()` upon exiting its run loop will then call appbase's `shutdown()`. `application_base::shutdown()` will then effectively call `http_plugin::plugin_shutdown()` (stopping http threads) and then destroy `http_plugin`. However, `priority_queue_executor` will still be keeping alive a `beast_http_session` via some pending callback. As appbase continues to destruct eventually `priority_queue_executor` is destroyed which destroys all pending callbacks. It's at that time the last reference to the `shared_ptr<beast_http_session>` is removed causing it to be destroyed and expecting access to `http_plugin_impl::plugin_state`.

But wait! When appbase's io_context stops it first clears its pending queue before `application_base::shutdown()`, right?
https://github.com/AntelopeIO/appbase/blob/b75b31e14f966fa3de6246e120dcba36c6ce5264/include/appbase/application_base.hpp#L148-L150
afaict this is racy: there is nothing to stop code in some other non-main thread (like http_plugin's threads!) from queueing more callbacks on to appbase's executor between the time `exec.clear()` is called and, say, `http_plugin::shutdown()` is called which halts its threads.

Not immediately seeing any good approaches to resolve this issue (especially in a patch release) beyond migrating `http_plugin_impl::plugin_state` to a `shared_ptr` so that every `beast_http_session` can keep it alive as long as it needs to. The `socket_` needing `http_plugin_impl::plugin_state`'s `thread_pool` to remain alive is seemingly what makes this non-negotiable without more significant refactoring.

I am considering this resolving #2022 because while I can't get the exact fatal malloc error to print, the failure I can reproduce here is at the same point in time.

Was unable to reproduce the issue in 3.2 or 4.0. Curiously, those versions already do have a `std::shared_ptr<http_plugin_state>` so maybe that's why.

### DIFF
--- a/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
@@ -69,6 +69,8 @@ std::string to_log_string(const T& req, size_t max_size = 1024) {
 template <class Socket>
 class beast_http_session : public detail::abstract_conn,
                            public std::enable_shared_from_this<beast_http_session<Socket>> {
+
+   std::shared_ptr<http_plugin_state> plugin_state_;
    Socket             socket_;
    api_category_set   categories_;
    beast::flat_buffer buffer_;
@@ -83,7 +85,6 @@ class beast_http_session : public detail::abstract_conn,
    // HTTP response object
    std::optional<http::response<http::string_body>> res_;
 
-   http_plugin_state& plugin_state_;
    std::string remote_endpoint_;
    std::string local_address_;
 
@@ -112,12 +113,12 @@ class beast_http_session : public detail::abstract_conn,
       res_->version(req.version());
       res_->set(http::field::content_type, "application/json");
       res_->keep_alive(req.keep_alive());
-      if(plugin_state_.server_header.size())
-         res_->set(http::field::server, plugin_state_.server_header);
+      if(plugin_state_->server_header.size())
+         res_->set(http::field::server, plugin_state_->server_header);
 
       // Request path must be absolute and not contain "..".
       if(req.target().empty() || req.target()[0] != '/' || req.target().find("..") != beast::string_view::npos) {
-         fc_dlog( plugin_state_.get_logger(), "Return bad_reqest:  ${target}",  ("target", std::string(req.target())) );
+         fc_dlog( plugin_state_->get_logger(), "Return bad_reqest:  ${target}",  ("target", std::string(req.target())) );
          error_results results{static_cast<uint16_t>(http::status::bad_request), "Illegal request-target"};
          send_response( fc::json::to_string( results, fc::time_point::maximum() ),
                         static_cast<unsigned int>(http::status::bad_request) );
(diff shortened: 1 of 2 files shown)


plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
```diff
@@ -251,13 +252,13 @@ class beast_http_session : public detail::abstract_conn,
    // shared_from_this() requires default constructor
    beast_http_session() = default;
 
-   beast_http_session(Socket&& socket, http_plugin_state& plugin_state, std::string remote_endpoint,
+   beast_http_session(Socket&& socket, std::shared_ptr<http_plugin_state> plugin_state, std::string remote_endpoint,
                       api_category_set categories, const std::string& local_address)
-       : socket_(std::move(socket)), categories_(categories), plugin_state_(plugin_state),
+       : plugin_state_(plugin_state), socket_(std::move(socket)), categories_(categories),
```
heifner, line 257: This can be `std::move(plugin_state)`

heifner (approved):

linh2931 (approved):
Like your detailed analysis.

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Extend life of `http_plugin_state` to ensure no invalid memory accesses during nodeos shutdown.
Note:end

plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
spoonincode: I usually always copy `shared_ptr`s, and it feels weird to move something in the initializer list that isn't (and can't be) moved in the ctor call,
https://github.com/AntelopeIO/leap/blob/5fcf5b49548e58e8bb01b50b6d7f2a9458602b0f/plugins/http_plugin/http_plugin.cpp#L230-L232

But the code generation is surprisingly better when adding the `move` here, so maybe it's worth doing

heifner: Move of `shared_ptr` is nice as it avoid the `std::atomic` operation I believe.

greg7mdp (approved):

heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: BUG
summary: Prevent error on shutdown by extending the life of `http_plugin_state` thereby ensuring no invalid memory accesses.
Note:end

Issue #2028, opened 2024-01-03
IF: use the same way to determine if instant finality is active


linh2931:
https://github.com/AntelopeIO/leap/pull/2024/files#r1440617405

The presence of `instant_finality_extension` should be used as an indication of instant finality activated. But doing that fails unit test producer_schedule_hs_tests/verify_producer_schedule_after_hotstuff_activation with

```
3030000 block_validate_exception: Block exception
invalid block transaction merkle root 0cb3d1d90b7b4d61b7db5cb3730bee81f98d339cc1ac5fb0c1b50ddc12e9e108 != b26dc703930e4f88b4dffa3a24af3a580009ffaa83a8d50e4bdbc397db729d8f
    {"b":"0cb3d1d90b7b4d61b7db5cb3730bee81f98d339cc1ac5fb0c1b50ddc12e9e108","c":"b26dc703930e4f88b4dffa3a24af3a580009ffaa83a8d50e4bdbc397db729d8f"}
    chain-1  controller.cpp:2226 create_block_state_i
rethrow
```

Additional work is required for the transition from DPOS to Instant Finality.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Unified implementation to detect activation of new faster finality. 
Note:end

PR #2043, opened 2024-01-05
IF: use the same way to determine if instant finality is active

linh2931:
https://github.com/AntelopeIO/leap/issues/2028 reports `transaction_mroot` difference between when it is first calculated and when it is recreated in `create_block_state_i`. The difference  is caused by the way how `hs_active` is determined for mroot calculation. When `transaction_mroot`  is first created, `hs_lib > 0` is used as an indication of `hs_active`; in `create_block_state_i`, only presence of `instant_finality_extension` is used.

This inconsistency causes integration test failures too.

This PR changes to use  the same `hs_lib > 0` as the indication.

Resolved https://github.com/AntelopeIO/leap/issues/2028

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -2799,8 +2799,13 @@ struct controller_impl {
       uint32_t hs_lib = hs_irreversible_block_num.load();
       const bool hs_active = hs_lib > 0;
       auto trx_mroot = calculate_trx_merkle( b->transactions, hs_active );
-      EOS_ASSERT( b->transaction_mroot == trx_mroot, block_validate_exception,
-                  "invalid block transaction merkle root ${b} != ${c}", ("b", b->transaction_mroot)("c", trx_mroot) );
+      if( b->transaction_mroot != trx_mroot ) {
+         // Call of create_block_state_i can happen right before hs_irreversible_block_num
+         // is set. Fall back to verify in the other way.
+         trx_mroot = calculate_trx_merkle( b->transactions, !hs_active );
+         EOS_ASSERT( b->transaction_mroot == trx_mroot, block_validate_exception,
+                     "invalid block transaction merkle root ${b} != ${c}", ("b", b->transaction_mroot)("c", trx_mroot) );
+      }
 
       const bool skip_validate_signee = false;
       auto bsp = std::make_shared<block_state_legacy>(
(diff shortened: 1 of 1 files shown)


libraries/chain/controller.cpp
```diff
@@ -2796,16 +2796,8 @@ struct controller_impl {
...
-      if (!b->header_extensions.empty()) {
-         std::optional<block_header_extension> instant_finality_ext = b->extract_header_extension(instant_finality_extension::extension_id());
-#warning change to use instant_finality_ext https://github.com/AntelopeIO/leap/issues/1508
-         if (instant_finality_ext) {
-            const auto& ext = std::get<instant_finality_extension>(*instant_finality_ext);
-            hs_active = !!ext.new_proposer_policy;
-         }
-      }
-
+      uint32_t hs_lib = hs_irreversible_block_num.load();
```
heifner, line 2799: We need a better way to do this. We need to make this determination via `prev` as timing of this call can happen right before `hs_irreversible_block_num` is set. We could just live with that as validation will fail, but it would fall back to trying again.

linh2931: How can `prev` be used to tell if current block `b` is constructed under instant finality. I think the fall back way is better -- if validation fail, we try the other way.

libraries/chain/controller.cpp
```diff
@@ -2796,19 +2796,16 @@ struct controller_impl {
...
       auto trx_mroot = calculate_trx_merkle( b->transactions, hs_active );
-      EOS_ASSERT( b->transaction_mroot == trx_mroot, block_validate_exception,
-                  "invalid block transaction merkle root ${b} != ${c}", ("b", b->transaction_mroot)("c", trx_mroot) );
+      if( b->transaction_mroot != trx_mroot ) {
+         // Call of create_block_state_i can happen right before hs_irreversible_block_num
+         // is set. Fall back to verify in the other way.
+         trx_mroot = calculate_trx_merkle( b->transactions, !hs_active );
+         EOS_ASSERT( b->transaction_mroot == trx_mroot, block_validate_exception,
+                     "invalid block transaction merkle root ${b} != ${c}", ("b", b->transaction_mroot)("c", trx_mroot) );
+      }
```
heifner, line 2808: This opens up an attack vector as you can create a block with invalid mroot that will pass this validation. I think we can go with your first implementation. There is a small race condition at time of activation, but even if that is tripped and the block is considered invalid, the node will attempt to sync the block in the future and succeed; that is what I meant by re-try.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Unified implementation to detect activation of new faster finality. 
Note:end

libraries/chain/controller.cpp
linh2931: Thanks for your insight!

heifner (approved):

greg7mdp (approved):

Issue #2044, opened 2024-01-05
IF: remove unneeded files


linh2931:
* `finalizer_policy.cpp` only exists for the following reason, which is no longer applicable
```
  /**
    * These definitions are all here to avoid including bls_public_key.hpp which includes <bls12-   381/bls12-381.hpp>
    * and pulls in bls12-381 types. This keeps bls12-381 out of libtester.
    */
```
* `proposal_info.hpp` was for `proposal_info_extension` which has been folded into `instant_finality_extension`.

Remove them from code base to have less cluters.

PR #2049, opened 2024-01-05
IF: Remove unneeded files

linh2931:
* `finalizer_policy.cpp` only exists for the following reason, which is no longer applicable
```
/**

These definitions are all here to avoid including bls_public_key.hpp which includes <bls12- 381/bls12-381.hpp>
and pulls in bls12-381 types. This keeps bls12-381 out of libtester.
*/
```
* `proposal_info.hpp` was for `proposal_info_extension` which has been folded into `instant_finality_extension`.

Remove them from code base to have less clutters.

Resolved https://github.com/AntelopeIO/leap/issues/2044


heifner (approved):

greg7mdp:
I seem to remember that @heifner had created this on purpose to avoid including the bls12 headers somewhere (was it contracts?).

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: CHORE
summary: Consolidation and refactoring created some un-needed files. This PR removed those un-needed files.  
Note:end

PR #2052, opened 2024-01-07
IF: Changes in controller to complete unification

### LINKED ISSUE #2034: IF: Unification: Changes in controller to complete unification
Finish data structures in `building_block` and the `block_header_state_input` needed to transition to `assembled_block`. Prioritize this first as the first PR.

Then complete changes to the following in controller to work post-IF transition, including (but not necessarily limited to):
* building_block, assembled_block, completed_block structures
* apply_block function
* push_block function

greg7mdp:
Will eventually address #2034 .

This is a work in progress... but still would like to check it in to unblock Lin and Kevin.


libraries/chain/controller.cpp
```diff
@@ -416,7 +571,7 @@ struct building_block {
    bool is_dpos() const { return std::holds_alternative<building_block_dpos>(v); }
 
    // if constructor
-   building_block(const block_header_state& prev, const building_block_input& bbi) :
+   building_block(const block_header_state& prev, const building_block_input bbi) :
```
heifner, line 574: Looks like removed by accident

greg7mdp: Right, thanks Kevin, fixing.

heifner (approved):

heifner (approved):

libraries/chain/block_header_state.cpp
```diff
@@ -1,210 +1,95 @@
...
-         result.producer_to_last_produced                           = producer_to_last_produced;
-         result.producer_to_last_produced[proauth.producer_name]    = result.block_num;
-         result.producer_to_last_implied_irb                        = producer_to_last_implied_irb;
-         result.producer_to_last_implied_irb[proauth.producer_name] = dpos_proposed_irreversible_blocknum;
-      }
-   } // !hotstuff_activated
-#endif
    
+   emplace_extension(result.header.header_extensions, instant_finality_extension::extension_id(),
+                     fc::raw::pack(instant_finality_extension{qc_info,
```
linh2931, line 89: std::move?

linh2931 (approved):

libraries/chain/block_header_state.cpp
greg7mdp: The `qc_info` is just an `int` and a `bool`, so `std::move` doesn't make a difference, which is why I didn't add it.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Refactor of controller to support both Faster Finality and Legacy Block Finalization.
Note: end

PR #2054, opened 2024-01-08
IF: Fix double confirming test failure

heifner:
Reference contracts: https://github.com/AntelopeIO/reference-contracts eosio system tests were failing do a double confirmation error. Port of tests from reference contracts to provide same coverage in leap.

Revert unneeded changes in `block_header_state_legacy`  since it no longer contains instant finality logic.
Add `controller::head_active_producers()` needed by `tester`.


libraries/chain/block_header_state_legacy.cpp
```diff
@@ -107,37 +92,39 @@ namespace eosio { namespace chain {
       int32_t i = (int32_t)(result.confirm_count.size() - 1);
       uint32_t blocks_to_confirm = num_prev_blocks_to_confirm + 1; /// confirm the head block too
       while( i >= 0 && blocks_to_confirm ) {
-         --result.confirm_count[i];
-         //idump((confirm_count[i]));
-         if( result.confirm_count[i] == 0 )
-         {
+        --result.confirm_count[i];
```
linh2931, line 95: Why is the space removed?

linh2931 (approved):

libraries/chain/block_header_state_legacy.cpp
heifner: Just copied over the 5.0 impl, including spaces

libraries/chain/block_header_state.cpp
```diff
@@ -4,6 +4,16 @@
...
+#warning Add last_proposed_finalizer_policy_generation to snapshot_block_header_state_v3, see header file TODO
+
+   namespace detail {
+      uint32_t get_next_next_round_block_num( block_timestamp_type t, uint32_t block_num ) {
+         auto index = t.slot % config::producer_repetitions; // current index in current round
+         //                 (increment to the end of this round  ) + next round
+         return block_num + (config::producer_repetitions - index) + config::producer_repetitions;
+      }
+   }
+
```
greg7mdp, line 16: I moved this in a new file in my PR so as not to duplicate, however we can checkin as-is and I'll remove in my PR.

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Better support for system contracts by consolidating confirmations and exposing `head_active_producers`
Note: end

PR #2066, opened 2024-01-10
IF: Update signature-provider parsing for base64 BLS public keys

### LINKED ISSUE #2060: IF: signature-provider for BLS keys
Currently `signature-provider` for BLS keys are specified as:
```
--signature-provider PUB_BLS_PAD80INyOVD3K0KJdl/7b5+UQGT0nrr8MDxO0tB80HMmaQS+X8YWRrYyyb7slAYFbOV69KN/skAFJfIFgdEEvD00MNhyaSJaJ1NGJjdKfq+OARJKoMBoBvj5/78n6BgLDpoJ7A===KEY:PVT_BLS_cvOsv1fAUDVgcK/J1j3U1DfQWMrmSFZDMiu4E0yJCEDWWZl3
```
The separator between the public key and private key is `=KEY:`. This can cause confusion since the base64 encoded public key also often ends in `=`.  This follows the current separator between EOS keys:
```
--signature-provider EOS5VLjLL9nY5b2QymL8QVpwRnUDEhaedeHvVJ4vU8tpaEbxAwDc3=KEY:5JpgeNa7fAf4mXukVHmnNPqcN7qHbrp4NkMJQp32c1MbrLLdwYz
```
Consider a different separator for BLS keys provided to `signature-provider`.

Also the current implementation finds the first `=` and uses it to separate out the public key from the private key. This is clearly wrong for base64 encoded keys.
https://github.com/AntelopeIO/leap/blob/hotstuff_integration/plugins/signature_provider_plugin/signature_provider_plugin.cpp#L43

If the existing separator of `=KEY:` is maintained the parsing needs to be fixed.

Currently malformed signature-provider specifications do not stop nodeos. It ignores the error and continues. Seems this should stop nodeos from starting.
```
error 2024-01-09T22:46:19.780 nodeos    producer_plugin.cpp:1143      plugin_initialize    ] Malformed signature provider: "PUB_BLS_PAD80INyOVD3K0KJdl/7b5+UQGT0nrr8MDxO0tB80HMmaQS+X8YWRrYyyb7slAYFbOV69KN/skAFJfIFgdEEvD00MNhyaSJaJ1NGJjdKfq+OARJKoMBoBvj5/78n6BgLDpoJ7A===KEY:PVT_BLS_cvOsv1fAUDVgcK/J1j3U1DfQWMrmSFZDMiu4E0yJCEDWWZl3": {"code":10,"name":"assert_exception","message":"Assert Exception","stack":[{"context":{"level":"error","file":"common.hpp","line":61,"method":"apply","hostname":"","thread_name":"nodeos","timestamp":"2024-01-09T22:46:19.780"},"format":"false: No matching suite type for ${prefix}_${data}","data":{"prefix":"BLS","data":"PAD80INyOVD3K0KJdl/7b5+UQGT0nrr8MDxO0tB80HMmaQS+X8YWRrYyyb7slAYFbOV69KN/skAFJfIFgdEEvD00MNhyaSJaJ1NGJjdKfq+OARJKoMBoBvj5/78n6BgLDpoJ7A"}}]}, ignoring!
```

heifner:
BLS public keys are base64 encoded which include trailing `=`. Update the `signature-provider` parsing to expect these "extra" `=`s. 
Shutdown on startup for malformed `signature-provider` arguments. Previously nodeos would log an error and continue.

Resolves #2060

### DIFF
--- a/plugins/signature_provider_plugin/signature_provider_plugin.cpp
+++ b/plugins/signature_provider_plugin/signature_provider_plugin.cpp
@@ -42,6 +42,9 @@ class signature_provider_plugin_impl {
       std::tuple<std::string, std::string, std::string> parse_spec(const std::string& spec) const {
          auto delim = spec.find("=");
          EOS_ASSERT(delim != std::string::npos, chain::plugin_config_exception, "Missing \"=\" in the key spec pair");
+         // public_key can be base64 encoded with trailing `=`
+         while( spec.size() > delim+1 && spec[delim+1] == '=' )
+            ++delim;
          auto pub_key_str = spec.substr(0, delim);
          auto spec_str = spec.substr(delim + 1);
 
--- a/plugins/signature_provider_plugin/signature_provider_plugin.cpp
+++ b/plugins/signature_provider_plugin/signature_provider_plugin.cpp
@@ -43,8 +43,10 @@ class signature_provider_plugin_impl {
          auto delim = spec.find("=");
          EOS_ASSERT(delim != std::string::npos, chain::plugin_config_exception, "Missing \"=\" in the key spec pair");
          // public_key can be base64 encoded with trailing `=`
+         // e.g. --signature-provider PUB_BLS_Fmgk<snip>iuA===KEY:PVT_BLS_NZhJ<snip>ZHFu
          while( spec.size() > delim+1 && spec[delim+1] == '=' )
             ++delim;
+         EOS_ASSERT(delim < spec.size() + 1, chain::plugin_config_exception, "Missing spec data in the key spec pair");
          auto pub_key_str = spec.substr(0, delim);
          auto spec_str = spec.substr(delim + 1);
 
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -1139,10 +1139,13 @@ void producer_plugin_impl::plugin_initialize(const boost::program_options::varia
             }
          } catch(secure_enclave_exception& e) {
             elog("Error with Secure Enclave signature provider: ${e}; ignoring ${val}", ("e", e.top_message())("val", key_spec_pair));
+            throw;
          } catch (fc::exception& e) {
             elog("Malformed signature provider: \"${val}\": ${e}, ignoring!", ("val", key_spec_pair)("e", e));
+            throw;
          } catch (...) {
             elog("Malformed signature provider: \"${val}\", ignoring!", ("val", key_spec_pair));
+            throw;
          }
       }
    }
--- a/tests/TestHarness/launcher.py
+++ b/tests/TestHarness/launcher.py
@@ -514,7 +514,7 @@ class cluster_generator:
             a(a(eosdcmd, '--plugin'), 'eosio::producer_plugin')
             producer_keys = list(sum([('--signature-provider', f'{key.pubkey}=KEY:{key.privkey}') for key in instance.keys], ()))
             eosdcmd.extend(producer_keys)
-            finalizer_keys = list(sum([('--signature-provider', f'{key.blspubkey}=KEY:{key.blsprivkey}') for key in instance.keys], ()))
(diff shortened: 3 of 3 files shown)


plugins/signature_provider_plugin/signature_provider_plugin.cpp
```diff
@@ -42,6 +42,9 @@ class signature_provider_plugin_impl {
       std::tuple<std::string, std::string, std::string> parse_spec(const std::string& spec) const {
          auto delim = spec.find("=");
          EOS_ASSERT(delim != std::string::npos, chain::plugin_config_exception, "Missing \"=\" in the key spec pair");
+         // public_key can be base64 encoded with trailing `=`
+         while( spec.size() > delim+1 && spec[delim+1] == '=' )
+            ++delim;
          auto pub_key_str = spec.substr(0, delim);
          auto spec_str = spec.substr(delim + 1);
```
linh2931, line 49: Will this be out of range if `=` is the last char in `spec` and `delim` points to `=`?

linh2931 (approved):
Later tests can be added.

plugins/signature_provider_plugin/signature_provider_plugin.cpp
```diff
@@ -42,6 +42,9 @@ class signature_provider_plugin_impl {
       std::tuple<std::string, std::string, std::string> parse_spec(const std::string& spec) const {
          auto delim = spec.find("=");
          EOS_ASSERT(delim != std::string::npos, chain::plugin_config_exception, "Missing \"=\" in the key spec pair");
+         // public_key can be base64 encoded with trailing `=`
```
greg7mdp, line 45: Can you add in the comment what `spec` looks like when the public_key has trailing `=`?

plugins/signature_provider_plugin/signature_provider_plugin.cpp
```diff
@@ -42,6 +42,11 @@ class signature_provider_plugin_impl {
...
+         // e.g. --signature-provider PUB_BLS_Fmgk<snip>iuA===KEY:PVT_BLS_NZhJ<snip>ZHFu
+         while( spec.size() > delim+1 && spec[delim+1] == '=' )
+            ++delim;
+         EOS_ASSERT(delim < spec.size() + 1, chain::plugin_config_exception, "Missing spec data in the key spec pair");
          auto pub_key_str = spec.substr(0, delim);
```
greg7mdp, line 50: won't `pub_key_str` potentially end with multiple `=` characters? Maybe it is not an issue?

heifner: Yes, which like the example comment shows is rather ugly. We will likely move to base64url in the future partly do to this.

greg7mdp: So should we do something like this:
```
         auto delim_end = delim;
         while( spec.size() > delim_end+1 && spec[delim_end+1] == '=' )
            ++delim_end;
         EOS_ASSERT(delim_end < spec.size() + 1, chain::plugin_config_exception, "Missing spec data in the key spec pair");
         auto pub_key_str = spec.substr(0, delim);
         auto spec_str = spec.substr(delim_end + 1);
```

heifner: No, The public key is `PUB_BLS_Fmgk<snip>iuA==` the separator is `=KEY:`. Also note the separator can also be `=KEOSD:` and `=SE:`.

greg7mdp: Oh, OK. No spaces possible, right?

greg7mdp (approved):

plugins/signature_provider_plugin/signature_provider_plugin.cpp
heifner: Correct, no spaces allowed.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Better support for BLS signatures by expecting and dealing correctly with '='. In addition, nodeos will shutdown during a startup, when invalid `signature-provider` arguments are detected.  
Note: end

Issue #2046, opened 2024-01-05
IF: Unification: Integrate vote message signatures into new fork database


arhag:
Vote messages received from net plugin should be processed by some code in libchain, which can be included in source code within the `finality` directory (renamed from `hotstuff`) within `chain`. This processing involves aggregating the vote signature into the pending QC for the block the vote corresponds to. This pending QC lives within the `block_state` kept within the new fork database.

PR #2067, opened 2024-01-10
IF:Unification: Integrate vote message signatures into new fork database

linh2931:
Process vote messages received from net plugin:
* added a new `finality_controller.cpp` within `hotstuff` directory (which will be renamed to `finality` directory) to contain control related code
* aggregated the vote signature into the pending QC for the block the vote corresponds to
* added tests

Resolved https://github.com/AntelopeIO/leap/issues/2046


libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -22,10 +22,14 @@ inline std::vector<uint32_t> bitset_to_vector(const hs_bitset& bs) {
...
 bool pending_quorum_certificate::votes_t::add_vote(const std::vector<uint8_t>& proposal_digest, size_t index,
                                                    const bls_public_key& pubkey, const bls_signature& new_sig) {
-   if (_bitset[index])
+   if (_bitset[index]) {
+      wlog( "finalizer ${i} has already voted", ("i", index) );
       return false; // shouldn't be already present
-   if (!fc::crypto::blslib::verify(pubkey, proposal_digest, new_sig))
+   }
+   if (!fc::crypto::blslib::verify(pubkey, proposal_digest, new_sig)) {
+      wlog( "sinature from finalizer ${i} cannot be verified", ("i", index) );
```
greg7mdp, line 30:
```suggestion
      wlog( "signature from finalizer ${i} cannot be verified", ("i", index) );
```

greg7mdp (commented):
I'm not understanding why we need `finality_controller`? It holds no data. 
Couldn't we just have a function in `block_state`, for example `block_state::aggregate_vote(const hs_vote_message& vote)`?

libraries/chain/controller.cpp
```diff
@@ -4007,8 +4020,13 @@ void controller::get_finalizer_state( finalizer_state& fs ) const {
 }
 
 // called from net threads
-void controller::notify_hs_message( const uint32_t connection_id, const hs_message& msg ) {
-   my->pacemaker->on_hs_msg(connection_id, msg);
+void controller::notify_vote_message( const hs_vote_message& vote ) {
+   auto bsp = fetch_block_state_by_id(vote.proposal_id);
+   if( bsp ) {
+      my->finality_control.aggregate_vote(bsp, vote);
+   } else {
```
greg7mdp, line 4027: Instead of adding `fetch_block_state_by_id` and `apply_block_data_new`, I suggest you add  a new method in `block_data_t` named:

`void aggregate_vote(const hs_vote_message& vote)` which would do nothing for the `dpos` case, and would fetch the block_state and apply the vote in the `if` case.

greg7mdp (changes requested):

linh2931:
> I'm not understanding why we need `finality_controller`? It holds no data. Couldn't we just have a function in `block_state`, for example `block_state::aggregate_vote(const hs_vote_message& vote)`?

We were talking about consolidate `hoststuff` directory into a new `finality` directory. I created `finality_controller` as a placeholder for that purpose. But we could do that in the future if needed. Will move to `block_state` for simplicity.

libraries/chain/controller.cpp
linh2931: Is `block_data_t` a general class and its main purpose to service fork_db APIs? `aggregate_vote` seems too specific for `block_data_t`.

greg7mdp: Block data and the other structs with variants should contain, as much as possible, any code that does something different for `dpos` and `if`, or accesses the block_state_* structs, so that the rest of controller doesn't see these differences.

It is perfectly fine to add new members like `aggregate_vote` which behave differently for `dpos` and `if`. 

The `apply` functions were mostly present to make controller work in the `dpos` mode while changes were not completed, ideally they should go away.

libraries/chain/controller.cpp
```diff
@@ -3889,7 +3902,7 @@ std::optional<signed_block_header> controller::fetch_block_header_by_number( uin
 
 
 block_state_legacy_ptr controller::fetch_block_state_by_id( block_id_type id )const {
-   // returns nullptr when in IF mode 
+   // returns nullptr when in in mode
```
greg7mdp, line 3905: ??

libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -134,10 +138,10 @@ bool pending_quorum_certificate::add_weak_vote(const std::vector<uint8_t>& propo
    return true;
 }
 
-bool pending_quorum_certificate::add_vote(bool strong, const std::vector<uint8_t>& proposal_digest, size_t index,
+bool pending_quorum_certificate::add_vote(bool strong, size_t index,
                                           const bls_public_key& pubkey, const bls_signature& sig) {
-   return strong ? add_strong_vote(proposal_digest, index, pubkey, sig)
-                 : add_weak_vote(proposal_digest, index, pubkey, sig);
+   return strong ? add_strong_vote(index, pubkey, sig)
+                 : add_weak_vote(index, pubkey, sig);
```
greg7mdp, line 144: as discussed in the meeting, you'll have to pass the `finalizer_digest` from the `block_state`, and we can remove the `_proposal_digest member`

greg7mdp (changes requested):

linh2931:
Replaced by simpler https://github.com/AntelopeIO/leap/pull/2073

PR #2073, opened 2024-01-11
IF:Unification: Process received Vote message

linh2931:
Process vote messages received from net plugin:
* aggregated the vote signature into the pending QC for the block the vote corresponds to
* added tests

Changing `proposal_id` to `block_id` in vote message and removing `proposal_digest` from pending_quorum_certificate will be done after unification as they should be part of the bigger cleaning up effort: https://github.com/AntelopeIO/leap/issues/2072

Resolved https://github.com/AntelopeIO/leap/issues/2046

Replaced https://github.com/AntelopeIO/leap/pull/2067

### DIFF
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -103,6 +103,7 @@ namespace eosio::chain {
    {}
 #endif
 
+   // Called from net threads
    bool block_state::aggregate_vote(const hs_vote_message& vote) {
       const auto& finalizers = finalizer_policy->finalizers;
       auto it = std::find_if(finalizers.begin(),
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -102,5 +102,32 @@ namespace eosio::chain {
    ,_cached_trxs( std::move(trx_metas) )
    {}
 #endif
+
+   bool block_state::aggregate_vote(const hs_vote_message& vote) {
+      const auto& finalizers = finalizer_policy->finalizers;
+      auto it = std::find_if(finalizers.begin(),
+                             finalizers.end(),
+                             [&](const auto& finalizer) { return finalizer.public_key == vote.finalizer_key; });
+
+      if (it != finalizers.end()) {
+         auto index = std::distance(finalizers.begin(), it);
+         if( vote.strong ) {
+            std::vector<uint8_t> d(strong_finalizer_digest.data(), strong_finalizer_digest.data() + strong_finalizer_digest.data_size());
+            return pending_qc.add_strong_vote( d,
+                                               index,
+                                               vote.finalizer_key,
+                                               vote.sig );
+         } else {
+            std::vector<uint8_t> d(weak_finalizer_digest.data(), weak_finalizer_digest.data() + weak_finalizer_digest.data_size());
+            return pending_qc.add_weak_vote( d,
+                                             index,
+                                             vote.finalizer_key,
+                                             vote.sig );
+         }
+      } else {
+         wlog( "finalizer_key (${k}) in vote is not in finalizer policy", ("k", vote.finalizer_key) );
+         return false;
+      }
+   }
    
 } /// eosio::chain
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -114,6 +114,7 @@ namespace eosio::chain {
          const digest_type& digest = vote.strong ? strong_finalizer_digest : weak_finalizer_digest;
 
          return pending_qc.add_vote(vote.strong,
+#warning TODO change to use std::span if possible
                                     std::vector<uint8_t>{digest.data(), digest.data() + digest.data_size()},
                                     index,
                                     vote.finalizer_key,
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -55,14 +55,14 @@ void block_state::set_trxs_metas( deque<transaction_metadata_ptr>&& trxs_metas, 
 
 // Called from net threads
 bool block_state::aggregate_vote(const hs_vote_message& vote) {
-   const auto& finalizers = finalizer_policy->finalizers;
+   const auto& finalizers = active_finalizer_policy->finalizers;
    auto it = std::find_if(finalizers.begin(),
                           finalizers.end(),
                           [&](const auto& finalizer) { return finalizer.public_key == vote.finalizer_key; });
 
    if (it != finalizers.end()) {
       auto index = std::distance(finalizers.begin(), it);
-      const digest_type& digest = vote.strong ? strong_finalizer_digest : weak_finalizer_digest;
+      const digest_type& digest = vote.strong ? strong_digest : weak_digest;
       return pending_qc.add_vote(vote.strong,
 #warning TODO change to use std::span if possible
                                  std::vector<uint8_t>{digest.data(), digest.data() + digest.data_size()},
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -111,19 +111,13 @@ namespace eosio::chain {
 
       if (it != finalizers.end()) {
          auto index = std::distance(finalizers.begin(), it);
-         if( vote.strong ) {
-            std::vector<uint8_t> d(strong_finalizer_digest.data(), strong_finalizer_digest.data() + strong_finalizer_digest.data_size());
-            return pending_qc.add_strong_vote( d,
-                                               index,
-                                               vote.finalizer_key,
-                                               vote.sig );
-         } else {
-            std::vector<uint8_t> d(weak_finalizer_digest.data(), weak_finalizer_digest.data() + weak_finalizer_digest.data_size());
-            return pending_qc.add_weak_vote( d,
-                                             index,
-                                             vote.finalizer_key,
-                                             vote.sig );
-         }
+         const digest_type& digest = vote.strong ? strong_finalizer_digest : weak_finalizer_digest;
+
+         return pending_qc.add_vote(vote.strong,
+                                    std::vector<uint8_t>{digest.data(), digest.data() + digest.data_size()},
+                                    index,
+                                    vote.finalizer_key,
+                                    vote.sig);
       } else {
          wlog( "finalizer_key (${k}) in vote is not in finalizer policy", ("k", vote.finalizer_key) );
          return false;
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -301,19 +301,16 @@ struct block_data_t {
       }, v);
    }
 
-   // called from net thread
+    // called from net thread
 	 bool aggregate_vote(const hs_vote_message& vote) {
 	    return std::visit(
 	       overloaded{[](const block_data_legacy_t&) {
-	                     EOS_ASSERT(false, misc_exception, "attempting to call aggregate_vote in legacy mode");
+                        // We can be late in switching to Instant Finality
+                        // and receive votes from those already having switched.
 	                     return false; },
 	                  [&](const block_data_new_t& bd) {
 	                     auto bsp = bd.fork_db.get_block(vote.proposal_id);
-	                     if (bsp) {
-	                        return bsp->aggregate_vote(vote);
-	                     } else {
-	                        return false;
-                        }; }
+	                     return bsp && bsp->aggregate_vote(vote); }
                     },
 	       v);
 	 }
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -249,6 +249,7 @@ struct block_data_t {
       }, v);
    }
 
+   // called from net thread
 	bool aggregate_vote(const hs_vote_message& vote) {
 	   return std::visit(
 	      overloaded{[](const block_data_legacy_t&) {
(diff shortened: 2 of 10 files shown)


libraries/chain/block_state.cpp
```diff
@@ -102,5 +102,32 @@ namespace eosio::chain {
...
+                                               index,
+                                               vote.finalizer_key,
+                                               vote.sig );
+         } else {
+            std::vector<uint8_t> d(weak_finalizer_digest.data(), weak_finalizer_digest.data() + weak_finalizer_digest.data_size());
+            return pending_qc.add_weak_vote( d,
+                                             index,
+                                             vote.finalizer_key,
+                                             vote.sig );
+         }
```
greg7mdp, line 126:
```suggestion
          auto                 index  = std::distance(finalizers.begin(), it);
          const digest_type&   digest = vote.strong ? strong_finalizer_digest : weak_finalizer_digest;
          std::vector<uint8_t> d(digest.data(), digest.data() + digest.data_size());
          return pending_qc.add_vote(vote.strong, d, index, vote.finalizer_key, vote.sig);
```

greg7mdp (changes requested):

greg7mdp (approved):

libraries/chain/block_state.cpp
```diff
@@ -102,5 +102,26 @@ namespace eosio::chain {
    ,_cached_trxs( std::move(trx_metas) )
    {}
 #endif
+
+   bool block_state::aggregate_vote(const hs_vote_message& vote) {
```
heifner, line 106: Add comment that this is called from net threads.
Also need to make thread safe. Currently we are not setting finalizer_policy, but seems like that can be set at construction so that we do not have to lock it. Likewise the strong_finalizer_digest and weak_finalizer_digest are not mutated after creation. If so we can likely get away with making `pending_qc` thread safe.

libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -22,10 +22,14 @@ inline std::vector<uint32_t> bitset_to_vector(const hs_bitset& bs) {
 
 bool pending_quorum_certificate::votes_t::add_vote(const std::vector<uint8_t>& proposal_digest, size_t index,
                                                    const bls_public_key& pubkey, const bls_signature& new_sig) {
-   if (_bitset[index])
+   if (_bitset[index]) {
+      wlog( "finalizer ${i} has already voted", ("i", index) );
```
heifner, line 26: I don't think this is a warning. This is what stops vote propagation on the network.

libraries/chain/controller.cpp
```diff
@@ -4007,8 +4024,8 @@ void controller::get_finalizer_state( finalizer_state& fs ) const {
 }
 
 // called from net threads
-void controller::notify_hs_message( const uint32_t connection_id, const hs_message& msg ) {
-   my->pacemaker->on_hs_msg(connection_id, msg);
+void controller::process_vote_message( const hs_vote_message& vote ) {
+   my->block_data.aggregate_vote(vote);
```
heifner, line 4028: If this returns `true` we need to call the bcast function of net_plugin.

libraries/chain/include/eosio/chain/block_state.hpp
```diff
@@ -11,7 +11,8 @@ namespace eosio::chain {
       // ------ data members -------------------------------------------------------------
       signed_block_ptr           block;
       bool                       validated;             // We have executed the block's trxs and verified that action merkle root (block id) matches.
-      digest_type                finalizer_digest;
+      digest_type                strong_finalizer_digest;
+      digest_type                weak_finalizer_digest;
```
heifner, line 15: These are not set.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -1095,7 +1095,7 @@ namespace eosio {
       void handle_message( const block_id_type& id, signed_block_ptr ptr );
       void handle_message( const packed_transaction& msg ) = delete; // packed_transaction_ptr overload used instead
       void handle_message( packed_transaction_ptr trx );
-      void handle_message( const chain::hs_message& msg );
+      void handle_message( const chain::hs_vote_message& msg );
```
heifner, line 1098: Need to update protocol.hpp

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3673,10 +3673,10 @@ namespace eosio {
...
    }
 
-   void connection::handle_message( const chain::hs_message& msg ) {
-      peer_dlog(this, "received hs: ${msg}", ("msg", msg));
+   void connection::handle_message( const chain::hs_vote_message& msg ) {
+      peer_dlog(this, "received vote: ${msg}", ("msg", msg));
       controller& cc = my_impl->chain_plug->chain();
-      cc.notify_hs_message(connection_id, msg);
+      cc.process_vote_message(msg);
    }
```
heifner, line 3680: Need to also update `bcast_hs_message` to broadcast `hs_vote_message` instead of `hs_message`. Go ahead and remove `hs_message` so nothing else is missed.

heifner (changes requested):

libraries/chain/controller.cpp
```diff
@@ -249,6 +249,23 @@ struct block_data_t {
...
+	   return std::visit(
+	      overloaded{[](const block_data_legacy_t&) {
+	                    EOS_ASSERT(false, misc_exception, "attempting to call aggregate_vote in legacy mode");
+	                    return false; },
+	                 [&](const block_data_new_t& bd) {
+	                    auto bsp = bd.fork_db.get_block(vote.proposal_id);
+	                    if (bsp) {
+	                       return bsp->aggregate_vote(vote);
+	                    } else {
+	                       wlog("no block exists for the vote (proposal_id: ${id}", ("id", vote.proposal_id));
```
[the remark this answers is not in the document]
linh2931: Removed the log.

libraries/chain/controller.cpp
```diff
@@ -249,6 +249,23 @@ struct block_data_t {
...
+	                       return false;
+	                    }}
+                   },
+	      v);
+	}
```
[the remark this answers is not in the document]
linh2931: I adjusted the indentation a bit.

libraries/chain/controller.cpp
```diff
@@ -249,6 +249,23 @@ struct block_data_t {
       }, v);
    }
 
+	bool aggregate_vote(const hs_vote_message& vote) {
```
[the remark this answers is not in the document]
linh2931: Do we need to make sure it runs on the main thread, or a mutex?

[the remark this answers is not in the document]
heifner: Would be nice to keep the bls verify and aggregate off the main thread.

libraries/chain/controller.cpp
linh2931: I asked at one of the meetings, @arhag said we just needed to aggregate the votes.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
linh2931: Oh I see your comments below.

libraries/chain/controller.cpp
heifner: Do we have an issue for it? We need that for the test net as vote messages will not propagate without it.

plugins/net_plugin/net_plugin.cpp
linh2931: Which `protocol.hpp`? The one `include/eosio/chain/protocol.hpp` contains only one line of `#include <eosio/chain/block.hpp>`.

heifner: [net_plugin](https://github.com/AntelopeIO/leap/blob/d2af8636c116a66d530af34bc3565f73b7ed29eb/plugins/net_plugin/include/eosio/net_plugin/protocol.hpp#L146-L146)

libraries/chain/include/eosio/chain/block_state.hpp
linh2931: I planned to add compute_finalizer_digest in a separate PR. Will put a TODO here for now.

libraries/chain/controller.cpp
linh2931: Will double check with him at tomorrow's meeting.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
linh2931: mutex added to pending_quorum_certificate.

libraries/chain/controller.cpp
linh2931: Will broadcast.

plugins/net_plugin/net_plugin.cpp
linh2931: `hs_message` is used in a number of places. Will remove it after unification.

libraries/chain/include/eosio/chain/block_state.hpp
linh2931: For the first prototype, it will be set to block_id in the constructor of block_state when the constructor code is ready.

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -160,16 +162,7 @@ namespace eosio::chain {
 
       void reset(const fc::sha256& proposal_id, const digest_type& proposal_digest, size_t num_finalizers, size_t quorum);
```
heifner, line 163: `reset(), is_quorum_met(), num_weak(), num_strong()` all need mutex protection. Careful as `num_weak()` and `num_strong()` are currently already called holding a mutex.

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -190,8 +183,23 @@ namespace eosio::chain {
       state_t              _state { state_t::unrestricted };
       size_t               _num_finalizers {0};
       size_t               _quorum {0};
+      std::unique_ptr<std::mutex> _mtx;  // protect both _strong_votes and _weak_votes
       votes_t              _weak_votes;
       votes_t              _strong_votes;
+
+   private:
+#warning TODO move members above to private after unification. Currently they are used by qc_chain.cpp directly
```
heifner, line 191: Can you make qc_chain a friend. Very hard to determine if this is thread-safe as-is.

libraries/chain/block_state.cpp
heifner: Please add the comment here as well about called from net threads. We will need to be careful that anything accessed in block_state is safe to access from multiple threads.

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
linh2931: Thanks!

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -169,6 +168,12 @@ namespace eosio::chain {
                     const bls_public_key& pubkey,
                     const bls_signature& sig);
 
+      fc::sha256 proposal_id()  const { std::lock_guard g(*_mtx); return _proposal_id; };
+      state_t    state()        const { std::lock_guard g(*_mtx); return _state; };
+      votes_t    weak_votes()   const { std::lock_guard g(*_mtx); return _weak_votes; };
+      votes_t    strong_votes() const { std::lock_guard g(*_mtx); return _strong_votes; };
+      std::vector<uint8_t> proposal_digest() const { std::lock_guard g(*_mtx); return _proposal_digest; };
```
heifner, line 175: These all need to be in-sync. Pulling them out individually means they can be modified separately . I think instead you should add a method that returns a `valid_quorum_certificate` and remove all of these.

linh2931: Good idea! Thanks.

libraries/chain/controller.cpp
```diff
@@ -301,6 +301,23 @@ struct block_data_t {
       }, v);
    }
 
+   // called from net thread
+	 bool aggregate_vote(const hs_vote_message& vote) {
+	    return std::visit(
+	       overloaded{[](const block_data_legacy_t&) {
+	                     EOS_ASSERT(false, misc_exception, "attempting to call aggregate_vote in legacy mode");
```
greg7mdp, line 308: I think we should just silently return without an exception.. we could be late at switching to IF and receiving votes from other nodes which already switched.

libraries/chain/controller.cpp
```diff
@@ -301,6 +301,23 @@ struct block_data_t {
...
+	                     return false; },
+	                  [&](const block_data_new_t& bd) {
+	                     auto bsp = bd.fork_db.get_block(vote.proposal_id);
+	                     if (bsp) {
+	                        return bsp->aggregate_vote(vote);
+	                     } else {
+	                        return false;
+                        }; }
```
greg7mdp, line 316:
```suggestion
	                    return bsp && bsp->aggregate_vote(vote); }
```

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary:  Process vote messages received from net plugin by aggregating signatures into the pending QC for the corresponding block.
Note: end

greg7mdp (approved):

heifner (approved):

Issue #2075, opened 2024-01-11
IF: Unification: implement get_best_qc in block_state


linh2931:
`get_best_qc` is needed when finding the most recent ancestor QC.

If `pending_qc` does not have a valid QC, return valid_qc. Otherwise, extract the valid QC from `pending_qc`. Compare that to `valid_qc` to determine which is better: Strong beats Weak. Break tie with highest accumulated weight.  Return the better one.

PR #2076, opened 2024-01-11
IF: Unification: Implement get_best_qc in block_state

linh2931:
If `pending_qc` does not have a valid QC, return valid_qc. Otherwise, extract the valid QC from `pending_qc`. Compare that to `valid_qc` to determine which is better: Strong beats Weak. Break tie with `valid_qc`.  Return the better one.

Resolved https://github.com/AntelopeIO/leap/issues/2075

### DIFF
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -102,5 +102,45 @@ namespace eosio::chain {
    ,_cached_trxs( std::move(trx_metas) )
    {}
 #endif
+
+   std::optional<qc_data_t> block_state::get_best_qc() const {
+      auto block_number = block_num();
+
+      // if pending_qc does not have a valid QC, consider valid_qc only
+      if( !pending_qc.is_valid() ) {
+         if( valid_qc ) {
+            return qc_data_t{ quorum_certificate{ block_number, valid_qc.value() },
+                              qc_info_t{ block_number, valid_qc.value().is_strong() }};
+         } else {
+            return std::nullopt;;
+         }
+      }
+
+      // extract valid QC from pending_qc
+      valid_quorum_certificate valid_qc_from_pending(pending_qc);
+
+      // if valid_qc does not have value, consider valid_qc_from_pending only
+      if( !valid_qc ) {
+         return qc_data_t{ quorum_certificate{ block_number, valid_qc_from_pending },
+                           qc_info_t{ block_number, valid_qc_from_pending.is_strong() }};
+      }
+
+      // Both valid_qc and valid_qc_from_pending have value. Compare them and select a better one.
+      // Strong beats weak. Break tie with highest accumulated weight.
+      auto valid_qc_is_better = false;
+      if( valid_qc.value().is_strong() && !valid_qc_from_pending.is_strong() ) {
+         valid_qc_is_better = true;
+      } else if( !valid_qc.value().is_strong() && valid_qc_from_pending.is_strong() ) {
+         valid_qc_is_better = false;
+      } else if( valid_qc.value().accumulated_weight() >= valid_qc_from_pending.accumulated_weight() ) {
+         valid_qc_is_better = true;
+      } else {
+         valid_qc_is_better = false;
+      }
+
+      const auto& qc = valid_qc_is_better ? valid_qc.value() : valid_qc_from_pending;
+      return qc_data_t{ quorum_certificate{ block_number, qc },
+                        qc_info_t{ block_number, qc.is_strong() }};
+   }
    
 } /// eosio::chain
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -107,10 +107,10 @@ namespace eosio::chain {
       auto block_number = block_num();
 
       // if pending_qc does not have a valid QC, consider valid_qc only
-      if( !pending_qc.is_valid() ) {
+      if( !pending_qc.is_quorum_met() ) {
          if( valid_qc ) {
-            return qc_data_t{ quorum_certificate{ block_number, valid_qc.value() },
-                              qc_info_t{ block_number, valid_qc.value().is_strong() }};
+            return qc_data_t{ quorum_certificate{ block_number, *valid_qc },
+                              qc_info_t{ block_number, valid_qc->is_strong() }};
          } else {
             return std::nullopt;;
          }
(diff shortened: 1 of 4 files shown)


libraries/chain/block_state.cpp
```diff
@@ -102,5 +102,45 @@ namespace eosio::chain {
...
 #endif
+
+   std::optional<qc_data_t> block_state::get_best_qc() const {
+      auto block_number = block_num();
+
+      // if pending_qc does not have a valid QC, consider valid_qc only
+      if( !pending_qc.is_valid() ) {
+         if( valid_qc ) {
+            return qc_data_t{ quorum_certificate{ block_number, valid_qc.value() },
+                              qc_info_t{ block_number, valid_qc.value().is_strong() }};
```
greg7mdp, line 113:
```suggestion
            return qc_data_t{ quorum_certificate{ block_number, *valid_qc },
                              qc_info_t{ block_number, valid_qc->is_strong() }};
```

libraries/chain/block_state.cpp
```diff
@@ -102,5 +102,45 @@ namespace eosio::chain {
...
+      // Strong beats weak. Break tie with highest accumulated weight.
+      auto valid_qc_is_better = false;
+      if( valid_qc.value().is_strong() && !valid_qc_from_pending.is_strong() ) {
+         valid_qc_is_better = true;
+      } else if( !valid_qc.value().is_strong() && valid_qc_from_pending.is_strong() ) {
+         valid_qc_is_better = false;
+      } else if( valid_qc.value().accumulated_weight() >= valid_qc_from_pending.accumulated_weight() ) {
+         valid_qc_is_better = true;
+      } else {
+         valid_qc_is_better = false;
```
greg7mdp, line 138: For modularity, I believe it would be better to have a comparison operator in `valid_quorum_certificate`, such as:

```c++
      bool operator<(const valid_quorum_certificate& o) {
         if (is_strong() != o.is_strong())
            return o.is_strong();
         if (accumulated_weight() != o.accumulated_weight()) 
            return accumulated_weight() <  o.accumulated_weight();
         return false;
      }
```

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -211,6 +212,8 @@ namespace eosio::chain {
       bool is_weak()   const { return !!_weak_votes; }
       bool is_strong() const { return !_weak_votes; }
 
+      uint32_t accumulated_weight() const { return (_strong_votes ? _strong_votes.value().count() : 0) + (_weak_votes ? _weak_votes.value().count() : 0); }
+
```
greg7mdp, line 216: I believe the `accumulated_weight` function should not count the weak votes when the qc is strong.

greg7mdp (changes requested):

libraries/chain/block_state.cpp
linh2931: Thanks!

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
linh2931: When qc is strong, weak count is 0. No need to have a separate expression. Will add a comment.

heifner: Curious why your preference for `->` and `*` over `.value()` ?

linh2931: `value()` is clearer while `->` is shorter? Either way is fine with me.

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -211,6 +211,12 @@ namespace eosio::chain {
       bool is_weak()   const { return !!_weak_votes; }
       bool is_strong() const { return !_weak_votes; }
 
+      bool     operator<(const valid_quorum_certificate& other) const;
+
+      // When it is strong, _weak_votes does not exist. No need to have a separate
+      // calculation for strong
+      uint32_t accumulated_weight() const { return (_strong_votes ? _strong_votes->count() : 0) + (_weak_votes ? _weak_votes->count() : 0); }
```
arhag, line 218: For now, ignore the `accumulated_weight` and just focus on strong QCs beating weak QCs. Otherwise, they are equally good enough.

libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -184,6 +184,18 @@ valid_quorum_certificate::valid_quorum_certificate(
       _weak_votes = vector_to_bitset(weak_votes);
 }
 
+// Strong beats weak. Break tie with highest accumulated weight.
+bool valid_quorum_certificate::operator<(const valid_quorum_certificate& other) const {
```
arhag, line 188: Remove this comparison function as well. There is no clear way to consider one QC better than the other as a total order. And we don't need a total order to get a good enough implementation for `get_best_qc` for now.

libraries/chain/include/eosio/chain/block_state.hpp
```diff
@@ -29,6 +29,7 @@ namespace eosio::chain {
       bool                   is_valid()          const { return validated; } 
       void                   set_valid(bool b)         { validated = b; }
       uint32_t               irreversible_blocknum() const { return 0; } // [greg todo] equivalent of dpos_irreversible_blocknum
+      std::optional<qc_data_t> get_best_qc()     const;
```
arhag, line 32: We should return `std::optional<quorum_certificate>`.

I don't see the purpose of `qc_data_t` at all. It should be possible to derive a `qc_info_t` from a `quorum_certificate`.

libraries/chain/block_state.cpp
```diff
@@ -102,5 +102,34 @@ namespace eosio::chain {
...
+         if( valid_qc ) {
+            return qc_data_t{ quorum_certificate{ block_number, *valid_qc },
+                              qc_info_t{ block_number, valid_qc->is_strong() }};
+         } else {
+            return std::nullopt;;
+         }
+      }
+
+      // extract valid QC from pending_qc
+      valid_quorum_certificate valid_qc_from_pending(pending_qc);
```
arhag, line 120: I don't like that we use a constructor that has an `assert(0)` inside of it if it is called incorrectly. I get that it shouldn't happen in this case since you have already ensured that by this point `pending_qc.is_quorum_met() == true`. But this is a fragile pattern. It would be better to have a member function in `pending_quorum_certificate` to extract an optional `valid_quorum_certificate`.

Anyway, this is just a note for the future. I don't expect this to change in this PR.

libraries/chain/block_state.cpp
```diff
@@ -102,5 +102,34 @@ namespace eosio::chain {
...
+
+      // if valid_qc does not have value, consider valid_qc_from_pending only
+      if( !valid_qc ) {
+         return qc_data_t{ quorum_certificate{ block_number, valid_qc_from_pending },
+                           qc_info_t{ block_number, valid_qc_from_pending.is_strong() }};
+      }
+
+      // Both valid_qc and valid_qc_from_pending have value. Compare them and select a better one.
+      // Strong beats weak. Break tie with highest accumulated weight.
```
arhag, line 129: Don't yet break the tie with the highest accumulated weight. Just pick `valid_qc` if both are strong.

arhag (changes requested):

libraries/chain/block_state.cpp
linh2931: Added a TODO warning.

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
greg7mdp: > Curious why your preference for `->` and `*` over `.value()` ?

I believe it is simpler, more compact, and more idiomatic. For example the example in cppreference uses `*` and not `value()`. Also it looks better to me because it works just like a pointer, and I can tell with a quick look that it is correct, while the `value()` version takes more effort to read.

I guess it is a matter of preference. Personally, everything else (like legibility) being equal, I prefer the most concise way of expressing something.

greg7mdp: Oh and I forgot `*` could actually be more efficient because it doesn't check that the optional contains a value (while `value()` throws if it doesn't).

arhag (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Returns the most significant QC from the ancestors. Takes many factors into account when determining the QC to return. 
Note: end

PR #2077, opened 2024-01-11
IF: Unification: Changes in controller to complete unification - Part 2

### LINKED ISSUE #2034: IF: Unification: Changes in controller to complete unification

### LINKED ISSUE #2047: IF: Unification: Implement block_header_state::next and related functions

greg7mdp:
Resolves #2034.
Resolves #2047.


libraries/chain/controller.cpp
```diff
@@ -930,7 +939,7 @@ struct controller_impl {
         cfg.state_size, false, cfg.db_map_mode ),
     blog( cfg.blocks_dir, cfg.blog ),
     block_data(block_data_t::block_data_variant{
-         std::in_place_type<block_data_legacy_t>, // [greg todo] create correct type depending on whether IF activated
+         std::in_place_type<block_data_legacy_t>, // initial state is always dpos
```
heifner, line 942: Not sure that will be true after IF is enabled on startup, but fine for now.

libraries/chain/controller.cpp
```diff
@@ -2917,7 +2931,8 @@ struct controller_impl {
          }
          
          auto do_push = [&](auto& fork_db, auto& head) {
-            fork_db.add( bsp );
+            if constexpr (std::is_same_v<BSP, typename std::decay_t<decltype(head)>>)
```
heifner, line 2934: That is really ugly, would be nice to find a way not to have these `if constexpr`.

libraries/chain/controller.cpp
```diff
@@ -3680,13 +3694,23 @@ block_state_legacy_ptr controller::create_block_state( const block_id_type& id,
...
                              const block_state_legacy_ptr& bsp,
+                             const forked_branch_callback_legacy& forked_branch_cb,
+                             const trx_meta_cache_lookup& trx_lookup )
+{
+   validate_db_available_size();
+   my->push_block( br, bsp, forked_branch_cb, trx_lookup );
+}
+
+void controller::push_block( controller::block_report& br,
+                             const block_state_ptr& bsp,
```
heifner, line 3705: Don't think we will want to public push_block like this. Users of controller shouldn't need to know which one to call.

heifner (approved):

libraries/chain/controller.cpp
greg7mdp: OK. I thought I asked and we said we would always go through the transition, but I was thinking it might not be true when we start from a snapshot.

heifner: We will always go through the transition when starting a new chain. We have no immediate plans to add the feature of enabling protocol feature activation in genesis. 
Startup of nodeos after transition will start with state (either from chainbase or snapshot) that is in IF mode.

libraries/chain/controller.cpp
greg7mdp: Sure, I agree and ideally the `block_state` struct should not be visible outside `libraries/chain`. But I thought that if `create_block_state` and `create_block_state_future` were needed maybe `push_block` as well? We can certainly remove these if not needed.

heifner: Worse case we have some type that controller provides via something like create_block_start_state() that `controller::push_block` breaks apart and calls the appropriate `controller_impl::push_block`.

heifner: This is fine for now until we get the create_block_state() and create_block_state_future() fixed.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -21,11 +21,13 @@ namespace eosio::chain {
     *
     * An internal mutex is used to provide thread-safety.
     */
-   template<class bsp, class bhsp>  // either [block_state_legacy_ptr, block_state_ptr], same with block_header_state_ptr
+   template<class bsp>  // either block_state_legacy_ptr or block_state_ptr
```
linh2931, line 24: Conventionally template type should be capitalized as `BSP`.

linh2931 (approved):

libraries/chain/block_header_state.cpp
```diff
@@ -1,32 +1,37 @@
...
+}
+
+const vector<digest_type>& block_header_state::get_new_protocol_feature_activations()const {
+   return detail::get_new_protocol_feature_activations(header_exts);
 }
 
 #warning Add last_proposed_finalizer_policy_generation to snapshot_block_header_state_v3, see header file TODO
    
-block_header_state_core block_header_state_core::next(const uint32_t last_qc_block_num, bool is_last_qc_strong) const {
+block_header_state_core block_header_state_core::next(qc_info_t incoming) const {
```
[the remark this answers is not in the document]
greg7mdp: passing by reference (indirection with a pointer) would actually be slower since the whole struct is 8 bytes.

libraries/chain/controller.cpp
```diff
@@ -201,6 +207,29 @@ struct block_data_t {
       return std::visit([](const auto& bd) -> const signed_block_ptr& { return bd.head->block; }, v);
    }
 
+   void replace_producer_keys( const public_key_type& key ) {
+      ilog("Replace producer keys with ${k}", ("k", key));
```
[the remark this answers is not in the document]
greg7mdp: I think @heifner added this, but I don't see the harm in logging it.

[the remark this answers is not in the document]
heifner: It is public keys so no harm. Useful to see it working.

libraries/chain/controller.cpp
```diff
@@ -42,6 +43,7 @@
 #include <new>
 #include <shared_mutex>
 #include <utility>
+#include <boost/phoenix/core/nothing.hpp>
```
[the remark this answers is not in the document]
greg7mdp: Kevin added this, but it doesn't appear to be needed so I'll remove it.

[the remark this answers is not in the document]
heifner: Sorry about that. Editor auto adds includes and I think I did an auto complete that pulled that in.

libraries/chain/include/eosio/chain/fork_database.hpp
linh2931: No need to change.

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Finish data structures to allow assembling of blocks under new Faster Finality. 
Note: end

PR #2078, opened 2024-01-11
IF: Implement proposer policy change

heifner:
- Implement proposer policy change design: https://hackmd.io/@9xzIZ7ZaRwegCHgQW02r3Q/BkQ4Oc4Up
- Producer schedule is changed after `set_proposed_producer` in the next, next producer round.
  - For example: Producer A [ b1, b2, .. b12 ], Producer B [ b1, b2, .. b12 ], Producer X [ b1, b2, .. b12 ]
    - If `set_proposed_producer` is called in trx in any blocks of Producer A [ b1, b2, .. b12], the schedule change happens at Producer X [b1]. 
- public controller `pending_producers()` and `proposed_producers` renamed to `_legacy`.
- Added `next_producers()` for use by external enties.

### DIFF
--- a/libraries/chain/block_header_state.cpp
+++ b/libraries/chain/block_header_state.cpp
@@ -1,13 +1,14 @@
 #include <eosio/chain/block_header_state.hpp>
 #include <eosio/chain/block_header_state_utils.hpp>
 #include <eosio/chain/hotstuff/instant_finality_extension.hpp>
+#include <eosio/chain/hotstuff/proposer_policy.hpp>
 #include <eosio/chain/exceptions.hpp>
 #include <limits>
 
 namespace eosio::chain {
 
 producer_authority block_header_state::get_scheduled_producer(block_timestamp_type t) const {
-   return detail::get_scheduled_producer(proposer_policy->proposer_schedule.producers, t);
+   return detail::get_scheduled_producer(active_proposer_policy->proposer_schedule.producers, t);
 }
 
 #warning Add last_proposed_finalizer_policy_generation to snapshot_block_header_state_v3, see header file TODO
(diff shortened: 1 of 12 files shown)


greg7mdp (approved):

libraries/chain/block_header_state.cpp
```diff
@@ -61,13 +62,29 @@ block_header_state block_header_state::next(block_header_state_input& input) con
...
       .previous          = input.parent_id,
       .transaction_mroot = input.transaction_mroot,
       .action_mroot      = input.action_mroot,
-      //.schedule_version = ?, [greg todo]
-      //.new_producers = ?     [greg todo]
+      .schedule_version  = header.schedule_version
    };
 
+   if(!proposer_policies.empty()) {
+      auto it = proposer_policies.begin();
```
greg7mdp, line 73: Do we want the earliest one?

libraries/chain/block_header_state.cpp
```diff
@@ -61,13 +62,29 @@ block_header_state block_header_state::next(block_header_state_input& input) con
...
+      if (it->first <= input.timestamp) {
+         result.active_proposer_policy = it->second;
+         result.header.schedule_version = header.schedule_version + 1;
+         result.active_proposer_policy->proposer_schedule.version = result.header.schedule_version;
+         result.proposer_policies = { ++it, proposer_policies.end() };
+      } else {
+         result.proposer_policies = proposer_policies;
```
greg7mdp, line 80: should we also have here:
`result.proposer_policy = proposer_policy;`

libraries/chain/block_header_state.cpp
heifner: They are ordered by timestamp. We want the earliest timestamp for our comparison with current (input.timestamp).

libraries/chain/block_header_state.cpp
heifner: I could have swore I had that in there. Yes, we want `result=active_proposer_policy = active_proposer_policy`. I can push a commit with that fix.

heifner: pushed

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Implement proposer policy for Faster Finality. Existing proposal policy renamed to reflect they are legacy.
Note: end

Issue #2083, opened 2024-01-12
IF: fork database transition not thread safe


heifner:
Currently `fork_database` is thread safe via internal mutex. During the transition of dpos fork database to instant finality fork database threads from API calls and `net_plugin` can be running. Need to provide a wrapper around the two fork databases that is thread safe and can provide the interfaces needed by API calls, `net_plugin`, and controller.

Also handle `[greg todo]` leftover from https://github.com/AntelopeIO/leap/issues/1941 & #2045 in leap-util blocklog. The `fork_database` wrapper can examine file to determine which internal implementation is in use.

greg7mdp:
>  Need to provide a wrapper around the two fork databases that is thread safe and can provide the interfaces needed by API calls, net_plugin, and controller.

Is it really necessary? I would not have thought so because:
1. only the main thread modifies fork_database, right?
2. only one of the two fork_database is valid at any one time
3. the main thread can convert `fork_db<legacy>` -> `fork_db<new>` while net_plugin and APIs still access `fork_db<legacy>`, and then we have to make an atomic switch where maybe we freeze `net_plugin`?

heifner:
1. Correct.
2. Correct.
3. Correct. The way we freeze net_plugin currently during snapshot would not prevent a call into fork database. It might be possible via freezing net_plugin and API calls, but would have to verify all currently running calls are complete. Not really sure how you do that without a mutex or atomic counter somewhere.  There is already a mutex in fork database that provides the currency projection. I think we can just move that mutex up to a wrapper that also does the switch from one to the other.

greg7mdp:
Oh wait, because there is a  mutex in fork_db I think we might be fine without the wrapper. What about:
1. store a `shared_ptr` to fork_db in controller (instead of the struct itself).
2. when time comes for the conversion, lock the mutex on `fork_db<legacy>`
3. once we have the lock, do the conversion ` fork_db<legacy>` -> `fork_db<new>`, but keep a shared_ptr to ` fork_db<legacy>` so it stays alive. update controller to point to `fork_db<new>`.
4. release the mutex of `fork_db<legacy>`. net_plugin and APIs would finish their work using the old fork_db, but that should be fine.

heifner:
That doesn't work without an atomic shared ptr. Boost has an atomic shared ptr we could use for this.

Without atomic shared ptr:
Net thread begins access of control->shared_ptr->fork_db<legacy>
[...]

PR #2085, opened 2024-01-14
IF: Transition from dpos to instant-finality

### LINKED ISSUE #2045: IF: Unification: IF transition of fork database usage
Support transition from reversible blocks in old fork database to blocks in new fork database. Convert block_header_state_legacy to block_header_state for the appropriate block. Handle transition at the appropriate time. ~~Consider how a fork re-org would be handled in the middle of the transition.~~

Also handle any remaining small TODOs leftover from #1941 (e.g. in leap-util). leap-util todo fix moved to #2083 since not needed for testnet.

~~As soon as the transition begins, producer schedule changes will be rejected until the transition completes. The host function would not fail. It would just silently drop the new proposed producer schedules. This allows us to avoid needing to update proposer policy for the blocks created during the transition.~~

For this issue, we want a simpler (but unsafe) transition in which the block that first proposes the finalizer policy immediately becomes irreversible and atomically transitions from the old algorithm to the IF algorithm going forward.

This should also have a Boost test to show the transition works correctly.

heifner:
- Transition from dpos to instant-finality on the block after `set_finalizer` host function call. This is a temporary method of transition until we work out the algorithm to use in production.
  - `fork_database` switched over after block with `set_finalizer` call.
- Add `block_token` as a wrapper of `block_state_ptr` and `block_state_legacy_ptr`.
  - `block_state` no longer used outside of chain library except for leap-util `blocklog_actions::read_log` which will be fixed by #2083 
- Enable `instant-finality` `nodeos_run_test.py` integration test which tests transition.
- Performance improvement via `controller::block_exists` instead of always pulling the complete block out of the block log.
- Fixes producer schedule switch bug after instant-finality is enabled.
- LIB does not advance after switching to instant-finality. This is expected as we are not generating vote messages yet.

Resolves #2045

### DIFF
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -21,20 +21,20 @@ block_state::block_state(const block_header_state& bhs, deque<transaction_metada
    block->transactions = std::move(trx_receipts);
 }
 
-// Used for transition from dbpos to instant-finality
+// Used for transition from dpos to instant-finality
 block_state::block_state(const block_state_legacy& bsp) {
    block_header_state::id = bsp.id();
    header = bsp.header;
    activated_protocol_features = bsp.activated_protocol_features;
    std::optional<block_header_extension> ext = bsp.block->extract_header_extension(instant_finality_extension::extension_id());
-   assert(ext); // required by current transistion mechanism
+   assert(ext); // required by current transition mechanism
    const auto& if_extension = std::get<instant_finality_extension>(*ext);
-   assert(if_extension.new_finalizer_policy); // required by current transistion mechanism
+   assert(if_extension.new_finalizer_policy); // required by current transition mechanism
    active_finalizer_policy = std::make_shared<finalizer_policy>(*if_extension.new_finalizer_policy);
    active_proposer_policy = std::make_shared<proposer_policy>();
    active_proposer_policy->active_time = bsp.timestamp();
    active_proposer_policy->proposer_schedule = bsp.active_schedule;
-   header_exts = bsp.header_exts; // not needed, but copy over just in case
+   header_exts = bsp.header_exts;
    block = bsp.block;
    validated = bsp.is_valid();
    pub_keys_recovered = bsp._pub_keys_recovered;
--- a/libraries/chain/block_state.cpp
+++ b/libraries/chain/block_state.cpp
@@ -1,5 +1,6 @@
 #include <eosio/chain/block_state.hpp>
 #include <eosio/chain/block_header_state_utils.hpp>
+#include <eosio/chain/block_state_legacy.hpp>
 #include <eosio/chain/exceptions.hpp>
 
 namespace eosio::chain {
@@ -16,7 +17,29 @@ block_state::block_state(const block_header_state& bhs, deque<transaction_metada
    , block(std::make_shared<signed_block>(signed_block_header{bhs.header})) // [greg todo] do we need signatures?
    , pub_keys_recovered(true) // probably not needed
    , cached_trxs(std::move(trx_metas))
-{}
+{
+   block->transactions = std::move(trx_receipts);
+}
+
+// Used for transition from dbpos to instant-finality
+block_state::block_state(const block_state_legacy& bsp) {
+   block_header_state::id = bsp.id();
+   header = bsp.header;
+   activated_protocol_features = bsp.activated_protocol_features;
+   std::optional<block_header_extension> ext = bsp.block->extract_header_extension(instant_finality_extension::extension_id());
+   assert(ext); // required by current transistion mechanism
+   const auto& if_extension = std::get<instant_finality_extension>(*ext);
+   assert(if_extension.new_finalizer_policy); // required by current transistion mechanism
+   active_finalizer_policy = std::make_shared<finalizer_policy>(*if_extension.new_finalizer_policy);
+   active_proposer_policy = std::make_shared<proposer_policy>();
+   active_proposer_policy->active_time = bsp.timestamp();
+   active_proposer_policy->proposer_schedule = bsp.active_schedule;
+   header_exts = bsp.header_exts; // not needed, but copy over just in case
+   block = bsp.block;
+   validated = bsp.is_valid();
+   pub_keys_recovered = bsp._pub_keys_recovered;
+   cached_trxs = bsp._cached_trxs;
+}
 
 deque<transaction_metadata_ptr> block_state::extract_trxs_metas() {
    pub_keys_recovered = false;
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -2835,7 +2835,7 @@ struct controller_impl {
                auto producer_block_id = bsp->id();
                start_block( b->timestamp, b->confirmed, new_protocol_feature_activations, s, producer_block_id, fc::time_point::maximum() );
 
-               // validated in create_block_state_future()
+               // validated in create_block_token()
                std::get<building_block>(pending->_block_stage).trx_mroot_or_receipt_digests() = b->transaction_mroot;
 
                const bool existing_trxs_metas = !bsp->trxs_metas().empty();
@@ -2956,14 +2956,8 @@ struct controller_impl {
 
 
    // thread safe, expected to be called from thread other than the main thread
-   block_state_legacy_ptr create_block_state_i( const block_id_type& id, const signed_block_ptr& b, const block_header_state_legacy& prev ) {
-      // There is a small race condition at time of activation where create_block_state_i
-      // is called right before hs_irreversible_block_num is set. If that happens,
-      // the block is considered invalid, and the node will attempt to sync the block
-      // in the future and succeed
-      uint32_t instant_finality_lib = hs_irreversible_block_num.load();
-      const bool instant_finality_active = instant_finality_lib > 0;
-      auto trx_mroot = calculate_trx_merkle( b->transactions, instant_finality_active );
+   block_token create_block_state_i( const block_id_type& id, const signed_block_ptr& b, const block_header_state_legacy& prev ) {
+      auto trx_mroot = calculate_trx_merkle( b->transactions, false );
       EOS_ASSERT( b->transaction_mroot == trx_mroot, block_validate_exception,
                   "invalid block transaction merkle root ${b} != ${c}", ("b", b->transaction_mroot)("c", trx_mroot) );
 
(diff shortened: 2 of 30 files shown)


greg7mdp (approved):

libraries/chain/controller.cpp
```diff
@@ -3757,10 +3772,10 @@ void controller::start_block( block_timestamp_type when,
                     bs, std::optional<block_id_type>(), deadline );
 }
 
-void controller::finalize_block( block_report& br, const signer_callback_type& signer_callback ) {
+void controller::finish_block( block_report& br, const signer_callback_type& signer_callback ) {
```
greg7mdp, line 3775: I think I would have called it `complete_block` as it switches to the `completed_block` stage.

libraries/chain/controller.cpp
```diff
@@ -3994,48 +4009,45 @@ std::optional<signed_block_header> controller::fetch_block_header_by_id( const b
...
-   auto blk_state = fetch_block_state_by_number( block_num );
-   if( blk_state ) {
-      return blk_state->header;
-   }
+   auto fetch_block = [&](auto& fork_db, auto& head) -> signed_block_ptr {
+      auto bsp = fork_db.search_on_branch( fork_db.head()->id(), block_num);
+      if (bsp) return bsp->block;
+      return {};
+   };
+   auto b = my->block_data.apply<signed_block_ptr>(fetch_block);
```
greg7mdp, line 4030: Because this is duplicated here and above in `fetch_block_by_number`, and might be useful elsewhere, I think we should define a function in `struct block_data_t `:

```
   signed_block_ptr fetch_block(uint32_t block_num) const {
      return std::visit([&](const auto& bd) -> signed_block_ptr {
         auto bsp = bd.fork_db.search_on_branch(bd.fork_db.head()->id(), block_num);
         if (bsp) return bsp->block;
         return {};
      }, v);
   }
```

libraries/chain/controller.cpp
```diff
@@ -3994,48 +4009,45 @@ std::optional<signed_block_header> controller::fetch_block_header_by_id( const b
...
 
    if( !find_in_blog ) {
-      auto bsp = fetch_block_state_by_number( block_num );
-      if( bsp ) return bsp->id();
+      auto fetch_block_id = [&](auto& fork_db, auto& head) -> std::optional<block_id_type> {
+         auto bsp = fork_db.search_on_branch( fork_db.head()->id(), block_num);
+         if (bsp) return bsp->id();
+         return {};
+      };
+      auto id = my->block_data.apply<std::optional<block_id_type>>(fetch_block_id);
```
greg7mdp, line 4049: Same here, could define in `struct block_data_t`:

```
   block_id_type fetch_block_id(uint32_t block_num) const {
      return std::visit([&](const auto& bd) -> block_id_type {
         auto bsp = bd.fork_db.search_on_branch(bd.fork_db.head()->id(), block_num);
         if (bsp) return bsp->id();
         return {};
      }, v);
   }
```

libraries/chain/controller.cpp
heifner: @arhag mentioned calling it `finish_block` so just went with that. I'm ok with either `finish_block` or `complete_block`.

libraries/chain/controller.cpp

libraries/chain/controller.cpp

libraries/chain/block_state.cpp
```diff
@@ -16,7 +17,29 @@ block_state::block_state(const block_header_state& bhs, deque<transaction_metada
    , block(std::make_shared<signed_block>(signed_block_header{bhs.header})) // [greg todo] do we need signatures?
    , pub_keys_recovered(true) // probably not needed
    , cached_trxs(std::move(trx_metas))
-{}
+{
+   block->transactions = std::move(trx_receipts);
+}
+
+// Used for transition from dbpos to instant-finality
```
greg7mdp, line 24:
```suggestion
// Used for transition from dpos to instant-finality
```

libraries/chain/block_state.cpp
```diff
@@ -16,7 +17,29 @@ block_state::block_state(const block_header_state& bhs, deque<transaction_metada
...
+block_state::block_state(const block_state_legacy& bsp) {
+   block_header_state::id = bsp.id();
+   header = bsp.header;
+   activated_protocol_features = bsp.activated_protocol_features;
+   std::optional<block_header_extension> ext = bsp.block->extract_header_extension(instant_finality_extension::extension_id());
+   assert(ext); // required by current transistion mechanism
```
greg7mdp, line 30:
```suggestion
   assert(ext); // required by current transition mechanism
```

libraries/chain/block_state.cpp
```diff
@@ -16,7 +17,29 @@ block_state::block_state(const block_header_state& bhs, deque<transaction_metada
...
+   const auto& if_extension = std::get<instant_finality_extension>(*ext);
+   assert(if_extension.new_finalizer_policy); // required by current transistion mechanism
```
greg7mdp, line 32:
```suggestion
   assert(if_extension.new_finalizer_policy); // required by current transition mechanism
```

greg7mdp (approved):
Nice, thanks for all the fixes Kevin.

Naming personal perference again, instead of `block_token`, I would prefer `block_handle`, I think it is somewhat more accurate,

heifner:
> Nice, thanks for all the fixes Kevin.
> 
> Naming personal perference again, instead of `block_token`, I would prefer `block_handle`, I think it is somewhat more accurate,

Hmm, either nomenclature works I think. Not sure which I would consider more accurate.

libraries/chain/block_state.cpp
heifner: Hmm, I wonder if my editor last update broke the spell-checker. It is not flagging that.

libraries/chain/block_state.cpp
heifner: Fixed

libraries/chain/block_state.cpp
heifner: Fixed

libraries/chain/block_state.cpp
heifner: Fixed

libraries/chain/controller.cpp
```diff
@@ -177,6 +177,19 @@ struct block_data_t {
...
+
+   void transition_fork_db_to_if(const auto& vbsp) {
+      std::visit([](auto& bd) { bd.fork_db.close(); }, v);
+      auto bsp = std::make_shared<block_state>(*std::get<block_state_legacy_ptr>(vbsp));
+      v.emplace<block_data_new_t>(std::visit([](const auto& bd) { return bd.fork_db.get_data_dir(); }, v));
+      std::visit(overloaded{
+                    [](block_data_legacy_t&) {},
+                    [&](block_data_new_t& bd) {
+                       bd.head = bsp;
+                       bd.fork_db.reset(*bd.head);
```
linh2931, line 189: Do you need to re-open fork_db?

libraries/chain/controller.cpp
```diff
@@ -278,6 +291,22 @@ struct block_data_t {
...
 
+   signed_block_ptr fork_db_fetch_block_by_num(uint32_t block_num) const {
+      return std::visit([&](const auto& bd) -> signed_block_ptr {
+         auto bsp = bd.fork_db.search_on_branch(bd.fork_db.head()->id(), block_num);
+         if (bsp) return bsp->block;
+         return {};
+      }, v);
+   }
+
+   std::optional<block_id_type> fork_db_fetch_block_id(uint32_t block_num) const {
```
linh2931, line 313: Can be named as `fork_db_fetch_block_id_by_num` to match `fork_db_fetch_block_by_num`.

libraries/chain/controller.cpp
linh2931: I think `finish_block` highlights the last step of those block building stages. Sounds more natural.

libraries/chain/controller.cpp
```diff
@@ -3979,63 +4023,47 @@ signed_block_ptr controller::fetch_block_by_id( const block_id_type& id )const {
    return signed_block_ptr();
 }
 
+bool controller::block_exists(const block_id_type&id) const {
+   auto sb_ptr = my->block_data.fork_db_fetch_block_by_id(id);
```
linh2931, line 4035: What is `sb` in `sb_ptr`?

libraries/chain/controller.cpp
```diff
@@ -2707,8 +2718,28 @@ struct controller_impl {
...
+         auto transition = [&](auto& fork_db, auto& head) -> bool {
+            const auto& bsp = std::get<std::decay_t<decltype(head)>>(cb.bsp);
+            std::optional<block_header_extension> ext = bsp->block->extract_header_extension(instant_finality_extension::extension_id());
+            if (ext) {
+               const auto& if_extension = std::get<instant_finality_extension>(*ext);
+               if (if_extension.new_finalizer_policy) {
+                  ilog("Transition to instant finality happening after block ${b}", ("b", bsp->block_num()));
+                  hs_irreversible_block_num = bsp->block_num();
+
+                  log_irreversible();
```
linh2931, line 2741: Will `log_irreversible` be executed twice as there was one above?

linh2931 (commented):
I like your combining `block_state_legacy` and `block_state` into `block_token`. The fewer the `_legacy`, the better.

libraries/chain/controller.cpp
```diff
@@ -2986,14 +3042,14 @@ struct controller_impl {
          } );
       };
 
-      return block_data.apply_dpos<std::future<block_state_legacy_ptr>>(f); // [greg todo] make it work with apply()
+      return block_data.apply<std::future<block_token>>(f); // [greg todo] make it work with apply()
```
linh2931, line 3045: Do you still need this todo comment?

linh2931:
> > Nice, thanks for all the fixes Kevin.
> > Naming personal perference again, instead of `block_token`, I would prefer `block_handle`, I think it is somewhat more accurate,
> 
> Hmm, either nomenclature works I think. Not sure which I would consider more accurate. Open to changing the name. Thoughts @arhag @linh2931 ?

Either is fine. Or `block_ptr` or `block_ctx` to be shorter?

libraries/chain/controller.cpp
heifner: No, nothing to read from disk and we are using the same path of the previous one. It would be a no-op. Although, might be clearer to call it anyway.

libraries/chain/controller.cpp
heifner: It will be executed twice. But `if_irreversible_block_num` is set before the second call. `log_irreversible` determines internally if it should do anything. Always safe to call it multiple times.

libraries/chain/controller.cpp
```diff
@@ -1196,8 +1196,8 @@ struct controller_impl {
                      ("lib_num", lib_num)("bn", fork_db_root_block_num()) );
       }
 
-      const uint32_t hs_lib = hs_irreversible_block_num;
-      const uint32_t new_lib = hs_lib > 0 ? hs_lib : fork_db_head_irreversible_blocknum();
+      const uint32_t if_lib = if_irreversible_block_num;
+      const uint32_t new_lib = if_lib > 0 ? if_lib : fork_db_head_irreversible_blocknum();
```
linh2931, line 1200: This `if_lib > 0` really looks like a regular `if` statement😺 We can try to reduce `if_` prefix.

libraries/chain/controller.cpp
heifner: No, removed.

libraries/chain/controller.cpp
heifner: signed_block_header. Added type to make it clear.

libraries/chain/controller.cpp
heifner: What about we standardize on `inf`: `inf_lib`, `set_inf_irreversible_block_number` ? @greg7mdp @linh2931

linh2931: @arhag mentioned we should try to avoid `instant` and just say `finality`.

linh2931: We can clean up those later.

greg7mdp: I'm OK with `inf_lib`, although I thought `if_lib` was fine too, and I don't think `inf` is clearer.
But then I think the set function should be `set_inf_lib`.

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Initial early version of transition from legacy finality to new Faster Finality. Additional work required on transition to address all scenarios.  
Note: end

Issue #2069, opened 2024-01-10
IF: Unification: Sign finalizer_digest of received block header and send out vote


arhag:
First validate the block header and use it to compute the next `block_header_state`. The `block_header_state` should compute the `finalizer_digest`. This should be done as part of receiving the block.

Afterward, without blocking the signal of receiving the block, the configured finalizers of the node should attempt to sign that block asynchronously, either the `finalizer_digest` or the weak digest computed from the `finalizer_digest` and a tag indicating WEAK, depending on whether the finalizer chooses to sign strong or weak, respectively. A new signal will need to be added to let the net_plugin know there is a new vote message that is ready to broadcast.

For this issue, the finalizer should always sign strongly. This will be refined in a later issue.

Issue #2071, opened 2024-01-10
IF: Reject blocks with unsupported claims of last QC in block header extension


arhag:
The block header extension for instant finality makes a claim of the block number of the last block in its ancestry for which it has seen a QC. It also makes a claim as to whether this QC was strong or weak.

The node should not simply trust these claims. It needs to validate them somehow.

One way of validating this is if the claim is the same as the claim in the prior block that was already validated. This can be seen by looking at the header extension within the header stored in the `block_header_state` that the block builds upon.

But often the claim will be new. And in this case, to verify the claim, the node must verify that a valid QC (of the appropriate type) exists on the claimed block. Note on appropriate type: if the claim is that a strong QC exists, the node must verify it has a strong QC on the block; if the claim is that a weak QC exists, the node must verify it has some QC (strong or weak) on the block.

The block should include the necessary QC in its block extension. So one way to carry out this verification process is to first integrate the QC provided from that block (after checking it is valid) into the `block_state` associated with the claimed block. This should be done anyway because the QC may be valuable information to keep around even if the received (new) block is abandoned anyway due to a fork reorganization or because we later realize the received block is invalid and needs to be thrown away.

Note: When integrating the QC into the `block_state` of the claimed block, it may provide the information/proof needed to allow LIB to advance. LIB should be advanced at this point (or shortly after) without having to wait for a new block to be built/completed or for a new vote message to be received.

Then validating the claim in the received block always involves looking up the `block_state` for the claimed block and ensuring it has a valid QC on it of the appropriate type.

A note on integrating the QC into the `block_state`:
The `block_state` may have a pending quorum certificate that the node is collecting votes for. This should not be touched when integrating a received QC into the `block_state`; there is a separate optional field for the `valid_qc`. The received QC should be placed in the field for the `valid_qc` but only if is better than any existing one in `valid_qc`. A strong valid QC is better than a weak valid QC. For now, do not bother comparing two different valid QC of the same type; they can be treated as equally good.
When observing the QC available for a particular block, we want to pick the best one available. This means that an accessor function `get_best_qc` should compare the pending quorum certificate (assuming it is valid) against the `valid_qc` and choose the better of the two (again better in this sense can just consider strong vs weak).

PR #2087, opened 2024-01-15
IF: Unification: Sign finalizer_digest of received block header and send out vote

linh2931:
* Controller signs `finalizer_digest` of received block and emits a signal to net plugin
* Net plugin broadcasts the vote on receiving the signal

Note: 
* This PR only signs strong finalizer_digest.
* Calling of `create_and_send_vote_msg` will be in another final PR.

Resolved https://github.com/AntelopeIO/leap/issues/2069

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -2832,21 +2832,16 @@ struct controller_impl {
 
       // A vote is created and signed by each finalizer configured on the node that
       // in active finalizer policy
-#warning possible optimization: when setting finalizer policy, update finalizer_keys_on_the_node to indicate if it is active to save going over bhs.finalizer_policy->finalizers
-      // go over each active finalizer and check if it is configured on the node
       for (const auto& f: bhs.finalizer_policy->finalizers) {
          auto it = finalizer_keys_on_the_node.find( f.public_key );
          if( it != finalizer_keys_on_the_node.end() ) {
             const auto& private_key = it->second;
+            const auto& digest = bhs.compute_finalizer_digest();
 
-            // signing can take long time, do it asynchronously
-            auto sig_fut = post_async_task(thread_pool.get_executor(),
-                                           [&]() {
-                                              const auto& digest = bhs.compute_finalizer_digest();
-                                              return private_key.sign(std::vector<uint8_t>(digest.data(), digest.data() + digest.data_size())); });
+            auto sig =  private_key.sign(std::vector<uint8_t>(digest.data(), digest.data() + digest.data_size()));
 
             // construct the vote message
-            hs_vote_message vote{ bhs.id, strong, private_key.get_public_key(), sig_fut.get() };
+            hs_vote_message vote{ bhs.id, strong, private_key.get_public_key(), sig };
 
             // net plugin subscribed this signal. it will broadcast the vote message
             // on receiving the signal
@@ -3436,7 +3431,7 @@ struct controller_impl {
       wasmif.code_block_num_last_used(code_hash, vm_type, vm_version, block_num);
    }
 
-   void set_finalizer_keys_on_the_node(const bls_pub_priv_key_map_t finalizer_keys) {
+   void set_finalizer_keys_on_the_node(const bls_pub_priv_key_map_t& finalizer_keys) {
       for (const auto& k : finalizer_keys) {
          finalizer_keys_on_the_node[bls_public_key{k.first}] = bls_private_key{k.second};
       }
@@ -4496,7 +4491,7 @@ void controller::code_block_num_last_used(const digest_type& code_hash, uint8_t 
    return my->code_block_num_last_used(code_hash, vm_type, vm_version, block_num);
 }
 
-void controller::set_finalizer_keys_on_the_node(const bls_pub_priv_key_map_t finalizer_keys) {
+void controller::set_finalizer_keys_on_the_node(const bls_pub_priv_key_map_t& finalizer_keys) {
    my->set_finalizer_keys_on_the_node(finalizer_keys);
 }
 
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -884,7 +884,7 @@ struct controller_impl {
    named_thread_pool<chain>        thread_pool;
    deep_mind_handler*              deep_mind_logger = nullptr;
    bool                            okay_to_print_integrity_hash_on_stop = false;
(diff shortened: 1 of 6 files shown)


libraries/chain/controller.cpp
```diff
@@ -3407,6 +3436,12 @@ struct controller_impl {
       wasmif.code_block_num_last_used(code_hash, vm_type, vm_version, block_num);
    }
 
+   void set_finalizer_keys_on_the_node(const bls_pub_priv_key_map_t finalizer_keys) {
```
heifner, line 3439: `const&`

libraries/chain/controller.cpp
```diff
@@ -4461,6 +4496,10 @@ void controller::code_block_num_last_used(const digest_type& code_hash, uint8_t
    return my->code_block_num_last_used(code_hash, vm_type, vm_version, block_num);
 }
 
+void controller::set_finalizer_keys_on_the_node(const bls_pub_priv_key_map_t finalizer_keys) {
```
heifner, line 4499: `const&`

libraries/chain/controller.cpp
```diff
@@ -2825,6 +2826,34 @@ struct controller_impl {
...
+            const auto& private_key = it->second;
+
+            // signing can take long time, do it asynchronously
+            auto sig_fut = post_async_task(thread_pool.get_executor(),
+                                           [&]() {
+                                              const auto& digest = bhs.compute_finalizer_digest();
+                                              return private_key.sign(std::vector<uint8_t>(digest.data(), digest.data() + digest.data_size())); });
+
+            // construct the vote message
+            hs_vote_message vote{ bhs.id, strong, private_key.get_public_key(), sig_fut.get() };
```
heifner, line 2849: The `sig_fut.get()` blocks until async task is complete. You should loop through twice. Once to create a vector of futures and second time to create vote and emit.

Although in practice nodes will only have one finalizer key configured (except for our integration tests). So this optimization is not useful. Also unless we can do something else there is no point in doing the signing async and this extra post async is just added overhead.

libraries/chain/controller.cpp
```diff
@@ -2825,6 +2826,34 @@ struct controller_impl {
       }
    } FC_CAPTURE_AND_RETHROW() } /// apply_block
 
+   void create_and_send_vote_msg(const block_header_state& bhs) {
+#warning use decide_vote() for strong after it is implementd by https://github.com/AntelopeIO/leap/issues/2070
+      bool strong = true;
+
+      // A vote is created and signed by each finalizer configured on the node that
+      // in active finalizer policy
+#warning possible optimization: when setting finalizer policy, update finalizer_keys_on_the_node to indicate if it is active to save going over bhs.finalizer_policy->finalizers
```
heifner, line 2835: Not sure how that would work. The active finalizer_policy will constantly be changing.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -61,7 +61,7 @@ struct block_header_state {
    flat_map<uint32_t, finalizer_policy_ptr> finalizer_policies;
 
    // ------ functions -----------------------------------------------------------------
-   digest_type           compute_finalizer_digest() const;
+   digest_type           compute_finalizer_digest() const { return id; };
```
heifner, line 64: Add a TODO that this is temporary.

libraries/chain/controller.cpp
heifner: Add call in `producer_plugin` startup.

heifner (changes requested):

libraries/chain/controller.cpp
greg7mdp: Why don't we construct the ` hs_vote_message` in the async_task, and `emit` the `future ` as a parameter to `self.voted_block`? `net_plugin` can do the wait for the future in one of its many threads.

heifner: Would have to be a `shared_future`, but seem like a great idea.

libraries/chain/controller.cpp
linh2931: Thanks! I missed that.

libraries/chain/controller.cpp
linh2931: Thanks. For the first prototype, I change to do signing directly.

libraries/chain/controller.cpp
linh2931: Will remove the warning.

libraries/chain/controller.cpp
greg7mdp: The issue mentions that the digest should be signed asynchronously. If we merge this implementation it won't fully resolve the issue.

linh2931: Why `shared_future`? Because signal can only be copied?

I will ask @aheg if asynchronous signing is required for the prototype tomorrow.

heifner: Because signals are sent to multiple connections to the signal. You will need to be able to `.get()` from multiple.

libraries/chain/controller.cpp
```diff
@@ -2825,6 +2826,29 @@ struct controller_impl {
       }
    } FC_CAPTURE_AND_RETHROW() } /// apply_block
 
+   void create_and_send_vote_msg(const block_header_state& bhs) {
```
greg7mdp, line 2829: I don't see where this function is called.

linh2931: This PR was created before code for processing IF received block was ready. The PR provides the plumbing. The function will be called in create_block_state_i after the block is validated (https://github.com/AntelopeIO/leap/issues/2071)

heifner (approved):

libraries/chain/controller.cpp
```diff
@@ -884,6 +884,7 @@ struct controller_impl {
    named_thread_pool<chain>        thread_pool;
    deep_mind_handler*              deep_mind_logger = nullptr;
    bool                            okay_to_print_integrity_hash_on_stop = false;
+   bls_key_map_t                   finalizer_keys_on_the_node;
```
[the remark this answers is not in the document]
linh2931: Changed to `node_finalizer_keys`.

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Code to sign finalizer digest and send a signal to net plugin. Additional work needed to handle additional scenarios. 
Note: end

PR #2011, opened 2023-12-20
IF: add latest commit QC known to proposer to block extension (part 1)

linh2931:
This is the first part of adding QC to block extension, implementing the definition and serialization. The remaining work is the insert of QC, which will be done after new `block_header_state` is finished (#1941).

Partly resolved https://github.com/AntelopeIO/leap/issues/1586

PR #2090, opened 2024-01-16
IF: Unification: add latest commit QC known to proposer to block extension (final part)

linh2931:
Continuation of https://github.com/AntelopeIO/leap/pull/2011. Add QC to block extension if needed.

Resolved https://github.com/AntelopeIO/leap/issues/1586


libraries/chain/controller.cpp
```diff
@@ -548,8 +548,13 @@ struct assembled_block {
...
                                    },
                                    [&](assembled_block_if& ab) {
+                                      std::optional<quorum_certificate> qc;
+                                      if( ab.qc && ab.bhs.is_needed(*ab.qc ) ) {
+                                         qc = *ab.qc;
+                                      }
+
                                       auto bsp = std::make_shared<block_state>(ab.bhs, std::move(ab.trx_metas),
-                                                                               std::move(ab.trx_receipts));
+                                                                               std::move(ab.trx_receipts), qc);
```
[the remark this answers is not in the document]
heifner: Actually, I see no reason to copy it at all. Pass the `ab.qc` to `block_state` constructor as `const std::optional<quorum_certificate>&`  and do the `is_needed` check in block_state constructor and pack `*qc` there.

heifner (changes requested):

libraries/chain/controller.cpp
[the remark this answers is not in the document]
linh2931: Great!

libraries/chain/controller.cpp
```diff
@@ -549,7 +549,7 @@ struct assembled_block {
                                    },
                                    [&](assembled_block_if& ab) {
                                       auto bsp = std::make_shared<block_state>(ab.bhs, std::move(ab.trx_metas),
-                                                                               std::move(ab.trx_receipts));
+                                                                               std::move(ab.trx_receipts), std::move(ab.qc));
```
heifner, line 552: This should not be a `std::move`

greg7mdp: > This should not be a `std::move`

Why not? I think the constructor should take a rvalue and it should be a std::move.

heifner: If we cache it in `block_state` then makes sense to move it, otherwise faster to just pass by value.

linh2931: We don't store it in `bloclk_state`. Do you want me to change to pass by value?

greg7mdp: You can leave it as is, thanks Lin. We can change it if the QC needs to be stored in block_state.

greg7mdp (approved):

heifner (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Continuing PR #2011 add quorum certificate to block header extension.
Note:end

PR #2094, opened 2024-01-16
IF: Remove `fork_database::get_block_header` and simplify `create_block_s…

### LINKED ISSUE #2283: IF: Change Fork_db root to store block state
Currently fork_db's root only stores `block_header_state`. https://github.com/AntelopeIO/leap/issues/2080 requires copying `valid` structure from previous `block_state` to current one. This causes problems when previous block is the root.

* Change fork_db root to store `block_state`.
* Provide a new API to return `block_state` including root while keeping existing APIs unchanged
* Make sure old-version fork_db with the root storing `block_header_state` can be opened.

greg7mdp:
Resolves #2283 .

Remove `fork_database::get_block_header` and simplify `create_block_state_i`.

Make sure we can still load Leap V5 fork databases from disk.

The only difference between `fork_database::get_block_header` and `fork_database::get_block` was that the former checked the `root`. It was also relying on the fact that `block_state` is `public block_header_state` to return the same pointer downcasted.

I have added an optional `check_root` parameter to `fork_database::get_block` instead.

This allowed to make `create_block_state_i` a template instead of having two separate implementations.

### DIFF
--- a/libraries/chain/fork_database.cpp
+++ b/libraries/chain/fork_database.cpp
@@ -77,8 +77,7 @@ namespace eosio::chain {
       void             close_impl();
       void             add_impl( const bsp& n, bool ignore_duplicate, bool validate, validator_t& validator );
 
-      bhsp             get_block_header_impl( const block_id_type& id ) const;
-      bsp              get_block_impl( const block_id_type& id ) const;
+      bsp              get_block_impl( const block_id_type& id, bool check_root = false ) const;
       void             reset_impl( const bhs& root_bhs );
       void             rollback_head_to_root_impl();
       void             advance_root_impl( const block_id_type& id );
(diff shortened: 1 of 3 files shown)


libraries/chain/fork_database.cpp
```diff
@@ -589,13 +569,17 @@ namespace eosio::chain {
...
       std::shared_lock g( my->mtx );
-      return my->get_block_impl(id);
+      return my->get_block_impl(id, check_root);
    }
 
    template<class bsp>
-   bsp fork_database_impl<bsp>::get_block_impl(const block_id_type& id) const {
+   bsp fork_database_impl<bsp>::get_block_impl(const block_id_type& id, bool check_root /* = false */) const {
+      if( check_root && root->id() == id ) {
+         return root;
```
heifner, line 580: This is not valid. `root` can just contain `block_header` when it is read from disk. I guess we will see how good of test coverage we have. Should hopefully get some test failures on this PR.

heifner (changes requested):

libraries/chain/fork_database.cpp
greg7mdp: > root can just contain block_header when it is read from disk.

Is there a reason that the root is different? Why not store the full block_state? Also I see we do `head = root` so maybe it is the same for `head`? Also the type for `head` and `root` in fork_db is `block_state`, not `block_header_state`.

Also we call `fork_db::reset()` with a `block_state` even though it takes a `block_header_state`. This is the kind of issues which make it difficult to not have `block_state` inherit from `block_header_state`.

I think clearly this PR will not cause any issue, since the two calls to `fork_db::get_block_header()`  did not (and could not) use any member that is not part of `block_header_state`. I think it would be better if fork_db uniforly stored `block_state` structs and not one or the other depending on where you look.

heifner: I don't disagree, but we have to support the legacy format which does store only the `block_header_state_legacy`.

heifner: I think the issue would be if someone passes `true` to `check_root` when they are expecting a full `block_state`.

greg7mdp: > I think the issue would be if someone passes `true` to `check_root` when they are expecting a full `block_state`.

Yes, someone doing that would quickly realize they don't get what they expect. The usage is so limited though (just in create_block_token_*) that I wonder if this is a big issue.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Consolidate `fork_database` block data structures to simplify code pathways. 
Note: end

PR #2105, opened 2024-01-18
IF: Fixes for IF to advance LIB

### LINKED ISSUE #2100: IF: Update integration test that checks LIB advances after IF transition
The test is `nodeos_run_test`.

heifner:
New integration test that verifies that LIB advances on all nodes under instant-finality.

Fixes for instant-finality:
- Fix vote message processing in `net_plugin`
- Add missing `strong` serialization in vote message.
- Update `fork_database` `block_state` implementation to use `core.last_final_block_num` for `irreversible_blocknum()`
- Update LIB according to vote processing.
- Vote when we produce a block.

Reviewers: Pay close attention to validation logic for `qc_info`. Not sure this is the best way to handle block validation with `qc_info`.

Works:  `./tests/nodeos_lib_test.py -n 4 -p 3 -s mesh --activate-if --keep-logs -v`
Works: `./tests/nodeos_lib_test.py -n 4 -p 3 -s ring --activate-if --keep-logs -v`

Resolves #2100


libraries/chain/block_header_state.cpp
```diff
@@ -117,9 +118,8 @@ block_header_state block_header_state::next(block_header_state_input& input) con
    result.active_finalizer_policy = active_finalizer_policy;
 
    // [greg todo] correct support for new finalizer_policy activation using finalizer_policies map
-
-   if (input.new_finalizer_policy)
-      ++input.new_finalizer_policy->generation;
+   //   if (input.new_finalizer_policy)
```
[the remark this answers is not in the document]
heifner: Yes. We have to design the finalizer policy change algorithm.

libraries/chain/block_header_state.cpp
```diff
@@ -131,7 +131,9 @@ block_header_state block_header_state::next(block_header_state_input& input) con
    instant_finality_extension new_if_ext {if_ext.qc_info,
                                           std::move(input.new_finalizer_policy),
                                           std::move(input.new_proposer_policy)};
-   if (input.qc_info)
+   if (input.validating)
+      new_if_ext.qc_info = input.qc_info;
```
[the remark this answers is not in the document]
heifner: Because if it is empty `std::optional` then that is what we want.

libraries/chain/controller.cpp
```diff
@@ -4266,7 +4322,11 @@ void controller::get_finalizer_state( finalizer_state& fs ) const {
 
 // called from net threads
 bool controller::process_vote_message( const hs_vote_message& vote ) {
-   return my->block_data.aggregate_vote(vote);
+   auto [valid, new_lib] = my->block_data.aggregate_vote(vote);
+   if (new_lib) {
+      my->if_irreversible_block_num = *new_lib;
```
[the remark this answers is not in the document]
heifner: We `db.commit()` in `log_irreversible`, so don't want that called from net thread.

libraries/chain/controller.cpp
```diff
@@ -913,13 +912,17 @@ struct building_block {
...
-               for( auto it = branch.begin(); it != branch.end(); ++it ) {
-                  auto qc = (*it)->get_best_qc();
-                  if( qc ) {
-                     EOS_ASSERT( qc->block_height <= block_header::num_from_id(parent_id()), block_validate_exception, "most recent ancestor QC block number (${a}) cannot be greater than parent's block number (${p})", ("a", qc->block_height)("p", block_header::num_from_id(parent_id())) );
-                     qc_data = qc_data_t{ *qc, qc_info_t{ qc->block_height, qc->qc.is_strong() }};
-                     break;
+               if (!validating) {
+                  auto branch = fork_db.fetch_branch(parent_id());
+                  for( auto it = branch.begin(); it != branch.end(); ++it ) {
+                     auto qc = (*it)->get_best_qc();
```
greg7mdp, line 918: If we are validating, we are not retrieving the qc here, so the one we pass to the `assembled_block_if` below is not  initialized.

libraries/chain/block_state.cpp
```diff
@@ -9,12 +9,18 @@ block_state::block_state(const block_header_state& prev, signed_block_ptr b, con
...
+   , pending_qc(prev.active_finalizer_policy->finalizers.size(), prev.active_finalizer_policy->threshold)
 {}
 
 block_state::block_state(const block_header_state& bhs, deque<transaction_metadata_ptr>&& trx_metas,
                          deque<transaction_receipt>&& trx_receipts, const std::optional<quorum_certificate>& qc)
    : block_header_state(bhs)
    , block(std::make_shared<signed_block>(signed_block_header{bhs.header})) // [greg todo] do we need signatures?
+   , strong_digest(compute_finalizer_digest())
+   , weak_digest(compute_finalizer_digest())
+   , pending_qc(bhs.active_finalizer_policy->finalizers.size(), bhs.active_finalizer_policy->threshold)
```
greg7mdp, line 23: We'll also need to pass the weights, so it probably would be better to have the `pending_quorum_certificate` constructor take a `finalizer_policy`. But we can make that change later.

libraries/chain/block_state.cpp
```diff
@@ -29,6 +35,7 @@ block_state::block_state(const block_header_state& bhs, deque<transaction_metada
 block_state::block_state(const block_state_legacy& bsp) {
    block_header_state::id = bsp.id();
    header = bsp.header;
+   core.last_final_block_num = bsp.block_num();
```
greg7mdp, line 38: Probably add a `todo` because this will change when we dont transition immediately when executing set_finalizers.

libraries/chain/block_state.cpp
```diff
@@ -67,15 +74,16 @@ bool block_state::aggregate_vote(const hs_vote_message& vote) {
...
 #warning TODO change to use std::span if possible
                                  std::vector<uint8_t>{digest.data(), digest.data() + digest.data_size()},
                                  index,
                                  vote.finalizer_key,
                                  vote.sig);
+      return {valid, strong ? core.final_on_strong_qc_block_num : std::optional<uint32_t>{}};
    } else {
       wlog( "finalizer_key (${k}) in vote is not in finalizer policy", ("k", vote.finalizer_key) );
-      return false;
+      return {false, {}};
```
greg7mdp, line 86: small detail, `return {};` is equivalent.

libraries/chain/controller.cpp
```diff
@@ -857,7 +854,9 @@ struct building_block {
 
    assembled_block assemble_block(boost::asio::io_context& ioc,
                                   const protocol_feature_set& pfs,
-                                  const block_data_t& block_data) {
+                                  const block_data_t& block_data,
+                                  bool validating,
+                                  std::optional<qc_info_t> validating_qc_info) {
```
greg7mdp, line 859: should take `qc_data_t`.

Instead of two parameters `validating` and `validating_qc_info`, what about we pass a single parameter:

`std::optional<incoming_block_data>` with

```
struct incoming_block_data {
   std::optional<qc_data_t> qc_data;
};
```

That way if we need to pass more stuff from the incoming block we can add it to `incoming_block_data`. It will simplify the `finish_block() interface as well.

If you want I'm happy to make the change in a derived branch.

libraries/chain/controller.cpp
```diff
@@ -2875,6 +2887,23 @@ struct controller_impl {
       EOS_REPORT( "new_producers", b.new_producers, ab.new_producers )
       EOS_REPORT( "header_extensions", b.header_extensions, ab.header_extensions )
 
+      if (b.header_extensions != ab.header_extensions) {
+         {
```
greg7mdp, line 2891: I don't think we need this extra block

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: TEST
summary: Test to ensure LIB advances.
Note: end

Issue #2104, opened 2024-01-18
`read-only-threads` is required to submit `/v1/chain/send_read_only_transaction`


aaroncox:
On v5.0.0, without providing `read-only-theads` in the config.ini - the `/v1/chain/send_read_only_transaction` API endpoint does not work. 

Against a Jungle4 testnet node, if I execute the following query:

```
curl http://localhost:8888/v1/chain/send_read_only_transaction -d '{"transaction":{"signatures":[],"compression":1,"packed_context_free_data":"789c63000000010001","packed_trx":"789c636040068c06c71c8e2f48be60c830a52fedfceacdbb1838c0e2528bb31800a31509a1"}}'
```

The response from curl is:

```
curl: (52) Empty reply from server
```

Now if I modify the config and add `read-only-theads = 4`, and rerun the same curl command, I get a response:

```
[...]

PR #2106, opened 2024-01-18
[5.0] more robust `chain_api_plugin` check in `producer_plugin`

spoonincode:
Make this check more robust to handle the esoteric config pattern in #2104

Tests needed?

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -1214,7 +1214,7 @@ void producer_plugin_impl::plugin_initialize(const boost::program_options::varia
    } else if (_producers.empty()) {
       //appbase initializes configured plugins before auto-start plugins, so if chain_api_plugin is enabled it's
       // initialized before producer_plugin (i.e. before this code here)
-      if (app().get_plugin("eosio::chain_api_plugin").get_state() == abstract_plugin::initialized) {
+      if (abstract_plugin* capi = app().find_plugin("eosio::chain_api_plugin"); capi && capi->get_state() == abstract_plugin::initialized) {
          // default to 3 threads for non producer nodes running chain_api_plugin if not specified
          _ro_thread_pool_size = 3;
          ilog("chain_api_plugin configured, defaulting read-only-threads to ${t}", ("t", _ro_thread_pool_size));
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -1212,14 +1212,12 @@ void producer_plugin_impl::plugin_initialize(const boost::program_options::varia
    if (options.count("read-only-threads")) {
       _ro_thread_pool_size = options.at("read-only-threads").as<uint32_t>();
    } else if (_producers.empty()) {
-      if (options.count("plugin")) {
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1212,14 +1212,12 @@ void producer_plugin_impl::plugin_initialize(const boost::program_options::varia
...
-      if (options.count("plugin")) {
-         const auto& v = options.at("plugin").as<std::vector<std::string>>();
-         auto        i = std::find_if(v.cbegin(), v.cend(), [](const std::string& p) { return p == "eosio::chain_api_plugin"; });
-         if (i != v.cend()) {
-            // default to 3 threads for non producer nodes running chain_api_plugin if not specified
-            _ro_thread_pool_size = 3;
-            ilog("chain_api_plugin configured, defaulting read-only-threads to ${t}", ("t", _ro_thread_pool_size));
-         }
+      //appbase initializes configured plugins before auto-start plugins, so if chain_api_plugin is enabled it's
+      // initialized before producer_plugin (i.e. before this code here)
```
spoonincode, line 1216: https://github.com/AntelopeIO/appbase/blob/b75b31e14f966fa3de6246e120dcba36c6ce5264/application_base.cpp#L397-L416

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1212,14 +1212,12 @@ void producer_plugin_impl::plugin_initialize(const boost::program_options::varia
...
+      if (app().get_plugin("eosio::chain_api_plugin").get_state() == abstract_plugin::initialized) {
+         // default to 3 threads for non producer nodes running chain_api_plugin if not specified
+         _ro_thread_pool_size = 3;
+         ilog("chain_api_plugin configured, defaulting read-only-threads to ${t}", ("t", _ro_thread_pool_size));
```
spoonincode, line 1220: Another approach would have been to change the previous check from `==` to `find()`, but I like asking appbase more directly instead of fishing in config items that this plugin doesn't "own".

The downside is that the check is reliant on plugin initialization ordering semantics that are arguably an implementation detail of appbase, and that `producer_plugin` will always remain an auto-start plugin. (Not for 5.0 but possibly in main) appbase might should have a new plugin state: `enabled`, and all plugins get marked enabled before they get `initalize()`ed, to avoid this dependency on ordering. (It's possible this isn't a plugin `state()` but rather some other new `is_enabled()` property on the plugin)

heifner: How about adding `(producer_plugin)` to `APPBASE_PLUGIN_REQUIRES((chain_plugin)(http_plugin))` of `chain_api_plugin.hpp`.

spoonincode: That would require adding `producer_plugin` as a build dependency of `chain_api_plugin` -- `target_link_libraries()` etc. kinda gnarly when they really don't depend on each other.

heifner (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: STABILITY
category: BUG
summary: Add more robust configuration parsing and enhance check for `chain_api_plugin`. This change addresses failure to provide read-only transaction when using esoteric config pattern.
Note:end

PR #2111, opened 2024-01-19
IF: Simplify code path when validating a received block.

greg7mdp:


heifner (approved):

libraries/chain/block_state.cpp
```diff
@@ -26,7 +26,7 @@ block_state::block_state(const block_header_state& bhs, deque<transaction_metada
 {
    block->transactions = std::move(trx_receipts);
 
-   if( qc && bhs.is_needed(*qc) ) {
+   if( qc ) {
```
linh2931, line 29: Why is `is_needed` removed? We now always include QC block extension?

linh2931 (approved):

libraries/chain/block_state.cpp
greg7mdp: > Why is `is_needed` removed? We now always include QC block extension?

The parameter passed is a `std::optional<quorum_certificate>`, and the optional is empty when the qc should not be included. Whenever a qc is passed, it should always be included. 
With the check in this branch, the test doesn't pass.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Code simplification when validating a received block
Note: end

PR #2113, opened 2024-01-19
IF: Support load of fork database after instant-finality is enabled

### LINKED ISSUE #2048: IF: Persist post-IF fork database to disk on shutdown

### LINKED ISSUE #2083: IF: fork database transition not thread safe

heifner:
- Supports writing/reading of `fork_database` in legacy and after instant-finality is activated.
- Moves mutex out of fork database implementation and into `fork_database` to protect transition of impl from legacy to instant finality.
- Includes some refactoring:
  - Moves `controller::head` which was in `controller block_data_gen_t::head` to `fork_database_t::chain_head`
  - Removes `block_data_gen_t` and `block_data_t`

Resolves #2048 
Resolves #2083


libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -95,15 +98,77 @@ namespace eosio::chain {
...
-   
+   using fork_database_legacy_t = fork_database_t<block_state_legacy_ptr>;
+   using fork_database_if_t     = fork_database_t<block_state_ptr>;
+
+   /**
+    * Provides thread safety on fork_database_t and provide mechanism for opening the correct type
+    * as well as switching from legacy to instant-finality.
+    */
+   class fork_database {
+      mutable std::recursive_mutex m;
```
heifner, line 113: I'm not a fan of recursive mutexes. However, currently controller uses nested calls of `fork_database.apply`. We can look at refactoring those so that no `apply` calls functions which also calls `apply` but that will take some work.

libraries/chain/fork_database.cpp
```diff
@@ -48,9 +39,9 @@ namespace eosio::chain {
       using bhsp             = bs::bhsp_t;
       using bhs              = bhsp::element_type;
       
-      using fork_database_t  = fork_database<bsp>;
-      using branch_type      = fork_database_t::branch_type;
-      using branch_type_pair = fork_database_t::branch_type_pair;
+      using fork_db_t  = fork_database_t<bsp>;
```
greg7mdp, line 42: very minor ... align `=`

libraries/chain/fork_database.cpp
```diff
@@ -251,20 +226,14 @@ namespace eosio::chain {
...
+               ("filename", fork_db_file) );
       }
 
       index.clear();
    }
 
    template<class bsp>
-   fork_database<bsp>::~fork_database() {
-      my->close_impl();
-   }
```
greg7mdp, line 263: Why do we not need the destructor anymore?

libraries/chain/controller.cpp
```diff
@@ -904,26 +664,24 @@ struct building_block {
...
-                  auto branch = fork_db.fetch_branch(parent_id());
-                  for( auto it = branch.begin(); it != branch.end(); ++it ) {
-                     auto qc = (*it)->get_best_qc();
-                     if( qc ) {
-                        EOS_ASSERT( qc->block_height <= block_header::num_from_id(parent_id()), block_validate_exception,
-                                    "most recent ancestor QC block number (${a}) cannot be greater than parent's block number (${p})",
-                                    ("a", qc->block_height)("p", block_header::num_from_id(parent_id())) );
-                        qc_data = qc_data_t{ *qc, qc_info_t{ qc->block_height, qc->qc.is_strong() }};
-                        break;
+               // get fork_database so that we can search for the best qc to include in this block.
```
greg7mdp, line 671: indent comment

libraries/chain/controller.cpp
```diff
@@ -2538,21 +2466,20 @@ struct controller_impl {
...
+      fork_db.apply_dpos<void>([&](auto& forkdb) {
+         maybe_session session = self.skip_db_sessions(s) ? maybe_session() : maybe_session(db);
+         pending.emplace(std::move(session), *forkdb.chain_head, when, confirm_block_count, new_protocol_feature_activations);
+      });
+      fork_db.apply_if<void>([&](auto& forkdb) {
+         maybe_session session = self.skip_db_sessions(s) ? maybe_session() : maybe_session(db);
+         building_block_input bbi{ forkdb.chain_head->id(), when, forkdb.chain_head->get_scheduled_producer(when).producer_name,
+                                   new_protocol_feature_activations };
+         pending.emplace(std::move(session), *forkdb.chain_head, bbi);
+      });
```
greg7mdp, line 2482: I wonder if we should have an `apply` which takes two lambdas, one for the `dpos` code, and one for the `if` code.

greg7mdp (commented):
Nice change. I'm really glad we got rid of block_data, this is much nicer.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -95,15 +98,86 @@ namespace eosio::chain {
...
+            return std::visit(overloaded{[&](fork_database_legacy_t& forkdb) -> R {
+                                            std::lock_guard g(m);
+                                            return f(forkdb);
+                                         },
+                                         [&](fork_database_if_t&) -> R {
+                                            return {};
+                                         }}, v);
+      }
+
+      // if we every support more than one version then need to save min/max in fork_database_t
```
greg7mdp, line 179: Comment doesn't read right. Maybe this
```suggestion
      // if we ever support more than one version, then we need to save min/max in fork_database_t
```

greg7mdp (approved):

libraries/chain/block_header_state.cpp
```diff
@@ -7,6 +7,11 @@
 
 namespace eosio::chain {
 
+// moved this warning out of header so it only uses once
+#warning TDDO https://github.com/AntelopeIO/leap/issues/2080
```
linh2931, line 11: Now it occurs to me, if we already have an issue, we don't need a warning, because the issue will remind us.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -11,33 +11,33 @@ namespace eosio::chain {
...
     *
-    * An internal mutex is used to provide thread-safety.
+    * Not thread safe, thread safety provided by fork_database below.
     */
    template<class bsp>  // either block_state_legacy_ptr or block_state_ptr
-   class fork_database {
+   class fork_database_t {
    public:
+      static constexpr uint32_t legacy_magic_number = 0x30510FDB;
+      static constexpr uint32_t magic_number = 0x4242FDB;
```
linh2931, line 28: Just curious. How were the magic numbers picked? Just random numbers?

libraries/chain/fork_database.cpp
```diff
@@ -602,9 +558,80 @@ namespace eosio::chain {
...
+            uint32_t totem = 0;
+            fc::raw::unpack( ds, totem );
+            EOS_ASSERT( totem == fork_database_legacy_t::legacy_magic_number || totem == fork_database_if_t::magic_number, fork_database_exception,
+                        "Fork database file '${filename}' has unexpected magic number: ${actual_totem}. Expected ${t1} or ${t2}",
+                        ("filename", fork_db_file)
+                        ("actual_totem", totem)("t1", fork_database_legacy_t::legacy_magic_number)("t2", fork_database_if_t::magic_number)
+            );
+
+            if (totem == fork_database_legacy_t::legacy_magic_number) {
+               apply_dpos<void>([&](auto& forkdb) {
```
linh2931, line 595: No need to store the magic number?

linh2931 (approved):
Like the restructuring.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Supports writing/reading of `fork_database` in legacy and after instant-finality is activated. Specific to Faster Finality Includes work to persist `fork_database` on shutdown.
Note: end

libraries/chain/fork_database.cpp
heifner: Because it is handled by `fork_database` now instead of `fork_database_t`.

libraries/chain/include/eosio/chain/fork_database.hpp
heifner: Yes, just random. I did stick with it ending in `FDB` which is cute.

libraries/chain/fork_database.cpp
heifner: We are not creating a `fork_database_t` here so it maintains the `magic_number` it was created with which was `legacy_magic_number`. Would be nice to create `fork_database_t` here, however that is not possible without adding a `std::monostate` to the variant or making the variant optional.

This inspired me to use a `std::optional` which helped point out that the current impl methods `apply_if` and `apply_dpos` were not thread safe.

libraries/chain/controller.cpp
heifner: I removed `apply_if` and `apply_dpos` and started to use just the one that took two lambdas. But that makes it not at all obvious what is happening when looking at the calls in controller. I think I will add an apply that takes two but also keep the `apply_if` and `apply_legacy` (renmaing `apply_dpos` to `apply_legacy`)

libraries/chain/fork_database.cpp
heifner: :( optional doesn't work because we expect a `nullptr` `chain_head` before fork database is opened.

greg7mdp (approved):

linh2931 (approved):

greg7mdp (approved):

PR #2114, opened 2024-01-19
minor cleanups to `memfd_create()`

spoonincode:
Two small tweaks,

When `memfd_create()` was first used in EOSIO 2.0, RHEL7 was a targeted platform. Despite RHEL7's kernel shipping with memfd support, its glibc did not as it was too old. So instead, a direct `syscall()` was used. Now that we don't care about RHEL7 any longer we can just use `memfd_create()` proper; both Ubuntu 18.04 and RHEL8 have a new enough glibc that includes support.

As of Linux 6.3 the kernel starts to log a warning about nodeos,
```
nodeos[266224]: memfd_create() called without MFD_EXEC or MFD_NOEXEC_SEAL set
```
This warning has actually been walked back and completely removed as of Linux 6.7. Still, using the new `MFD_NOEXEC_SEAL` seems like good behavior for future proofing as environments may one day set the new `memfd_noexec` knob to enforce `MFD_NOEXEC_SEAL` usage. Sadly, the `MFD_NOEXEC_SEAL` define is only available out of the box on the very latest glibc 2.38 (released July 2023), and Linux prior to 6.3 will `EINVAL` on `MFD_NOEXEC_SEAL` so we need a little bit of fluff to gracefully handle these variations.

### DIFF
--- a/libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/ipc_helpers.hpp
+++ b/libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/ipc_helpers.hpp
@@ -6,7 +6,6 @@
 
 #include <vector>
 
-#include <sys/syscall.h>
 #include <linux/memfd.h>
 
 namespace eosio { namespace chain { namespace eosvmoc {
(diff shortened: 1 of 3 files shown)


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Make a direct call to create a RAM based file. Previously lack of platform support required indirect `syscall()`
Note: end

PR #2116, opened 2024-01-20
IF - addon to PR 2113 - Whitespace change to avoid very long lines.

greg7mdp:


libraries/chain/fork_database.cpp
```diff
@@ -331,15 +332,22 @@ namespace eosio::chain {
...
-               const auto& new_protocol_features = std::get<protocol_feature_activation>(exts.lower_bound(protocol_feature_activation::extension_id())->second).protocol_features;
-               validator( n->timestamp(), static_cast<bs*>(prev_bh.get())->get_activated_protocol_features()->protocol_features, new_protocol_features );
+            if (exts.count(protocol_feature_activation::extension_id()) > 0) {
+               const auto& new_protocol_features =
+                  std::get<protocol_feature_activation>(
+                     exts.lower_bound(protocol_feature_activation::extension_id())->second)
+                     .protocol_features;
+               validator(n->timestamp(),
+                         static_cast<bs*>(prev_bh.get())->get_activated_protocol_features()->protocol_features,
+                         new_protocol_features);
```
heifner, line 346: I can see the original being too long, but this is too short and kind of a mess.

heifner (changes requested):

libraries/chain/fork_database.cpp
greg7mdp: Right, fixed it!

heifner (approved):

PR #2117, opened 2024-01-20
improve `authority_checker`'s performance: don't use exceptions as flow control for non-existing permissions

spoonincode:
Periodically I've seen significant overhead handling exceptions in nodeos while it's live on EOS. It can even approach nearly 10% of main thread time in some of the worst cases. And these exceptions come about even when just applying blocks -- not processing speculative transactions.

What appears to be happening is contracts performing a `send_inline` causing the `authority_checker` to in some situations generate a fury of exceptions for missing `eosio.code` as it counts up weights.

This refactors the `authority_checker` to not lean on an exception to indicate a non-existing permission, a nullptr is returned indicating a non-existing permission instead.

### DIFF
--- a/libraries/chain/include/eosio/chain/authority_checker.hpp
+++ b/libraries/chain/include/eosio/chain/authority_checker.hpp
@@ -236,7 +236,7 @@ namespace detail {
 
                      auto res = cached_permissions.emplace( permission.permission, being_evaluated );
                      itr = res.first;
-                     r = checker.satisfied( std::forward<decltype(*auth)>(*auth), cached_permissions, recursion_depth + 1 );
+                     r = checker.satisfied( *auth, cached_permissions, recursion_depth + 1 );
 
                      if( r ) {
                         total_weight += permission.weight;
--- a/libraries/chain/include/eosio/chain/authority_checker.hpp
+++ b/libraries/chain/include/eosio/chain/authority_checker.hpp
@@ -28,8 +28,8 @@ namespace detail {
     * then determine whether that list of keys is sufficient to satisfy the authority. This class takes a list of keys and
     * provides the @ref satisfied method to determine whether that list of keys satisfies a provided authority.
     *
-    * @tparam F A callable which takes a single argument of type @ref AccountPermission and returns the corresponding
-    * authority
+    * @tparam F A callable which takes a single argument of type @ref AccountPermission and returns a pointer to
+    * the corresponding authority if it exists, or a nullptr.
     */
    template<typename PermissionToAuthorityFunc>
    class authority_checker {
@@ -225,19 +225,18 @@ namespace detail {
                      bool r = false;
                      typename permission_cache_type::iterator itr = cached_permissions.end();
 
-                     bool propagate_error = false;
+                     std::invoke_result_t<decltype(checker.permission_to_authority), const permission_level> auth = nullptr;
                      try {
-                        auto&& auth = checker.permission_to_authority( permission.permission );
-                        propagate_error = true;
-                        auto res = cached_permissions.emplace( permission.permission, being_evaluated );
-                        itr = res.first;
-                        r = checker.satisfied( std::forward<decltype(auth)>(auth), cached_permissions, recursion_depth + 1 );
-                     } catch( const permission_query_exception& ) {
-                        if( propagate_error )
-                           throw;
-                        else
-                           return total_weight; // if the permission doesn't exist, continue without it
+                        auth = checker.permission_to_authority( permission.permission );
                      }
+                     catch( const permission_query_exception& ) {}
+                     //permission was either invalid (threw permission_query_exception), or wasn't found
+                     if(!auth)
+                        return total_weight;
+
+                     auto res = cached_permissions.emplace( permission.permission, being_evaluated );
+                     itr = res.first;
+                     r = checker.satisfied( std::forward<decltype(*auth)>(*auth), cached_permissions, recursion_depth + 1 );
 
                      if( r ) {
                         total_weight += permission.weight;
--- a/libraries/chain/authorization_manager.cpp
+++ b/libraries/chain/authorization_manager.cpp
@@ -479,7 +479,12 @@ namespace eosio { namespace chain {
 
       auto effective_provided_delay =  (provided_delay >= delay_max_limit) ? fc::microseconds::maximum() : provided_delay;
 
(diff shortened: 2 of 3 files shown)


libraries/chain/include/eosio/chain/authority_checker.hpp
```diff
@@ -225,19 +225,18 @@ namespace detail {
                      bool r = false;
                      typename permission_cache_type::iterator itr = cached_permissions.end();
 
-                     bool propagate_error = false;
+                     std::invoke_result_t<decltype(checker.permission_to_authority), const permission_level> auth = nullptr;
```
spoonincode, line 228: This is pretty weird; it comes about because `permission_to_authority` could return either a `shared_authority` or an `authority` (well, now either a `shared_authority*` or `authority*`). I want the scope to be outside the `try` to make handling the null check more straightforward so I couldn't use `auto` in this particular pattern.

libraries/chain/include/eosio/chain/authority_checker.hpp
```diff
@@ -225,19 +225,18 @@ namespace detail {
...
+                        auth = checker.permission_to_authority( permission.permission );
                      }
+                     catch( const permission_query_exception& ) {}
+                     //permission was either invalid (threw permission_query_exception), or wasn't found
+                     if(!auth)
+                        return total_weight;
+
+                     auto res = cached_permissions.emplace( permission.permission, being_evaluated );
+                     itr = res.first;
+                     r = checker.satisfied( std::forward<decltype(*auth)>(*auth), cached_permissions, recursion_depth + 1 );
```
greg7mdp, line 239: I don't believe the `std::forward` does anything with `auth` declared as it is now.

libraries/chain/include/eosio/chain/authority_checker.hpp
```diff
@@ -225,19 +225,18 @@ namespace detail {
...
-                        if( propagate_error )
-                           throw;
-                        else
-                           return total_weight; // if the permission doesn't exist, continue without it
+                        auth = checker.permission_to_authority( permission.permission );
                      }
+                     catch( const permission_query_exception& ) {}
+                     //permission was either invalid (threw permission_query_exception), or wasn't found
+                     if(!auth)
+                        return total_weight;
```
greg7mdp, line 235: Couldn't you just have added these two lines after line 230 in the original version?

ericpassmore:
Note:start
group: PERFORMANCE
category: CONTRACTS
summary: Improve overall performance of `authority_checker` by removing exceptions. At times up to 10% of time on main thread spent dealing with exceptions. Most frequent exceptions are from `authority_checker`.  
Note: end

libraries/chain/include/eosio/chain/authority_checker.hpp
spoonincode: :+1: indeed

libraries/chain/include/eosio/chain/authority_checker.hpp
spoonincode: In the original 230, `permission_to_authority` calls `get_permission()` which throws when it can't find the permission. So, there is no execution after 230 until the `catch` (which then does the `return total_weight` if `propagate_error` is false).

greg7mdp: But then, how does this PR reduce the overhead of handling exceptions, if they still occur and we catch them as before?

spoonincode: The original code was using `get_permission()`,
https://github.com/AntelopeIO/leap/blob/8a9fa7689aebb0d6df276b8e4c31d3bf64800efe/libraries/chain/authorization_manager.cpp#L259-L263
This throws on either an invalid permission or when chainbase's `get()` throws due to it not existing.

The new code uses `find_permission()`,
https://github.com/AntelopeIO/leap/blob/8a9fa7689aebb0d6df276b8e4c31d3bf64800efe/libraries/chain/authorization_manager.cpp#L253-L257
This only throws on an invalid permission -- chainbase's `find()` just returns a `nullptr` if it can't find the permission.

That's why the new code still handles a `permission_query_exception`,
https://github.com/AntelopeIO/leap/blob/7ab4d4c10c563b291d5f2d3ae2774c696cdf0bf4/libraries/chain/include/eosio/chain/authority_checker.hpp#L228-L232
`permission_to_authority()` can still throw via `find_permission()`'s "Invalid permission" check. I'm not actually sure under what conditions that comes in to play (I've not seen it), but of course any refactor here needs to be exact 1:1 behavior unless we do a significant audit and the previous behavior was to treat that identically to a failed lookup (return `total_weight`).

It does appear like `find_permission()` maybe could be refactored to be `nothrow` by also returning a `nullptr` when an invalid permission is used too. But again, we'd have to investigate all uses of `find_permission()` very closely.

greg7mdp: I'm just wondering about the benefit of moving  lines 234 to 239 out of the `try` block, which requires the awkward declaration of `auth` on line 228. My apologies if I'm still missing something obvious.

greg7mdp: > I'm just wondering about the benefit of moving lines 234 to 239 out of the `try` block, which requires the awkward declaration of `auth` on line 228. My apologies if I'm still missing something obvious.

Never mind, got it, thanks Matt for the thorough explanation!

greg7mdp (approved):

libraries/chain/include/eosio/chain/authority_checker.hpp
```diff
@@ -225,19 +225,18 @@ namespace detail {
...
+
+                     auto res = cached_permissions.emplace( permission.permission, being_evaluated );
+                     itr = res.first;
+                     r = checker.satisfied( *auth, cached_permissions, recursion_depth + 1 );
```
greg7mdp, line 239: Wondering if it would be valid and/or useful to  do `std::move(*auth)` here?

heifner (approved):

libraries/chain/include/eosio/chain/authority_checker.hpp
spoonincode: I don't think so since `satisfied()` is just taking a reference anyways -- we're ultimately just plumbing through a reference/pointer to the object owned by the DB. Plus, the return of `permission_to_authority()` is a `const`

PR #2119, opened 2024-01-21
IF: Misc nomenclature changes

heifner:
- `hs_vote_message` => `vote_message`
- `dpos` => `legacy`
- Remove `chain_pacemaker` connections as it is not needed


greg7mdp (approved):
I wish we had found something shorter than `_legacy`! 

I would have preferred `_f0` (meaning finality version 0) for `_legacy` and `_f1` for instant finality, but I have a feeling that I would not get a consensus on this proposal.

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Nomenclature changes for more readable code.
Note: end

PR #2121, opened 2024-01-22
IF: Rename block_token to block_handle

heifner:
- Block handle seems a better name than block token.
  -  Thanks @greg7mdp for suggestion
  - https://www.quora.com/What-is-the-difference-between-a-token-and-an-object-handle


greg7mdp (approved):
Thanks!

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Nomenclature change to `block_handle` for better readability.
Note: end

PR #2122, opened 2024-01-22
IF: confirmed==0 after instant finality enabled

### LINKED ISSUE #1580: IF: Ensure confirmed count in block header is zero after HotStuff transition
Depends on https://github.com/AntelopeIO/leap/issues/1523.

After HotStuff is enabled and the network has transition to it, proposers should create blocks with a `confirmed` value of zero, and validators should check that the value is zero.

heifner:
Resolves #1580


greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Set `block_header.confirmed` to zero after transition to faster finality.
Note: end

Issue #2084, opened 2024-01-13
IF: accumulated weight calculation


linh2931:
From @arhag in reviewing https://github.com/AntelopeIO/leap/pull/2076:

_I'm realizing there is a problem with the weak and strong vote calculations. It is assuming they are all of equal weight. But that isn't necessarily true. It needs to refer to the weights in the finalizer policy to determine how to accumulate the weight._

PR #2131, opened 2024-01-24
IF: accumulate the weight in vote calculation

linh2931:
Currently in the weak and strong vote calculations, it is assuming they are all of equal weight. It needs to refer to the weights in the finalizer policy to determine how to accumulate the weight.

Resolved https://github.com/AntelopeIO/leap/issues/2084


libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -84,21 +77,20 @@ bool pending_quorum_certificate::add_strong_vote(const std::vector<uint8_t>& pro
...
    if (!_strong_votes.add_vote(proposal_digest, index, pubkey, sig))
       return false;
-   size_t weak   = num_weak();
-   size_t strong = num_strong();
+   _strong_accumulated_weight += _finalizer_weights[index];
 
    switch (_state) {
    case state_t::unrestricted:
    case state_t::restricted:
-      if (strong >= _quorum) {
```
greg7mdp, line 93: I really prefer the short names. Could we keep `_strong` and `_weak`?

linh2931: Those are class members and used across the file. Longer names help readability. `_strong` and `_weak` also imply state.

greg7mdp: > Those are class members and used across the file. Longer names help readability. `_strong` and `_weak` also imply state.

Can we call them `_strong_sum` and `weak_sum`?

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -223,15 +218,14 @@ namespace eosio::chain {
       std::vector<uint8_t> _proposal_digest;
       state_t              _state { state_t::unrestricted };
       size_t               _num_finalizers {0};
-      size_t               _quorum {0};
+      std::vector<uint64_t> _finalizer_weights; // weight of each finalizer
```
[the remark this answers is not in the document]
linh2931: If passing in the weight, every time it needs to be passed from `block_state::aggregate_vote` to `pending_quorum_certificate::add_vote` to `pending_quorum_certificate::add_strong_vote` or `pending_quorum_certificate::add_weak_vote`. Plus we have already stored `quorum` and `num_finalizers` as members.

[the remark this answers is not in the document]
greg7mdp: > If passing in the weight, every time it needs to be passed from `block_state::aggregate_vote` to `pending_quorum_certificate::add_vote` to `pending_quorum_certificate::add_strong_vote` or `pending_quorum_certificate::add_weak_vote`. Plus we have already stored `quorum` and `num_finalizers` as members.

Still, this would make the change much simpler, and there is no reason to store a whole vector of weights, recreated every time we create a quorum certificate, just because we need to weight of the current finalizer in `add_vote`.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: When checking if quorum has been reached, use the finalizer policy to determine how to accumulate vote weights.
Note: end

unittests/block_state_tests.cpp
```diff
@@ -96,7 +191,7 @@ BOOST_AUTO_TEST_CASE(verify_qc_test) try {
    digest_type strong_digest(fc::sha256("0000000000000000000000000000002"));
    std::vector<uint8_t> strong_digest_data(strong_digest.data(), strong_digest.data() + strong_digest.data_size());
    digest_type weak_digest(fc::sha256("0000000000000000000000000000003"));
-   std::vector<uint8_t> weak_digest_data(weak_digest.data(), weak_digest.data() + weak_digest.data_size());
+   std::vector<uint8_t> weak_digest_data(weak_digest.data(), weak_digest.data() + weak_digest.        data_size());
```
greg7mdp, line 194: the spaces after `weak_digest.` seem to be a mistake.

greg7mdp (approved):

heifner:
Not related to your changes, but looks like we have a real issue with this test: `producer_schedule_if_unit_test_eos-vm (SEGFAULT)`. I have seen it fail a number of times on recent runs.

heifner (approved):

libraries/chain/include/eosio/chain/hotstuff/finalizer_policy.hpp
```diff
@@ -9,6 +9,16 @@ namespace eosio::chain {
       uint32_t                         generation = 0; ///< sequentially incrementing version number
       uint64_t                         threshold = 0;  ///< vote weight threshold to finalize blocks
       std::vector<finalizer_authority> finalizers; ///< Instant Finality voter set
+
+      // max accumulated weak weight before becoming weak_final
+      uint64_t max_weak_sum_before_weak_final() const {
+         auto n = finalizers.size();
+         uint64_t sum = 0;
+         for( size_t i = 0; i < n; ++i ) {
```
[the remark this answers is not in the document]
greg7mdp: or `std::accumulate` with a lambda such as `[](uint64_t sum, const auto& fa) { return sum + fa.weight; }`

unittests/block_state_tests.cpp
linh2931: Thanks for noticing this!

libraries/chain/include/eosio/chain/hotstuff/finalizer_policy.hpp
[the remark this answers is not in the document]
linh2931: Thanks for both suggestions! I will go with `std::accumulate` as it matches what it is doing here.

heifner (approved):

greg7mdp (approved):

PR #2133, opened 2024-01-24
[5.0] Also need to switch to lib_catchup if in head_catchup when starting to sync

heifner:
If you are starting to sync with a node then you expect to be in `lib_catchup`. Need to switch to `lib_catchup` even if you are in `head_catchup`. Observed node not syncing when LIB was stuck in an integration test. The node would remain in `head_catchup` and never switch to `lib_catchup`. Since the `sync-fetch-span` was set to 5, it would keep syncing the same 5 blocks and never leave `head_catchup`.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -2214,7 +2214,7 @@ namespace eosio {
          return;
       }
 
-      if( sync_state == in_sync ) {
+      if( sync_state != lib_catchup ) {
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

spoonincode (approved):

ericpassmore:
Note:start
group: STABILITY
category: BUG
summary: Check and switch to `lib_catchup` when starting to sync. 
Note:end

PR #2132, opened 2024-01-24
[5.0] clear the DB dirty bit if `pinnable_mapped_file` fails to fully start

spoonincode:
Bumps the chainbase submodule to chainbase's leap-5.0 branch head with AntelopeIO/chainbase#38 fixing #2086

PR #2134, opened 2024-01-24
[5.0 -> main] clear the DB dirty bit if `pinnable_mapped_file` fails to fully start

### LINKED ISSUE #2086: When mapped_private doesn't work, database is left dirty
When I tried to flip a server running 5.0 to the new "mapped_private" memory mode, I encountered the following error (because my chain state db setting was higher than mem on the server):

```error 2024-01-15T14:16:13.557 nodeos    main.cpp:209                  main                 ] 13 N5boost12interprocess22interprocess_exceptionE: Cannot allocate memory
rethrow Cannot allocate memory:
    {"what":"Cannot allocate memory"}
    nodeos  chain_plugin.cpp:1102 plugin_initialize
```

The problem is that when i flipped the mode back to the default "mapped" mode, it left the state dirty and I had to use a snapshot to recover it (which Kevin said is not the expected behavior).

spoonincode:
main merge of #2132 but moving chainbase's submodule to top of main, resolves #2086


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: BUG
summary: Revs chainbase version to include a bug fix. Leap will now start up cleanly in non-mapped_private mode after failing to startup in mapped_private mode due to lack of swap and/or physical memory.
Note:end

PR #2135, opened 2024-01-24
IF: Update finalizer safety information and rules for how a finalizer signs

### LINKED ISSUE #2070: IF: Update finalizer safety information and rules for how a finalizer signs
The finalizer safety information should be updated to track the following data per each finalizer:
* The range of timestamps covered by their last vote.
* The block ID of the last block that they voted on.
* The block ID of the block that they are locked on.

Whenever a finalizer signs a new block, the finalizer safety information must be updated and durably committed before propagating the signature in a vote message.

Whenever a finalizer considers a new block to sign, it must consult its existing finalizer safety information to determine whether it should sign, and if so whether it should sign strongly or weakly. The rules for this are captured in the pseudo-code: https://github.com/AntelopeIO/leap/blob/59532dfa676a9ab6a0b9c54d5e9843f0ab71fe56/libraries/chain/hotstuff/hs_pseudo#L283-L341.

Related to https://github.com/AntelopeIO/leap/issues/2069.

For this issue, update the finalizer signing process to consider the changes described above and in the pseudo-code. When signing weakly, the digest to sign should be a hash of the concatenation of the `finalizer_digest` and the string `WEAK`.

greg7mdp:
resolves #2070.

This PR implements support for managing a set of stateful finalizers (the local finalizers that this node is running). 

It is important that the safety information for each finalizer (last vote and proposal lock) is preserved across nodeos invocations, since incorrect voting can lead to finality violations. 

The safety information for all finalizers is saved in a file named `safety.dat`, which is by default located in `{data_dir}/finalizers`. This file preserves the safety information for not only the currently running finalizers, but for all thefinalizers who ever ran on this node with the same specified safety file location.

The currently running finalizers are always the first ones stored in the file, followed by all the inactive ones. After each vote, only the information for the currently running finalizers is written to the file, which is then flushed. This file update occurs *before* the vote is propagated.

The way the finalizer safety information is initialized (if not found in the persistent `safety.dat` file) is important, and described [here](https://github.com/AntelopeIO/leap/issues/2070#issuecomment-1941901836). The initial configuration must be such that it will prevent safety violations, while still allowing liveness to eventually happen, so the finalizer can vote.

This specification may yet evolve, as the document is not 100% completed.

This PR also includes some other misceallenous changes:
- update `fc/crypto` bls primitives to take `std::span` instead of `std::vector`, so we don't have to do unnecessary copies.
- correct support for weak digest (which is the strong digest with the 4 characters "WEAK" appended at the end). since we couldn't use `digest_type`, a custom type `weak_digest_t` was defined. 
- extract a couple of utility functions from `tester.cpp` into a new file `bls_utils.hpp`, so that they can be reused in boost tests.


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Track the following data for each finalizer: timestamp of last vote, last block ID voted on, last block ID locked on.
Note:end

heifner (changes requested):

libraries/libfc/include/fc/crypto/sha256.hpp
```diff
@@ -22,6 +23,10 @@ class sha256
     char*       data();
     size_t      data_size() const { return 256 / 8; }
 
+    std::span<const uint8_t> to_span() const {
+       return {(const uint8_t*)data(),  (const uint8_t*)data() + data_size()};
```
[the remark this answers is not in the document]
greg7mdp: sure, updated.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,155 @@
...
+   // ----------------------------------------------------------------------------------------
+   struct qc_chain_t {
+      block_state_ptr b2; // first phase,  prepare
+      block_state_ptr b1; // second phase, precommit
+      block_state_ptr b;  // third phase,  commit
+   };
+
+   // ----------------------------------------------------------------------------------------
+   struct finalizer {
+      enum class VoteDecision { StrongVote, WeakVote, NoVote };
```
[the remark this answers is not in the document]
greg7mdp: updated.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+      struct proposal_ref {
+         block_id_type         id;
+         block_timestamp_type  timestamp;
+
+         proposal_ref() = default;
+
+         template<class bsp>
```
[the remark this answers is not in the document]
greg7mdp: I don't really like `Bsp` because it is already an abbreviation. Whould `BSP`  be ok with you?

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,155 @@
...
+         proposal_ref(const bsp& p) :
+            id(p->id()),
+            timestamp(p->timestamp())
+         {}
+
+         proposal_ref(const block_id_type&id, block_timestamp_type t) :
```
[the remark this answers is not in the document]
greg7mdp: added space.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,155 @@
...
+               lock == o.lock;
+         }
+      };
+
+      bls_private_key           priv_key;
+      safety_information        fsi;
+
+   private:
+      using branch_type = fork_database_if_t::branch_type;
+      qc_chain_t     get_qc_chain(const block_state_ptr&  proposal, const branch_type& branch) const;
```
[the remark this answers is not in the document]
greg7mdp: fixed.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,155 @@
...
+      mutable fc::datastream<fc::cfile> persist_file;          // we want to keep the file open for speed
+      std::map<bls_public_key, finalizer>  finalizers;         // the active finalizers for this node
+      fsi_map                           inactive_safety_info;  // loaded at startup, not mutated afterwards
+      fsi_t                             default_fsi = fsi_t::unset_fsi(); // default provided at leap startup
+      mutable bool                      inactive_safety_info_written{false};
+
+      ~finalizer_set();
+
+      template<class F>
+      void maybe_vote(const finalizer_policy &fin_pol,
```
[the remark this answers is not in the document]
greg7mdp: fixed.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,155 @@
...
+            }
+         }
+         // then save the safety info and, if successful, gossip the votes
+         if (!votes.empty()) {
+            try {
+               save_finalizer_safety_info();
+               for (const auto& vote : votes)
+                  std::forward<F>(process_vote)(vote);
+            } catch(...) {
+               throw;
```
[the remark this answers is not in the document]
greg7mdp: right, removed!

libraries/testing/include/eosio/testing/bls_utils.hpp
```diff
@@ -0,0 +1,25 @@
+#pragma once
+
+#include <fc/crypto/bls_private_key.hpp>
+#include <fc/crypto/bls_public_key.hpp>
+#include <eosio/chain/name.hpp>
+
+namespace eosio::testing {
+
+   using eosio::chain::name;
```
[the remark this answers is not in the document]
greg7mdp: makes sense, updated.

tests/CMakeLists.txt
```diff
@@ -333,8 +333,8 @@ set_property(TEST cli_test PROPERTY LABELS nonparallelizable_tests)
 
 add_test(NAME larger_lib_test COMMAND tests/large-lib-test.py ${UNSHARE} WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 set_property(TEST larger_lib_test PROPERTY LABELS nonparallelizable_tests)
-add_test(NAME larger_lib_if_test COMMAND tests/large-lib-test.py --activate-if ${UNSHARE} WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
-set_property(TEST larger_lib_if_test PROPERTY LABELS nonparallelizable_tests)
+#add_test(NAME larger_lib_if_test COMMAND tests/large-lib-test.py --activate-if ${UNSHARE} WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
```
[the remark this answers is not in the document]
greg7mdp: I added a comment. I also commented out the same test for dpos, which I think was what Areg concluded we should do.

libraries/chain/block_state.cpp
```diff
@@ -154,7 +154,7 @@ void block_state::verify_qc(const valid_quorum_certificate& qc) const {
 
    if( qc._weak_votes ) {
```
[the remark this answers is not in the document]
greg7mdp: I'm not following your suggestion. What would this span be constructed from?
But maybe you mean that `aggregate_verify` should take a `std::span<std::span<const uint8_t>>` as a second argument. I think that would make sense and indeed avoid that copy.

libraries/chain/controller.cpp
```diff
@@ -1555,6 +1565,41 @@ struct controller_impl {
...
+      // At Leap startup, we want to provide to our local finalizers the correct safety information
+      // to use if they don't already have one.
+      // If we start at a block prior to the IF transition, that information will be provided when
+      // we create the new `fork_db_if`.
+      // If we start at a block during or after the IF transition, we need to provide this information
+      // at startup.
+      // ---------------------------------------------------------------------------------------------
+      if (fork_db.fork_db_if_present()) {
+         // we are already past the IF transition point where we create the updated fork_db.
+         // so we can't rely on the finalizer safety information update happening duting the transition.
```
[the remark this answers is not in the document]
greg7mdp: updated.

libraries/chain/controller.cpp
```diff
@@ -1555,6 +1565,41 @@ struct controller_impl {
...
+         // see https://hackmd.io/JKIz2TWNTq-xcWyNX4hRvw for details
```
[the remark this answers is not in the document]
greg7mdp: I don't mind doing this, but I'm worried the comment and the hackmd document may get out of sync.

libraries/chain/controller.cpp
```diff
@@ -1555,6 +1565,41 @@ struct controller_impl {
...
+            };
+            fork_db.apply_if<void>(set_finalizer_defaults);
+         } else {
+            // we are past the IF transition.
+            auto set_finalizer_defaults = [&](auto& forkdb) -> void {
+               auto lib = forkdb.root();
+               my_finalizers.set_default_safety_information(
+                  finalizer::safety_information{ .last_vote_range_start = block_timestamp_type(0),
+                                                 .last_vote = {},
+                                                 .lock      = finalizer::proposal_ref(lib) });
```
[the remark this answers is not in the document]
greg7mdp: Yes, I know, but I think we may eventually have a different initialization for these two cases and the if was there to imply this (obviously not very effectively).

libraries/chain/controller.cpp
```diff
@@ -2998,52 +3058,42 @@ struct controller_impl {
...
-         if( it != node_finalizer_keys.end() ) {
-            const auto& private_key = it->second;
-            const auto& digest = bsp->compute_finalizer_digest();
-
-            auto sig =  private_key.sign(std::vector<uint8_t>(digest.data(), digest.data() + digest.data_size()));
+   }
 
-            // construct the vote message
-            vote_message vote{ bsp->id(), strong, f.public_key, sig };
+   void create_and_send_vote_msg(const block_state_ptr& bsp, const fork_database_if_t& fork_db) {
```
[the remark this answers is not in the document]
greg7mdp: It is just one vote, so I think it is less important than doing the validation of the 20 other votes we receive on the thread pool, and additionally the aggregation of the vote onto our qc is already done on the thread pool. So we are just talking about just one signature per block, and probably we can save this optimization for later if we feel it is needed.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
+#include <eosio/chain/hotstuff/finalizer.hpp>
+#include <eosio/chain/exceptions.hpp>
+#include <fc/log/logger_config.hpp>
+
+namespace eosio::chain {
+
+// ----------------------------------------------------------------------------------------
+block_state_ptr get_block_by_height(const fork_database_if_t::branch_type& branch, uint32_t block_num) {
```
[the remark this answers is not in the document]
greg7mdp: done

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+   while (cur) {
+      if (cur->id() == ancestor)
+         return true;
+      cur = fork_db.get_block_header(cur->previous());
+   }
+   return false;
+}
+
+// ----------------------------------------------------------------------------------------
+finalizer::VoteDecision finalizer::decide_vote(const block_state_ptr& p, const fork_database_if_t& fork_db) {
```
linh2931: maybe use `bsp` for `p` as parameter name for clearer understanding.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+   qc_chain_t chain = get_qc_chain(p, p_branch);
+
+   // we expect last_qc_block_num() to always be found except at bootstrap
+   // in `assemble_block()`, if we don't find a qc in the ancestors of the proposed block, we use block_num
+   // from fork_db.root(), and specify weak.
+   auto bsp_last_qc = p->last_qc_block_num() ? get_block_by_height(p_branch, *p->last_qc_block_num()) : block_state_ptr{};
+
+   bool monotony_check = !fsi.last_vote || p->timestamp() > fsi.last_vote.timestamp;
+   // !fsi.last_vote means we have never voted on a proposal, so the protocol feature just activated and we can proceed
```
linh2931, line 75: Change `proposal` to `block`?

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+   if (monotony_check && (liveness_check || safety_check)) {
+      auto [p_start, p_end] = std::make_pair(bsp_last_qc ? bsp_last_qc->timestamp() : p->timestamp(),
+                                             p->timestamp());
+
+      bool time_range_disjoint    = fsi.last_vote_range_start >= p_end || fsi.last_vote.timestamp <= p_start;
+
+      bool enough_for_strong_vote = time_range_disjoint || extends(fork_db, p, fsi.last_vote.id);
+
+      fsi.last_vote             = proposal_ref(p);          // v_height
```
linh2931, line 113: remove the comment or update it

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+      fsi.last_vote_range_start = p_start;
+
+      if (chain.b1 && chain.b1->timestamp() > fsi.lock.timestamp)
+         fsi.lock = proposal_ref(chain.b1);                // commit phase on b1
```
linh2931, line 117: remove the comment or update it.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+      decision = enough_for_strong_vote ? VoteDecision::StrongVote : VoteDecision::WeakVote;
+   } else {
+      dlog("bsp_last_qc=${bsp}, last_qc_block_num=${lqc}, fork_db root block_num=${f}",
+           ("bsp", !!bsp_last_qc)("lqc",!!p->last_qc_block_num())("f",fork_db.root()->block_num()));
+      if (p->last_qc_block_num())
```
linh2931, line 123: embed the if check inside the dlog

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+         dlog("last_qc_block_num=${lqc}", ("lqc", *p->last_qc_block_num()));
+   }
+   if (decision != VoteDecision::NoVote)
+      dlog("Voting ${s}", ("s", decision == VoteDecision::StrongVote ? "strong" : "weak"));
```
linh2931: embed the `if` check inside the dlog so it is not always executed.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+}
+
+// ----------------------------------------------------------------------------------------
+qc_chain_t finalizer::get_qc_chain(const block_state_ptr& proposal, const branch_type& branch) const {
+   qc_chain_t res;
+
+   // get b2
+   // ------
+   auto it2 = std::find_if(branch.begin(), branch.end(),
+        [t=proposal->core.last_qc_block_num](const block_state_ptr& bsp) { return bsp->block_num() == t; });
```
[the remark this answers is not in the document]
greg7mdp: Not quite that easy, because we don't search the whole branch for `b1` and `b` (we start from the previously returned iterator).

libraries/chain/controller.cpp
```diff
@@ -236,7 +245,7 @@ struct assembled_block {
    const block_id_type& id() const {
       return std::visit(
          overloaded{[](const assembled_block_legacy& ab) -> const block_id_type& { return ab.id; },
-                    [](const assembled_block_if& ab)   -> const block_id_type& { return ab.bhs.id; }},
+               [](const assembled_block_if& ab)   -> const block_id_type& { return ab.bhs.id(); }},
```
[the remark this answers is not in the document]
greg7mdp: indeed, thanks!

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+   }
+   try {
+      persist_file.seek(0);
+      fc::raw::pack(persist_file, fsi_t::magic);
+      fc::raw::pack(persist_file, (uint64_t)(finalizers.size() + inactive_safety_info.size()));
+      for (const auto& [pub_key, f] : finalizers) {
+         fc::raw::pack(persist_file, pub_key);
+         fc::raw::pack(persist_file, f.fsi);
+      }
+      if (!inactive_safety_info_written) {
```
linh2931, line 170: Can we have a separate file for inactive safety info so it won't be written every time?

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+      for (size_t i=0; i<num_finalizers; ++i) {
+         fsi_map::value_type entry;
+         fc::raw::unpack(persist_file, const_cast<bls_public_key&>(entry.first));
+         fc::raw::unpack(persist_file, entry.second);
+         res.insert(entry);
+      }
+      persist_file.close();
+   } catch (const fc::exception& e) {
+      edump((e.to_detail_string()));
+      // std::filesystem::remove(persist_file_path); // don't remove file we can't load
```
linh2931, line 237: Remove the code but keep the comment

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+      throw;
+   } catch (const std::exception& e) {
+      edump((e.what()));
+      // std::filesystem::remove(persist_file_path); // don't rremove file we can't load
```
linh2931, line 241: Remove the commented-out code but keep the comment.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,155 @@
...
+#include <fc/io/cfile.hpp>
+#include <fc/reflect/reflect.hpp>
+#include <fc/crypto/sha256.hpp>
+#include <compare>
+#include <utility>
+
+namespace eosio::chain {
+   // ----------------------------------------------------------------------------------------
+   struct qc_chain_t {
+      block_state_ptr b2; // first phase,  prepare
```
linh2931, line 15: Use our own terminology.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+void finalizer_set::set_keys(const std::map<std::string, std::string>& finalizer_keys) {
+   assert(finalizers.empty()); // set_keys should be called only once at startup
+   if (finalizer_keys.empty())
+      return;
+
+   fsi_map safety_info = load_finalizer_safety_info();
+   for (const auto& [pub_key_str, priv_key_str] : finalizer_keys) {
+      auto public_key {bls_public_key{pub_key_str}};
+      auto it  = safety_info.find(public_key);
+      auto fsi = it != safety_info.end() ? it->second : default_fsi;
```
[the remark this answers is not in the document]
greg7mdp: done, thanks.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+// --------------------------------------------------------------------------------------------
+// Can be called either:
+//   - when transitioning to IF (before any votes are to be sent)
+//   - at leap startup, if we start at a block which is either within or past the IF transition.
+// In either case, we are never updating existing finalizer safety information. This is only
+// to ensure that the safety information will have defaults that ensure safety as much as
+// possible, and allow for liveness which will allow the finalizers to eventually vote.
+// --------------------------------------------------------------------------------------------
+void finalizer_set::set_default_safety_information(const fsi_t& fsi) {
+   //assert(fsi.last_vote.id.empty() || t_startup < fsi.last_vote.timestamp);
```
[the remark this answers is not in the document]
greg7mdp: done

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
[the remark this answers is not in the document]
heifner: I think the proper CamelCase for `bsp` would be `BSP`, as you say it is an abbreviation.

libraries/chain/block_state.cpp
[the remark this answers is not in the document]
heifner: Actually, never mind.
`std::span<const std::span<const uint8_t>>` would require changing `bls12_381::aggregate_verify`. 
I keep thinking you can use a range for `std::span`. It is really unfortunate that `bls12_381` uses `std::span` and not ranges.

libraries/chain/controller.cpp
```diff
@@ -1083,6 +1092,7 @@ struct controller_impl {
     chain_id( chain_id ),
     read_mode( cfg.read_mode ),
     thread_pool(),
+    my_finalizers{ .t_startup = cfg.node_startup_time, .persist_file_path = cfg.data_dir / "finalizers" / "safety.dat" },
```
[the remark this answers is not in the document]
heifner: If it helps with testing then that is fine. Otherwise, seems unneeded.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
heifner: I thought it was odd they were the same. I'm ok with leaving as is.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
heifner: I'm ok with doing the optimization later. I do think it is useful. Can you add a TODO for the optimization.

libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -91,6 +91,8 @@ namespace eosio::chain {
             flat_set<account_name>   contract_blacklist;
             flat_set< pair<account_name, action_name> > action_blacklist;
             flat_set<public_key_type> key_blacklist;
+            block_timestamp_type     node_startup_time      { fc::time_point::now() };
+            path                     data_dir               =  std::filesystem::current_path();
```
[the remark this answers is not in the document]
greg7mdp: done

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,155 @@
...
+         {}
+
+         void reset() {
+            id = block_id_type();
+            timestamp = block_timestamp_type();
+         }
+
+         bool empty() const { return id.empty(); }
+
+         operator bool() const { return !id.empty(); }
```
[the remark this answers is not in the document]
greg7mdp: done

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,155 @@
...
+      qc_chain_t     get_qc_chain(const block_state_ptr&  proposal, const branch_type& branch) const;
+      VoteDecision   decide_vote(const block_state_ptr& proposal, const fork_database_if_t& fork_db);
+
+   public:
+      std::optional<vote_message> maybe_vote(const bls_public_key& pub_key, const block_state_ptr& bsp,
+                                             const digest_type& digest, const fork_database_if_t& fork_db);
+   };
+
+   // ----------------------------------------------------------------------------------------
+   struct finalizer_set {
```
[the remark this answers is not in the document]
linh2931: Yeah, don't use the post-fix `finalizer_set`; plural is more accurate than `_set`. Use some kind of `_finalizers`.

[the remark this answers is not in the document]
greg7mdp: I renamed it to `my_finalizers_t`.

tests/TestHarness/queries.py
```diff
@@ -688,10 +688,11 @@ def getHeadBlockNum(self):
             headBlockNumTag = "head_block_num"
             return info[headBlockNumTag]
 
-    def getIrreversibleBlockNum(self):
+    def getIrreversibleBlockNum(self, printIt=True):
         info = self.getInfo(exitOnError=True)
         if info is not None:
-            Utils.Print("current lib: %d" % (info["last_irreversible_block_num"]))
+            if printIt:
+                Utils.Print("current lib: %d" % (info["last_irreversible_block_num"]))
```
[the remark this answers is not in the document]
greg7mdp: sure, I like that better but I didn't want to change the behavior of existing tests.

libraries/chain/controller.cpp
```diff
@@ -2730,6 +2775,21 @@ struct controller_impl {
                   ilog("Transition to instant finality happening after block ${b}", ("b", forkdb.chain_head->block_num()));
                   if_irreversible_block_num = forkdb.chain_head->block_num();
 
+                  {
+                     // If leap started at a block prior to the IF transition, it needs to provide a default safety
```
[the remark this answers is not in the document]
greg7mdp: updated.

libraries/chain/controller.cpp
```diff
@@ -2730,6 +2775,21 @@ struct controller_impl {
...
+                     // information for those finalizer that don't already have one. This typically should be done when
```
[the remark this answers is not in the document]
greg7mdp: updated.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+// ----------------------------------------------------------------------------------------
+block_state_ptr get_block_by_height(const fork_database_if_t::branch_type& branch, uint32_t block_num) {
+   auto it = std::find_if(branch.begin(), branch.end(),
+                          [&](const block_state_ptr& bsp) { return bsp->block_num() == block_num; });
+   return it == branch.end() ? block_state_ptr{} : *it;
+}
+
+// ----------------------------------------------------------------------------------------
+qc_chain_t finalizer::get_qc_chain(const block_state_ptr& proposal, const branch_type& branch) const {
```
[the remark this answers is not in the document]
greg7mdp: I think proposal adds information here.

libraries/chain/hotstuff/finalizer.cpp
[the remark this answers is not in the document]
greg7mdp: done.

greg7mdp: renamed it proposal.

libraries/chain/hotstuff/finalizer.cpp
greg7mdp: I don't know. We have used proposal a lot when talking about hotstuff. We could talk about it at the next meeting and ask Areg if we should use block exclusively instead of proposal. My personal feeling is that, in finality related code, using proposal makes it clear what info from the block we are interested in, and the type makes it clear it is a block we are looking at.

libraries/chain/hotstuff/finalizer.cpp
greg7mdp: Do you have a suggestion on how to do that cleanly?

libraries/chain/hotstuff/finalizer.cpp
greg7mdp: It is not written every time.

libraries/chain/hotstuff/finalizer.cpp
greg7mdp: done

libraries/chain/hotstuff/finalizer.cpp
greg7mdp: done

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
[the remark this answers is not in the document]
greg7mdp: updated!

libraries/chain/block_state.cpp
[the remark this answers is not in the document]
greg7mdp: I added a comment in `bls_utils.hpp`:

std::span<const std::vector<uint8_t>> messages, // should be `std::span<const std::span<const uint8_t>>`

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: removed it as you suggested.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: done.

libraries/chain/controller.cpp
```diff
@@ -2730,6 +2775,21 @@ struct controller_impl {
...
+                     // we create the non-legacy fork_db, as from this point we may need to cast votes to participate
+                     // to the IF consensus.
+                     // See https://hackmd.io/JKIz2TWNTq-xcWyNX4hRvw -  [if todo] set values accurately
```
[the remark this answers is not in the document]
greg7mdp: done.

tests/CMakeLists.txt
```diff
@@ -205,8 +205,8 @@ set_property(TEST terminate-scenarios-if-test-resync PROPERTY LABELS nonparallel
 # requires https://github.com/AntelopeIO/leap/issues/2141
 #add_test(NAME terminate-scenarios-if-test-replay COMMAND tests/terminate-scenarios-test.py -c replay --terminate-at-block 10 --kill-sig term --activate-if ${UNSHARE} WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 #set_property(TEST terminate-scenarios-if-test-replay PROPERTY LABELS nonparallelizable_tests)
-add_test(NAME terminate-scenarios-if-test-hard_replay COMMAND tests/terminate-scenarios-test.py -c hardReplay --terminate-at-block 10 --kill-sig term --activate-if ${UNSHARE} WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
-set_property(TEST terminate-scenarios-test-hard_replay PROPERTY LABELS nonparallelizable_tests)
+#add_test(NAME terminate-scenarios-if-test-hard_replay COMMAND tests/terminate-scenarios-test.py -c hardReplay --terminate-at-block 10 --kill-sig term --activate-if ${UNSHARE} WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
+#set_property(TEST terminate-scenarios-if-test-hard_replay PROPERTY LABELS nonparallelizable_tests)
```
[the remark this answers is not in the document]
greg7mdp: Added the note to #2141

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: Added the TODO comment.

libraries/chain/hotstuff/finalizer.cpp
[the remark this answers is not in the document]
greg7mdp: [done](https://github.com/AntelopeIO/leap/blob/f073ee7634b28ac246c880f13d28acba400a7688/libraries/chain/hotstuff/finalizer.cpp#L7-L39)

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
+#include <eosio/chain/hotstuff/finalizer.hpp>
+#include <eosio/chain/exceptions.hpp>
+#include <fc/log/logger_config.hpp>
+
+namespace eosio::chain {
+
+// ----------------------------------------------------------------------------------------
+block_state_ptr get_block_by_height(const fork_database_if_t::branch_type& branch, uint32_t block_num) {
+   auto it = std::find_if(branch.begin(), branch.end(),
+                          [&](const block_state_ptr& bsp) { return bsp->block_num() == block_num; });
```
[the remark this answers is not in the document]
greg7mdp: [done](https://github.com/AntelopeIO/leap/blob/f073ee7634b28ac246c880f13d28acba400a7688/libraries/chain/hotstuff/finalizer.cpp#L7-L39)

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+        [t=res.b1->core.last_qc_block_num](const block_state_ptr& bsp) { return bsp->block_num() == t; });
+   if (it == branch.end())
+      return res;
+   res.b = *it;
+
+   return res;
+}
+
+// ----------------------------------------------------------------------------------------
+bool extends(const fork_database_if_t& fork_db, const block_state_ptr& descendant, const block_id_type& ancestor) {
```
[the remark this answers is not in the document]
greg7mdp: [done](https://github.com/AntelopeIO/leap/blob/1400f531022788b1f450d3ed9310021dce56082d/libraries/chain/fork_database.cpp#L436-L453)

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+void finalizer_set::save_finalizer_safety_info() const {
+
+   if (!persist_file.is_open()) {
+      EOS_ASSERT(!persist_file_path.empty(), finalizer_safety_exception,
+                 "path for storing finalizer safety information file not specified");
+      if (!std::filesystem::exists(persist_file_path.parent_path()))
+          std::filesystem::create_directories(persist_file_path.parent_path());
+      persist_file.set_file_path(persist_file_path);
+      persist_file.open(fc::cfile::truncate_rw_mode);
+      EOS_ASSERT(persist_file.is_open(), finalizer_safety_exception,
```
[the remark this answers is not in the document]
greg7mdp: removed assert.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+   } catch(std::exception& e) {
+      elog( "unable to open finalizer safety persistence file ${p}, using defaults. Exception: ${e}",
+            ("p", persist_file_path)("e", e.what()));
+      return res;
+   } catch(...) {
+      elog( "unable to open finalizer safety persistence file ${p}, using defaults", ("p", persist_file_path));
+      return res;
+   }
+   EOS_ASSERT(persist_file.is_open(), finalizer_safety_exception,
+              "unable to open finalizer safety persistence file: ${p}", ("p", persist_file_path));
```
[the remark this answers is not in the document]
greg7mdp: removed assert.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+   try {
+      persist_file.seek(0);
+      uint64_t magic = 0;
+      fc::raw::unpack(persist_file, magic);
+      EOS_ASSERT(magic == fsi_t::magic, finalizer_safety_exception,
+                 "bad magic number in finalizer safety persistence file: ${p}", ("p", persist_file_path));
+      uint64_t num_finalizers {0};
+      fc::raw::unpack(persist_file, num_finalizers);
+      for (size_t i=0; i<num_finalizers; ++i) {
+         fsi_map::value_type entry;
```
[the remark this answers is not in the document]
greg7mdp: [done](https://github.com/AntelopeIO/leap/blob/a120592ac5b11cb0da09c0aabc3d769a15530436/libraries/chain/hotstuff/finalizer.cpp#L205-L211)

libraries/chain/controller.cpp
```diff
@@ -814,7 +823,7 @@ struct controller_impl {
    named_thread_pool<chain>        thread_pool;
    deep_mind_handler*              deep_mind_logger = nullptr;
    bool                            okay_to_print_integrity_hash_on_stop = false;
-   bls_key_map_t                   node_finalizer_keys;
+   finalizer_set                   my_finalizers;
```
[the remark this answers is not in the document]
greg7mdp: It is updated to `my_finalizers_t`

libraries/chain/hotstuff/finalizer.cpp
linh2931: OK. Not change it at the moment.

libraries/chain/hotstuff/finalizer.cpp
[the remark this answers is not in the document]
greg7mdp: Actually I undid this as per our conversation (change will be easier with your adapter).

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <eosio/chain/fork_database.hpp>
+#include <fc/crypto/bls_utils.hpp>
+#include <fc/io/cfile.hpp>
+#include <fc/reflect/reflect.hpp>
+#include <fc/crypto/sha256.hpp>
+#include <compare>
+#include <utility>
+
+namespace eosio::chain {
+   // ----------------------------------------------------------------------------------------
```
ericpassmore, line 13: This is a new file, can we add a comment(s) that helps define what changes go into finalizer.hpp/finalizer.cpp by explaining the purpose or role of finalizer structs? This would help future developers understand what changes should go here vs other files like the controller.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,259 @@
...
+      // --------------------------------------------------------------------------
+      if (extends(fork_db, proposal, fsi.lock.id))
+         safety_check = true;
+
+      // Liveness check : check if the height of this proposal's justification is higher
+      // than the height of the proposal I'm locked on.
+      // This allows restoration of liveness if a replica is locked on a stale proposal
+      // -------------------------------------------------------------------------------
+      if (proposal->last_qc_block_timestamp() > fsi.lock.timestamp)
+         liveness_check = true;
```
heifner, line 52: If `liveness_check` is true no reason to do the expensive safety check.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,259 @@
...
+   }
+   return false;
+}
+
+// ----------------------------------------------------------------------------------------
+finalizer::vote_decision finalizer::decide_vote(const block_state_ptr& proposal, const fork_database_if_t& fork_db) {
+   bool safety_check   = false;
+   bool liveness_check = false;
+
+   auto p_branch = fork_db.fetch_branch(proposal->id());
```
heifner, line 36: No reason to grab the branch unless `monotony_check && (liveness_check || safety_check)`, this can be moved down into the `if`.

libraries/chain/include/eosio/chain/exceptions.hpp
```diff
@@ -668,4 +668,9 @@ namespace eosio { namespace chain {
                                     3250002, "Protocol feature exception (invalid block)" )
       FC_DECLARE_DERIVED_EXCEPTION( protocol_feature_iterator_exception, protocol_feature_exception,
                                     3250003, "Protocol feature iterator exception" )
+
+   FC_DECLARE_DERIVED_EXCEPTION( finalizer_exception,    chain_exception,
+                                 3260000, "Finalizer exception" )
+      FC_DECLARE_DERIVED_EXCEPTION( finalizer_safety_exception, finalizer_exception,
+                                    3260001, "Protocol feature validation exception" )
```
heifner, line 675: Update description.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,302 @@
...
+   auto it = std::find_if(branch.begin(), branch.end(),
+                          [&](const block_state_ptr& bsp) { return bsp->block_num() == block_num; });
+   return it == branch.end() ? block_state_ptr{} : *it;
+}
+
+// ----------------------------------------------------------------------------------------
+qc_chain_t finalizer::get_qc_chain(const block_state_ptr& proposal, const branch_type& branch) const {
+   qc_chain_t res;
+
+   // get b2
```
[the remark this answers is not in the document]
greg7mdp: done.

libraries/chain/hotstuff/finalizer.cpp
greg7mdp: done

libraries/chain/hotstuff/finalizer.cpp
greg7mdp: done

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
greg7mdp: removed.

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include "eosio/chain/block_timestamp.hpp"
```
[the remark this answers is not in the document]
greg7mdp: done.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -39,14 +33,15 @@ struct block_header_state_core {
    uint32_t                last_final_block_num = 0;       // last irreversible (final) block.
    std::optional<uint32_t> final_on_strong_qc_block_num;   // will become final if this header achives a strong QC.
    std::optional<uint32_t> last_qc_block_num;              //
-   uint32_t                finalizer_policy_generation;    // 
+   block_timestamp_type    last_qc_block_timestamp;        // The block timestamp of the most recent ancestor block that has a QC justification
+   uint32_t                finalizer_policy_generation;    //
```
linh2931, line 37: Initialize to `0`.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+         auto operator==(const proposal_ref& o) const {
+            return id == o.id && timestamp == o.timestamp;
+         }
+      };
+
+      struct safety_information {
+         block_timestamp_type last_vote_range_start;
+         proposal_ref         last_vote;          // v_height under hotstuff
+         proposal_ref         lock;               // b_lock under hotstuff
```
linh2931, line 57: maybe no need to mention hotstuff.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -0,0 +1,259 @@
...
+         fsi.lock = proposal_ref(bsp_final_on_strong_qc);
+
+      decision = enough_for_strong_vote ? vote_decision::strong_vote : vote_decision::weak_vote;
+   } else {
+      dlog("last_qc_block_num=${lqc}, fork_db root block_num=${f}",
+           ("lqc",!!proposal->last_qc_block_num())("f",fork_db.root()->block_num()));
+      if (proposal->last_qc_block_num())
+         dlog("last_qc_block_num=${lqc}", ("lqc", *proposal->last_qc_block_num()));
+   }
+   if (decision != vote_decision::no_vote)
```
linh2931, line 90: It might be useful to to log no_vote too.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,149 @@
...
+      qc_chain_t     get_qc_chain(const block_state_ptr& proposal, const branch_type& branch) const;
+      vote_decision  decide_vote(const block_state_ptr& proposal, const fork_database_if_t& fork_db);
+
+   public:
+      std::optional<vote_message> maybe_vote(const bls_public_key& pub_key, const block_state_ptr& bsp,
+                                             const digest_type& digest, const fork_database_if_t& fork_db);
+   };
+
+   // ----------------------------------------------------------------------------------------
+   struct my_finalizers_t {
```
linh2931, line 84: I thought this name again. It seems `my_fianlizers` is more accurate, as we are talking about the finalizers on the local node. In configuration instructions to BPs, local finalizer sounds more natural than my finalizers. `my` implies something at implementation level like a class or a file. @heifner @arhag ?

libraries/chain/include/eosio/chain/exceptions.hpp
greg7mdp: updated.

libraries/chain/block_state.cpp
```diff
@@ -154,20 +154,19 @@ void block_state::verify_qc(const valid_quorum_certificate& qc) const {
...
    }
 
    // validate aggregated signature
-   EOS_ASSERT( fc::crypto::blslib::aggregate_verify( pubkeys, digests, qc._sig ),  invalid_qc_claim, "signature validation failed" );
+   EOS_ASSERT( fc::crypto::blslib::aggregate_verify( pubkeys, digests, qc._sig ),
+               invalid_qc_claim, "signature validation failed" );
 }
 
 std::optional<quorum_certificate> block_state::get_best_qc() const {
-   auto block_number = block_num();
```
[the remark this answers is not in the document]
greg7mdp: I was not seeing the benefit of adding a variable, but I don't mind adding it back.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
greg7mdp: removed comments.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
greg7mdp: This is a class name at the implementation level.

libraries/chain/hotstuff/finalizer.cpp
greg7mdp: yes, updated.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -0,0 +1,149 @@
...
+      enum class vote_decision { strong_vote, weak_vote, no_vote };
+
+      struct proposal_ref {
+         block_id_type         id;
+         block_timestamp_type  timestamp;
+
+         proposal_ref() = default;
+
+         template<class BSP>
+         proposal_ref(const BSP& p) :
```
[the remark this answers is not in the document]
greg7mdp: done.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -86,6 +87,12 @@ namespace eosio::chain {
        */
       branch_type fetch_branch( const block_id_type& h, uint32_t trim_after_block_num = std::numeric_limits<uint32_t>::max() ) const;
 
+      /**
+       *  Returns full branch of `block_header_state` pointers all the way to the fork_db root.
```
heifner, line 91: I think this should say "Returns full branch of `block_header_state` pointers including the `root`". It was not clear to me that this included the root.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -34,9 +34,10 @@ namespace eosio::chain {
       using bhsp             = bs::bhsp_t;
       using bhs              = bhsp::element_type;
       using bsp_t            = bsp;
-      using branch_type      = deque<bsp>;
+      using branch_type      = std::vector<bsp>;
+      using full_branch_type = std::vector<bhsp>;
```
heifner, line 38: Is the idea that we should use vector because these are expected to be small? That probably makes sense, but curious why the change.

libraries/chain/fork_database.cpp
```diff
@@ -412,6 +410,7 @@ namespace eosio::chain {
    fork_database_t<bsp>::branch_type
    fork_database_impl<bsp>::fetch_branch_impl(const block_id_type& h, uint32_t trim_after_block_num) const {
       branch_type result;
+      result.reserve(8);
```
heifner, line 413: `8` seems like an odd choice if you are going to use non-default.
Seems like we could use `index.size()+1` assuming our total fork count would never be too large.
Same comment for the reserve below.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: sure!

unittests/finality_tests.cpp
```diff
@@ -453,6 +454,7 @@ BOOST_AUTO_TEST_CASE(delayed_strong_weak_lost_vote) { try {
 
    BOOST_REQUIRE(cluster.produce_blocks_and_verify_lib_advancing());
 } FC_LOG_AND_RETHROW() }
+#endif
```
linh2931, line 457: I believe in Boost unit tests, when safety file is the default, those simulated weak vote tests should have passed. Should come back to revisit those later.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -58,29 +60,29 @@ finalizer::vote_decision finalizer::decide_vote(const block_state_ptr& proposal,
       safety_check   = false;
    }
 
-   dlog("liveness_check=${l}, safety_check=${s}, monotony_check=${m}",
-        ("l",liveness_check)("s",safety_check)("m",monotony_check));
+   dlog("liveness_check=${l}, safety_check=${s}, monotony_check=${m}, ${can} vote",
+        ("l",liveness_check)("s",safety_check)("m",monotony_check)("can",(liveness_check || safety_check)?"can":"cannot"));
```
heifner, line 60: I actually prefer `vote=` instead of the "fancy", `can/cannot`. Easier to see in the log output.

unittests/finality_tests.cpp
```diff
@@ -227,6 +227,7 @@ BOOST_AUTO_TEST_CASE(lost_votes) { try {
    BOOST_REQUIRE(cluster.produce_blocks_and_verify_lib_advancing());
 } FC_LOG_AND_RETHROW() }
 
+#if 0
```
heifner, line 230: A #warning would be good here.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: Sure, since they are expected to be small, I think vector is more appropriate for sure. 

But even for hundreds  or thousands typically I think we should prefer vectors to `deque`, because we can `reserve` the appropriate size (fewer memory allocations) and subsequent access is cache friendly because memory is contiguous.

IMO we significantly overuse `deque` in our code base.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -33,23 +33,25 @@ finalizer::vote_decision finalizer::decide_vote(const block_state_ptr& proposal,
...
-      // --------------------------------------------------------------------------
-      if (extends(fork_db, proposal, fsi.lock.id))
-         safety_check = true;
+   if (!monotony_check) {
+      dlog("monotony check failed for proposal ${p}, cannot vote", ("p", proposal->id()));
+      return vote_decision::no_vote;
+   }
 
+   if (!fsi.lock.empty()) {
       // Liveness check : check if the height of this proposal's justification is higher
```
linh2931, line 40: We don't use height any more. Maybe change to block number?

libraries/chain/fork_database.cpp
greg7mdp: I don't think the size rreally matters, as long as it is big enough for 99% of the cases.
Realistically it makes no difference in speed wether we allocate space for 5, 8, or 10 pointers, it is still one heap allocation. So I just picked 8 as I thought it would be large enough. It is just a reserve so it is not critical.

We could use `index.size()` but I'm not sure if that is just a lookup of a member variable, and even then it is probably slower than just hardcoding a size.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: If you don't agree I don't mind changing `branch_type` back to `deque`.

libraries/chain/block_state.cpp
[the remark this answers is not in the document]
linh2931: Oh, don't revert.

unittests/finality_tests.cpp
greg7mdp: added

libraries/chain/hotstuff/finalizer.cpp
greg7mdp: While we standardized on `block_num`, I think using height in this sentence makes it clearer. What do you think @heifner ?

libraries/chain/hotstuff/finalizer.cpp
greg7mdp: At this point I don't know what the vote will be, just that we can or cannot vote.
If we can vote, the vote is spelled out [below](https://github.com/AntelopeIO/leap/blob/a512f1793e9bcc48ecf4407ca5e2675cd5128228/libraries/chain/hotstuff/finalizer.cpp#L94-L95)

libraries/chain/include/eosio/chain/fork_database.hpp
heifner: I think vector is fine for these `shared_ptr`s as the size is small for each one and our fork db should never be very deep. Even in dpos it is typically only about 12.
As for `significantly overuse deque in our code base`, we demonstrated significate performance improvement by using `deque` instead of `vector` in controller. Those are also for small types, although the size was ~10K. ~20K TPS.
On other projects, I have also seen issues with `vector` not providing the performance benefit often discussed.

heifner: deque is less like list and more like vector. You get groups of locality of reference unlike wth list.

libraries/chain/fork_database.cpp
heifner: Looks like `index.size()` does return a cached size value.

libraries/chain/hotstuff/finalizer.cpp
heifner: Personally, I think height in comments if fine.

libraries/chain/hotstuff/finalizer.cpp
heifner: I just meant `vote=true`, `vote=false`

greg7mdp: I actually think this is less clear. What about `can vote = true`, `can vote = false`?

heifner: Yes, that is fine, or even just leave it how you had it, not a big deal.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: Yes, indeed, but the size of the group is implementation-defined and unreliable, as we see with the Microsoft STL version. What benefit do you see with `deque` in this use case?

libraries/chain/fork_database.cpp
greg7mdp: OK I'll use that.

libraries/chain/include/eosio/chain/fork_database.hpp
heifner: I don't see any in this use-case. Our branches should be very small.

heifner: Our `deque` is defined as:
```
   using block_1024_option_t = boost::container::deque_options< boost::container::block_size<1024u> >::type;
   template<typename T>
   using deque = boost::container::deque< T, void, block_1024_option_t >;
```

libraries/chain/include/eosio/chain/block_header_state.hpp
greg7mdp: done.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: Yes, using this deque definition will prevent the pathological poor behavior of Microsoft's version.
Still, it seems to me that a  `1024` block size is, more often than not, less optimal than using a correctly `reserve'd` vector.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
greg7mdp: added.

libraries/chain/include/eosio/chain/fork_database.hpp
heifner: When you know the size ahead of time, vector is clear winner in all cases. It is when you don't know and you vector has to grow multiple times that it can kill performance.

heifner (approved):

linh2931 (approved):

Issue #2136, opened 2024-01-24
IF: Bring back is_needed to avoid adding redundant QCs to blocks; enforce sensible finalizer policy in setfinalizers


arhag:
If the QC claim in the to-be-assembled block is the same (same block number and strong/weak status) as the QC claim in the previous block, then the QC should not be included in the block extension.

Also, within the `setfinalizers` host function, assert that the sum of the weights is greater than or equal to the threshold.

PR #2124, opened 2024-01-22
IF: Integration tests enabled for instant-finality (Part 1)

heifner:
Adding additional integration tests for instant-finality
- block_log_util_if_test
- block_log_retain_blocks_if_test
- cluster_launcher_if
- distributed-transactions-if-test
- nodeos_forked_chain_if_lr_test
- nodeos_contrl_c_if_test
- nodeos_chainbase_allocation_if_test
- nodeos_producer_watermark_if_lr_test
- nodeos_high_transaction_if_lr_test
- larger_lib_if_test
- light_validation_sync_if_test
- auto_bp_peering_if_test

There are more that need to be added, but this set gives us much more coverage than we currently have. Merging this now, and will add additional tests in a (Part 2).

PR #2137, opened 2024-01-25
IF: Bring back is_needed to avoid adding redundant QCs to blocks; enforce sensible finalizer policy in setfinalizers host function

linh2931:
* If the QC claim in the to-be-assembled block is the same (same block number and strong/weak status) as the QC claim in the previous block, then the QC should not be included in the block extension. `is_needed` was removed by https://github.com/AntelopeIO/leap/pull/2111 due to test failure. There was actually a bug in the original use of `is_needed`. It should have been used against parent block. This PR brings back `is_needed` and fixes the bug.

* If the sum of weights is less than threshold, the quorum will never be met. This PR asserts that the sum of the weights is greater than or equal to the threshold within the setfinalizers host function, 

Resolved https://github.com/AntelopeIO/leap/issues/2136

### DIFF
--- a/libraries/chain/webassembly/privileged.cpp
+++ b/libraries/chain/webassembly/privileged.cpp
@@ -196,7 +196,7 @@ namespace eosio { namespace chain { namespace webassembly {
                                                                 .public_key{fc::crypto::blslib::bls_public_key{*pk}}});
       }
 
-      EOS_ASSERT( weight_sum >= finpol.threshold && finpol.threshold > weight_sum / 2, wasm_execution_error, "Finalizer policy threshold (${t}) cannot be met by finalizer weights (${w})", ("t", finpol.threshold)("w", weight_sum) );
+      EOS_ASSERT( weight_sum >= finpol.threshold && finpol.threshold > weight_sum / 2, wasm_execution_error, "Finalizer policy threshold (${t}) must be greater than half of the sum of the weights (${w}), and less than or equal to the sum of the weights", ("t", finpol.threshold)("w", weight_sum) );
 
       context.control.set_proposed_finalizers( finpol );
    }
--- a/libraries/chain/webassembly/privileged.cpp
+++ b/libraries/chain/webassembly/privileged.cpp
@@ -196,7 +196,7 @@ namespace eosio { namespace chain { namespace webassembly {
                                                                 .public_key{fc::crypto::blslib::bls_public_key{*pk}}});
       }
 
-      EOS_ASSERT( finpol.threshold > weight_sum / 2, wasm_execution_error, "Finalizer policy threshold cannot be met by finalizer weights" );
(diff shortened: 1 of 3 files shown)


heifner (approved):

libraries/chain/webassembly/privileged.cpp
```diff
@@ -196,7 +196,7 @@ namespace eosio { namespace chain { namespace webassembly {
                                                                 .public_key{fc::crypto::blslib::bls_public_key{*pk}}});
       }
 
-      EOS_ASSERT( finpol.threshold > weight_sum / 2, wasm_execution_error, "Finalizer policy threshold cannot be met by finalizer weights" );
+      EOS_ASSERT( weight_sum >= finpol.threshold && finpol.threshold > weight_sum / 2, wasm_execution_error, "Finalizer policy threshold (${t}) cannot be met by finalizer weights (${w})", ("t", finpol.threshold)("w", weight_sum) );
```
greg7mdp, line 199: If we have 2 finalizers with `weight==1` each, that is a `weight_sum==2`. Isn't a threshold of 1 OK?
With the test as is: `finpol.threshold > weight_sum / 2`, threshold would have to be 2.

linh2931: @heifner and @arhag ?

heifner: I believe threshold of 2 is desired for 2 finalizers. Otherwise the one finalizer is all that would be needed to finalize the block.

greg7mdp: It is not only 2. Assuming weights of 1, the current check requires a minimum threshold of 3 for 4 finalizers, 4 for 6, ...

heifner:
Seems we need https://github.com/AntelopeIO/reference-contracts/issues/41
```
/__w/leap/leap/reference-contracts/tests/main.cpp(19): fatal error: in "eosio_bios_if_tests/set_1_finalizer": Caught Unexpected Exception
3070002 wasm_execution_error: Runtime Error Processing WASM
Finalizer policy threshold (4) cannot be met by finalizer weights (3)
    {"t":4,"w":3}
```
Before libtester tests will pass.

heifner:
This does seem to fix the test failures I was seeing, here is a run of #2124 with this PR merged into it: https://github.com/AntelopeIO/leap/actions/runs/7649477360

libraries/chain/webassembly/privileged.cpp
linh2931: The hotstuff paper uses `2f + 1` as threshold where `n = 3f + 1`. Our version of minimum `weight_sum/2` looks reasonable.

arhag: `threshold > weight_sum / 2` is correct.

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Avoid adding redundant QCs to blocks, by bringing back `is_needed`
Note: end

PR #2140, opened 2024-01-25
upgrade to `actions/upload-artifact@v4`

spoonincode:
Upgrade to `actions/upload-artifact@v4`. This has a couple nice benefits (as advertised by v4),

Upload performance is substantially better than v3, so much so that we can get rid of my `upload-artifact-large-chunks-action` and just rely on the canonical `upload-artifact` going forward.

More interestingly, artifacts are viewable and downloadable before the workflow completes. This has long been an annoyance to me so great to see this resolved. It's nice to, for example, grab the build or the leap-dev or some logs before the entire workflow has completed (i.e. before LR tests take.. a long time to complete). It'll also be hugely beneficial to usages of `asset-artifact-download-action` for similar reasons. For example, a workflow that's still in progress but has a leap-dev.deb completed can be used immediately instead of waiting for LR tests to complete. Sadly, I believe, `asset-artifact-download-action` will need to be refactored to deal with this new possibility (I believe it waits for the workflow to be completed currently). Maybe one day..

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -93,7 +93,7 @@ jobs:
         with:
           submodules: recursive
       - name: Download builddir
-        uses: actions/download-artifact@v3
+        uses: actions/download-artifact@v4
         with:
           name: ${{matrix.platform}}-build
       - name: Build packages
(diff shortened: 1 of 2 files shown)


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: RELEASE
category: CHORE
summary: Upgrade to github `actions/upload-artifact@v4` which fixes large uploads, and makes artifacts viewable before workflow completes.
Note:end

Issue #2123, opened 2024-01-22
IF: Remove chain_pacemaker and qc_chain related files and code


linh2931:
`chain_pacemaker.(hpp/cpp)` and `qc_chain.(hpp/cpp)` are no longer needed. Remove them to save maintenance.

~~Ensure tests using `qc_chain` and `chain_pacemaker` have been ported to work with the new implementation.~~

As it is a significant work to port the large number of the tests, https://github.com/AntelopeIO/leap/issues/2147 is created to track the work.

arhag:
Also avoid using `hotstuff` in the code. And prefer using just `finality` rather than `instant_finality` if possible.

PR #2146, opened 2024-01-26
IF: remove chain_pacemaker and qc_chain related files

linh2931:
Remove chain_pacemaker and qc_chain related files and tests. Do the removal now so that we don't need to maintain the tests using qc_chain and chain_pacemaker while making changes to common files. 

Porting tests to work with the new implementation is tracked by https://github.com/AntelopeIO/leap/issues/2147.

Resolved https://github.com/AntelopeIO/leap/issues/2123.


libraries/chain/hotstuff/test/hotstuff_tools.cpp
```diff
@@ -1,63 +1,14 @@
-#include <boost/test/unit_test.hpp>
+#define BOOST_TEST_MODULE hotstuff
+
```
heifner, line 2: `hotstuff_tools.cpp` seems like an odd name for this file. Maybe rename to `hotstuff_misc_tests.cpp`

heifner (changes requested):

libraries/chain/hotstuff/test/hotstuff_tools.cpp
linh2931: Changed to `finality_misc_tests.cpp` for eventual renaming `hotstuff` directory to `finality` directory.

heifner (approved):

greg7mdp:
lgtm, but I'm wondering if before removing all the tests that existed for the qc_chain implementation, we should not at least review them, and convert those which make sense for our new implementation of IF?

heifner:
> lgtm, but I'm wondering if before removing all the tests that existed for the qc_chain implementation, we should not at least review them, and convert those which make sense for our new implementation of IF?

That is what #2147 is for.

greg7mdp (approved):

linh2931:
The adapting of the tests will be done by https://github.com/AntelopeIO/leap/issues/2147 (It will be a significant work).  The tests being removed do not cover any new implementation; removing them does not reduce our existing test coverage but saves us from updating them just for compile whenever we change some common data types.

ericpassmore:
Note:start
group: IF
category: TEST
summary: Remove original pacemaker code and update tests to match current working code. 
Note: end

PR #2148, opened 2024-01-27
[5.0] Report read-only API not enabled to caller

heifner:
Report `read-only transactions execution not enabled on API node. Set read-only-threads > 0` to caller when `/v1/chain/send_read_only_transaction` is not supported because of no configured `read-only-threads`. Note this is handled correctly in 4.0.x.

Includes some misc fixes found while debugging this issue.
- Log http errors before calling the callback `cb`, as `cb` may throw; e.g. `cb` not set.
- Add missing `}` in a `producer_plugin` error log, so the error is reported correctly.
- Add the response to the `trx_provider` error log when the `http` call fails.


greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Now report read-only API not enabled to caller, when `read-only-threads` is not configured. 
Note: end

Issue #2095, opened 2024-01-16
IF: Extend Boost test to show finality advancing


arhag:
Extend Boost test introduced in https://github.com/AntelopeIO/leap/issues/2045 for transition to IF to keep it going for longer and show finality advances.

This will require adding a mechanism in libtester to allow a finalizer to vote on blocks.

PR #2154, opened 2024-01-29
IF: extend Boost test to show finality LIB advancing

linh2931:
Extend Boost test to show finality advances after transition to Instant Finality. This is achieved by configuring finalizer private keys on the Controller to allow a finalizer to vote on blocks.

Resolved https://github.com/AntelopeIO/leap/issues/2095


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: TEST
summary: Test to ensure finality advances along as LIB moves forward.
Note: end

PR #2160, opened 2024-01-30
IF: Increase LR test timeout from 30 minutes to 45 minutes

heifner:
LR tests taking longer than 30 minutes. 
For example: https://github.com/AntelopeIO/leap/actions/runs/7704500906/job/20997994089

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -229,7 +229,7 @@ jobs:
           error-log-paths: '["build/etc", "build/var", "build/leap-ignition-wd", "build/TestLogs"]'
           log-tarball-prefix: ${{matrix.cfg.name}}
           tests-label: long_running_tests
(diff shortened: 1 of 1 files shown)


linh2931 (approved):

greg7mdp (approved):

spoonincode:
I don't think this will help -- the longest test runs in something like 21-23 minutes when everything passes. It looks like something in the test framework that is preventing it from exiting when it's complete.

ericpassmore:
Note:start
group: IF
category: TEST
summary: Increase timeout for long running tests.
Note: end

heifner:
> I don't think this will help -- the longest test runs in something like 21-23 minutes when everything passes. It looks like something in the test framework that is preventing it from exiting when it's complete.

Thanks. Looks like timeout is due to: #2165

PR #2177, opened 2024-01-31
[5.0] Report snapshot rename location on failure

heifner:
Provide the full path of the snapshot file that can't be renamed if the rename fails. This provides useful information to the user so they can investigate r/w capabilities of the reported paths.


greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Provide the full path of the snapshot file that can't be renamed if the rename fails.
Note: end

Issue #1569, opened 2023-08-26
IF: Verify net messages for validity


heifner:
Depends on #1881 and #1657.

- Verify `hs_vote_message` finalizer is valid one to use. Verify public key is part of the sig. (This should be done once in qc_chain.)
- Verify `quorum_certificate_message` `active_finalizers` public keys are part of the sig. (This should be done once in qc_chain.)
- Verify length of `quorum_certificate_message` `active_finalizers` does not exceed maximum size calculated based on maximum number of allowed finalizers (2^16) and the constraint that each finalizer has a description that does not exceed 256 bytes.

arhag:
So far only for Proposal messages (P) and still not complete. Need to expand for Vote messages (V) and QC messages (Q).

Three categories of message validation failure:

1. Immediately disconnect peer that sent the message:

- Proposal has slot/view number that violates monotonicity (or other objectively verifiable constraints). (P2)
- Proposal is for a slot/view number that is very far in the future. (P4)
- Proposal by leader not allowed to propose for that slot/view number.

2. Track by updating per-connection abuse detector (which may eventually disconnect peer):

- Unlinkable proposal. (P1)
- Proposal is for a slot/view number that is a little bit in the future. (P3)

3. Drop message and do nothing else other than maybe logging:

PR #2179, opened 2024-02-01
IF: Net plugin takes appropriate actions on invalid QC claims and invalid vote messages

linh2931:
When an invalid QC claim or an invalid vote message is received, Net plugin needs to disconnect the peer immediately, keep track of the occurrence, or do nothing, based on the nature of validation failures. This PR classifies validation failures and implements the corresponding actions.

Resolved https://github.com/AntelopeIO/leap/issues/1569.


libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -92,6 +92,14 @@ namespace eosio::chain {
       invalid                  // invalid message (other reason)
    };
 
+   enum class vote_status {
+      succeeded,
```
greg7mdp, line 96: no big deal, but I think `success` is more typical as a status code than `succeeded`

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Take appropriate action when invalid claim or vote is received. 
Note: end

plugins/net_plugin/net_plugin.cpp
```diff
@@ -261,13 +261,17 @@ namespace eosio {
       bool verify_catchup( const connection_ptr& c, uint32_t num, const block_id_type& id ); // locks mutex
 
    public:
+      enum class closing_mode {
+         immediately,
+         back_off
```
heifner, line 266: This `closing_mode` is confusing to me. It is used to determine if you close or handshake. 
Not sure what `backoff` implies. It currently means: don't close but rather send a handshake.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -910,6 +914,7 @@ namespace eosio {
       std::atomic<bool>               block_sync_throttling{false};
       std::atomic<std::chrono::nanoseconds>   last_bytes_sent{0ns};
       std::atomic<boost::asio::ip::port_type> remote_endpoint_port{0};
+      std::atomic<uint64_t>           invalid_finality_msg_total{0};
```
heifner, line 917: This appears to not be used. It is only incremented and set to 0, never checked. Seems like the `block_status_monitor` is doing everything you need.
We may decide we want to increase `max_consecutive_rejected_windows_` if we find we see a lot of `vote_status::unknown_block`. Or we may wish to track them separately.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3691,8 +3701,22 @@ namespace eosio {
...
+      switch( cc.process_vote_message(msg) ) {
+         case vote_status::success:
+            my_impl->bcast_vote_message(connection_id, msg);
+            break;
+         case vote_status::unknown_public_key:
+         case vote_status::invalid_signature: // close peer immediately
+            close( false ); // do not reconnect after closing
+            break;
+         case vote_status::unknown_block: // keep tracking
+            ++invalid_finality_msg_total; // atomic
```
heifner, line 3714: Should this call `c->block_status_monitor_.rejected();` ?

heifner (changes requested):

plugins/net_plugin/net_plugin.cpp
linh2931: https://github.com/AntelopeIO/leap/issues/1569 specifies to track the failures and indicates eventually we may decide to disconnect. The counter is implemented for that purpose. Will ask @arhag in today's meeting to clarify.

linh2931: Use block_status_monitor per discussion. Thanks.

heifner (approved):

PR #2180, opened 2024-02-01
Optimizations for supporting large smart contracts

on-a-t-break:
Increased the default_max_transaction_net_usage to just the default_max_block_net_usage, allowing large smart contracts to be deployed, supporting raw abi / wasm files up to 1,024kb (1mb) without compression & up to 2,048kb (2mb) with compression. This aligns with the default_max_wasm_code_bytes size of 20mb.

Reduced default_min_transaction_cpu_usage from 100 to 90, lowering the floor on transaction CPU costs.

Increased the default_max_wasm_nested_structures from 1024 to 2048 to facilitate more functionality within each Action.


ericpassmore:
These values can be changed on chain with agreement of block producers. The updates would only impacts new chains starting from genesis.

PR #2182, opened 2024-02-01
IF: Fix fork database deadlock

heifner:
Fix for deadlock caused by holding a fork database lock while posting tasks to chain thread pool which also tried to access fork database.

- log_irreversible
  - Grabs fork database lock in `fork_db.apply`
    - In the apply, posts a task to chain thread pool
    - Waits on task via `future.get()`
  - Meanwhile a vote comes in that posts a task to chain thread pool that tries to access the fork database
    - If you use up all your thread pool threads trying to get a lock on fork database then you deadlock

There is a reason I am not a fan of recursive mutexes. https://github.com/AntelopeIO/leap/pull/2113#discussion_r1459796518
I should have listened to my own advice. Recursive mutexes are an indication that your design is wrong. In this case it provided a crutch that let us hit a deadlock in a much more obscure way.

This PR removes the fork database recursive mutex.
- Moves fork database mutex back down into the individual fork database implementations.
- Uses an atomic bool for determining which fork database implementation to use.

Also enable tests:
- auto_bp_peering_if_test 

Fix for flaky test:
- trx_finality_status_forked_test.py

### DIFF
--- a/libraries/chain/fork_database.cpp
+++ b/libraries/chain/fork_database.cpp
@@ -10,7 +10,7 @@
 #include <fc/io/fstream.hpp>
 #include <fc/io/cfile.hpp>
 #include <fstream>
-#include <shared_mutex>
+#include <mutex>
 
 namespace eosio::chain {
    using boost::multi_index_container;
@@ -57,7 +57,7 @@ namespace eosio::chain {
                                          BOOST_MULTI_INDEX_CONST_MEM_FUN(bs, const block_id_type&, id)>,
                            composite_key_compare<std::greater<bool>, std::greater<uint32_t>, std::greater<uint32_t>, sha256_less>>>>;
 
-      std::shared_mutex      mtx;
+      std::mutex             mtx;
       fork_multi_index_type  index;
       bsp                    root; // Only uses the block_header_state_legacy portion
       bsp                    head;
@@ -313,7 +313,7 @@ namespace eosio::chain {
 
    template<class bsp>
    fork_database_t<bsp>::bhsp fork_database_t<bsp>::get_block_header( const block_id_type& id ) const {
-      std::shared_lock g( my->mtx );
+      std::lock_guard g( my->mtx );
       return my->get_block_header_impl( id );
    }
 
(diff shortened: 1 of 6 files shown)


libraries/chain/fork_database.cpp
```diff
@@ -10,6 +10,7 @@
 #include <fc/io/fstream.hpp>
 #include <fc/io/cfile.hpp>
 #include <fstream>
+#include <shared_mutex>
```
greg7mdp, line 13: I think it most likely would be more efficient to make it a `<mutex>` and take exclusive locks everywhere. 
The shared mutex may help if we were doing mostly long reads and an occasional quick mutation here and there. 

But here is is the opposite, the `shared_lock` accesses are very quick so we are not saving much by using read locks, and we pay because the `shared_mutex` is slower than the exclusive mutex.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -108,86 +109,111 @@ namespace eosio::chain {
...
+            } else {
+               f(*fork_db_if);
+            }
+         } else {
+            if (legacy) {
+               return f(*fork_db_legacy);
+            } else {
+               return f(*fork_db_if);
+            }
+         }
```
greg7mdp, line 150: Personal preference, I prefer the more compact version (esp because we have many variations of `apply`), but fine either way.

```suggestion
         if constexpr (std::is_same_v<void, R>) {
            legacy ? f(*fork_db_legacy) : f(*fork_db_if);
         } else {
            return legacy ? f(*fork_db_legacy) :  return f(*fork_db_if);
         }
```

greg7mdp (approved):

libraries/chain/include/eosio/chain/fork_database.hpp
heifner: I like compact code, but the current way allows for easier setting of breakpoints.

greg7mdp: Yes, that's true! I'm fine with leaving as is.

greg7mdp (approved):

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Make code safer by removing the fork database recursive mutex. Resolves deadlock issue.
Note: end

PR #2185, opened 2024-02-01
IF: Fix sigsegv in the use of process_vote_message

linh2931:
Thanks @spoonincode for getting us the core backtrace and @heifner for suggesting the fix.

A reference of `control` caught by a lamda in `process_vote_message` lived past its destruction and caused the crash.

Resolved https://github.com/AntelopeIO/leap/issues/2184

PR #2190, opened 2024-02-02
IF: Shutdown thread pool in controller destructor

heifner:
After discussing https://github.com/AntelopeIO/leap/pull/2185 with @greg7mdp decided to move the shutdown of the `controller_impl` thread pool to the `controller` destructor. The real issue here, I believe, is that we have a circular reference between `controller` <-> `controller_impl`. I think we should break that circular reference, but until we do, this seems like a much safer approach.

This PR takes a step in the direction of removing `controller_impl::self`. It moves the signals to `controller_impl` and provides accessors on controller. It also removes all internal controller uses of `self` except for passing it on to `transaction_context`. Changing `transaction_context` and `apply_context` to use `controller_impl` instead of `controller` is a bigger change left for later. This PR does make the breaking change of signals being accessors instead of direct members on `controller`. Since we have changed what many of the signals emit in 6.0, seems like a good time to make this interface change. The removal of usage of `self` in `transaction_context` and `apply_context` can happen later without impacting external users of `controller`.

Depends on: https://github.com/AntelopeIO/reference-contracts/pull/50 for the `libtester` tests.


greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Better management of `controller_impl` thread pool.
Note: end

linh2931 (approved):
Nice clean up!

greg7mdp (approved):
can you remove the member `controller&  self;` at line 793?

greg7mdp (approved):

PR #2192, opened 2024-02-02
Tests: Explicit shutdown since atexit not called on error

### LINKED ISSUE #2165: Python test shutdown on errors
Been seeing timeouts in our integration tests. These timeouts appear to be caused by not shutting down the cluster of nodeos that are running. This PR https://github.com/AntelopeIO/leap/pull/1182 moved shutdown of `Cluster` to a `atexit.register` call. Unfortunately, that is not called if a test exits due to an error or if `os.exit()` is called directly. 

Seems we need to revert back to `TestHarness.shutdown()` calling `Cluster.shutdown()` instead of using `atexit.register`.

heifner:
Python does not call `atexit` on error. Revert back to calling `shutdown` explicitly from `TestHelper.shutdown`.
`trace_plugin_test.py` was the only test not already explicitly calling `TestHelper.shutdown`.

Resolves #2165

### DIFF
--- a/tests/trace_plugin_test.py
+++ b/tests/trace_plugin_test.py
@@ -117,6 +117,7 @@ class TraceApiPluginTest(unittest.TestCase):
     @classmethod
     def tearDownClass(self):
         TraceApiPluginTest.cluster.testFailed = not testSuccessful
+        TraceApiPluginTest.cluster.shutdown()
 
 if __name__ == "__main__":
     unittest.main()
--- a/tests/TestHarness/Cluster.py
+++ b/tests/TestHarness/Cluster.py
@@ -1,4 +1,3 @@
-import atexit
 import copy
 import subprocess
 import time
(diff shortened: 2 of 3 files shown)


greg7mdp (approved):

tests/trace_plugin_test.py
```diff
@@ -117,6 +117,7 @@ def setUpClass(self):
     @classmethod
     def tearDownClass(self):
         TraceApiPluginTest.cluster.testFailed = not testSuccessful
+        TraceApiPluginTest.cluster.shutdown()
```
linh2931, line 120: Why does this test need to call cluster shutdown explicitly?

linh2931 (approved):

tests/trace_plugin_test.py
heifner: Because it does not call TestHelper shutdown

linh2931: Oh, only it missed calling the shutdown method.

ericpassmore:
Note:start
group: IF
category: TEST
summary: Update python scripts to properly exit on error.
Note: end

PR #2200, opened 2024-02-05
[5.0] make `read-only-threads` defaults clearer in help text

### LINKED ISSUE #2103: Default read-only-threads could be clearer in the help
The default number of read-only-threads could be clearer in the help

```
# Number of worker threads in read-only execution thread pool. Max 8. (eosio::producer_plugin)
# read-only-threads =
```

https://github.com/AntelopeIO/leap/blob/4531e9751a8d6d712801ffab1e2b24fe37dc9ca8/plugins/chain_plugin/chain_plugin.cpp#L2544

spoonincode:
Resolves #2103

### DIFF
--- a/plugins/producer_plugin/producer_plugin.cpp
+++ b/plugins/producer_plugin/producer_plugin.cpp
@@ -590,6 +590,7 @@ public:
    // threads use 11TB (128 * 11 * 8GB)). It is about 11.7% of total VM space
    // in a 64-bit Linux machine (about 128TB).
    static constexpr uint32_t         _ro_max_threads_allowed{128};
+   static constexpr uint32_t         _ro_default_threads_nonproducer{3};
    named_thread_pool<struct read>    _ro_thread_pool;
    fc::microseconds                  _ro_write_window_time_us{200000};
    fc::microseconds                  _ro_read_window_time_us{60000};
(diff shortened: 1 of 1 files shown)


plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1071,7 +1074,7 @@ void producer_plugin::set_program_options(
          ("snapshots-dir", bpo::value<std::filesystem::path>()->default_value("snapshots"),
           "the location of the snapshots directory (absolute path or relative to application data dir)")
          ("read-only-threads", bpo::value<uint32_t>(),
-          "Number of worker threads in read-only execution thread pool. Max 8.")
+         ("Number of worker threads in read-only execution thread pool. Defaults to 0 if configured as producer, otherwise defaults to "s + std::to_string(producer_plugin_impl::_ro_default_threads_nonproducer) + ". Max "s + std::to_string(producer_plugin_impl::_ro_max_threads_allowed) + "."s).c_str())
```
spoonincode, line 1077: > `.c_str()`

`option_description` makes a copy of the `char*` passed in so this should be okay; not sure what a better pattern may be

heifner (approved):

greg7mdp (approved):

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: If `option_description` did not makes a copy of the char* passed, you could return `s.c_str()`, `s` being a `static std::string` declared in a lambda called immediately, but probably fine as is.

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Clearer help text for `read-only-threads` defaults.
Note: end

PR #2210, opened 2024-02-05
[5.0 -> main] change HTTP code on exceeding `http-max-bytes-in-flight-mb` or `http-max-in-flight-requests` to 503

### LINKED ISSUE #2129: Review status code when http-max-bytes-in-flight-mb is reached
Review status code when `http-max-bytes-in-flight-mb` is reached. Also if the code is not 503, fix the help.

Requesting change for 5.0.1.

Copying chat from telegram:

Aaron Cox (jesta) — Greymass, [2024-01-23 11:46 A.M.]
I just had a couple API nodes that entered a state where this check was repeatedly being hit:

https://github.com/AntelopeIO/leap/blob/8a9fa7689aebb0d6df276b8e4c31d3bf64800efe/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp#L234

Aaron Cox (jesta) — Greymass, [2024-01-23 11:47 A.M.]
Traffic wasn’t super heavy, no noticeable errors in nodeos, but every API response was hitting this.

There a configuration value I can change to prevent this, or is there a bug where this is accumulating and not reseting? Not sure what happened here. I haven’t investigated that deeply yet, but anyone else experienced this?

Matt Witherspoon, [2024-01-23 11:49 A.M.]
http-max-bytes-in-flight-mb is the knob to set the limit

Aaron Cox (jesta) — Greymass, [2024-01-23 12:00 P.M.]
I’ll have to study these nginx logs more, but it looks like a barrage of get_table_rows calls came in and that’s when it started.

Aaron Cox (jesta) — Greymass, [2024-01-23 12:02 P.M.]
And I don’t think our failover rules were expecting a 429 response from nodeos, so they didn’t reroute. 

Normally nginx is the one handing out the 429 responses to clients 😂

Aaron Cox (jesta) — Greymass, [2024-01-23 6:07 P.M.]
After finally getting a bit more time to look at this and find a workaround, I’m not sure I can actually failover to alternative nodeos instances when nodeos is stuck returning a 429 error code (at least not with nginx). 

Should it be returning a 429? The help command seems to indicate it would have been a 503, which would have successfully knocked our failing upstreams out of rotation and kept the services running.

https://github.com/AntelopeIO/leap/blob/cf09e01336a4436aed11fe9403c0546b65b19f62/docs/03_keosd/10_usage.md?plain=1#L58-L61

Aaron Cox (jesta) — Greymass, [2024-01-23 6:09 P.M.]
I’m not sure how to reproduce at this point, but since I can’t automatically failover when this occurs, I’m going to set http-max-bytes-in-flight-mb = -1 for the time being to ensure these nodes don’t end up in a deadlock without the automatic failure.

Matthew Darwin | Pinax | Nation, [2024-01-23 6:33 P.M.]
HTTP 429 (too many requests) seems like wrong return code to me.

Matt Witherspoon, [2024-01-23 7:01 P.M.]
yeah disappointing from 2.0 -> 3.1 that was changed from 503 to 429

spoonincode:
Merges #2199, #2205 & #2206 in to main

> 4xx response codes are for errors caused by the client where 5xx is for errors on the server side. The server exceeding its configured resource quota should be a 5xx error not a 4xx error. Change this from 429 to 503 (which is what it was in 2.0, fwiw). A 5xx error in this condition makes it easier for failover tooling, among other things. Resolves #2129

### DIFF
--- a/plugins/http_plugin/http_plugin.cpp
+++ b/plugins/http_plugin/http_plugin.cpp
@@ -357,9 +357,9 @@ namespace eosio {
             ("max-body-size", bpo::value<uint32_t>()->default_value(my->plugin_state->max_body_size),
              "The maximum body size in bytes allowed for incoming RPC requests")
             ("http-max-bytes-in-flight-mb", bpo::value<int64_t>()->default_value(500),
-             "Maximum size in megabytes http_plugin should use for processing http requests. -1 for unlimited. 429 error response when exceeded." )
+             "Maximum size in megabytes http_plugin should use for processing http requests. -1 for unlimited. 503 error response when exceeded." )
             ("http-max-in-flight-requests", bpo::value<int32_t>()->default_value(-1),
-             "Maximum number of requests http_plugin should use for processing http requests. 429 error response when exceeded." )
+             "Maximum number of requests http_plugin should use for processing http requests. 503 error response when exceeded." )
             ("http-max-response-time-ms", bpo::value<int64_t>()->default_value(15),
              "Maximum time on main thread for processing a request, -1 for unlimited")
             ("verbose-http-errors", bpo::bool_switch()->default_value(false),
--- a/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
+++ b/plugins/http_plugin/include/eosio/http_plugin/beast_http_session.hpp
@@ -219,18 +219,18 @@ public:
 
    virtual void send_busy_response(std::string&& what) final {
       error_results::error_info ei;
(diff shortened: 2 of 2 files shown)


greg7mdp (approved):

ericpassmore:
Note:start
group: PROTOCOL
category: HTTP
summary: 503 is now the HTTP error return code for exceeding `http-max-bytes-in-flight-mb` or `http-max-in-flight-requests`. Previous code was 429.
Note: end

Issue #2147, opened 2024-01-26
IF: port tests using qc_chain and chain_pacemaker to work with the new implementation


linh2931:
Depends on https://github.com/AntelopeIO/leap/issues/2193.

https://github.com/AntelopeIO/leap/issues/2123 removes qc_chain and chain_pacemaker related code.

There are a large number of comprehensive tests using qc_chain and chain_pacemaker. They should be reviewed; those appropriate should be ported to work with the new implementation when appropriate.

linh2931:
## Review results

3 files in `hotstuff` test directory contain tests: `test_hotstuff.cpp`, `hotstuff_tools.cpp`, and `test_hotstuff_state.cpp`. 

1.  `hotstuff_tools.cpp` tests votes and is kept. 
2.  `test_hotstuff_state.cpp` was intended to test safety file; it is unfinished. When new finalizer safety file work (https://github.com/AntelopeIO/leap/issues/2070) is finished, new tests should be added.
3. `test_hotstuff.cpp` has 9 tests:
- `hotstuff_1` -- test optimistic responsiveness (3 confirmations per block):  precommit, precommitQC, commit all come within one block. Not applicable. 
- `hotstuff_2` -- test slower network (1 confirmation per block): precommit comes in one block a time. Existing tests cover this. 
- `hotstuff_3` -- test leader rotation: not applicable
- `hotstuff_4` -- test loss and recovery of liveness on new block
- `hotstuff_5` -- test finality violation: This is covered by fork integration tests
- no `hotstuff_6`
- `hotstuff_7` -- test leader rotation with a non-complete connection graph: not applicable
- `hotstuff_8` -- same as hotstuff_1 with duplicate vote messages
- `hotstuff_9` -- test leader rotation with a star toplogy: not applicable
- `hotstuff_10` -- test leader rotation with a ring topology: not applicable

greg7mdp:
Renable weak vote related tests which were disabled in order to merge #2135

linh2931:
Previous hotstuff tests did not have tests for weak votes. Each of the 9 tests has either a similar test now or not applicable. Close this and reopen https://github.com/AntelopeIO/leap/issues/2193 for weak related tests (which cannot be done by simply changing vote messages in the middle of transmission).

PR #2223, opened 2024-02-07
IF: add a test for loss and recovery of finality liveness 

linh2931:
Add a test to verify loss and recovery of finality liveness are handled correctly.

Resolved https://github.com/AntelopeIO/leap/issues/2147


greg7mdp (approved):

heifner (approved):

ericpassmore:
Note:start
group: IF
category: TEST
summary: Add a test to verify loss and recovery of finality liveness are handled correctly. Part of the work for removing `qc_chain` and `chain_pacemaker` structs.
Note:end

PR #2224, opened 2024-02-08
IF: Test: Transition to instant-finality with multiple producers

### LINKED ISSUE #1512: IF: Add integration test for transition from legacy consensus to IF consensus
Depends on https://github.com/AntelopeIO/leap/issues/1522 and https://github.com/AntelopeIO/leap/issues/1548.

Many of the tests transition from pre-IF to IF to carry out the tests (see https://github.com/AntelopeIO/leap/issues/1510). But they transition with 1 BP.

This integration test should test the transition with multiple BPs. To keep test runtime short, stick with 4 BPs.

heifner:
- Adds integration test that verifies transition to instant-finality works with multiple producers producing before transition
- Adds integration test for `setprods` after transition to instant-finality
- Fix issue found with `setprods` to activate new proposer policy at correct block (the first block of the producer)

Resolves #1512

### DIFF
--- a/tests/TestHarness/Cluster.py
+++ b/tests/TestHarness/Cluster.py
@@ -482,6 +482,7 @@ class Cluster(object):
                         Path(instance.config_dir_name), eosdcmd, unstarted=instance.dont_start,
                         launch_time=launcher.launch_time, walletMgr=self.walletMgr, nodeosVers=self.nodeosVers)
             node.keys = instance.keys
+            node.isProducer = len(instance.producers) > 0
             if nodeNum == Node.biosNodeId:
                 self.biosNode = node
             else:
@@ -1000,7 +1001,9 @@ class Cluster(object):
         for n in (self.nodes + [self.biosNode]):
             if not n or not n.keys or not n.keys[0].blspubkey:
                 continue
-            if n.nodeId == Node.biosNodeId and not biosFinalizer:
+            if not n.isProducer:
+                continue
+            if n.nodeId == 'bios' and not biosFinalizer:
                 continue
             numFins = numFins + 1
 
@@ -1016,7 +1019,9 @@ class Cluster(object):
         for n in (self.nodes + [self.biosNode]):
             if not n or not n.keys or not n.keys[0].blspubkey:
                 continue
-            if n.nodeId == Node.biosNodeId and not biosFinalizer:
+            if not n.isProducer:
+                continue
+            if n.nodeId == 'bios' and not biosFinalizer:
                 continue
             setFinStr += f'    {{"description": "finalizer #{finNum}", '
             setFinStr += f'     "weight":1, '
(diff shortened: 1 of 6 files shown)


tests/TestHarness/Cluster.py
```diff
@@ -993,34 +995,33 @@ def parseClusterKeys(totalNodes):
...
-        # pnodes does not include biosNode
-        if Utils.Debug: Utils.Print(f"threshold: {threshold}, numFins: {numFins}, pnodes: {pnodes}")
+        if Utils.Debug: Utils.Print(f"threshold: {threshold}, numFins: {numFins}")
         setFinStr =  f'{{"finalizer_policy": {{'
         setFinStr += f'  "threshold": {threshold}, '
         setFinStr += f'  "finalizers": ['
         finNum = 1
-        for n in launcher.network.nodes.values():
-            if n.index == Node.biosNodeId and not biosFinalizer:
+        for n in (self.nodes + [self.biosNode]):
```
linh2931, line 1019: Can this for loop be combined with the one above? They are in similar shape.

tests/TestHarness/Cluster.py
```diff
@@ -1457,6 +1461,30 @@ def cleanup(self):
             os.remove(f)
 
     # Create accounts, if account does not already exist, and validates that the last transaction is received on root node
```
linh2931, line 1463: This comment should be moved down. Add a new comment about setProds if needed.

tests/transition_to_if.py
```diff
@@ -0,0 +1,81 @@
...
+
+try:
+    TestHelper.printSystemInfo("BEGIN")
+
+    cluster.setWalletMgr(walletMgr)
+
+    Print(f'producing nodes: {pnodes}, topology: {topo}, delay between nodes launch: {delay} second{"s" if delay != 1 else ""}')
+
+    Print("Stand up cluster")
+    if args.plugin:
```
linh2931, line 44: Will we ever test with extra plugins? If so, add the plugins to the test command in CMakeLists.txt.

tests/TestHarness/Cluster.py
```diff
@@ -1044,6 +1045,7 @@ def activateInstantFinality(self, launcher, biosFinalizer, pnodes):
         if not self.biosNode.waitForTransFinalization(transId, timeout=21*12*3):
```
linh2931, line 1045: So is the host function sent to the biosNode first even if biosFinalizer is false?

tests/TestHarness/Cluster.py
heifner: I don't think so, need to have `threshold` calculated first.

tests/TestHarness/Cluster.py
heifner: Yes. Doesn't matter which node you look for the transaction on. Maybe we should rename this python method now. It has been there for a long time. It means is the transaction in a LIB block.

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: TEST
summary: Integration test that verifies transition to instant-finality works with multiple producers.
Note:end

PR #2229, opened 2024-02-08
[5.0] improve reliability of `http_plugin_unit_tests` via port workaround

spoonincode:
For unknown reasons, sometimes when `http_plugin_test_fixture` is destroyed `http_plugin`'s `listener`s are not destroyed despite `http_plugin` itself being destroyed. When this occurs the port that `listener` was listening on gets tied up for the remainder of `http_plugin_unit_tests` causing failures for the remaining test cases.

It's not clear to me why this is occurring. I suspect either some `static` somewhere or something I don't expect holding `listener`'s `shared_ptr`.

In the mean time just don't reuse the same port numbers among all the different test cases. It is rather curious that the `valid_category_addresses` test case already avoided 8888 that the prior `http_plugin_unit_tests` test case used.


linh2931 (approved):

heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: STABILITY
category: TEST
summary: Improve http plugin unit tests by avoiding dangling listeners.
Note: end

Issue #2193, opened 2024-02-02
IF: add more flexibility for IF transition to Boost libtester


linh2931:
Currently `set_finalizers` of libtester takes only a vector of producer names for IF transision. `set_finalizer` hardcodes the threshold and weights for each producer and sets up all producers to vote in the node.

To facilitate more complex tests, `set_finalizer`  or another function like `transition_to_if` should take threshold, weights, and voting producers as parameters to set up IF testing.

Tests using the new function will be added in the same PR resolving the issue.

arhag:
Libtester should have functionality to allow for tests with explicit control of votes: who votes, how they vote (strong/weak), where the votes arrive (which nodes or instances of libtester), and when the votes arrive.

linh2931:
Manipulating finalizer safety info is now in https://github.com/AntelopeIO/leap/issues/2233.

greg7mdp:
weak vote tests are not passing and were disabled. Reopening until these tests can be re-enabled.

PR #2234, opened 2024-02-12
IF: Implement capabilities to modify, reorder, drop, and duplicate votes and add comprehensive vote processing Boost unit tests

linh2931:
Implemented  a test cluster consisting of 3 nodes (controllers) with capabilities to modify, reorder, drop, and duplicate votes:
* Added 18 tests to cover vote processing, including IF transitions, long run of 1000 blocks, conflicting votes, delayed votes, dropped votes, out of order votes.
* Fixed a mutex lockup issue.
* Fixed a missed voting status check issue.

Resolved https://github.com/AntelopeIO/leap/issues/2193 and https://github.com/AntelopeIO/leap/issues/2147


libraries/chain/include/eosio/chain/controller.hpp
```diff
@@ -270,6 +270,7 @@ namespace eosio::chain {
          // Called by qc_chain to indicate the current irreversible block num
          // After hotstuff is activated, this should be called on startup by qc_chain
          void set_if_irreversible_block_num(uint32_t block_num);
+         uint32_t if_lib_num() const;
```
heifner, line 273: Should match the name of the `set`. `if_irreversible_block_num()`

unittests/finality_tests.cpp
```diff
@@ -0,0 +1,739 @@
...
+      BOOST_TEST(fin_policy->finalizers.size() == account_names.size());
+      BOOST_TEST(fin_policy->generation == 1);
+   }
+};
+
+// verify LIB advances with 2 finalizers voting and for a long time
+BOOST_AUTO_TEST_CASE(two_votes_large_number_blocks) { try {
+   tester_cluster cluster;
+
+   for (auto i = 0; i < 1000; ++i) {
```
heifner, line 200: Seems pointless to loop so many times. How could anything be different after say the 2nd loop?

unittests/finality_tests.cpp
```diff
@@ -0,0 +1,739 @@
...
+      BOOST_REQUIRE(cluster.node2_lib_advancing());
+      BOOST_REQUIRE(cluster.node3_lib_advancing());
+   }
+} FC_LOG_AND_RETHROW() }
+
+// verify LIB advances with all of the three finalizers voting
+BOOST_AUTO_TEST_CASE(all_votes) { try {
+   tester_cluster cluster;
+
+   for (auto i = 0; i < 50; ++i) {
```
heifner, line 217: Don't see why this needs to be tested 50 times. 2 or 3 seems plenty. Same for other tests.

unittests/finality_tests.cpp
```diff
@@ -0,0 +1,739 @@
...
+   }
+} FC_LOG_AND_RETHROW() }
+
+// Verify a delayed vote works
+BOOST_AUTO_TEST_CASE(one_delayed_votes) { try {
+   tester_cluster cluster;
+
+   // hold the vote for the first block to simulate delay
+   cluster.produce_and_push_block();
+   cluster.produce_and_push_block();
```
heifner, line 267: Verify lib is not advancing here.

unittests/finality_tests.cpp
```diff
@@ -0,0 +1,739 @@
...
+// Verify 3 consecutive delayed votes work
+BOOST_AUTO_TEST_CASE(three_delayed_votes) { try {
+   tester_cluster cluster;
+
+   // produce 4 blocks and hold the votes for the first 3 to simulate delayed votes
+   // The 4 blocks have the same QC claim as no QCs are created because missing one vote
+   for (auto i = 0; i < 4; ++i) {
+      cluster.produce_and_push_block();
+   }
+
```
heifner, line 298: Verify lib not advancing here.

heifner (changes requested):

libraries/chain/include/eosio/chain/controller.hpp
linh2931: Thanks. Initially I used `if_irreversible_block_num` and then thought it was too long. I will change back to `if_irreversible_block_num` to be consistent.

unittests/finality_tests.cpp
linh2931: Will reduce to 3 for this and others.

unittests/finality_tests.cpp
heifner: I just don't see how it could make any difference how long the test runs.

linh2931: Was thinking about testing a long duration of block production. Here is probably not a right place to do it. After reducing all to 3, running time of all tests in this suite is reduced from 56 seconds to 12 seconds.

heifner (approved):

unittests/finality_tests.cpp
```diff
@@ -0,0 +1,745 @@
...
+   uint32_t                 node1_prev_lib_bum {0};
+   uint32_t                 node2_prev_lib_bum {0};
+   uint32_t                 node3_prev_lib_bum {0};
+   std::vector<eosio::chain::vote_message> carol_votes;
+
+   void setup_node(eosio::testing::tester& node, eosio::chain::account_name local_finalizer) {
+      std::vector<eosio::chain::account_name> account_names = { "alice"_n, "bob"_n, "carol"_n };
+
+      node.produce_block();
+      node.create_accounts(account_names);
```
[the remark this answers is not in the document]
heifner: Also the interface in `tester.hpp`, should use `name` not `account_name`. That is my mistake. Actually we might be better to just use `std::string` as it is only used for the description.

heifner (changes requested):

unittests/finality_tests.cpp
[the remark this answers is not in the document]
linh2931: You are right. We don't need accounts to run those tests. Will change them and update the interface in tester.hpp.

unittests/finality_tests.cpp
```diff
@@ -0,0 +1,745 @@
...
+// After starup up, IF are activated on both nodes.
+//
+// APIs are provided to modify/delay/reoder/remove votes from bob and carol to node1.
+
+enum class vote_mode {
+   strong,
+   weak,
+};
+
+class tester_cluster {
```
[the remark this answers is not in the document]
linh2931: Thanks for the suggestion. I was thinking about this too. Where is the best place for it? Inside `tester.hpp`? Another header in `testing/include/eosio/testing` or just in `unittests`?

unittests/finality_tests.cpp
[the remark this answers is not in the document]
linh2931: They are also used to create deterministic private keys.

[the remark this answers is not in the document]
linh2931: I will remove `create_accounts` call and not use `account_names` to remove the confusion. They are only finalizer names. Will change them to `"node1"_n`, `"node2_n`, and `"node3"_n`.

unittests/finality_tests.cpp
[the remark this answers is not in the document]
linh2931: I will create `finality_tests.hpp` in `unittests` directory for now.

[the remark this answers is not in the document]
heifner: I think `finality_cluster` would be a better name than `finality_tester`. I think all our other testers are singular testers.

ericpassmore:
Note:start
group: IF
category: TEST
summary: Test cluster with additional coverage for vote processing, including IF transitions, long run of 1000 blocks, conflicting votes, delayed votes, dropped votes, and out of order votes.
Note:end

heifner (approved):

unittests/finality_tests.hpp
```diff
@@ -0,0 +1,188 @@
...
+      prev_lib_num = curr_lib_num;
+      return advancing;
+   }
+
+   eosio::testing::tester   node1;
+   eosio::testing::tester   node2;
+   eosio::testing::tester   node3;
+   uint32_t                 node1_prev_lib_bum {0};
+   uint32_t                 node2_prev_lib_bum {0};
+   uint32_t                 node3_prev_lib_bum {0};
```
greg7mdp, line 160: Having a `std::array<node_info, 3>` with 
```
struct node_info {
    eosio::testing::tester   node;
    uint32_t                 node_prev_lib_bum;
};
```

would remove code duplication (in member functions and elsewhere).

greg7mdp (changes requested):

unittests/finality_tests.hpp
linh2931: Thanks. Done

libraries/chain/block_state.cpp
```diff
@@ -88,7 +88,9 @@ std::pair<vote_status, std::optional<uint32_t>> block_state::aggregate_vote(cons
                                  vote.finalizer_key,
                                  vote.sig,
                                  finalizers[index].weight);
-      return {status, strong ? core.final_on_strong_qc_block_num : std::optional<uint32_t>{}};
+      return {status,
+              (status == vote_status::success && strong) ?
+                 core.final_on_strong_qc_block_num : std::optional<uint32_t>{}};
```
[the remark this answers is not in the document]
linh2931: When it returns a non-null-optional, the caller will go ahead to call `set_if_irreversible_block_num` which is a waste. I also think it is cleaner: if a vote is bad, both return values (`status` and `strong`) should be set properly so that they cannot be used. @heifner ?

unittests/finality_tests.hpp
linh2931: Also move the implementation out of .hpp file.

libraries/chain/block_state.cpp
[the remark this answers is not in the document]
heifner: In my branch I have this as:
```cpp
      std::optional<uint32_t> new_lib{};
      if (status == vote_status::success && state == pending_quorum_certificate::state_t::strong)
         new_lib = core.final_on_strong_qc_block_num;
      return {status, state, new_lib};

```

unittests/finality_tests.hpp
heifner: Somewhat minor, but I wonder if you shouldn't name them `node0, node1, node2` if you use the array approach so they match index.

linh2931: Yes, I will make the change. That reduces confusion about which node is which. Thanks.

libraries/chain/block_state.cpp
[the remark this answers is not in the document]
greg7mdp: I'm fine with Kevin's version.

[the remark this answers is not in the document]
greg7mdp: Honestly, I liked it better before any change. The meaning was clear. We're asking `block_state` to aggregate a vote into its `pending_qc`, and it returns a pair where:

`pair.first` is the status of applying the vote (so we know whether to propagate the vote)
`pair.second` is the new `lib` that `pending_qc` certifies (if it achieved strong status). The caller decides what to do with this info.

I'm not understanding why it helps to make it more complicated than that.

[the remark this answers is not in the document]
heifner: This is changing in my PR. We have to let fork_db tell us new lib.

libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -55,15 +56,17 @@ pending_quorum_certificate::pending_quorum_certificate(size_t num_finalizers, ui
...
+   return is_quorum_met_no_lock();
 }
 
 // called by add_vote, already protected by mutex
 vote_status pending_quorum_certificate::add_strong_vote(const std::vector<uint8_t>& proposal_digest, size_t index,
                                                         const bls_public_key& pubkey, const bls_signature& sig,
                                                         uint64_t weight) {
-   if (auto s = _strong_votes.add_vote(proposal_digest, index, pubkey, sig); s != vote_status::success)
+   if (auto s = _strong_votes.add_vote(proposal_digest, index, pubkey, sig); s != vote_status::success) {
+      dlog("add_strong_vote returned failure");
```
greg7mdp, line 67: why is this `dlog` useful since we have a non-conditional `dlog` in `add_vote`?

unittests/finality_test_cluster.hpp
```diff
@@ -0,0 +1,101 @@
...
+   void node1_corrupt_vote_signature();
+
+   // Restore node1's original vote
+   void node1_restore_to_original_vote();
+
+private:
+
+   static constexpr size_t node0 = 0;  // node0 index in nodes array
+   static constexpr size_t node1 = 1;  // node1 index in nodes array
+   static constexpr size_t node2 = 2;  // node2 index in nodes array
```
greg7mdp, line 79: Code will be cleaner if you define these instead:

```
   node_info& node0 = node[0];
   node_info& node1 = node[1];
   node_info& node2 = node[2];
```

and:

```
   void setup_node(node_info& node, eosio::chain::account_name local_finalizer);
   bool lib_advancing(node_info& node);
   eosio::chain::vote_status process_vote(node_info& node, size_t vote_index, vote_mode mode);
   eosio::chain::vote_status process_vote(node_info& node, vote_mode mode);
```

Code will look like this:

```
finality_test_cluster::finality_test_cluster() {
   using namespace eosio::testing;

   setup_node(node0, "node0"_n);
   setup_node(node1, "node1"_n);
   setup_node(node2, "node2"_n);

   // collect node1's votes
   node1.node.control->voted_block().connect( [&]( const eosio::chain::vote_message& vote ) {
      node1.votes.emplace_back(vote);
   });
   // collect node2's votes
   node2.node.control->voted_block().connect( [&]( const eosio::chain::vote_message& vote ) {
      node2.votes.emplace_back(vote);
   });
...
```

greg7mdp (changes requested):

libraries/chain/hotstuff/hotstuff.cpp
linh2931: This is only happening when duplicate occurs; useful for debugging.

unittests/finality_test_cluster.cpp
```diff
@@ -0,0 +1,188 @@
...
+   produce_and_push_block();
+   process_node1_vote();
+   FC_ASSERT(node1_lib_advancing(), "LIB has not advanced on node1");
+   FC_ASSERT(node2_lib_advancing(), "LIB has not advanced on node2");
+
+   // clean up processed votes
+   for (size_t i = 0; i < node.size(); ++i) {
+      node[i].votes.clear();
+      node[i].prev_lib_num = node[i].node.control->if_irreversible_block_num();
+   }
```
greg7mdp, line 38:
```suggestion
   for (auto& n : nodes {
      n.votes.clear();
      n.prev_lib_num = n.node.control->if_irreversible_block_num();
   }
```

unittests/finality_test_cluster.hpp
```diff
@@ -0,0 +1,101 @@
...
+   eosio::chain::vote_status process_node1_vote(uint32_t vote_index, vote_mode mode = vote_mode::strong);
+
+   // send node1's latest vote
+   eosio::chain::vote_status process_node1_vote(vote_mode mode = vote_mode::strong);
+
+   // send node2's vote identified by "index" in the collected votes
+   eosio::chain::vote_status process_node2_vote(uint32_t vote_index, vote_mode mode = vote_mode::strong);
+
+   // send node2's latest vote
+   eosio::chain::vote_status process_node2_vote(vote_mode mode = vote_mode::strong);
```
greg7mdp, line 47: Why not a single finction `process_node_vote` taking a node parameter?

greg7mdp, line 47: Why not a single finction process_node_vote taking a node parameter?

unittests/finality_test_cluster.hpp
```diff
@@ -0,0 +1,101 @@
...
+
+   // returns true if node0's LIB has advanced
+   bool node0_lib_advancing();
+
+   // returns true if node1's LIB has advanced
+   bool node1_lib_advancing();
+
+   // returns true if node2's LIB has advanced
+   bool node2_lib_advancing();
```
greg7mdp, line 56: same here, single function.

unittests/finality_test_cluster.hpp
linh2931: Those are provided as a convenient method for the users and hide the implementation details from them. This way the actual test cases do not need t be changed even if the implementation of test cluster changes again.

greg7mdp (approved):

libraries/chain/block_state.cpp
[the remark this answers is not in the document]
greg7mdp: Isn't still fine to let `aggregate_vote` tell you which block its `pending_qc` is making final, even if that's not the lib that the controller will end up storing?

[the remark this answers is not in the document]
heifner: Potentially, but as is it is not telling you the correct answer because it is not correctly taking previous votes into account. It is only looking at `core` not `updated_core`.

heifner (approved):

libraries/chain/block_state.cpp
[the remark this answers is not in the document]
greg7mdp: I'm not understanding what `updated_core` is about? I thought you may need to update the core of other block state (the descendants), but why the one we are voting on?

Issue #2244, opened 2024-02-20
IF: Integrate new finality core algorithm


linh2931:
@arhag proposed a new finality core algorithm, which is simpler and clearer, and makes other issues simpler to implement, for example, https://github.com/AntelopeIO/leap/issues/2125 and https://github.com/AntelopeIO/leap/issues/2080.

This issue tracks the task to replace block_headter_state_core with the new core.

Issue #2125, opened 2024-01-22
IF: Update fork-choice rule for fork database


arhag:
Depends on https://github.com/AntelopeIO/leap/issues/2244.

Update index for picking the best branch (rename `by_lib_block_num` to `by_best_branch`) to order based on the computed fields derived from data in `block_state`.

This issue assumes the following structures exist:
```cpp
struct qc_claim
{
   uint32_t  block_num;
   bool      is_strong_qc;

   auto operator<=>(const qc_claim&) const = default;
};

struct core_metadata
{
   uint32_t  last_final_block_num;
   uint32_t  final_on_strong_qc_block_num;
   uint32_t  latest_qc_claim_block_num;
};
```

And that there is a member function `next_metadata` in the block header state core with the following function signature:
```cpp
   /**
    *  @pre this->latest_qc_claim().block_num <= most_recent_ancestor_with_qc.block_num <= this->current_block_num()
    *  @pre this->latest_qc_claim() <= most_recent_ancestor_with_qc
    *
    *  @post returned core_metadata has last_final_block_num <= final_on_strong_qc_block_num <= latest_qc_claim_block_num
    *  @post returned core_metadata has latest_qc_claim_block_num == most_recent_ancestor_with_qc.block_num
    *  @post returned core_metadata has final_on_strong_qc_block_num >= this->final_on_strong_qc_block_num
    *  @post returned core_metadata has last_final_block_num >= this->last_final_block_num()
    */
   core_metadata next_metadata(const qc_claim& most_recent_ancestor_with_qc) const;
```

The idea is to keep a cache of a `qc_claim` called `most_recent_ancestor_with_qc` within each `block_state`  in the fork database which makes a QC claim on its most recent ancestor that has a valid QC. And in addition to keep a cache of the `next_core_metadata` that is derived from the `core` of that `block_state` by calling `core.next_metadata(most_recent_ancestor_with_qc)`. Any time `most_recent_ancestor_with_qc` is mutated for a `block_state`, the corresponding `next_core_metadata` should also be recomputed.

If a new QC (weak of strong) is achieved on a `block_state`, Leap must not only update the `most_recent_ancestor_with_qc` of that `block_state` but it must also traverse to the descendant blocks to determine whether their corresponding `most_recent_ancestor_with_qc`s should also be updated. If a `block_state` has a better `qc_claim` (as determined by `operator<`) then its `most_recent_ancestor_with_qc` should not be updated nor should those of any of its descendant blocks.

[...]

Issue #2080, opened 2024-01-12
IF: Implement correct compute_finality_digest


linh2931:
`compute_finality_digest` needs to be implemented in `block_header_state`.

This should include the proper commitment structure that meets the needs of IBC and finality violations proofs.

## Concepts and notation

**Two fork databases exist:** the legacy fork database and the new fork database. Prior to the start of the IF transition, only the legacy fork database is active. After the IF transition has been completed, only the new fork database is active. During the IF transition, both the legacy and new fork databases are simultaneously active. Some blocks will exist in both fork databases during the transition, but there will also be blocks that only exit in one and not the other.

**Four classifications of blocks exist:**
1. *Pre-IF Legacy Block*: A Pre-IF Legacy Block is a block in which `set_finalizers` was never called in it or any of its ancestor blocks. It does not have a finality extension in its block header. A `block_state_legacy` (with a base of `block_header_state_legacy`) must be used to represent the state just after applying the Pre-IF Legacy Block. A block classified as a Pre-IF Legacy Block cannot be classified as anything else. A reversible Pre-IF Legacy Block can only be found in the legacy fork database.
[...]

PR #2245, opened 2024-02-20
IF: Integrate new finality core algorithm

linh2931:
@arhag proposed a new finality core algorithm, which is simpler and clearer, and makes other issues simpler to implement, for example, https://github.com/AntelopeIO/leap/issues/2125 and https://github.com/AntelopeIO/leap/issues/2080.

This PR replaced existing `block_header_state_core` with the new algorithm and updated relevant code to use it.

Resolved https://github.com/AntelopeIO/leap/issues/2244


libraries/chain/finality_core.cpp
```diff
@@ -0,0 +1,362 @@
...
+         size_t links_index = 0; // Default to no garbage collection (if last_final_block_num does not change).
+
+         if (last_final_block_num() < new_last_final_block_num) {
+            // new_blocks_nums found the new_last_final_block_num from a link that had a source_block_num
+            // equal to new_final_on_strong_qc_block_num.
+            // The index within links was (new_final_on_strong_qc_block_num - last_final_block_num).
+            // All prior links can be garbage collected.
+
+            links_index = new_last_final_block_num - last_final_block_num();
+            while ( links_index < links.size() && links[links_index].target_block_num == last_final_block_num() ) {
```
linh2931, line 167: This `while` is needed for cases where multiple links point to the same target.

libraries/chain/finality_core.cpp
```diff
@@ -0,0 +1,362 @@
...
+    *  @post returned core has current_block_num() == this->current_block_num() + 1
+    *  @post returned core has latest_qc_claim() == most_recent_ancestor_with_qc
+    *  @post returned core has final_on_strong_qc_block_num >= this->final_on_strong_qc_block_num
+    *  @post returned core has last_final_block_num() >= this->last_final_block_num()
+    */
+   finality_core finality_core::next(const block_ref& current_block, const qc_claim& most_recent_ancestor_with_qc) const
+   {
+      assert(current_block.block_num() == current_block_num()); // Satisfied by precondition 1.
+      
+      assert(refs.empty() || (refs.back().timestamp <= current_block.timestamp)); // Satisfied by precondition 2.
```
linh2931, line 92: In Areg's original algorithm, the timestamp comparison was `<`. I run into cases where timestamp were the same. I am not sure if the change of '<=' was correct.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Simplification of finality core algorithm
Note:end

libraries/chain/finality_core.cpp
```diff
@@ -0,0 +1,411 @@
...
+         // earlier (and still remain satisfied since next_core.links and next_core.final_on_strong_qc_block_num have not changed).
+      }
+
+      return next_core;
+      // Invariants 1 to 7 were verified to be satisfied for the current value of next_core at various points above. 
+      // (And so, the remaining invariants for next_core are also automatically satisfied.)
+   }
+} /// eosio::chain
+
+#if 0
```
heifner, line 299: Should we remove this for this PR?

libraries/chain/finality_core.cpp
```diff
@@ -0,0 +1,302 @@
+#include <eosio/chain/finality_core.hpp>
+#include <eosio/chain/block_header.hpp>
+
+#warning Remove undef NDEBUG for assert before RC
+//Undefine NDEBUG to enable assertions in CICD.
+#undef NDEBUG
+#include <cassert>
```
heifner, line 7: I would think this would need to be the first file included as the files above this could have already included `<cassert>`.

libraries/chain/controller.cpp
```diff
@@ -666,22 +666,29 @@ struct building_block {
...
                            } else {
-                              qc_data = qc_data_t{ {},  qc_claim };
+                              qc_data = qc_data_t{ {},  claim };
                            }
                            break;
                         }
                      }
+
+                     if (!qc) {
+                        dlog("IF genesis Block");
```
heifner, line 687: I think either remove this debug log or make it a `ilog` with more useful information.

libraries/chain/controller.cpp
```diff
@@ -666,22 +666,29 @@ struct building_block {
...
+                        qc_data = qc_data_t{ {}, bb.parent.core.latest_qc_claim() };
```
heifner, line 688: Could this not be stored on the IF genesis block so that it can be found above?

libraries/chain/controller.cpp
```diff
@@ -691,10 +698,14 @@ struct building_block {
                   .new_protocol_feature_activations = new_protocol_feature_activations()
                };
 
+               // get current block reference
+               block_ref current_block {parent_id(), timestamp()};
```
heifner, line 702: This doesn't look correct. How can current block have parent id and current timestamp?

heifner (changes requested):

libraries/chain/block_header_state.cpp
```diff
@@ -213,11 +151,16 @@ block_header_state block_header_state::next(const signed_block_header& h, const
       .new_protocol_feature_activations = std::move(new_protocol_feature_activations)
    };
 
+   block_ref current_block{
+      .block_id  = block_id,
+      .timestamp = h.timestamp
+   };
```
heifner, line 157: This is previous block ref, not current block ref.

libraries/chain/finality_core.cpp
linh2931: Thanks. You are correct. Will change.

libraries/chain/controller.cpp
linh2931: I will remove the dlog but add a comment.

libraries/chain/controller.cpp
linh2931: This happens when current block (bb.parent) is the IF genesis block. It does not have a most recent ancestor which has a QC. We need to construct a default QC claim.

libraries/chain/controller.cpp
linh2931: Thank you Kevin for catching this! I have fixed it and also renamed it to parent_block to reduce confusion.

libraries/chain/finality_core.cpp
linh2931: Moved to the very top.

libraries/chain/controller.cpp
```diff
@@ -691,10 +700,19 @@ struct building_block {
                   .new_protocol_feature_activations = new_protocol_feature_activations()
                };
 
+               // Get parent block reference.
+               auto parent_timestamp = timestamp();
+               parent_timestamp.slot--;
```
heifner, line 705: The parent timestamp is not necessarily the previous time slot. Instead I think this should be `chain_head->timestamp()`

linh2931: Thanks. I should have used `bb.parent.timestamp()`; it was too late last night to go through the code.

libraries/chain/block_header_state.cpp
heifner: It is current of this, so I think it is fine as is.

heifner (approved):

greg7mdp (changes requested):

greg7mdp:
Shouldn't the `finality_core` files be in the `hotstuff` directories?

libraries/chain/include/eosio/chain/finality_core.hpp
```diff
@@ -0,0 +1,140 @@
...
+   finality_core next(const block_ref& current_block, const qc_claim& most_recent_ancestor_with_qc) const;
+};
+
+} /// eosio::chain
+
+FC_REFLECT( eosio::chain::block_ref, (block_id)(timestamp) )
+FC_REFLECT( eosio::chain::qc_link, (source_block_num)(target_block_num)(is_link_strong) )
+FC_REFLECT( eosio::chain::qc_claim, (block_num)(is_strong_qc) )
+FC_REFLECT( eosio::chain::finality_core,
+      (links)(refs)(final_on_strong_qc_block_num))
```
greg7mdp, line 140: Why a line break here?

libraries/chain/block_header_state.cpp
```diff
@@ -213,11 +151,16 @@ block_header_state block_header_state::next(const signed_block_header& h, const
...
 
+   block_ref current_block{
+      .block_id  = block_id,
+      .timestamp = timestamp()
+   };
+
    block_header_state_input bhs_input{
       bb_input,      h.transaction_mroot, h.action_mroot, if_ext.new_proposer_policy, if_ext.new_finalizer_policy,
-      if_ext.qc_claim };
+      current_block, if_ext.new_qc_claim };
```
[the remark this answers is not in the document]
linh2931: The two timestamps are different. For finality_core's next function, it requires the timestamp of the block where the next core is constructed from, ie, the parent's timestamp.

linh2931:
> Shouldn't the `finality_core` files be in the `hotstuff` directories?

I think hotstuff directory will eventually go away and `finality_core` is mostly for block_header_state.

libraries/chain/block_header_state_legacy.cpp
```diff
@@ -210,9 +210,8 @@ namespace eosio::chain {
       if (new_finalizer_policy) {
          new_finalizer_policy->generation = 1;
          // set current block_num as qc_claim.last_qc_block_num in the IF extension
-         qc_claim_t initial_if_claim { .last_qc_block_num = block_num,
-                                       .last_qc_block_timestamp = timestamp,
-                                       .is_last_qc_strong = false };
+         qc_claim initial_if_claim { .block_num = block_num,
```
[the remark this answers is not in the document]
linh2931: Was trying to be consistent with other types in finality_core.hpp which do not have `_t`. Was struggling with naming the qc claim in the `instant_finality_extension`. Do you and @heifner have any better name for that field?

greg7mdp:
Was file `block_header_state_tests.cpp` deleted and not replaced?

greg7mdp:
> I think hotstuff directory will eventually go away.

Maybe so, but not having it in `hotstuff` is inconsistent for now I think.

libraries/chain/block_header_state_legacy.cpp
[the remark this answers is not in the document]
greg7mdp: It would be better to add `_t`  to the types in `finality_core.hpp`  rather than struggling to find variable or members name when `qc_claim` is ideal.

libraries/chain/block_header_state.cpp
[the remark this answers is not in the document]
greg7mdp: So when we call finality_core's next function, shouldn't the timetstamp always be the timestamp of the block where the next core is constructed from? It seems that there is something wrong in the second link I provided.

libraries/chain/finality_core.cpp
```diff
@@ -0,0 +1,338 @@
...
+ */
+const block_ref& finality_core::get_block_reference(block_num_type block_num) const
+{
+   assert(last_final_block_num() <= block_num); // Satisfied by precondition.
+   assert(block_num < current_block_num()); // Satisfied by precondition.
+
+   // If refs.empty() == true, then by invariant 3, current_block_num() == last_final_block_num(), 
+   // and therefore it is impossible to satisfy the precondition. So going forward, it is safe to assume refs.empty() == false.
+
+   const size_t ref_index = block_num - last_final_block_num();
```
greg7mdp, line 94: While it works, it is a little odd to use links here. Since we know the refs are monotonically increasing block numbers, I think the following is more natural:
```suggestion
   const size_t ref_index = block_num - refs.front().block_num();
```

libraries/chain/block_header_state_legacy.cpp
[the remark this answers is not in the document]
greg7mdp: As a note, `finality_core.hpp` defines `block_num_type` and `block_time_type`, so it is altready inconsistent with the naming convention for types.
I propose using `_t` for all types, like the STL does for example in `uint32_t` or `std::invoke_result_t,`.

libraries/chain/finality_core.cpp
```diff
@@ -0,0 +1,338 @@
...
+
+/**
+ *  @pre c.latest_qc_claim().block_num <= most_recent_ancestor_with_qc.block_num <= c.current_block_num()
+ * 
+ *  @post std::get<0>(returned_value) <= std::get<1>(returned_value) <= std::get<2>(returned_value) <= most_recent_ancestor_with_qc.block_num
+ *  @post c.last_final_block_num() <= std::get<0>(returned_value)
+ *  @post c.links.front().source_block_num <= std::get<1>(returned_value)
+ *  @post c.final_on_strong_qc_block_num <= std::get<2>(returned_value)
+ */
+std::tuple<block_num_type, block_num_type, block_num_type> get_new_block_numbers(const finality_core& c, const qc_claim& most_recent_ancestor_with_qc)
```
greg7mdp, line 136: Let's make `get_new_block_numbers` a member function of `finality_core`, rather than a free function taking a `const finality_core&`.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -25,27 +26,18 @@ struct block_header_state_input : public building_block_input {
    digest_type                       action_mroot;         // Compute root from  building_block::action_receipt_digests
    std::shared_ptr<proposer_policy>  new_proposer_policy;  // Comes from building_block::new_proposer_policy
    std::optional<finalizer_policy>   new_finalizer_policy; // Comes from building_block::new_finalizer_policy
-   std::optional<qc_claim_t>         qc_claim;             // Comes from traversing branch from parent and calling get_best_qc()
+   block_ref                         current_block;
```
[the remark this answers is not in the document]
linh2931: Made the change. Thanks.

linh2931:
> Was file `block_header_state_tests.cpp` deleted and not replaced?

It is deleted as the tests there were only for testing previous `block_header_state_core`'s `next`. New tests can be added later in a new `finality_core_tests.cpp`.

libraries/chain/include/eosio/chain/finality_core.hpp
linh2931: Will move them to a single line

libraries/chain/finality_core.cpp
```diff
@@ -0,0 +1,338 @@
...
+   // By the precondition, link0.source_block_num <= most_recent_ancestor_with_qc.block_num.
+   // If c.links.size() > 1, then by invariant 7 of core, link0.target_block_num <= link1.target_block_num.
+   // Otherwise if c.links.size() == 1, then link0 == link1 and so link0.target_block_num == link1.target_block_num.
+   // Therefore, c.final_on_strong_qc_block_num <= link1.target_block_num.
+
+   assert(c.final_on_strong_qc_block_num <= link1.target_block_num); // Satisfied by justification above.
+
+   // Finality does not advance if a better 3-chain is not found.
+   if (!link1.is_link_strong || (link1.target_block_num < c.links.front().source_block_num)) {
+      return {c.last_final_block_num(), c.links.front().source_block_num, link1.target_block_num};
```
greg7mdp, line 182: Why `link1.target_block_num` and not `c.final_on_strong_qc_block_num`?

libraries/chain/block_header_state_legacy.cpp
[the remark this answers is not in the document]
linh2931: I'll change to qc_claim_t

greg7mdp:
> It is deleted as the tests there were only for testing previous `block_header_state_core`'s `next`. New tests can be added later in a new `finality_core_tests.cpp`.

In that case we need a new issue to make sure we don't forget to add these tests.

libraries/chain/block_header_state_legacy.cpp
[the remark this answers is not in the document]
linh2931: Done.

libraries/chain/include/eosio/chain/finality_core.hpp
linh2931: Fixed.

libraries/chain/block_header_state.cpp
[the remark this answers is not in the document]
linh2931: For the second link, we are building the new core for the signed block. The the current block is the "parent" in a sense.

libraries/chain/finality_core.cpp
greg7mdp: OK, Areg answered this.

libraries/chain/finality_core.cpp
greg7mdp: OK. maybe later!

libraries/chain/finality_core.cpp
linh2931: oh we cannot do that as refs can be empty while `last_final_block_num()` always returns a valid value.

unittests/block_header_tests.cpp
```diff
@@ -19,21 +18,21 @@ BOOST_AUTO_TEST_CASE(instant_finality_extension_with_empty_values_test)
 {
    block_header       header;
    constexpr uint32_t last_qc_block_num {0};
-   constexpr bool     is_last_qc_strong {false};
+   constexpr bool     is_strong_qc {false};
```
greg7mdp, line 21: Why this name change? I think the previous one matched `last_qc_block_num` better.

linh2931:
> In that case we need a new issue to make sure we don't forget to add these tests.

https://github.com/AntelopeIO/leap/issues/2259 was created for that.

libraries/chain/block_header_state.cpp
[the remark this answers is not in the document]
greg7mdp: OK, my bad, looks fine now!

libraries/chain/finality_core.cpp
greg7mdp: line above says: 
` So going forward, it is safe to assume refs.empty() == false.`

unittests/block_header_tests.cpp
linh2931: Done

linh2931:
> Maybe so, but not having it in `hotstuff` is inconsistent for now I think.

I think for new files we can start to stay away from that directory.

greg7mdp (approved):

Issue #2061, opened 2024-01-09
IF: BLS key/signature encoding


heifner:
Currently BLS public key and signatures are encoded as base64. This makes it difficult to copy/paste from terminals. Does this provide enough heart-ache to use a different encoding?

base64url would provide an easier to use format. We would want to change our base64url encoder not use the optional trailing `=` padding to make it easier to use. Specifically with `--signature-provider` which uses `=KEY:` as a separator.

PR #2255, opened 2024-02-21
IF: change BLS key/signature encoding to base64url for easier copy/paste

linh2931:
* changed to use @heifner 's base64 library
* changed BLS key/signature encoding to base64url for easier copy/paste, with a single copy/paste instead of multiple ones using original base64 endcoding.
* updated tests and test configs to use the new encoding. While updating the keys and signatures, I felt the pains @heifner went through preparing keys/signatures for tests and test nodes.

Reference Contracts will be updated in a separate PR in reference contract repo.

Resolved https://github.com/AntelopeIO/leap/issues/2061


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: For ease of use encode BLS key/signature in base64url.
Note:end

libraries/libfc/test/test_bls.cpp
```diff
@@ -294,54 +294,54 @@ BOOST_AUTO_TEST_CASE(bls_regenerate_check) try {
 BOOST_AUTO_TEST_CASE(bls_prefix_encoding_check) try {
 
   //test no_throw for correctly encoded keys
-  BOOST_CHECK_NO_THROW(bls_private_key("PVT_BLS_O+gS5yNxVtSwL0/Uhl1IVqu/Y7Gq4qdrtB66EIb192ENfK8D"));
-  BOOST_CHECK_NO_THROW(bls_public_key("PUB_BLS_dEvut0ydHevDGP6Ef3O4Iq6QXf9jUcMUT1nCJRX+JRYlFYrO/qKt/x439vUJ2DkZ32Od6AdJZ+S9dWRE9Sy+7Q6bNjpoIOP0cWzkKC1DqmhfE3paW+KThA3noLkV8SsILcfxpQ=="));
-  BOOST_CHECK_NO_THROW(bls_signature("SIG_BLS_prHR3PtNGJQLEhqcuHqj5Ty6FkNFo+ih32+ZHh6LH74+SKlTgq4PWtudoYt8heEZjPyNDRrqfRoYoTlAZ1mpW0QzgyGRXU+lfZ27M9Bg1mNS0MI6wWL4ZG9E8bchiMUVWijpX66sc11t60m/g8/vJIf1tIuFLhKCcX57OVCoXisciI7D21b3tKjb7VAlc2oNEoJx17XOafIWvcH1YKAc2uv9T/ocAlE3VQNprXKuGaZYA9Q5yzaOhVgGYxrjv/wNv0DlzA=="));
+  BOOST_CHECK_NO_THROW(bls_private_key("PVT_BLS_vh0bYgBLOLxs_h9zvYNtj20yj8UJxWeFFAtDUW2_pG44e5yc"));
+  BOOST_CHECK_NO_THROW(bls_public_key("PUB_BLS_82P3oM1u0IEv64u9i4vSzvg1-QDl4Fb2n50Mp8Sk7Fr1Tz0MJypzL39nSd5VPFgFC9WqrjopRbBm1Pf0RkP018fo1k2rXaJY7Wtzd9RKlE8PoQ6XhDm4PyZlIupQg_gOuiMhcg=="));
```
heifner, line 298: One reason for using base64url was to avoid the `=` padding characters. However, our implementation of base64url does still add the padding characters. Worse yet, the implementation it is based on https://github.com/ReneNyffenegger/cpp-base64 latest code used `.` as a padding character for base64url which is clearly wrong. The spec https://datatracker.ietf.org/doc/html/rfc4648#section-5 actually says that `.` was not used because `The remaining unreserved URI character is ".",
   but some file system environments do not permit multiple "." in a
   filename, thus making the "." character unattractive as well.`

I would say to use my version https://github.com/heifner/base64 but it also currently has the incorrect implementation for base64url since it is based off of ReneNyffenegger version. I will fix that now that I realize it is a broken.

linh2931: Thanks. @arhag yesterday said we don't need to add padding. We can talk it at today's meeting.

spoonincode: Our implementation of base64url encode adds the padding but that can be changed not to without affecting consensus. Our implementation of base64url decode, which is consensus, does not care about padding, see for example this test that deliberately removes the padding,
https://github.com/AntelopeIO/leap/blob/6d188675ac97ef00aa48fe9e64e3e73a1ada09c4/libraries/libfc/test/crypto/test_webauthn.cpp#L210-L216

So if you just change base64url encode to not add padding it seems like it'd work.

linh2931: Thanks. We only change base64url.

spoonincode: That new implementation breaks consensus due to how it changes base64url decode.

The existing implementation probably just needed something simple like
```c++
while(ret.ends_with('='))
   ret.pop_back();
```

linh2931: Should we go back to original implementation. One thing about the new implementation is the same decode works on both url-safe and url-non-safe encoding; it was at first a surprise but correct.

libraries/libfc/test/crypto/test_webauthn.cpp
```diff
@@ -183,38 +185,55 @@ BOOST_AUTO_TEST_CASE(challenge_wrong_base64_chars) try {
    std::vector<uint8_t> auth_data(37);
    memcpy(auth_data.data(), origin_hash.data(), sizeof(origin_hash));
 
-   BOOST_CHECK_EXCEPTION(make_webauthn_sig(priv, auth_data, json).recover(d, true), fc::exception, [](const fc::exception& e) {
-      return e.to_detail_string().find("encountered non-base64 character") != std::string::npos;
-   });
+   BOOST_CHECK_NO_THROW(make_webauthn_sig(priv, auth_data, json).recover(d, true));
```
spoonincode, line 188: This change really looks like a consensus change: what threw an exception before now doesn't

linh2931: Thanks Matt. We can add a check in the base64url_decode wrapper in the new implementation for that but to be safe we should just switch back to the existing implementation. @heifner ?

linh2931: My latest commit https://github.com/AntelopeIO/leap/pull/2255/commits/00f2ce0066aafd60d99efacaef9f10444860a7d1 tries to behave in the same way as the existing base64url decode; the existing tests are only modified to accommodate the fact that trailing separators are no longer added.

libraries/libfc/test/crypto/test_webauthn.cpp
```diff
@@ -188,13 +188,12 @@ BOOST_AUTO_TEST_CASE(challenge_wrong_base64_chars) try {
...
 
-//valid signature but replace the padding '=' with '.'
+//valid signature but replace the last character '=' with '.'
 BOOST_AUTO_TEST_CASE(challenge_base64_dot_padding) try {
    webauthn::public_key wa_pub(pub.serialize(), webauthn::public_key::user_presence_t::USER_PRESENCE_NONE, "fctesting.invalid");
    std::string b64 = fc::base64url_encode(d.data(), d.data_size());
    char& end = b64.back();
 
-   BOOST_REQUIRE(end == '=');
    end = '.';
```
spoonincode, line 197: This test was originally changing the padding from a valid character to an invalid character. Now it's changing a non-padding character from a valid character to an invalid character. I think maybe this should just append a `.` (or really any other invalid character) to keep the spirit of the original test; we already have `challenge_non_base64` to cover validity of non-padded characters.

libraries/libfc/test/crypto/test_webauthn.cpp
```diff
@@ -207,14 +206,11 @@ BOOST_AUTO_TEST_CASE(challenge_base64_dot_padding) try {
    });
 } FC_LOG_AND_RETHROW();
 
-//valid signature but remove padding
+//valid signature
```
spoonincode, line 209: This test is now just repeating what the `good` test does, but more importantly I believe it means we no longer have a test with valid `=` padding. I think instead this test should be modified to append valid padding to ensure that base64urldecode ignores padding as required.

spoonincode (commented):
I think it'd be best to tweak these tests a wee bit

spoonincode:
webauthn test changes LGTM

heifner (approved):

greg7mdp (approved):

PR #2262, opened 2024-02-26
upgrade libsecp256k1 to 0.4.1

spoonincode:
Upgrade libsecp256k1 from 0.4.0 to 0.4.1. Release notes state,
> Optional handwritten x86_64 assembly for field operations was removed because modern C compilers are able to output more efficient assembly. This change results in a significant speedup of some library functions when handwritten x86_64 assembly is enabled (`--with-asm=x86_64` in GNU Autotools, `-DSECP256K1_ASM=x86_64` in CMake), which is the default on x86_64. Benchmarks with GCC 10.5.0 show a 10% speedup for `secp256k1_ecdsa_verify` and `secp256k1_schnorrsig_verify`.

Indeed, the performance for a k1 key recovery improves from about 33µs to a little under 30µs on my hardware. A nice free performance boost.


heifner (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: PERFORMANCE
category: INTERNALS
summary: Upgrade `libsecp256k1` library for faster operations on keys. 
Note:end

Issue #2141, opened 2024-01-25
IF: replay from before instant-finality transition


heifner:
When replaying: `nodeos` option `--replay-blockchain` we clear state database and replay all blocks including reversible blocks in the fork database. The fork database is opened and consumed before replay. This causes an issue as the fork database needed during replay (until instant-finality transition) requires the legacy fork database. After instant-finality transition, it requires the new fork database. 

On replay, need to delay consuming the fork database until after all blocks are replayed so the fork database can be opened and consumed in the correct mode with the expected data (`block_state_legacy` vs `block_state`).

Currently `nodeos_irreversible_mode_test.py` fails due to this with instant-finality enabled.

heifner:
Immediate fix for `hotstuff_integration` branch. Update replay logic to not use `fork_database::chain_head` and instead pass current and previous to `replay_push_block`. `replay_push_block` can return the new pushed block state. Likely better to create a `replay_push_irreversible_block` (returns newly created block state) and `replay_push_reveserable_block` (just updates chain_head).

heifner:
Long term fix: https://github.com/AntelopeIO/leap/issues/2143

greg7mdp:
Uncomment test `terminate-scenarios-if-test-hard_replay` in `tests/CMakeLists.txt` when  it works.

heifner:
Resolved by #2287

PR #2265, opened 2024-02-27
IF: Update fork database fork-choice rule - complex

### LINKED ISSUE #2125: IF: Update fork-choice rule for fork database

heifner:
Update `fork_database` for-choice rule.
3. block timestamp

Potentially switches forks before starting to produce a block. Does not interrupt producing blocks once block production has started.

Includes some misc performance improvements to `fork_database`.

Resolves #2125


libraries/chain/block_state.cpp
```diff
@@ -92,20 +99,16 @@ std::pair<vote_status, std::optional<uint32_t>> block_state::aggregate_vote(cons
...
                                  finalizers[index].weight);
-
-      std::optional<uint32_t> new_lib{};
-      if (status == vote_status::success && strong)
-         new_lib = core.final_on_strong_qc_block_num;
-      return {status, new_lib};
    } else {
       wlog( "finalizer_key (${k}) in vote is not in finalizer policy", ("k", vote.finalizer_key) );
-      return {vote_status::unknown_public_key, {}};
+      return {vote_status::unknown_public_key, pending_quorum_certificate::state_t::unrestricted};
```
greg7mdp, line 111: why not?
```suggestion
      return {vote_status::unknown_public_key, pending_qc.state();
```

Seems more correct for the function signature.

libraries/chain/controller.cpp
```diff
@@ -1440,7 +1440,7 @@ struct controller_impl {
          initialize_blockchain_state(genesis); // sets head to genesis state
 
          if( !forkdb.head() ) {
-            forkdb.reset( *forkdb.chain_head );
+            forkdb.reset_root( *forkdb.chain_head );
```
linh2931, line 1443: Why not call `fork_db_reset_root_to_head()` here?

We have `forkdb.reset_root( *head )`, and `forkdb.reset_root( *forkdb.chain_head );` which is equivalent to `fork_db_reset_root_to_head()`. Should `fork_db_reset_root_to_head()` be renamed to a more clearer name?

libraries/chain/controller.cpp
```diff
@@ -2780,10 +2780,8 @@ struct controller_impl {
             const auto& bsp = std::get<std::decay_t<decltype(forkdb.chain_head)>>(cb.bsp);
 
             if( s == controller::block_status::incomplete ) {
-               forkdb.add( bsp );
-               forkdb.mark_valid( bsp );
+               forkdb.add( bsp, true, false );
```
linh2931, line 2783: Can a comment added to indicate what true and false mean?

libraries/chain/controller.cpp
```diff
@@ -2780,10 +2780,8 @@ struct controller_impl {
...
                emit( accepted_block_header, std::tie(bsp->block, bsp->id()) );
-               EOS_ASSERT( bsp == forkdb.head(), fork_database_exception, "committed block did not become the new head in fork database");
```
linh2931, line 2786: Why is this assert removed?

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Update for database for-choice rules with the following computed fields, listed in order of importance, next_core_metadata.last_final_block_num, next_core_metadata.latest_qc_claim_block_num, block timestamp, block ID
Note:end

libraries/chain/controller.cpp
heifner: Because it is no longer always true. A better head can be realized.

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -93,14 +93,18 @@ namespace eosio::chain {
 
       // thread safe
       bool is_quorum_met() const;
+      static bool is_quorum_met(state_t s) {
+         return s == state_t::weak_achieved || s == state_t::weak_final || s == state_t::strong;
```
linh2931, line 97: In normal operations, `s == state_t::strong` should be more common. Can it be placed at the beginning?

libraries/chain/controller.cpp
heifner: Renamed to `fork_db_reset_root_to_chain_head`. This location can't use that because it is inside a lambda that is using the specific `forkdb` type. This is being refactored in the PR for #2141 as `chain_head` is moving from `fork_database` back to `controller_impl`.

libraries/chain/fork_database.cpp
```diff
@@ -21,79 +20,154 @@ namespace eosio::chain {
...
+         return bs.updated_core.latest_qc_claim_block_num;
+      }
+      static bool qc_claim_update_needed(block_state& bs, const qc_claim_t& most_recent_ancestor_with_qc) {
+         return bs.most_recent_ancestor_with_qc < most_recent_ancestor_with_qc;
+      }
+      static bool qc_claim_update_needed(block_state& bs, const block_state& prev) {
+         return bs.most_recent_ancestor_with_qc < prev.most_recent_ancestor_with_qc;
+      }
+      static void update_best_qc(block_state& bs, const qc_claim_t& most_recent_ancestor_with_qc) {
+         assert(bs.most_recent_ancestor_with_qc < most_recent_ancestor_with_qc);
```
linh2931, line 47: Is this assert always true when this method is called? You can also just check if `<` then return.

libraries/chain/fork_database.cpp
```diff
@@ -21,79 +20,154 @@ namespace eosio::chain {
...
+         bs.updated_core = bs.core.next_metadata(most_recent_ancestor_with_qc);
+         bs.most_recent_ancestor_with_qc = most_recent_ancestor_with_qc;
+      }
+      static void update_best_qc(block_state& bs, const block_state& prev) {
+         assert(bs.most_recent_ancestor_with_qc < prev.most_recent_ancestor_with_qc);
+         update_best_qc(bs, prev.most_recent_ancestor_with_qc);
+      }
+
+      // thread safe
+      static uint32_t block_height(const block_state& bs) {
```
linh2931, line 57: `block_height` might imply it is block_num. Change to another name?

libraries/chain/fork_database.cpp
```diff
@@ -571,28 +679,75 @@ namespace eosio::chain {
...
+      auto& previdx = index.template get<by_prev>();
+      for( uint32_t i = 0; i < descendants.size(); ++i ) {
+         auto previtr = previdx.lower_bound( descendants[i] );
+         while( previtr != previdx.end() && (*previtr)->previous() == descendants[i] ) {
+            if (BSAccessor::qc_claim_update_needed(**previtr, most_recent_ancestor_with_qc)) {
+               previdx.modify( previtr, [&](auto& i) {
+                  BSAccessor::update_best_qc(*i, most_recent_ancestor_with_qc);
+               });
+            } else {
+               break;
```
linh2931, line 725: Is this `break` too early? Should next two statements be always executed?

libraries/chain/fork_database.cpp
```diff
@@ -571,28 +679,75 @@ namespace eosio::chain {
...
+            BSAccessor::update_best_qc(*i, most_recent_ancestor_with_qc);
+         } );
+      }
+
+      // process descendants
+      vector<block_id_type> descendants;
+      descendants.reserve(index.size());
+      descendants.push_back(id);
+      auto& previdx = index.template get<by_prev>();
+      for( uint32_t i = 0; i < descendants.size(); ++i ) {
```
linh2931, line 717: The end condition `i < descendants.size()` can change. Not sure which is the best way to do this kind of loop.

libraries/chain/fork_database.cpp
linh2931: My bad, if a block does not need to be updated, neither do its descendants.

libraries/chain/fork_database.cpp
heifner: Yes, the expectation is that `qc_claim_update_needed` is called first. This is internal to `fork_database.cpp`, so used an `assert` instead of an if-check.

libraries/chain/fork_database.cpp
heifner: I don't think so. If not applicable to this (previtr) then not going to be applicable to its descendants, so no need to add it to the list to be processed.

libraries/chain/fork_database.cpp
heifner: Granted it is a bit complex. It is based on the logic found in the existing `remove_impl`.

libraries/chain/controller.cpp
```diff
@@ -2780,10 +2780,10 @@ struct controller_impl {
             const auto& bsp = std::get<std::decay_t<decltype(forkdb.chain_head)>>(cb.bsp);
 
             if( s == controller::block_status::incomplete ) {
-               forkdb.add( bsp );
-               forkdb.mark_valid( bsp );
+               const bool mark_valid = true;
+               const bool ignore_duplicate = false;
+               forkdb.add( bsp, mark_valid, ignore_duplicate );
```
greg7mdp, line 2785: I'd like it even better if fork_database exported the types:

```
enum class mark_valid{ off, on };
enum class ignore_duplicate { off, on };
```

The signature of add was:
`add(const BSP& next_block, mark_valid do_mark_valid, ignore_duplicate do_ignore_duplicate);`

and it was called as:

` forkdb.add( bsp, mark_valid::on, ignore_duplicate::off);`

Not only it is clearer, but you cannot pass the parameters in the wrong order by mistake.

heifner: Much better. Seems like our canonical use is `{ no, yes }`

heifner: Added:
```cpp
   enum class mark_valid_t { no, yes };
   enum class ignore_duplicate_t { no, yes };
```

greg7mdp (changes requested):

libraries/chain/include/eosio/chain/block_state.hpp
```diff
@@ -20,46 +20,62 @@ inline weak_digest_t create_weak_digest(const digest_type& digest) {
...
-   bool                       validated = false;     // We have executed the block's trxs and verified that action merkle root (block id) matches.
    digest_type                strong_digest;         // finalizer_digest (strong, cached so we can quickly validate votes)
    weak_digest_t              weak_digest;           // finalizer_digest (weak, cached so we can quickly validate votes)
    pending_quorum_certificate pending_qc;            // where we accumulate votes we receive
    std::optional<valid_quorum_certificate> valid_qc; // best qc received from the network inside block extension
 
+   // ------ updated for votes, used for fork_db ordering ------------------------------
+private:
+   bool                       validated = false;            // We have executed the block's trxs and verified that action merkle root (block id) matches.
+   qc_claim_t                 most_recent_ancestor_with_qc; // only modify/access while holding forkdb lock
```
greg7mdp, line 36: I think the name `best_qc_claim` would be better. I was not sure we necessarily want to update to a higher block_num with a weak qc if the one we have is is strong.

libraries/chain/fork_database.cpp
```diff
@@ -321,99 +395,133 @@ namespace eosio::chain {
...
             const auto& exts = n->header_exts;
 
-            if (exts.count(protocol_feature_activation::extension_id()) > 0) {
-               const auto& pfa = exts.lower_bound(protocol_feature_activation::extension_id())->second;
-               const auto& new_protocol_features = std::get<protocol_feature_activation>(pfa).protocol_features;
-               validator(n->timestamp(),
-                         static_cast<bs*>(prev_bh.get())->get_activated_protocol_features()->protocol_features,
-                         new_protocol_features);
+            if( exts.count(protocol_feature_activation::extension_id()) > 0 ) {
+               const auto& new_protocol_features = std::get<protocol_feature_activation>(exts.lower_bound(protocol_feature_activation::extension_id())->second).protocol_features;
```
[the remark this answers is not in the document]
heifner: I was just reverting to 5.0 as there is no reason for any change here.

libraries/chain/controller.cpp
```diff
@@ -3112,10 +3112,21 @@ struct controller_impl {
...
+          if (bsp) {
+             auto [status, state] = bsp->aggregate_vote(vote);
+             std::optional<uint32_t> new_lib{};
+             if (status == vote_status::success && pending_quorum_certificate::is_quorum_met(state)) {
+                if (state == pending_quorum_certificate::state_t::strong) {
+                   new_lib = bsp->core.final_on_strong_qc_block_num;
+                   forkdb.update_best_qc(bsp->id(), {.block_num = bsp->block_num(), .is_strong_qc = true});
+                } else {
+                   forkdb.update_best_qc(bsp->id(), {.block_num = bsp->block_num(), .is_strong_qc = false});
+                }
```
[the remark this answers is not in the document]
heifner: Do you propose having `pending_quorum_certificate::add_vote` return before/after state or maybe return a bool if state has changed?

libraries/chain/include/eosio/chain/finality_core.hpp
```diff
@@ -33,9 +33,9 @@ struct qc_claim_t
 
 struct core_metadata
 {
-   block_num_type  last_final_block_num;
-   block_num_type  final_on_strong_qc_block_num;
-   block_num_type  latest_qc_claim_block_num;
+   block_num_type  last_final_block_num {0};
+   block_num_type  final_on_strong_qc_block_num {0};
+   block_num_type  latest_qc_claim_block_num {0};
```
greg7mdp, line 38: I believe this is unneeded, int members are default initialized to 0. 
image

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: Sure, any of these two is fine. We don't really care about the result of adding the vote, just whether we have achieved a new state I think.

libraries/chain/fork_database.cpp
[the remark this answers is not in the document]
greg7mdp: haha the reason was just to avoid the excessively long line. But no big deal.

libraries/chain/include/eosio/chain/finality_core.hpp
linh2931: I was bitten by not initializing an integer member not long ago during local development. Plus explicit initialization marks intension clearer?

libraries/chain/fork_database.cpp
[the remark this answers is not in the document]
heifner: I can switch it back. I had made a change that I needed to revert, so reverted by copying over the 5.0 impl. Didn't realize there was a diff already there.

libraries/chain/include/eosio/chain/block_state.hpp
heifner: I like `best_qc_claim` as well. I was going with what @arhag had in #2125. If no objection from @arhag I'll change it to `best_qc_claim`.

libraries/chain/fork_database.cpp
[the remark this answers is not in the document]
greg7mdp: Thanks!

libraries/chain/controller.cpp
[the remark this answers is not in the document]
linh2931: net_plugin needs the voting status (success or not) to decide what to do to the connection.

libraries/chain/fork_database.cpp
```diff
@@ -571,28 +679,75 @@ namespace eosio::chain {
...
+                  BSAccessor::update_best_qc(*i, most_recent_ancestor_with_qc);
+               });
+            } else {
+               break;
+            }
+
+            descendants.emplace_back( (*previtr)->id() );
+            ++previtr;
+         }
+      }
```
greg7mdp, line 732: I assume this is to avoid a recursive implementation, but imo the added complexity and difficulty to ensure correctness makes me think a recursive solution would be better.

heifner: This implementation was chosen to match existing implementation in `remove_impl`. We know `remove_impl` works.

libraries/chain/include/eosio/chain/finality_core.hpp
heifner: I like the safety of explicit initialization. Shouldn't make any difference as our current only creation is:
```cpp
return core_metadata {
      .last_final_block_num = new_last_final_block_num,
      .final_on_strong_qc_block_num = new_final_on_strong_qc_block_num,
      .latest_qc_claim_block_num = most_recent_ancestor_with_qc.block_num,
   };
```

greg7mdp: OK, but I don't see how it is safer to re-specify the implicit initialization.

heifner: Not sure why you are claiming implicit initialization for ints.
https://godbolt.org/z/4Gd986Ea9

greg7mdp: I wonder if that's a gcc bug? weirdly, if you uncomment line 35 [here](https://godbolt.org/z/3vEdMc955) these are zero-initialized. Other compilers as well. 
In this particular case I am still not convinced the explicit initializers are needed, but I guess it is fair to argue the opposite!

heifner: Same for clang: https://godbolt.org/z/EWsGf15K6
Uncomment and run multiple times. It is indeterminate, so run it a few times.

heifner: https://stackoverflow.com/questions/706030/are-uninitialized-struct-members-always-set-to-zero

libraries/chain/fork_database.cpp
```diff
@@ -571,28 +680,75 @@ namespace eosio::chain {
...
+      descendants.push_back(id);
+      auto& previdx = index.template get<by_prev>();
+      for( uint32_t i = 0; i < descendants.size(); ++i ) {
+         auto previtr = previdx.lower_bound( descendants[i] );
+         while( previtr != previdx.end() && (*previtr)->previous() == descendants[i] ) {
+            if (BSAccessor::qc_claim_update_needed(**previtr, best_qc_claim)) {
+               previdx.modify( previtr, [&](auto& i) {
+                  BSAccessor::update_best_qc(*i, best_qc_claim);
+               });
+            } else {
```
greg7mdp, line 725: Doesn't this while loop iterate over all the immediate children of `descendants[i]`?

If that's the case, we don't want to `break;` and not visit the siblings. We want to not add this child to the `descendants` vector, so we won't consider its children.

greg7mdp (changes requested):

libraries/chain/fork_database.cpp
heifner: I think you are right. Good catch. Seems I should add a unittest for this.

heifner: Added a test that failed before fixing the logic.

unittests/fork_db_tests.cpp
```diff
@@ -0,0 +1,131 @@
...
+};
+
+} // namespace eosio::chain
+
+using namespace eosio::chain;
+
+BOOST_AUTO_TEST_SUITE(fork_database_tests)
+
+BOOST_AUTO_TEST_CASE(update_best_qc) try {
+
```
linh2931, line 59: Would you explain the high level setup of those blocks (bsps)?

libraries/chain/fork_database.cpp
linh2931: Glad it is fixed. I should have drew some links yesterday.

unittests/fork_db_tests.cpp
heifner: Added some comments

linh2931 (approved):

heifner:
This was not merged. Instead #2273 was merged. It was marked as merged because #2273 built off of it.

PR #2272, opened 2024-02-29
IF: Add more extensive tests for finalizer voting behavior

### LINKED ISSUE #2233: IF: Add more extensive tests for finalizer voting behavior
Depends on #2070.

- test decide_vote, validate monotony/liveness/safety checks.
- further safety file checks (verify it is updated after each strong vote but not a weak vote, ...)
- test `get_qc_chain`, `extends`.
- Verify initialization of new finalizer safety information is correct.

greg7mdp:
Resolves #2233.

This PR adds some unit test for finalizer voting that validate correct checks for monotony, liveness and safety.

It also adds to stream output functions for some types used by the finalizer, which allows to use these types in `BOOST_ASSERT_EQUAL` macros which provide better output when they fail than `BOOST_ASSERT`.

It also adds spaceship operator comparisons to some fc types (including `sha256`) and `block_timestamp`, which allows to define default spaceship comparison operators in new types containing these (as is the case for `block_ref`).


ericpassmore:
Note:start
group: IF
category: TEST
summary: Adds unit test for finalizer voting that validate correct checks for monotony, liveness and safety. Adds support for adds spaceship operator comparisons to some fc types, and support for `BOOST_ASSERT_EQUAL` to some types used by finalizer.
Note:end

libraries/chain/include/eosio/chain/block_timestamp.hpp
```diff
@@ -17,7 +17,10 @@ namespace eosio { namespace chain {
    template<uint16_t IntervalMs, uint64_t EpochMs>
    class block_timestamp {
       public:
-         explicit block_timestamp( uint32_t s=0 ) :slot(s){}
+
+         block_timestamp() : slot(0) {}
+
+         explicit block_timestamp( uint32_t s ) :slot(s){}
```
heifner, line 23: Why this change? What difference in behavior does this provide?

unittests/finalizer_tests.cpp
```diff
@@ -154,8 +168,308 @@ BOOST_AUTO_TEST_CASE( finalizer_safety_file_io ) try {
 
 } FC_LOG_AND_RETHROW()
```
heifner, line 170: Really seems like we are starting a whole diff section of tests. Maybe move to `finalizer_votes_tests.cpp`.

heifner (changes requested):

libraries/chain/include/eosio/chain/block_timestamp.hpp
greg7mdp: The default constructor is not `explicit` anymore, so structs containing `block_timestamp` can be constructed as aggregates.

unittests/finalizer_tests.cpp
greg7mdp: Will do!

libraries/chain/finality_core.cpp
```diff
@@ -78,6 +78,29 @@ qc_claim_t finality_core::latest_qc_claim() const
...
+}
+
+/**
+ *  @pre  all finality_core invariants
+ *  @post same
+ *  @returns boolean indicating whether `id` is an ancestor of this block
+ */
+bool finality_core::extends(const block_id_type& id) const {
+   uint32_t block_num = block_header::num_from_id(id);
+   if (block_num >= last_final_block_num() && block_num < current_block_num()) {
```
linh2931, line 97: When `block_num < last_final_block_num()`, it can be an ancestor. But it is too old and out of the links. Maybe add a comment.

unittests/finalizer_tests.cpp
linh2931, line 170: Do you want to add some tests for corrupted safety file?

libraries/chain/finality_core.cpp
greg7mdp: It doesn't matter that it is out of the link, it is a safety feature of the finalizer that we can only vote on proposals which are descendants of the one we are locked on.

unittests/finalizer_tests.cpp
```diff
@@ -1,3 +1,4 @@
+#include "eosio/chain/fork_database.hpp"
```
heifner, line 1: Use `<>`

greg7mdp: done

unittests/finalizer_tests.cpp
greg7mdp: done

heifner (approved):

unittests/finalizer_tests.cpp
greg7mdp: Yes, good suggestion, done.

linh2931 (approved):

Issue #2159, opened 2024-01-30
IF: Don't drop late blocks


heifner:
Currently, any blocks received while producing a block are dropped and not evaluated. IF includes QC info in block extensions. Do not drop blocks, but rather incorporate their QC info into the `fork_database`. Do not automatically abort block but incorporate the block into the `fork_database` for use in best fork evaluation. If LIB advances such that current block being produced is no longer built on LIB, abort the block immediate, re-evaluate best fork, and build block off the new best fork.

Related to #2125

heifner:
I believe `lib_advance_test` can be updated to expect no unlinkable blocks after this change.

Issue #2057, opened 2024-01-08
IF: Implement proper IF transition


arhag:
For some background on concepts related to IF see: https://github.com/AntelopeIO/leap/issues/2080.

Carry out proper (non-atomic) transition from pre-IF to post-IF that was avoided as part of https://github.com/AntelopeIO/leap/issues/2045.

After the IF hard fork, the old consensus algorithm rules are slightly modified to not allow `dpos_irreversible_block_num` to advance beyond an IF Genesis Block. This means the root of the legacy fork database can never be a descendant of the IF Genesis Block.

The root of the new fork database the first time it is used should always be an IF Genesis Block. 

Do not vote on (or collect votes on) Transition IF Blocks. So, the first time finality advances past the IF Genesis block (according to the new IF algorithm) it will also advance past all Transition IF Blocks and clear out the old fork database.

For a branch of Transition IF Blocks to exist in the new fork database, it must contain an IF Critical Block that makes the IF Genesis Block irreversible (according to the old consensus algorithm). Furthermore, the fork-choice rule in the new fork database guarantees that a Proper IF Block will not be built on any of the Transition Blocks in that branch other than the IF Critical Block. However, we should still validate that when a Proper IF Block builds on top of a Transition IF Block that the Transition Block is an IF Critical Block (this is determined by look at the `block_header_state_legacy` of the corresponding Transition Legacy Block in the legacy fork database).

A Proper IF Block must set the `schedule_version` in the block header to 2^31. If the `schedule_version` is less than 2^31 then it is not a Proper IF Block. 

As soon as there are blocks in the new fork database, the new fork database with its fork-choice rule should be used to determine the best head to build a new block on top of. All such blocks that are built would be Proper IF Blocks.

Transition Blocks should ignore any producer schedule changes.

Snapshots for Transition Blocks should include the data that would be included in the snapshot for both a Pre-IF Legacy Block and a Proper IF Block. This means including `block_header_state_legacy` but also including `block_header_state` plus the data in the `valid` structure of a `block_state`. When loading the snapshot for a Transition Block, both pieces of information should be used to construct a root for the legacy fork database and the new fork database.

Once the IF Genesis Block is identified (because it becomes irreversible according to the old algorithm), the currently uninitialized finalizer safety informations for any finalizers provided to nodeos should be initialized. Details for how this should be initialized are documented [here](https://hackmd.io/JKIz2TWNTq-xcWyNX4hRvw).

[...]

PR #2273, opened 2024-02-29
IF: Update fork database fork-choice rule

### LINKED ISSUE #2125: IF: Update fork-choice rule for fork database

heifner:
After this PR, we no longer update LIB via votes. LIB is only updated according to qc claims in block extensions.

Update `fork_database` fork-choice rule.
1. core.last_final_block_num
2. core.latest_qc_claim_block_num
3. block timestamp

Potentially switches forks before starting to produce a block. Does not interrupt producing blocks once block production has started.

Includes some misc performance improvements to `fork_database`.

Currently does not include voting on non-validated blocks. Would like to add the voting change in #2159 or as a separate PR.

Resolves #2125


libraries/chain/controller.cpp
```diff
@@ -3111,22 +3123,19 @@ struct controller_impl {
...
       // TODO: Do not aggregate votes on block_state if in legacy block fork_db
-      auto aggregate_vote_legacy = [](auto&) -> std::pair<vote_status, std::optional<uint32_t>> {
-         return {vote_status::unknown_block, {}};
+      auto aggregate_vote_legacy = [](auto&) -> vote_status {
+         return vote_status::unknown_block;
       };
-      auto [status, new_lib] = fork_db.apply<std::pair<vote_status, std::optional<uint32_t>>>(aggregate_vote_legacy, aggregate_vote);
-      if (new_lib) {
-         set_if_irreversible_block_num(*new_lib);
-      }
```
greg7mdp, line 3128: Maybe add in the comment PR that after this PR we no longer update lib when processing votes messages, but only when `commit`ing a block which has a claim in its IF extension which advances lib.

libraries/chain/controller.cpp
```diff
@@ -3111,22 +3123,19 @@ struct controller_impl {
...
-         return {vote_status::unknown_block, {}};
+      auto aggregate_vote = [&vote](auto& forkdb) -> vote_status {
+          auto bsp = forkdb.get_block(vote.proposal_id);
+          if (bsp) {
+             return bsp->aggregate_vote(vote);
+          }
+          return vote_status::unknown_block;
       };
       // TODO: https://github.com/AntelopeIO/leap/issues/2057
       // TODO: Do not aggregate votes on block_state if in legacy block fork_db
```
greg7mdp, line 3134: I think we can remove this comment, right?

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -1779,6 +1779,11 @@ producer_plugin_impl::start_block_result producer_plugin_impl::start_block() {
    if (!chain_plug->accept_transactions())
       return start_block_result::waiting_for_block;
 
+   abort_block();
+
+   chain.maybe_switch_forks([this](const transaction_metadata_ptr& trx) { _unapplied_transactions.add_forked(trx); },
+                            [this](const transaction_id_type& id) { return _unapplied_transactions.get_trx(id); });
```
greg7mdp, line 1785: Do we still need this?

libraries/chain/fork_database.cpp
```diff
@@ -21,79 +20,117 @@ namespace eosio::chain {
...
-                                         // see first_preferred comment
-                                         BOOST_MULTI_INDEX_CONST_MEM_FUN(bs, uint32_t, irreversible_blocknum),
-                                         BOOST_MULTI_INDEX_CONST_MEM_FUN(bs, uint32_t, block_num),
-                                         BOOST_MULTI_INDEX_CONST_MEM_FUN(bs, const block_id_type&, id)>,
-                           composite_key_compare<std::greater<bool>, std::greater<uint32_t>, std::greater<uint32_t>, sha256_less>>>>;
+            hashed_unique<tag<by_block_id>, BOOST_MULTI_INDEX_CONST_MEM_FUN(BS, const block_id_type&, id), std::hash<block_id_type>>,
+            ordered_non_unique<tag<by_prev>, const_mem_fun<BS, const block_id_type&, &BS::previous>>,
+            ordered_unique<tag<by_best_branch>,
+               composite_key<BS,
+                  global_fun<const BS&,             bool,     &BSAccessor::is_valid>,
```
greg7mdp, line 77: Why not use ` const_mem_fun<BS,  bool, &BS::is_valid>` instead of adding a member in the BSAccessors?

Same for `set_valid`, there was already a function available in both block_state versions.

greg7mdp: So as a result the `BSAccessor` would be needed only for the index definition.

Using the accessor is not optimal, because for the legacy version we compare twice on the same value `bs.irreversible_blocknum();`.

I think it would be better to have two separate definitions for the `by_best_branch` index, which would directly access their respective `block_state` and `block_state_legacy` members without accessors.

libraries/chain/fork_database.cpp
```diff
@@ -21,79 +20,117 @@ namespace eosio::chain {
...
    struct fork_database_impl {
-      using bs               = bsp::element_type;
-      using bhsp             = bs::bhsp_t;
-      using bhs              = bhsp::element_type;
-      
-      using fork_db_t        = fork_database_t<bsp>;
+      using BS               = BSP::element_type;
+      using BSAccessor       = BS::fork_db_block_state_accessor;
+      using BHSP             = BS::bhsp_t;
+      using BHS              = BHSP::element_type;
```
greg7mdp, line 63: these are not template parameters. I think it would be nicer and more idiomatic to have them in lowercase (like `std::vector<T>` defines `using value_type = T;`).

I'd be for somethink like this, or the types with `_t` suffix

```
      using bsp              = BSP;
      using bs               = bsp::element_type;
```

libraries/chain/hotstuff/block_construction_data_flow.md
```diff
@@ -72,7 +72,7 @@ The new storage for IF is:
 struct block_header_state_core {
```
greg7mdp, line 72: actually there is no `block_header_state_core` anymore :-)

unittests/fork_db_tests.cpp
```diff
@@ -0,0 +1,107 @@
...
+inline block_id_type make_block_id(block_num_type block_num) {
+   static uint32_t nonce = 0;
+   ++nonce;
+   block_id_type id = fc::sha256::hash(std::to_string(block_num) + "-" + std::to_string(nonce));
+   id._hash[0] &= 0xffffffff00000000;
+   id._hash[0] += fc::endian_reverse_u32(block_num); // store the block num in the ID, 160 bits is plenty for the hash
+   return id;
+}
+
+// Used to access privates of block_state
```
greg7mdp, line 19: I found a way to create `block_state_ptr` without accessing private members in my PR :-)

unittests/fork_db_tests.cpp
```diff
@@ -0,0 +1,107 @@
...
+   }
+
+   // test remove, should remove descendants
+   forkdb.remove(bsp12b->id());
+   BOOST_TEST(!forkdb.get_block(bsp12b->id()));
+   BOOST_TEST(!forkdb.get_block(bsp13b->id()));
+   BOOST_TEST(!forkdb.get_block(bsp14b->id()));
+   forkdb.add(bsp12b, mark_valid_t::no, ignore_duplicate_t::no); // will throw if already exists
+   forkdb.add(bsp13b, mark_valid_t::no, ignore_duplicate_t::no); // will throw if already exists
+   forkdb.add(bsp14b, mark_valid_t::no, ignore_duplicate_t::no); // will throw if already exists
```
greg7mdp, line 103: Nice test! The one in your other PR for `update_best_qc` is even better!

plugins/producer_plugin/producer_plugin.cpp
heifner: I think we do once #2159 is done. #2159 will put blocks into the fork database that could potentially be better forks. So a check here will allow us to switch to that fork if needed while producing.

unittests/fork_db_tests.cpp
heifner: Can you point me to what you are suggesting instead?

ericpassmore:
Nice to see the improvements. `fork_db_test.cpp` looks very nice.

libraries/chain/controller.cpp
greg7mdp: What I mean, is that with the current code in your PR we do not aggregate votes on block_states that are in legacy block fork_db because `aggregate_vote_legacy` is a no-op. So it seems that the comment specifies a TODO which is already done.

unittests/fork_db_tests.cpp
greg7mdp: I was referring to [this](https://github.com/AntelopeIO/leap/blob/4aa701fb52214b18ade27ef38503c0785f007156/unittests/finalizer_tests.cpp#L213-L235), but I have not looked at your use in great detail and maybe what I did would not work for your needs.

libraries/chain/controller.cpp
heifner: Wehn #2057 is done there will be two fork database active at once. When the change is made this is a reminder that this code needs to still not aggregate the votes into legacy. I think it best to leave for now and remove when #2057 is worked.

libraries/chain/fork_database.cpp
heifner: The BSAccessor for valid is used because I made it private. It really should be private as it is modified after other threads have access to the `block_state`.

I think to have two diff definitions of `by_best_branch` we would have to make `fork_multi_index_type` a template parameter of `fork_database_impl`. I'm not necessarily opposed to that, but I doubt the extra comparison will make much difference. If the BSAccessor and extra comparison is worth the effort, I can make the change. What do you think?

libraries/chain/fork_database.cpp
heifner: Interesting. I'm ok with that. I do think `bsp_t` `bs_t`, etc. is better.

libraries/chain/hotstuff/block_construction_data_flow.md
heifner: I'm not sure what we gain by maintaining this MD doc when it is just a mirror of the code. Should I remove it?

unittests/fork_db_tests.cpp
heifner: Hmm, I don't think that works. Although, the BSAccessor is really not needed for accessing privates anymore. It is used more as a way to normalize the interface. If we go with two diff `by_best_branch` then `BSAccessor` can be removed.

unittests/fork_db_tests.cpp
heifner: Don't need the `update_best_qc` anymore, but thanks.

libraries/chain/controller.cpp
greg7mdp: Oh yes this makes total sense, thanks.

plugins/producer_plugin/producer_plugin.cpp
greg7mdp: It makes sense that we have to do this, but I would have expected this deciding what is the best fork to be in `controller.cpp`, maybe in `start_block`. However I'm not very familiar with `producer_plugin`, or the architectural reason why it is separated into a plugin and not just in `libraries/chain`,  so there is probably a good reason why this has to be where you added it.

libraries/chain/fork_database.cpp
greg7mdp: I agree that the extra comparison is not a big issue. 
 I was just thinking that the resulting code with my suggestion would be simpler and cleaner (and I don't believe it would require making fork_multi_index_type a template parameter of fork_database_impl). Let me play with it to see if I can make it simpler and we can decide which way is the nicest.

plugins/producer_plugin/producer_plugin.cpp
heifner: Yeah, has to be here. We need to switch forks before determining who should produce which is determined in the `producer_plugin`. I have often said that `producer_plugin` should be in the chain library. However, I now know of a use case where it is really nice that it is a plugin. It allows for different producing logic while still using the core controller.

libraries/chain/hotstuff/block_construction_data_flow.md
greg7mdp: I think this .md (written by Areg btw) may still have some information (describing which post-if members are equivalent to the pre-if ones) which is not in the code, but it is true that I have not needed to look at it in some time. I agree that there is not need to keep it up-to-date, but maybe we should keep it a little longer and remove it before the release. Before removing it, I'd like to go through it and make sure that insightful info from it is replicated in code comments.

heifner: Sounds good to me.

libraries/chain/fork_database.cpp
greg7mdp: Perfect, let's go for that then, I like it too!

greg7mdp: Maybe even for consistency add `using bsp_t = BSP;`

libraries/chain/fork_database.cpp
greg7mdp: see [this](https://github.com/AntelopeIO/leap/compare/GH-2125-forkdb-simpler...by_best_branch)

libraries/chain/controller.cpp
```diff
@@ -2802,6 +2800,20 @@ struct controller_impl {
...
+               const auto& bsp = std::get<std::decay_t<decltype(forkdb.chain_head)>>(cb.bsp);
+
+               uint16_t if_ext_id = instant_finality_extension::extension_id();
+               assert(bsp->header_exts.count(if_ext_id) > 0); // in all instant_finality block headers
+               const auto& if_ext = std::get<instant_finality_extension>(bsp->header_exts.lower_bound(if_ext_id)->second);
+               if (if_ext.qc_claim.is_strong_qc) {
+                  auto claimed = forkdb.search_on_head_branch(if_ext.qc_claim.block_num);
+                  if (claimed) {
+                     set_if_irreversible_block_num(claimed->core.final_on_strong_qc_block_num);
+                  }
```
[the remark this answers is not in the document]
heifner: Actually when searching on head, yes. However, this should be searching on bsp branch not head which is not necessarily the same thing here. Even then I think it can be not found if LIB already advanced past the `claimed->core.final_on_strong_qc_block_num`.

libraries/chain/fork_database.cpp
```diff
@@ -341,114 +402,155 @@ namespace eosio::chain {
...
-   fork_database_impl<bsp>::fetch_branch_impl(const block_id_type& h, uint32_t trim_after_block_num) const {
-      branch_type result;
+   template <class BSP>
+   fork_database_t<BSP>::branch_t
+   fork_database_impl<BSP>::fetch_branch_impl(const block_id_type& h, uint32_t trim_after_block_num) const {
+      branch_t result;
+      result.reserve(index.size());
+      for (auto i = index.find(h); i != index.end(); i = index.find((*i)->previous())) {
+         if ((*i)->block_num() <= trim_after_block_num)
+            result.push_back(*i);
```
greg7mdp, line 482:
```suggestion
         if ((*i)->block_num() > trim_after_block_num)
            break;
         result.push_back(*i);
```

libraries/chain/fork_database.cpp
```diff
@@ -341,114 +402,155 @@ namespace eosio::chain {
...
+   block_branch_t
+   fork_database_impl<BSP>::fetch_block_branch_impl(const block_id_type& h, uint32_t trim_after_block_num) const {
+      block_branch_t result;
       result.reserve(index.size());
-      for (auto s = get_block_impl(h); s; s = get_block_impl(s->previous())) {
-         if (s->block_num() <= trim_after_block_num)
-            result.push_back(s);
+      for (auto i = index.find(h); i != index.end(); i = index.find((*i)->previous())) {
+         if ((*i)->block_num() <= trim_after_block_num)
+            result.push_back((*i)->block);
```
greg7mdp, line 502:
```suggestion
         if ((*i)->block_num() > trim_after_block_num)
            break;
         result.push_back((*i)->block);
```

libraries/chain/fork_database.cpp
```diff
@@ -341,114 +402,155 @@ namespace eosio::chain {
...
+   BSP fork_database_t<BSP>::search_on_head_branch( uint32_t block_num ) const {
+      std::lock_guard g(my->mtx);
+      return my->search_on_head_branch_impl(block_num);
+   }
+
+   template<class BSP>
+   BSP fork_database_impl<BSP>::search_on_head_branch_impl( uint32_t block_num ) const {
+      for (auto i = index.find(head->id()); i != index.end(); i = index.find((*i)->previous())) {
+         if ((*i)->block_num() == block_num)
+            return *i;
```
greg7mdp, line 553:
```suggestion
      return my->search_on_branch_impl(head->id(), block_num);
```

greg7mdp (changes requested):

linh2931 (approved):

libraries/chain/fork_database.cpp
heifner: This doesn't work. See new test case which fails with this implementation.

libraries/chain/controller.cpp
```diff
@@ -2807,7 +2807,8 @@ struct controller_impl {
                assert(bsp->header_exts.count(if_ext_id) > 0); // in all instant_finality block headers
                const auto& if_ext = std::get<instant_finality_extension>(bsp->header_exts.lower_bound(if_ext_id)->second);
                if (if_ext.qc_claim.is_strong_qc) {
-                  auto claimed = forkdb.search_on_head_branch(if_ext.qc_claim.block_num);
+                  // claim has already been verified
+                  auto claimed = forkdb.search_branch(bsp->id(), if_ext.qc_claim.block_num);
```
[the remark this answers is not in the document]
heifner: I don't want to reach outside the lambda to access fork_db variant again.

libraries/chain/fork_database.cpp
greg7mdp: I'm baffled! Do you know why my version doesn't work or should I look into it?

greg7mdp: Oh never mind I was again misled by the name of the parameter.

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: LIB is only updated according to qc claims in block extensions. Included updated fork choice rules.
Note:end

PR #2274, opened 2024-03-01
IF: Don't drop late blocks

### LINKED ISSUE #2159: IF: Don't drop late blocks

heifner:
- Instead of dropping blocks when producing, incorporate blocks received while producing into the fork database.
  - Potentially fork switch to newly incorporated blocks before producing a block
  - Use block id instead of block number for irreversible so that correct fork branch is made irreversible

Resolves #2159

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1229,6 +1229,12 @@ struct controller_impl {
                                     : forkdb.fetch_branch( fork_db_head(forkdb, irreversible_mode())->id(), new_lib );
          try {
             auto should_process = [&](auto& bsp) {
+               // Only make irreversible blocks that have been validated. Blocks in the fork database may not be on our current best head
+               // and therefore have not been validated.
+               // An alternative more complex implementation would be to do a fork switch here and validate all blocks so they can be then made
+               // irreversible. Instead this moves irreversible as much as possible and allows the next maybe_switch_forks call to apply these
+               // non-validated blocks. After the maybe_switch_forks call (before next produced block or on next received block), irreversible
+               // can then move forward on the then validated blocks.
                return read_mode == db_read_mode::IRREVERSIBLE || bsp->is_valid();
             };
 
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1010,6 +1010,12 @@ struct controller_impl {
       });
    }
 
+   bool fork_db_block_exists( const block_id_type& id ) const {
+      return fork_db.apply<bool>([&](const auto& forkdb) {
+         return forkdb.block_exists(id);
+      });
+   }
+
    signed_block_ptr fork_db_fetch_block_by_id( const block_id_type& id ) const {
       return fork_db.apply<signed_block_ptr>([&](const auto& forkdb) {
          auto bsp = forkdb.get_block(id);
@@ -3375,7 +3381,7 @@ struct controller_impl {
 
             auto prev = forkdb.get_block_header( b->previous );
             EOS_ASSERT( prev, unlinkable_block_exception,
-                        "unlinkable block ${id}", ("id", id)("previous", b->previous) );
+                        "unlinkable block ${id} previous ${p}", ("id", id)("p", b->previous) );
 
             return control->create_block_state_i( id, b, *prev );
          } );
@@ -3409,6 +3415,8 @@ struct controller_impl {
                     const forked_callback_t& forked_branch_cb,
                     const trx_meta_cache_lookup& trx_lookup )
    {
+      assert(bsp && bsp->block);
+
       // Save the received QC as soon as possible, no matter whether the block itself is valid or not
       if constexpr (std::is_same_v<BSP, block_state_ptr>) {
          integrate_received_qc_to_block(bsp);
@@ -3421,7 +3429,6 @@ struct controller_impl {
          trusted_producer_light_validation = old_value;
       });
       try {
-         EOS_ASSERT( bsp, block_validate_exception, "null block" );
          const auto& b = bsp->block;
 
          if( conf.terminate_at_block > 0 && conf.terminate_at_block <= head_block_num()) {
@@ -4315,12 +4322,16 @@ std::optional<block_handle> controller::create_block_handle( const block_id_type
 }
 
 void controller::push_block( block_report& br,
-                             const block_handle& bt,
+                             const block_handle& b,
                              const forked_callback_t& forked_cb,
                              const trx_meta_cache_lookup& trx_lookup )
 {
    validate_db_available_size();
-   std::visit([&](const auto& bsp) { my->push_block( br, bsp, forked_cb, trx_lookup); }, bt.bsp);
+   std::visit([&](const auto& bsp) { my->push_block( br, bsp, forked_cb, trx_lookup); }, b.bsp);
+}
+
+void controller::accept_block(const block_handle& b) {
+   std::visit([&](const auto& bsp) { my->accept_block(bsp); }, b.bsp);
 }
 
 transaction_trace_ptr controller::push_transaction( const transaction_metadata_ptr& trx,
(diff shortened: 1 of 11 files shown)


libraries/chain/controller.cpp
```diff
@@ -1212,23 +1218,28 @@ struct controller_impl {
...
          return;
 
       auto mark_branch_irreversible = [&, this](auto& forkdb) {
-         auto branch = forkdb.fetch_branch( fork_db_head(forkdb, irreversible_mode())->id(), new_lib );
+         auto branch = (if_lib > 0) ? forkdb.fetch_branch( if_irreversible_block_id, new_lib)
+                                    : forkdb.fetch_branch( fork_db_head(forkdb, irreversible_mode())->id(), new_lib );
          try {
+            auto should_process = [&](auto& bsp) {
+               return read_mode == db_read_mode::IRREVERSIBLE || bsp->is_valid();
+            };
```
greg7mdp, line 1233: Why is this necessary? I thought that lib and its ancestors are guaranteed to be valid.

libraries/chain/controller.cpp
```diff
@@ -1212,23 +1218,28 @@ struct controller_impl {
...
-      const uint32_t if_lib = if_irreversible_block_num;
+      uint32_t if_lib = block_header::num_from_id(if_irreversible_block_id);
       const uint32_t new_lib = if_lib > 0 ? if_lib : fork_db_head_irreversible_blocknum();
 
       if( new_lib <= lib_num )
          return;
 
       auto mark_branch_irreversible = [&, this](auto& forkdb) {
-         auto branch = forkdb.fetch_branch( fork_db_head(forkdb, irreversible_mode())->id(), new_lib );
+         auto branch = (if_lib > 0) ? forkdb.fetch_branch( if_irreversible_block_id, new_lib)
```
greg7mdp, line 1228: That's always going to be a single block, right, since `new_lib == num_from_id(if_irreversible_block_id);`

libraries/chain/controller.cpp
heifner: No, they are not guaranteed to be valid. We are putting non-validated blocks into the fork database. A different branch than our best head can become irreversible which means we have not necessarily validated them. The `lib_advance_test.py` actually triggered this case.

An alternative more complex implementation would be to do a fork switch so we can move irreversible all the way at this point. Instead this moves irreversible as much as possible right here and then waits for a `maybe_switch_forks` call to happen *soon* to switch to the new branch and then move irreversible after that. *soon* in this case should be before the next produced block or on the next received block.

I'll add a comment that explains this.

libraries/chain/controller.cpp
heifner: No, this returns all blocks from root up to `new_lib` on branch with `if_irreversible_block_id`.

greg7mdp: Aren't they the same though?

libraries/chain/controller.cpp
greg7mdp: Even if we have not validated the block, it became irreversible because we received a strong QC making it irreversible, so I am not sure what benefit there is to not accept it as irreversible. Even if it did not validate, I don't know what we can do since a quorum of finalizers already accepted it.

libraries/chain/controller.cpp
greg7mdp: Looking at the code I see:
```
if_lib = block_header::num_from_id(if_irreversible_block_id);
new_lib = if_lib;
```

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Incorporate blocks received while producing into fork database. Previously late blocks were dropped. 
Note:end

libraries/chain/controller.cpp
heifner: `new_lib` does equal `num_from_id(if_irreversible_block_id)`

However, `forkdb.fetch_branch( if_irreversible_block_id, new_lib)` returns all blocks from the `root` of the fork database up to and including `if_irreversible_block_id` trimming off any after `new_lib` which there will not be any. Could not pass in `new_lib` as it doesn't matter. However, might as well pass it in as it makes it same as pre-instant-finality.

Here is an example from a unittest where LIB moved by two blocks. You can see the branch is of size 2 in this case.

image

libraries/chain/controller.cpp
heifner: It is our job as a validating node to validate blocks. If some disaster scenario happens we want to be in a good state. Also our block log should not contain data that is not in our state. We need a consistent state and block log in case we have to shutdown. Not even sure what kind of strange corner cases there would be for a node that has blocks in the block log that are not incorporated into the state.

heifner: Yes, it would be split. But who is to say this isn't the correct chain. Manual intervention is needed.

libraries/chain/controller.cpp
greg7mdp: Got it, sorry, for some reason I keep misunderstanding that `fetch_branch` API.

libraries/chain/controller.cpp
greg7mdp: Do you agree that the node might as well terminate in that case?

greg7mdp: @heifner I modified your comment by mistake. Sorry!

greg7mdp: > But who is to say this isn't the correct chain.

A quorum of finalizers said so.

heifner: > @heifner I modified your comment by mistake. Sorry!

Done that many times myself. No worries.

greg7mdp: OK, I guess I understand the logic here. We expect that, even if the block made irreversible by the received QC has not been validated yet, we will validate it eventually and then be able to commit it to the state. Seems like a weird state to be in though.

libraries/chain/controller.cpp
```diff
@@ -816,7 +816,7 @@ struct controller_impl {
    block_log                       blog;
    std::optional<pending_state>    pending;
    fork_database                   fork_db;
-   std::atomic<uint32_t>           if_irreversible_block_num{0};
+   block_id_type                   if_irreversible_block_id;
```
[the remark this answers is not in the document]
heifner: Correct, only accessed by the main thread now.

libraries/chain/controller.cpp
```diff
@@ -3422,7 +3462,6 @@ struct controller_impl {
          trusted_producer_light_validation = old_value;
       });
       try {
-         EOS_ASSERT( bsp, block_validate_exception, "null block" );
```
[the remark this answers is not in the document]
heifner: It shouldn't as this is more of a true `assert` then an `EOS_ASSERT`.

plugins/producer_plugin/producer_plugin.cpp
```diff
@@ -667,38 +667,46 @@ class producer_plugin_impl : public std::enable_shared_from_this<producer_plugin
...
       }
 
+      if (in_producing_mode()) {
+         fc_ilog(_log, "producing, incoming block #${num} id: ${id}", ("num", blk_num)("id", id));
+         const block_handle& bh = obt ? *obt : btf.get();
+         chain.accept_block(bh);
+         return true; // return true because block was accepted
+      }
+
+      if (now - block->timestamp < fc::minutes(5) || (blk_num % 1000 == 0)) // only log every 1000 during sync
```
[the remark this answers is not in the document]
heifner: Sure, probably makes sense

linh2931 (approved):

greg7mdp (approved):

PR #2275, opened 2024-03-02
IF: Vote before applying/validating block if possible

### LINKED ISSUE #2125: IF: Update fork-choice rule for fork database

heifner:
Vote on blocks that are not yet validated. The rule for now is:

1. Get the `core.final_on_strong_qc_block_num` for the block you are considering to vote on and use that to find the actual block ID of the ancestor block that has that block number.
2. If that block ID is not an ancestor of the current head block, then do not vote for that block.
3. Otherwise, consider voting for that block according to the `decide_vote` rules.

Resolves #2125

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3172,43 +3172,6 @@ struct controller_impl {
           });
    }
 
-   // expected to be called from application thread as it modifies bsp->valid_qc and if_irreversible_block_id
-   void integrate_received_qc_to_block(const block_state_ptr& bsp_in) {
-      // extract QC from block extension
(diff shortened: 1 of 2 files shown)


greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Rules for voting on blocks that are not yet validated. Part of the improved core finality algorithm work. 
Note:end

linh2931 (approved):

PR #2276, opened 2024-03-02
don't do `set_new_handler()` workaround for LLVM 9+ where it's unnecessary

spoonincode:
See comment in code. There isn't really a problem with the existing code being there afaik, so I suppose this is just adding an indicator to remove this workaround completely once the LLVM requirement is strictly 9+

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -220,13 +220,18 @@ struct controller_impl {
              // Read-write tasks are not being executed.
    };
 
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

heifner (approved):

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Clean separation of code for LLVM versions before (<9) and after (9+). Once Leap's LLVM standard is moved to 9+ `set_new_handler()` call will be removed. 
Note:end

PR #2282, opened 2024-03-05
IF: Implement computing finality digest

linh2931:
A finality digest is a hash of a series sub-digests required for IBC verification. It is built from a hash of a serialization of `light_header_protocol_version`, `active_finalizer_policy_generation`, `finality_tree_digest`, and a combination of `base_digest` and `active_finalizer_policy_digest`. `base_digest` is a hash of a serialization of deterministic fields of `block_header_state`, including `header`, `core`, `finalizer_policies`, `active_proposer_policy`, `proposer_policies`, and `activated_protocol_features`.

`finality_tree` is defined as following:
 1. A Finality Merkle Tree is a Merkle tree over a sequence of Finality Leaf Nodes,
    one for each block starting from the IF Genesis Block and ending at some
   specified descendant block.
 2. The Validation Tree associated with a target block is the Finality Merkle
    Tree over Finality Leaf Nodes starting with the one for the IF Genesis Block
   and ending with the one for the block that is the parent of the the target Block.
 3. The Finality Tree ssociated with a target block is the Validation Tree of the
   block referenced by the target block's final_on_strong_qc_block_num.
  That is, validation_tree(core.final_on_strong_qc_block_num))

A `valid` structure is introduced into `block_state` to hold all necessary information.

This PR implemented computing finality digest based on finality merkle tree, validation tree, finality tree, and validated structure.

See https://github.com/AntelopeIO/leap/issues/2080 for design details.

Resolved https://github.com/AntelopeIO/leap/issues/2080


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Implementation of finality digest, a hash of a series sub-digests, required for IBC verification.
Note:end

libraries/chain/block_state.cpp
```diff
@@ -51,6 +64,19 @@ block_state::block_state(const block_state_legacy& bsp) {
    core = finality_core::create_core_for_genesis_block(bsp.block_num()); // [if todo] instant transition is not acceptable
    activated_protocol_features = bsp.activated_protocol_features;
 
+   // construct valid structure
+   valid = valid_t {
+      .finality_merkel_tree = incremental_merkle_tree{},
+      .finality_mroots      = { digest_type{} }, // for genesis block itself
+      .last_final_block_num = bsp.block_num()
+   };
+   valid_t::finality_leaf_node_t node {
```
systemzax, line 73: I think we also discussed adding a "finality leaf version number" here, which will help proper handling of different versions moving forward.

libraries/chain/block_state.cpp
```diff
@@ -51,6 +64,19 @@ block_state::block_state(const block_state_legacy& bsp) {
...
+      .block_num       = bsp.block_num(),
+      .finality_digest = digest_type{},
+      .finality_mroot  = digest_type{}
```
systemzax, line 76: Here should be the action_mroot.

linh2931: Thanks! I have rename it to `action_mroot` so it is clearer.

libraries/chain/block_state.cpp
linh2931: Does the `light_header_protocol_version` in `finality_digest_data_v0_t` of block_header_state.cpp serve that purpose?

systemzax: No, this here is a different version number, as it applies to the "dynamic data" portion of the leaf data. It will eventually be used when we incorporate state proofs at some point in the future. In the diagram, I called it "finality_tree_leaf_version".

libraries/chain/controller.cpp
```diff
@@ -354,14 +355,55 @@ struct assembled_block {
...
+   valid.finality_mroots.emplace_back(parent_bsp->valid->finality_merkel_tree.get_root());
+
+   // post condition of finality_mroots
+   assert(valid.finality_mroots.size() == (bhs.block_num() - bhs.core.last_final_block_num() + 1));
+
+   // construct block's finality leaf node.
+   valid_t::finality_leaf_node_t leaf_node{
+      .block_num       = bhs.block_num(),
+      .finality_digest = bhs.compute_finalizer_digest(),
+      .action_mroot    = bhs.finality_mroot()
```
systemzax, line 395: We also need to make sure the action_mroot here contains the merkle root of the action receipt digests tree.

This is calculated here :

https://github.com/AntelopeIO/leap/blob/main/libraries/chain/controller.cpp#L1874

libraries/chain/controller.cpp
```diff
@@ -658,27 +701,30 @@ struct building_block {
                };
             },
             [&](building_block_if& bb) -> assembled_block {
-               // compute the action_mroot and transaction_mroot
-               auto [transaction_mroot, action_mroot] = std::visit(
-                  overloaded{[&](digests_t& trx_receipts) { // calculate the two merkle roots in separate threads
+               // compute the transaction_mroot
+               auto transaction_mroot = std::visit(
+                  overloaded{[&](digests_t& trx_receipts) { // calculate the merkle root in a thread
```
systemzax, line 706: Is there a reason why the action_mroot calculation was removed from here? Is it calculated elsewhere now?

libraries/chain/controller.cpp
linh2931: oha, we still nee to calculate action mroot but just we don't eventually store it in block header.
Will revert back.

libraries/chain/controller.cpp
linh2931: Will revert back. Thanks for pointing this out!

libraries/chain/controller.cpp
```diff
@@ -3176,10 +3277,28 @@ struct controller_impl {
...
-               EOS_ASSERT(producer_block_id == ab.id(), block_validate_exception, "Block ID does not match",
+               EOS_ASSERT(producer_block_id == ab.id(), block_validate_exception, "Block ID does not match, ${producer_block_id} != ${validator_block_id}",
                           ("producer_block_id", producer_block_id)("validator_block_id", ab.id()));
             }
 
+            // verify received finality digest in action_mroot is the same as the computed
+            if constexpr (std::is_same_v<BSP, block_state_ptr>) {
+               if (!bsp->valid) { // no need to re-validate if it is already valid
+                  block_state_ptr parent_bsp = fork_db.apply_s<block_state_ptr> ([&](const auto& forkdb) {
+                      return forkdb.get_block(bsp->previous(), include_root_t::yes);
```
heifner, line 3288: Blocks are not stored in `forkdb` during replay.

linh2931: Thanks. Change to use `chain_head` as the parent of the block being built.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -54,13 +58,12 @@ struct block_header_state {
...
    // ------ functions -----------------------------------------------------------------
-   // [if todo] https://github.com/AntelopeIO/leap/issues/2080
-   const block_id_type&  id() const { return block_id; }
-   digest_type           compute_finalizer_digest() const { return block_id; };
-   block_timestamp_type  timestamp() const { return header.timestamp; }
-   account_name          producer() const  { return header.producer; }
-   const block_id_type&  previous() const  { return header.previous; }
-   uint32_t              block_num() const { return block_header::num_from_id(previous()) + 1; }
+   const block_id_type&  id()             const { return block_id; }
+   digest_type           finality_mroot() const { return header.action_mroot; }
```
heifner, line 62: Should return a `const digest_type&`

heifner (changes requested):

libraries/chain/controller.cpp
```diff
@@ -3167,7 +3257,18 @@ struct controller_impl {
                           ("lhs", r)("rhs", static_cast<const transaction_receipt_header&>(receipt)));
             }
 
-            assemble_block(true, extract_qc_data(b));
+            // save action_mroot for use in build_valid_structure
+            const auto& action_receipts = std::get<building_block>(pending->_block_stage).action_receipt_digests();
+            auto action_mroot = calculate_merkle(std::move(action_receipts));
```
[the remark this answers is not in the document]
linh2931: Thanks. Changed to cache it and reuse.

heifner (changes requested):

libraries/chain/controller.cpp
```diff
@@ -307,6 +309,13 @@ struct assembled_block {
          v);
    }
 
+   digest_type action_mroot() const {
+      return std::visit(
+         overloaded{[](const assembled_block_legacy& ab) -> digest_type { assert(false); return digest_type{}; },
+                    [](const assembled_block_if& ab)     -> digest_type { return ab.action_mroot; }},
+         v);
```
heifner, line 316: If this is needed, then it should return `const digest_type&`

libraries/chain/controller.cpp
```diff
@@ -725,14 +801,37 @@ struct building_block {
...
-                                                      std::move(bb.pending_trx_metas), std::move(bb.pending_trx_receipts),
-                                                      qc_data ? std::move(qc_data->qc) : std::optional<quorum_certificate>{}};
+               auto bhs = bb.parent.next(bhs_input);
+
+               std::optional<valid_t> valid;
+               if (!validating) {
+                  if (!parent_bsp) {
+                     parent_bsp = fork_db.apply_s<block_state_ptr> ([&](const auto& forkdb) {
+                        return forkdb.get_block(parent_id(), include_root_t::yes);
+                     });
```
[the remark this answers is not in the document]
linh2931: Yes I was thinking about that. It would require `chain_head` to building block or some plumbing. Will do it.

libraries/chain/controller.cpp
```diff
@@ -3176,10 +3283,25 @@ struct controller_impl {
...
+                  bsp->valid = build_valid_structure(parent_bsp, *bsp, ab.action_mroot());
+
+                  auto computed_finality_mroot = bsp->valid->get_finality_mroot(bsp->core.final_on_strong_qc_block_num);
+
+                  EOS_ASSERT(bsp->finality_mroot() == computed_finality_mroot,
+                        block_validate_exception,
+                        "finality_mroot does not match, received finality_mroot: ${r} != computed_finality_mroot: ${c}",
+                         ("r", bsp->finality_mroot())("c", computed_finality_mroot));
+               }
+            }
```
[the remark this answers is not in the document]
linh2931: `valid` is optional and was not created initially for blocks coming from peers. I could have building the valid when validating parameter is true in assemble block, store the valid in assembled_block, and update it here.

libraries/chain/controller.cpp
linh2931: I thought about `const digest_type&`. But for Legacy, it returns a reference to a temporary, which is ugly.

systemzax:
Small thing, but some variables are named "merkel" and others have "merkle" . Merkle is correct and should be used across the board for consistency.

libraries/chain/controller.cpp
```diff
@@ -712,27 +732,51 @@ struct building_block {
                         // Construct a default QC claim.
                         qc_data = qc_data_t{ {}, bb.parent.core.latest_qc_claim() };
                      }
-                  });
 
+                     finality_mroot_claim = get_finality_mroot_claim(bb.parent, qc_data->qc_claim);
+                  });
                }
```
heifner, line 738: I thought you were going to create a method for the above that returned the `qc_data` and `finality_mroot_claim` to avoid the use of the optionals.

libraries/chain/controller.cpp
```diff
@@ -627,12 +625,24 @@ struct building_block {
...
+   }
+
    assembled_block assemble_block(boost::asio::io_context& ioc,
                                   const protocol_feature_set& pfs,
                                   fork_database& fork_db,
                                   std::unique_ptr<proposer_policy> new_proposer_policy,
                                   bool validating,
-                                  std::optional<qc_data_t> validating_qc_data) {
+                                  std::optional<qc_data_t> validating_qc_data,
+                                  block_state_ptr validating_bsp) {
```
heifner, line 645: Should pass `validating_bsp` by `const &`

heifner (changes requested):

greg7mdp:
@linh2931 the `block_state::valid` member needs to be saved to the snapshot. 
Would you like me to add the changes to your branch?

libraries/chain/include/eosio/chain/block_state.hpp
```diff
@@ -22,13 +22,45 @@ inline weak_digest_t create_weak_digest(const digest_type& digest) {
...
 
+/*
+ * Important concepts:
+ * 1. A Finality Merkle Tree is a Merkle tree over a sequence of Finality Leaf Nodes,
+ *    one for each block starting from the IF Genesis Block and ending at some
+ *    specified descendant block.
+ * 2. The Validation Tree associated with a target block is the Finality Merkle
+ *    Tree over Finality Leaf Nodes starting with the one for the IF Genesis Block
+ *    and ending with the one for the target Block.
+ * 3. The Finality Tree ssociated with a target block is the Validation Tree of the
```
greg7mdp, line 33:
```suggestion
 * 3. The Finality Tree associated with a target block is the Validation Tree of the
```

libraries/chain/include/eosio/chain/block_header.hpp
```diff
@@ -52,7 +52,12 @@ namespace eosio::chain {
       block_id_type                    previous;
 
       checksum256_type                 transaction_mroot; /// mroot of cycles_summary
-      checksum256_type                 action_mroot; /// mroot of all delivered action receipts
+
+      // In Legacy, action_mroot is the mroot of all delivered action receipts.
+      // In Savanna, action_mroot is the digest of the root of the Finality Tree
+      // associated with the block, i.e. the digest of the root of
```
greg7mdp, line 58:
```suggestion
      // In Savanna, action_mroot is the root of the Finality Tree
      // associated with the block, i.e. the root of
```

libraries/chain/block_state.cpp
```diff
@@ -45,13 +50,28 @@ block_state::block_state(const block_header_state& bhs, deque<transaction_metada
    sign(signer, valid_block_signing_authority);
 }
 
-// Used for transition from dpos to instant-finality
-block_state::block_state(const block_state_legacy& bsp) {
+// Used for transition from dpos to Savanna.
+block_state::block_state(const block_state_legacy& bsp, const digest_type& action_mroot_svnn) {
```
greg7mdp, line 54: Not sure why, but I don't like the `_svnn`  suffix at all. Maybe because it is all consonnants?

libraries/chain/block_state.cpp
```diff
@@ -222,6 +240,52 @@ std::optional<quorum_certificate> block_state::get_best_qc() const {
...
+valid_t block_state::new_valid(const block_header_state& next_bhs, const digest_type& action_mroot) const {
+   assert(valid);
+   assert(next_bhs.core.last_final_block_num() >= core.last_final_block_num());
+
+   // Copy parent's validation_tree and validation_mroots.
+   auto start = next_bhs.core.last_final_block_num() - core.last_final_block_num();
+   valid_t next_valid {
+      .validation_tree = valid->validation_tree,
+      // Remove any roots from the front end to block whose block number is
+      // last_final_block_num - 1
```
greg7mdp, line 252:
```suggestion
      // Trim roots from the front end, up to block number `next_bhs.core.last_final_block_num() `
```

libraries/chain/include/eosio/chain/block_state.hpp
```diff
@@ -22,13 +22,45 @@ inline weak_digest_t create_weak_digest(const digest_type& digest) {
...
+      digest_type    action_mroot;    // digest of the root of the action Merkle tree of the block
+   };
+
+   // The Finality Merkle Tree, containing leaf nodes from IF genesis block to current block
+   incremental_merkle_tree validation_tree;
+
+   // The sequence of root digests of the validation trees associated
+   // to a unbroken sequence of blocks which consist of the block
+   // starting with the one that has a block number equal
+   // to core.last_final_block_num, and the current block
```
greg7mdp, line 52:
```suggestion
   // with an unbroken sequence of blocks consisting of the block
   // starting with the one that has a block number equal
   // to core.last_final_block_num, and ending with the current block
```

libraries/chain/controller.cpp
```diff
@@ -628,12 +626,61 @@ struct building_block {
...
+         if (!qc) {
+            // This only happens when parent block is the IF genesis block.
+            // There is no ancestor block which has a QC.
+            // Construct a default QC claim.
+            qc_data = qc_data_t{ {}, bb.parent.core.latest_qc_claim() };
+         }
+
+      });
+
+      return qc_data;
```
greg7mdp, line 674:
```suggestion

      return fork_db.apply_s<qc_data_t>([&](const auto& forkdb) {
         auto branch = forkdb.fetch_branch(bb.parent.id());
         std::optional<quorum_certificate> qc;
         for( auto it = branch.begin(); it != branch.end(); ++it ) {
            qc = (*it)->get_best_qc();
            if( qc ) {
               EOS_ASSERT( qc->block_num <= block_header::num_from_id(bb.parent.id()), block_validate_exception,
                           "most recent ancestor QC block number (${a}) cannot be greater than parent's block number (${p})",
                           ("a", qc->block_num)("p", block_header::num_from_id(bb.parent.id())) );
               auto qc_claim = qc_claim_t { qc->block_num, qc->qc.is_strong() };
               return bb.parent.is_needed(*qc) ? qc_data_t{ *qc, qc_claim } : qc_data_t{ {},  qc_claim };
            }
         }
         // This only happens when parent block is the IF genesis block or starting from a snapshot.
         // There is no ancestor block which has a QC.
         // Construct a default QC claim.
         return qc_data_t{ {}, bb.parent.core.latest_qc_claim() };
      });
```

libraries/chain/controller.cpp
```diff
@@ -3198,15 +3265,39 @@ struct controller_impl {
...
-               EOS_ASSERT(producer_block_id == ab.id(), block_validate_exception, "Block ID does not match",
+               EOS_ASSERT(producer_block_id == ab.id(), block_validate_exception, "Block ID does not match, ${producer_block_id} != ${validator_block_id}",
                           ("producer_block_id", producer_block_id)("validator_block_id", ab.id()));
             }
 
+            // verify received finality digest in action_mroot is the same as the actual one
+            if constexpr (std::is_same_v<BSP, block_state_ptr>) {
+               // For proper IF blocks that do not have an associated Finality Tree defined,
+               // its finality_mroot is empty
+               digest_type actual_finality_mroot{};
```
greg7mdp, line 3276:
```suggestion
               digest_type actual_finality_mroot;
```

libraries/chain/controller.cpp
```diff
@@ -3198,15 +3265,39 @@ struct controller_impl {
...
+            if constexpr (std::is_same_v<BSP, block_state_legacy_ptr>) {
+               // Store action_receipt_digests in completed_block for
+               // calculating action_mroot in the Savanna way during
+               // Legacy to Savanna transition
+               assert(std::holds_alternative<assembled_block::assembled_block_legacy>(ab.v));
+               auto legacy_ab = std::get<assembled_block::assembled_block_legacy>(ab.v);
+               pending->_block_stage = completed_block{ block_handle{bsp}, legacy_ab.action_receipt_digests };
+            } else {
+               pending->_block_stage = completed_block{ block_handle{bsp} };
+            }
```
greg7mdp, line 3300: This is awkward. What about having  both `assembled_block` versions have a member get_action_receipt_digests()`, so you can replace with:

```suggestion
               pending->_block_stage = completed_block{ block_handle{bsp}, ab.get_action_receipt_digests() };
```

libraries/chain/controller.cpp
```diff
@@ -3198,15 +3265,39 @@ struct controller_impl {
                report_block_header_diff(*b, ab.header());
 
                // this implicitly asserts that all header fields (less the signature) are identical
-               EOS_ASSERT(producer_block_id == ab.id(), block_validate_exception, "Block ID does not match",
+               EOS_ASSERT(producer_block_id == ab.id(), block_validate_exception, "Block ID does not match, ${producer_block_id} != ${validator_block_id}",
                           ("producer_block_id", producer_block_id)("validator_block_id", ab.id()));
             }
 
+            // verify received finality digest in action_mroot is the same as the actual one
+            if constexpr (std::is_same_v<BSP, block_state_ptr>) {
```
greg7mdp, line 3273: Can you do that check in the above `if constexpr` after the `assemble_block(true, extract_qc_data(b), bsp);`>

greg7mdp (changes requested):

linh2931:
> @linh2931 the `block_state::valid` member needs to be saved to the snapshot. Would you like me to add the changes to your branch?

Yes, please go ahead. I can do it too if you are busy.

libraries/chain/controller.cpp
linh2931: In earlier version data from assembled_block was required for the check. Not any more. Can move up to save an `if constexpr` and look nicer. Thanks.

libraries/chain/controller.cpp
linh2931: Thanks.

libraries/chain/block_state.cpp
linh2931: `_savanna` is too long while `_s` is too short. Plus some use `_s` to indicate a struct. I think `_svnn` is a good compromise.

greg7mdp:
> > @linh2931 the `block_state::valid` member needs to be saved to the snapshot. Would you like me to add the changes to your branch?
> 
> Yes, please go ahead. I can do it too if you are busy.

Thanks, I'll do the update tomorrow morning!

libraries/chain/block_state.cpp
greg7mdp: While I agree that `_savanna` is too long, I guess we can agree to disagree on `_svnn` being a good compromise.

linh2931:
> > > @linh2931 the `block_state::valid` member needs to be saved to the snapshot. Would you like me to add the changes to your branch?
> > 
> > 
> > Yes, please go ahead. I can do it too if you are busy.
> 
> Thanks, I'll do the update tomorrow morning!

Thank you for adding it!

libraries/chain/controller.cpp
```diff
@@ -628,12 +633,52 @@ struct building_block {
...
+                  return qc_data_t{ {},  parent.core.latest_qc_claim() };
+               }
+            }
+         }
+
+         // This only happens when parent block is the IF genesis block or starting from snapshot.
+         // There is no ancestor block which has a QC. Construct a default QC claim.
+         return qc_data_t{ {}, parent.core.latest_qc_claim() };
+      });
+   }
```
greg7mdp, line 673: I think these two should be `const` member functions of `block_state`.

heifner: I don't think `block_state` should be aware of `fork_database`.

linh2931: I agree. We should not make `block_state` aware of fork_db for `get_qc_data`. We could move `get_finality_mroot_claim` to `block_state`. But it is only used here. It might be clearer to implement  it along with `get_qc_data`.

greg7mdp: > I don't think `block_state` should be aware of `fork_database`.

otoh, I don't think `controller.cpp` should be 5,000 lines long :-).
Because `fork_database` is the way to resolve the parent id of a `block_state`, not allowing `block_state` to use `fork_database` seriously limits what it can do.

linh2931: In 6.1, we should consider splitting controller.cpp.

greg7mdp: I feel that just splitting it up into two files, but keeping the same organization would not help. I really feel that we should avoid free functions like the ones I pointed to above, and have these be members of the class that makes most sense. I feel this is a more `C++` way of coding, versus having data (struct) definitions and free functions like is done in `C`.

greg7mdp (approved):

libraries/chain/controller.cpp
linh2931: I managed to move the two functions to block_state without exposing forks_db to it.

libraries/chain/include/eosio/chain/block_state.hpp
```diff
@@ -63,7 +95,22 @@ struct block_state : public block_header_state {     // block_header_state provi
...
+   // Returns the root digest of the finality tree associated with the target_block_num
+   // [core.last_final_block_num, block_num]
+   digest_type get_validation_mroot( block_num_type target_block_num ) const;
+
+   // Returns finality_mroot_claim of the current block
+   digest_type get_finality_mroot_claim(const qc_claim_t& qc_claim) const;
+
+   // Returns the qc_data of the most ancestor block having a QC on the branch (including
+   // the current block)
+   qc_data_t get_most_ancestor_qc_data(std::vector<std::shared_ptr<block_state>> branch) const;
```
greg7mdp, line 112: make parameter a `const &`.

linh2931: Thanks! I missed that in a rash.

arhag (changes requested):

libraries/chain/include/eosio/chain/snapshot_detail.hpp
```diff
@@ -114,16 +114,13 @@ namespace eosio::chain::snapshot_detail {
...
-      incremental_merkle_tree                             finality_mtree;
       finalizer_policy_ptr                                active_finalizer_policy;
       proposer_policy_ptr                                 active_proposer_policy;
       flat_map<block_timestamp_type, proposer_policy_ptr> proposer_policies;
       flat_map<uint32_t, finalizer_policy_ptr>            finalizer_policies;
 
       // from block_state
-      using valid_t = uint32_t; // snapshot todo
-      std::optional<valid_t> valid;
+      std::optional<valid_t>                              valid;
```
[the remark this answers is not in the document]
greg7mdp: Yes, I remember you mentioned that it should always be present, and the std::optional was just because the `valid_t` is not be available when `block_state` is initially created. 

I assume that when the snapshot is initiated, and we save the chain head to the snapshot, we can `FC_ASSERT` that the `valid` member is populated, right?

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -58,13 +60,12 @@ struct block_header_state {
...
    // ------ functions -----------------------------------------------------------------
-   // [if todo] https://github.com/AntelopeIO/leap/issues/2080
-   const block_id_type&  id() const { return block_id; }
-   digest_type           compute_finalizer_digest() const { return block_id; };
-   block_timestamp_type  timestamp() const { return header.timestamp; }
-   account_name          producer() const  { return header.producer; }
-   const block_id_type&  previous() const  { return header.previous; }
-   uint32_t              block_num() const { return block_header::num_from_id(previous()) + 1; }
+   const block_id_type&  id()             const { return block_id; }
+   const digest_type&    finality_mroot() const { return header.action_mroot; }
```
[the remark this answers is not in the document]
linh2931: Thanks. Will do that.

libraries/chain/controller.cpp
```diff
@@ -672,66 +671,78 @@ struct building_block {
...
-
-                     if (!qc_data) {
-                        // This only happens when parent block is the IF genesis block or starting from snapshot.
-                        // There is no ancestor block which has a QC. Construct a default QC claim.
-                        qc_data = qc_data_t{ {}, bb.parent.core.latest_qc_claim() };
-                     }
+                  auto branch = fork_db.apply_s<std::vector<block_state_ptr>>([&](const auto& forkdb) {
+                     return forkdb.fetch_branch(bb.parent.id());
                   });
+                  qc_data = bb.parent.get_most_ancestor_qc_data(branch);
```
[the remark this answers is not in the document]
linh2931: So we will move this function back to `controller.cpp` because we don't want to expose forks_db to `block_state`.

libraries/chain/block_header_state.cpp
```diff
@@ -7,10 +7,62 @@
...
+
+   base_digest_data_t base_digest_data {
+      .header                      = header,
+      .core                        = core,
+      .finalizer_policies          = finalizer_policies,
+      .active_proposer_policy      = active_proposer_policy,
+      .proposer_policies           = proposer_policies,
+      .activated_protocol_features = activated_protocol_features
+   };
+   auto base_digest = fc::sha256::hash(base_digest_data);
```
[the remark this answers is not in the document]
linh2931: Thanks for your comments about the two pointers. Will correct them.

I used structs here so that all the data fields to be hashed are clearly spelled out.

By "explicitly serializing", do you mean using `fc:sha256`'s `enc` and `fc::raw::pack`?

libraries/chain/include/eosio/chain/block_header_state.hpp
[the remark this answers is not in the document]
linh2931: Should we name the function `is_proper_svnn_block`?

libraries/chain/block_header_state.cpp
```diff
@@ -156,10 +204,19 @@ block_header_state block_header_state::next(const signed_block_header& h, valida
...
 
    block_header_state_input bhs_input{
-      bb_input,      h.transaction_mroot, h.action_mroot, if_ext.new_proposer_policy, if_ext.new_finalizer_policy,
-      if_ext.qc_claim };
+      bb_input,
+      h.transaction_mroot,
+      if_ext.new_proposer_policy,
+      if_ext.new_finalizer_policy,
+      if_ext.qc_claim,
+      h.action_mroot // for finality_mroot_claim
```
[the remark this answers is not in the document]
linh2931: Done.

libraries/chain/include/eosio/chain/block_header_state.hpp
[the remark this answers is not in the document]
linh2931: Done.

libraries/chain/block_header_state.cpp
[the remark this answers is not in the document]
linh2931: For Proper IF block, is it also true if `h.action_mroot` is empty, then there must not be a Finality Tree Root associated with the block?

libraries/chain/block_header_state.cpp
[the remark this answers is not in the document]
linh2931: Done.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
linh2931: Talked to Areg. His concern was 

> The argument to the function is expected to be a branch that are the ancestor blocks of the block represented by that block state.
> 
> What if the caller doesn’t respect that precondition?

We agreed to move the function back to controller, taking parent as an argument and traversing the ancestor branch inside the function.

libraries/chain/block_header.cpp
```diff
@@ -25,6 +25,12 @@ namespace eosio { namespace chain {
       return result;
    }
 
+   bool block_header::is_proper_svnn_block() const {
+      auto exts = validate_and_extract_header_extensions();
```
arhag, line 29: I worry about doing this every time this function is called, since it isn't cached.

I think this should only do the faster `schedule_version == proper_svnn_schedule_version` check. It can be assumed that if that is the case, then the finality extension must also be present. Though that should be validated when validating the block header along with the other special block header validation rules for Proper IF Blocks such as the confirmed count being zero and the proposer signature being computed on the block ID.

libraries/chain/block_header_state.cpp
```diff
@@ -26,36 +26,39 @@ struct base_and_active_finalizer_policy_digest_data_t {
...
+// compute base_digest explicitly because of pointers involved.
+digest_type block_header_state::compute_base_digest() const {
+   digest_type::encoder enc;
+
+   fc::raw::pack( enc, header );
+   fc::raw::pack( enc, core );
+
+   for (const auto& fp_pair : finalizer_policies) {
+      fc::raw::pack( enc, *fp_pair.second );
+   }
```
arhag, line 38: Can you not just pack the `finalizer_policies`?

libraries/chain/block_header_state.cpp
```diff
@@ -26,36 +26,39 @@ struct base_and_active_finalizer_policy_digest_data_t {
...
+
+   fc::raw::pack( enc, *active_proposer_policy );
+
+   for (const auto& pp_pair : proposer_policies) {
+      fc::raw::pack( enc, *pp_pair.second );
+   }
```
arhag, line 44: Can we not just pack the `proposer_policies`?

libraries/chain/block_header_state.cpp
```diff
@@ -107,8 +110,7 @@ block_header_state block_header_state::next(block_header_state_input& input) con
       // +1 since this is called after the block is built, this will be the active schedule for the next block
       if (it->first.slot <= input.timestamp.slot + 1) {
          result.active_proposer_policy = it->second;
-         result.header.schedule_version = header.schedule_version + 1;
-         result.active_proposer_policy->proposer_schedule.version = result.header.schedule_version;
+         result.header.schedule_version = block_header::proper_svnn_schedule_version;
```
arhag, line 113: Isn't this redundant?

libraries/chain/block_header_state.cpp
```diff
@@ -203,20 +205,35 @@ block_header_state block_header_state::next(const signed_block_header& h, valida
       .new_protocol_feature_activations = std::move(new_protocol_feature_activations)
    };
 
+   digest_type action_mroot; // digest_type default value is empty
```
arhag, line 208: I think value initializing would be more appropriate pattern to follow here:
```cpp
digest_type action_mroot = {};
```

libraries/chain/block_header_state.cpp
```diff
@@ -203,20 +205,35 @@ block_header_state block_header_state::next(const signed_block_header& h, valida
...
+
+   if (h.is_proper_svnn_block()) {
+      // if there is no Finality Tree Root associated with the block,
+      // then this needs to validate that h.action_mroot is the empty digest
+      auto next_core_metadata = core.next_metadata(if_ext.qc_claim);
+      auto no_finality_tree_associated = core.is_genesis_block_num(next_core_metadata.final_on_strong_qc_block_num);
```
arhag, line 214: Could also just use `bool` instead of `auto` to make it more clear to the reader.

libraries/chain/include/eosio/chain/block_header_state.hpp
```diff
@@ -76,7 +75,11 @@ struct block_header_state {
    block_header_state next(block_header_state_input& data) const;
    block_header_state next(const signed_block_header& h, validator_t& validator) const;
 
-   digest_type           compute_finalizer_digest() const;
+   digest_type compute_base_digest() const;
+   digest_type compute_finalizer_digest() const;
```
arhag, line 79: This should be called `compute_finality_digest()`.

arhag (changes requested):

libraries/chain/include/eosio/chain/snapshot_detail.hpp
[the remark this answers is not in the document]
arhag: Yes, we can assert that the `valid` structure exists at the time the snapshot is being produced. It must always be the case.

In the snapshot itself, the `valid` structure must always be present, which is why I say it shouldn't be optional here. It should still be optional within `block_state`. But when loading the root `block_state` from a snapshot, `valid` should always end up populated.

libraries/chain/block_header_state.cpp
arhag: Never mind, I understand why you do not.

However, we should also pack the block number `fp_pair.first` before packing the finalizer policy.

arhag (approved):

unittests/producer_schedule_if_tests.cpp
```diff
@@ -150,13 +147,13 @@ BOOST_FIXTURE_TEST_CASE( proposer_policy_progression_test, validating_tester ) t
...
    // sch1  must become active no later than 2 rounds but sch2 cannot become active yet
-   BOOST_CHECK_EQUAL( control->active_producers().version, 1u );
+      BOOST_CHECK_EQUAL( control->active_producers().version, 1u );
    BOOST_CHECK_EQUAL( true, compare_schedules( sch1, control->active_producers() ) );
 
    produce_blocks(config::producer_repetitions);
 
    // sch2 becomes active
-   BOOST_CHECK_EQUAL( control->active_producers().version, 2u );
+      BOOST_CHECK_EQUAL( control->active_producers().version, 2u );
```
heifner, line 156: Fix indentation

libraries/chain/block_header.cpp
```diff
@@ -25,6 +25,14 @@ namespace eosio { namespace chain {
       return result;
    }
 
+   bool block_header::is_proper_svnn_block() const {
+      // We don't check whether finality extension exists here for performance reason.
+      // When block header is validated in block_header_state's next(),
+      // it is already validate if schedule_version == proper_svnn_schedule_version,
+      // finality extension must exist.
+      return ( schedule_version ==  proper_svnn_schedule_version );
```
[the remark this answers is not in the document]
linh2931: `is_proper_svnn_block()` was initially more complicated; that is why it was placed in .cpp. But now it is better to be in .hpp. And after I moved `proper_svnn_schedule_version` next to it, the code became more clearer.

Thanks.

libraries/chain/block_header_state.cpp
```diff
@@ -59,8 +117,6 @@ block_header_state block_header_state::next(block_header_state_input& input) con
       // +1 since this is called after the block is built, this will be the active schedule for the next block
       if (it->first.slot <= input.timestamp.slot + 1) {
          result.active_proposer_policy = it->second;
-         result.header.schedule_version = header.schedule_version + 1;
-         result.active_proposer_policy->proposer_schedule.version = result.header.schedule_version;
```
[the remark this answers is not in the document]
linh2931: Does https://github.com/AntelopeIO/leap/blob/7f3d5e6442d8d648f967a39a14f7c249227cb74a/libraries/chain/controller.cpp#L481-L484 ensure the version number not being skipped?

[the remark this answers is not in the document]
heifner: No, that doesn't help. We need to make sure that if a proposer policy is scheduled for the same `proposer_policy.active_time` that it replaces the current one there but doesn't increment the schedule version. I think we will need to change `controller::set_proposed_producers` to check if there is an existing one that would have the same `active_time` and replace it and use its version. We should return the correct version to the contract.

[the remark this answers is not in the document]
heifner: Feel free to create a follow-on issue for this.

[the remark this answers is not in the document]
linh2931: https://github.com/AntelopeIO/leap/issues/2313

Thanks.

heifner (approved):

PR #2284, opened 2024-03-06
IF: Change Fork_db root to store block state

### LINKED ISSUE #2283: IF: Change Fork_db root to store block state

greg7mdp:
Resolves #2283 .

Removes `fork_database::get_block_header` and simplifies `create_block_state_i`.

The only difference between `fork_database::get_block_header` and `fork_database::get_block` was that the former checked the `root`. It was also relying on the fact that `block_state` is `public block_header_state` to return the same pointer downcasted.

I have added an optional `check_root` parameter to `fork_database::get_block` instead.

This allowed to make `create_block_state_i` a template instead of having two separate implementations.

New disk format and compatibility mode will be addressed in #2285 .

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3351,7 +3351,7 @@ struct controller_impl {
             auto existing = forkdb.get_block( id );
             EOS_ASSERT( !existing, fork_database_exception, "we already know about this block: ${id}", ("id", id) );
 
-            auto prev = forkdb.get_block( b->previous, true );
+            auto prev = forkdb.get_block( b->previous, check_root_t::yes );
             EOS_ASSERT( prev, unlinkable_block_exception,
                         "unlinkable block ${id} previous ${p}", ("id", id)("p", b->previous) );
 
@@ -3372,7 +3372,7 @@ struct controller_impl {
          EOS_ASSERT( !existing, fork_database_exception, "we already know about this block: ${id}", ("id", id) );
 
          // previous not found could mean that previous block not applied yet
-         auto prev = forkdb.get_block( b->previous, true );
+         auto prev = forkdb.get_block( b->previous, check_root_t::yes );
          if( !prev ) return {};
 
          return create_block_state_i( id, b, *prev );
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1019,7 +1019,7 @@ struct controller_impl {
       fork_db.apply<void>([&](auto& forkdb) {
          std::visit([&](const auto& bsp) {
             if constexpr (std::is_same_v<std::decay_t<decltype(bsp)>, std::decay_t<decltype(forkdb.head())>>)
-               forkdb.reset_root(*bsp);
+               forkdb.reset_root(bsp);
          }, chain_head.internal());
       });
    }
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3310,43 +3310,23 @@ struct controller_impl {
    }
 
    // thread safe, expected to be called from thread other than the main thread
-   block_handle create_block_state_i( const block_id_type& id, const signed_block_ptr& b, const block_header_state& prev ) {
-      // Verify claim made by instant_finality_extension in block header extension and
-      // quorum_certificate_extension in block extension are valid.
-      // This is the only place the evaluation is done.
-      verify_qc_claim(id, b, prev);
+   template<class BS>
+   block_handle create_block_state_i( const block_id_type& id, const signed_block_ptr& b, const BS& prev ) {
+      constexpr bool savanna_mode = std::is_same_v<typename std::decay_t<BS>, block_state>;
+      if constexpr (savanna_mode) {
+         // Verify claim made by instant_finality_extension in block header extension and
+         // quorum_certificate_extension in block extension are valid.
+         // This is the only place the evaluation is done.
+         verify_qc_claim(id, b, prev);
+      }
 
-      auto trx_mroot = calculate_trx_merkle( b->transactions, true );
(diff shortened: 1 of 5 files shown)


libraries/chain/fork_database.cpp
```diff
@@ -18,6 +18,8 @@ namespace eosio::chain {
    /**
     * History:
     * Version 1: initial version of the new refactored fork database portable format
+    * Version 2: Savanna version, store either `block_state`, `block_state_legacy` or both versions,
+    *            root is full `block_state`, not just the header.
```
heifner, line 22: You didn't update `max_supported_version`. As implemented, it looks like this implementation can read version 1. However, a version 1 written by this implementation will not be readable by leap v5 or before. Not really sure we really need to support downgrading a node back to pre leap v6. A version dump would allow for a better error message in pre leap v6 (`Unsupported version of fork database file`).

libraries/chain/controller.cpp
```diff
@@ -3371,7 +3351,7 @@ struct controller_impl {
             auto existing = forkdb.get_block( id );
             EOS_ASSERT( !existing, fork_database_exception, "we already know about this block: ${id}", ("id", id) );
 
-            auto prev = forkdb.get_block_header( b->previous );
+            auto prev = forkdb.get_block( b->previous, true );
```
linh2931, line 3354: Do you want to define an enum instead of `true` or `false`?

linh2931 (approved):

libraries/chain/fork_database.cpp
```diff
@@ -298,16 +299,15 @@ namespace eosio::chain {
...
    }
 
    template<class BSP>
-   void fork_database_impl<BSP>::reset_root_impl( const bhs_t& root_bhs ) {
+   void fork_database_impl<BSP>::reset_root_impl( const bsp_t& root_bsp ) {
       index.clear();
-      root = std::make_shared<bs_t>();
-      static_cast<bhs_t&>(*root) = root_bhs;
+      root = root_bsp;
       bs_accessor_t::set_valid(*root, true);
```
heifner, line 311: Should only set valid if this was a version 1 `root_bsp` from `open_impl`. Otherwise, it should already be `valid`. Could assert it is valid and set the valid in `open_impl` if a `block_header_state` was read instead of a `block_state`.

libraries/chain/fork_database.cpp
heifner: Note I'm working on #2141 which has the test coverage fork database writing/reading. Would really like this tested with the test cases that are being restored in #2141.

heifner (changes requested):

libraries/chain/fork_database.cpp
greg7mdp: I'm not quite sure why, but it appears that `validated` is not set on root after loading a fork_db when running tests with this PR's code, se the `set_valid` is necessary.

libraries/chain/controller.cpp
greg7mdp: done

libraries/chain/fork_database.cpp
heifner: Set valid is only necessary when loading a version 1 fork database which does not have all the `block_state` (doesn't read in `validated`.). I think you should move the set valid into `open_impl` and assert in `reset_root_impl`.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Simplify fork db consolidating on `fork_database::get_block`
Note:end

libraries/chain/fork_database.cpp
greg7mdp: > Set valid is only necessary when loading a version 1 fork database which does not have all the `block_state` (doesn't read in `validated`.). I think you should move the set valid into `open_impl` and assert in `reset_root_impl`.

I tried that but many tests fail when I do that.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -11,6 +11,7 @@ namespace eosio::chain {
    using block_branch_t = std::vector<signed_block_ptr>;
    enum class mark_valid_t { no, yes };
    enum class ignore_duplicate_t { no, yes };
+   enum class check_root_t { no, yes };
```
heifner, line 14: Minor thing, but seems like this should be called `include_root_t`

libraries/chain/fork_database.cpp
heifner: Glancing at the tests, I think I know why. I can address that in #2141

heifner (approved):

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: That's actually the named I used first, but it felt wrong for some reason and I prefer `check_root_t`. If you want to change it later I'm fine with it.

libraries/chain/fork_database.cpp
greg7mdp: I'm curious to know what the issue could be?

heifner: Mock block_state rely on this to set valid. Those need to set valid outside.

greg7mdp: > Mock block_state rely on this to set valid. Those need to set valid outside.

Thanks, that makes sense.

PR #2287, opened 2024-03-06
IF: Support replay over instant finality transition

### LINKED ISSUE #2141: IF: replay from before instant-finality transition

heifner:
- Delay opening fork database until after replay of the block log.
- Change `complete_block` to hold a `block_handle`. This removes update code and allows access to block_handle methods like `id()` without using fork database variant.

Resolves #2141

### DIFF
--- a/libraries/chain/block_header_state_legacy.cpp
+++ b/libraries/chain/block_header_state_legacy.cpp
@@ -423,7 +423,7 @@ namespace eosio::chain {
       std::tie(is_satisfied, relevant_sig_count) = producer_authority::keys_satisfy_and_relevant(keys, valid_block_signing_authority);
 
       EOS_ASSERT(relevant_sig_count == keys.size(), wrong_signing_key,
-                 "block signed by unexpected key",
+                 "block signed by unexpected key: ${signing_keys}, expected: ${authority}. ${c} != ${s}",
                  ("signing_keys", keys)("authority", valid_block_signing_authority));
 
       EOS_ASSERT(is_satisfied, wrong_signing_key,
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -1124,11 +1124,6 @@ struct controller_impl {
     my_finalizers{ .t_startup = fc::time_point::now(), .persist_file_path = cfg.finalizers_dir / "safety.dat" },
     wasmif( conf.wasm_runtime, conf.eosvmoc_tierup, db, conf.state_dir, conf.eosvmoc_config, !conf.profile_accounts.empty() )
    {
-      fork_db.open([this](block_timestamp_type timestamp, const flat_set<digest_type>& cur_features,
-                          const vector<digest_type>& new_features) {
-         check_protocol_features(timestamp, cur_features, new_features);
-      });
-
       thread_pool.start( cfg.thread_pool_size, [this]( const fc::exception& e ) {
          elog( "Exception in chain thread pool, exiting: ${e}", ("e", e.to_detail_string()) );
          if( shutdown ) shutdown();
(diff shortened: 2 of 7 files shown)


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Support replay over instant finality transition by delaying consumption of the fork database until after all blocks are replayed.
Note:end

libraries/chain/block_header_state_legacy.cpp
```diff
@@ -423,7 +423,7 @@ namespace eosio::chain {
       std::tie(is_satisfied, relevant_sig_count) = producer_authority::keys_satisfy_and_relevant(keys, valid_block_signing_authority);
 
       EOS_ASSERT(relevant_sig_count == keys.size(), wrong_signing_key,
-                 "block signed by unexpected key",
+                 "block signed by unexpected key: ${signing_keys}, expected: ${authority}. ${c} != ${s}",
```
greg7mdp, line 426: Looks like `${c}` and `${s}` are not provided.

libraries/chain/controller.cpp
```diff
@@ -1290,97 +1302,147 @@ struct controller_impl {
...
+         except_ptr = replay_block_log(check_shutdown);
+      } else {
+         ilog( "no block log found" );
+      }
+
+      try {
+         if (startup != startup_t::existing_state)
+           open_fork_db();
+      } catch (const fc::exception& e) {
+         elog( "Update to open fork database, continueing without reversible blocks: ${e}", ("e", e));
```
greg7mdp, line 1394:
```suggestion
         elog( "Unable to open fork database, continuing without reversible blocks: ${e}", ("e", e));
```

greg7mdp (approved):

libraries/chain/block_header_state_legacy.cpp
heifner: Fixed.

libraries/chain/controller.cpp

libraries/chain/fork_database.cpp
```diff
@@ -666,15 +670,15 @@ namespace eosio::chain {
 
    template<class BSP>
    BSP fork_database_t<BSP>::get_block(const block_id_type& id,
-                                       check_root_t check_root /* = check_root_t::no */) const {
+                                       include_root_t include_root /* = include_root_t::no */) const {
```
linh2931, line 673: Remove the `/* = include_root_t::no */`

libraries/chain/fork_database.cpp
```diff
@@ -666,15 +670,15 @@ namespace eosio::chain {
...
       std::lock_guard g( my->mtx );
-      return my->get_block_impl(id, check_root);
+      return my->get_block_impl(id, include_root);
    }
 
    template<class BSP>
    BSP fork_database_impl<BSP>::get_block_impl(const block_id_type& id,
-                                               check_root_t check_root /* = check_root_t::no */) const {
-      if( check_root == check_root_t::yes && root->id() == id ) {
+                                               include_root_t include_root /* = include_root_t::no */) const {
```
linh2931, line 680: Remove the /* = include_root_t::no */

linh2931 (approved):

libraries/chain/fork_database.cpp
heifner: Why? It provides doc on what the default is.

libraries/chain/fork_database.cpp
heifner: Why? It provides doc on what the default is.

libraries/chain/fork_database.cpp
linh2931: Oh, I thought `include_root` was not a parameter with a default value. Don't make any changes for this PR. In general, should we stay away from default parameters?

heifner: In general, I'm not a fan of default parameters. Here I was just changing the name.

PR #2290, opened 2024-03-06
IF: Consider voting immediately if final on strong qc is validated

### LINKED ISSUE #2125: IF: Update fork-choice rule for fork database

heifner:
Consider voting immediately on a block if its final on strong qc block reference has been validated. The block does not necessarily need to be on our head branch as long as we have validated it.

The voting uncovered a number of problems that are fixed by this PR:
- The `parent.is_needed()` check of best qc in `assemble_block` was not correctly comparing the qc claim. Add a `to_qc_claim` to qc to make creating a qc claim from a qc easy.
- The fork switch logic for accepted blocks (non-applied blocks) was incorrect. It needed to check `pending_head` not `head`.
  - Most of the rest of the fixes are because fork switches are now possible as blocks are added to the fork database directly when received after header validation. This means that fork switches can happen much more rapidly as blocks are received. This is true in legacy as well as savanna. As such, this PR can also be thought of as part 2 of #2274.
- Existing `net_plugin` issues uncovered by faster switching enabled by the fork switch fix.
  - Use the provided `lib` of `irreversible` signal as the fork db is not updated until after the signal is emitted.
  - Received blocks that have already been added to the dispatcher block list are applied. Fix the call to `sync_recv_block` to pass `true` instead of `false`. The fast fork switching in tests exposed this issue. The blocks would be added to the fork database and fork switched out before they could be applied.
- Numerous fixes to `trx_finality_status_forked_test.py`. It is very difficult to catch a transaction in the forked-out state as fork switching is happing very rapidly with the fork switching as a result of blocks added to the fork database directly. This is true not just for instant finality but also for legacy consensus. 

Also:
- Additional debug logging was added to help diagnosis issues.
- Renamed `proposal_ref` to `block_ref`
- Renamed `proosal_id` to `block_id`

Resolves #2125


linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -3473,14 +3473,13 @@ struct controller_impl {
    void consider_voting(const block_state_ptr& bsp) {
       // 1. Get the `core.final_on_strong_qc_block_num` for the block you are considering to vote on and use that to find the actual block ID
       //    of the ancestor block that has that block number.
-      // 2. If that block ID is not an ancestor of the current head block, then do not vote for that block.
+      // 2. If that block ID is for a non validated block, then do not vote for that block.
       // 3. Otherwise, consider voting for that block according to the decide_vote rules.
 
       if (bsp->core.final_on_strong_qc_block_num > 0) {
```
greg7mdp, line 3479:
```suggestion
      if (bsp->is_valid() && bsp->core.final_on_strong_qc_block_num > 0) {
```

libraries/chain/controller.cpp
```diff
@@ -3473,14 +3473,13 @@ struct controller_impl {
...
          const auto& final_on_strong_qc_block_ref = bsp->core.get_block_reference(bsp->core.final_on_strong_qc_block_num);
-         auto final = fetch_bsp_on_head_branch_by_num(final_on_strong_qc_block_ref.block_num());
-         if (final) {
-            assert(final->is_valid()); // if found on head branch then it must be validated
+         auto final = fetch_bsp_on_branch_by_num(final_on_strong_qc_block_ref.block_id, bsp->core.final_on_strong_qc_block_num);
```
greg7mdp, line 3481: why `fetch_bsp_on_branch_by_num` and not `fork_db_fetch_block_by_id(final_on_strong_qc_block_ref.block_id)`?

greg7mdp (changes requested):

libraries/chain/controller.cpp
heifner: Because I want `block_state` not `signed_block`. Could add a `fetch_bsp(block_id)` for this, but figured it was not worth it.

libraries/chain/controller.cpp
heifner: No, don't need the block we vote on to be valid. It will not be valid at this point because we just received the block. The point of this is to vote early before we validate the block.

greg7mdp: I can't remember the reasoning as to why we don't need a block to be validated to vote on it? Don't we run the risk that is a bad block is gossipped, many finalizers will vote on it and it will get a QC even though it doesn't validate?

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Consider voting immediately on a block if its final on strong qc block reference has been validated.
Note:end

libraries/chain/controller.cpp
heifner: We don't vote to make something final unless it has been validated. So ultimately we do not contribute to finality on a block unless we have validated it. The voting on non-validated blocks just means we have verified the block header and we anticipate it being a valid block.

greg7mdp: > we anticipate it being a valid block.

which means we are not sure, right? So we would have voted to make an unvalidated block final.

heifner: We have not voted to make an unvalidated block final. We have voted to lock on a block that has not been validated. We  can switch to voting on a diff branch later. Although, I think the test failures on this PR may indicate that switching to a different branch to vote later is not working.

greg7mdp: Thanks for explaining. I finally get the intent of this PR. I am not convinced that it is a wise change to make (have we proven that it would work in all cases?). If no more QCs are produced I don't think we'll regain liveness or safety.

libraries/chain/controller.cpp
```diff
@@ -3447,15 +3454,22 @@ struct controller_impl {
...
       if (claimed->valid_qc && (claimed->valid_qc->is_strong() || received_qc.is_weak())) {
+         dlog("qc not better, claimed->valid: ${qbn} ${qid}, strong=${s}, received: ${rqc}, for block ${bn} ${id}",
+              ("qbn", claimed->block_num())("qid", claimed->id())("s", claimed->valid_qc->is_strong())
+              ("rqc", qc_ext.qc.to_qc_claim())("bn", bsp_in->block_num())("id", bsp_in->id()));
          return;
       }
 
       // Save the QC. This is safe as the function is called by push_block & accept_block from application thread.
+      dlog("setting valid qc: ${rqc} into claimed block ${bn} : ${id}",
+           ("rqc", qc_ext.qc.to_qc_claim())("rs", received_qc.is_strong())("bn", claimed->block_num())("id", claimed->id()));
```
greg7mdp, line 3472: "rs" is not used.

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -60,6 +61,10 @@ namespace eosio::chain {
    struct quorum_certificate {
       uint32_t                 block_num;
       valid_quorum_certificate qc;
+
+      qc_claim_t to_qc_claim() const {
```
greg7mdp, line 65: I'm not crazy about including `finality_core.hpp` just to define this function, but it would be equally awkward to include `hotstuff.hpp` into `finality_core.hpp`, so I don't really have a better solution.

greg7mdp (approved):

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
linh2931: This file should be eventually renamed and relocated to the chain include directory.

linh2931 (approved):

PR #2291, opened 2024-03-07
[4.0] fix appbase destruction order: destroy executor after destroying plugins

spoonincode:
See AntelopeIO/appbase#31 for a description of the problem since it's easier to inline code on that PR.

I've sent this against 4.0 to start with, since that's where this problem was first introduced, but it might be a stretch for inclusion in 4.0 since I'm not sure it quite meets the threshold for a bug fix to the "-1" release. imo it meets the threshold for 5.0 though. Granted, crashes this late typically don't result in any sort of data file corruption. I tried to make the change as surgical as possible.


ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Re-order operations when destroying plugins, to prevent crashes.
Note:end

greg7mdp (approved):

spoonincode:
I no longer feel comfortable sending this to a release branch because I've reminded myself of all sorts of shutdown sprinkles that play lifetime games, such as
https://github.com/AntelopeIO/leap/blob/968a0fa0d0aa5c36ea2f2ed3eb4469068a9afd68/plugins/net_plugin/net_plugin.cpp#L4393
I'll be integrating the change on a PR toward main instead

PR #2298, opened 2024-03-11
grab bag of small UB fixes

spoonincode:
Brings in,
AntelopeIO/abieos#25
AntelopeIO/appbase#31
AntelopeIO/bn256#20
AntelopeIO/eos-vm#22
along with some other noted changes below.

### DIFF
--- a/libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/config.hpp
+++ b/libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/config.hpp
@@ -9,6 +9,10 @@
 
 #include <sys/resource.h>
 
+#ifndef __has_feature
+#define __has_feature(x) 0
+#endif
+
 namespace eosio { namespace chain { namespace eosvmoc {
 
 struct config {
@@ -20,7 +24,11 @@ struct config {
    // libtester disables the limits in all tests, except enforces the limits
    // in the tests in unittests/eosvmoc_limits_tests.cpp.
    std::optional<rlim_t>   cpu_limit {20u};
+#if __has_feature(undefined_behavior_sanitizer) || __has_feature(address_sanitizer)
+   std::optional<rlim_t>   vm_limit; // UBSAN & ASAN can add massive virtual memory usage; don't enforce vm limits when either of them are enabled
+#else
    std::optional<rlim_t>   vm_limit  {512u*1024u*1024u};
+#endif
    std::optional<uint64_t> stack_size_limit {16u*1024u};
    std::optional<size_t>   generated_code_size_limit {16u*1024u*1024u};
 };
--- a/libraries/chain/webassembly/runtimes/eos-vm-oc/memory.cpp
+++ b/libraries/chain/webassembly/runtimes/eos-vm-oc/memory.cpp
@@ -40,7 +40,7 @@ memory::memory(uint64_t sliced_pages) {
    uintptr_t* const intrinsic_jump_table = reinterpret_cast<uintptr_t* const>(zeropage_base - first_intrinsic_offset);
    const intrinsic_map_t& intrinsics = get_intrinsic_map();
    for(const auto& intrinsic : intrinsics)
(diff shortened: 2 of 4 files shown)


libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/config.hpp
```diff
@@ -20,7 +24,11 @@ struct config {
    // libtester disables the limits in all tests, except enforces the limits
    // in the tests in unittests/eosvmoc_limits_tests.cpp.
    std::optional<rlim_t>   cpu_limit {20u};
+#if __has_feature(undefined_behavior_sanitizer) || __has_feature(address_sanitizer)
+   std::optional<rlim_t>   vm_limit; // UBSAN & ASAN can add massive virtual memory usage; don't enforce vm limits when either of them are enabled
```
spoonincode, line 28: Don't need this for tests to pass: OC in unit tests always has `vm_limit` disabled now (except for the one modified test case), and python integration tests never use OC. But I'd like to use OC + ASAN/UBSAN with nodeos. Such usage is probably "developer enough" where simply disabling this limit is safe when configured that way; i.e. won't trap unsuspecting users. Alternatively could guard this disablement behind `EOSVMOC_ENABLE_DEVELOPER_OPTIONS` but ultimately I want to remove that option.

libraries/chain/webassembly/runtimes/eos-vm-oc/memory.cpp
```diff
@@ -40,7 +40,7 @@ memory::memory(uint64_t sliced_pages) {
    uintptr_t* const intrinsic_jump_table = reinterpret_cast<uintptr_t* const>(zeropage_base - first_intrinsic_offset);
    const intrinsic_map_t& intrinsics = get_intrinsic_map();
    for(const auto& intrinsic : intrinsics)
-      intrinsic_jump_table[-intrinsic.second.ordinal] = (uintptr_t)intrinsic.second.function_ptr;
+      intrinsic_jump_table[-(int)intrinsic.second.ordinal] = (uintptr_t)intrinsic.second.function_ptr;
```
spoonincode, line 43: This is an interesting one. The error here gets reported somewhat confusingly as `addition of unsigned offset to 0x7aca262015f8 overflowed to 0x7aca26201570`. I believe what is occurring is that `ordinal` is a `size_t` thus `uint64_t`, and negating it is still a `uint64_t`: now a really large `uint64_t`. But pointer arithmetic is defined on integer math; `ptrdiff_t` is a signed integer for example. So I think this is causing some sort of wrap around on a large signed integer. Casting `ordinal` to an `int` before the negation ensures it remains a small value `int` before pointer arithmetic. I'd be curious on other takes though.

spoonincode:
whoops, I pushed b47a64c here to the wrong branch; I wanted to have a separate PR for that. Sorry for force pushing it away...

libraries/chain/webassembly/runtimes/eos-vm-oc/memory.cpp
spoonincode: `ordinal` is the host function index over in,
https://github.com/AntelopeIO/leap/blob/968a0fa0d0aa5c36ea2f2ed3eb4469068a9afd68/libraries/chain/include/eosio/chain/webassembly/eos-vm-oc/intrinsic_mapping.hpp#L18-L19
so `int` will always be plenty large enough (we won't have 2<sup>31</sup> host functions). My initial impression of `intptr_t` is that its purpose/connotation isn't really for indexes in to an array.

I do wonder if `ordinal` should just be an `int` instead; I don't think it'd be too hard to change it.

ericpassmore:
Note:start
group: CLEANCODE
category: INTERNALS
summary: Fixes several undefined behavior (UB) bugs. 
Note: end

heifner (approved):

greg7mdp (approved):

PR #2299, opened 2024-03-11
fix `net_plugin_impl` destruction order: destroy thread_pool's executor after users of it

spoonincode:
`net_plugin_impl` currently destroys its `named_thread_pool` before a number of users of it such as `expire_timer`. Those users can, and do, access their executor on their destruction which in this case is accessing an already destroyed executor.

Move the `named_thread_pool` up to the top so it's destroyed last. A nice side effect of this is being able to clean up a number of the `unique_ptr`s to just being a plain member.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -444,10 +444,13 @@ namespace eosio {
    class net_plugin_impl : public std::enable_shared_from_this<net_plugin_impl>,
                            public auto_bp_peering::bp_connection_manager<net_plugin_impl, connection> {
     public:
+      uint16_t                                    thread_pool_size = 4;
+      eosio::chain::named_thread_pool<struct net> thread_pool;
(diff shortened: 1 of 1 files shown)


plugins/net_plugin/net_plugin.cpp
```diff
@@ -3178,18 +3178,6 @@ namespace eosio {
...
-            fc::lock_guard g( expire_timer_mtx );
-            if( expire_timer )
-               expire_timer->cancel();
-         }
-         {
-            fc::lock_guard g( keepalive_timer_mtx );
-            if( keepalive_timer )
-               keepalive_timer->cancel();
-         }
-
```
spoonincode, line 3192: Didn't see a reason for these cancellations. The timers will effectively be stopped once the named_thread_pool is stopped.

spoonincode:
goofed up some branching & PRing; see replacement PR

PR #2301, opened 2024-03-11
Use `assert` instead of `FC_ASSERT` for conditions that are not progr…

greg7mdp:
…amatically possible.

In the sense that a user of the class cannot trigger these asserts unless there is a bug in `variant_object.[cpp,hpp]`

### DIFF
--- a/libraries/libfc/src/variant_object.cpp
+++ b/libraries/libfc/src/variant_object.cpp
@@ -9,7 +9,7 @@ namespace fc
 
    variant_object::entry::entry() {}
    variant_object::entry::entry( std::string k, variant v ) : _key(fc::move(k)),_value(fc::move(v)) {}
-   variant_object::entry::entry( entry&& e ) : _key(fc::move(e._key)),_value(fc::move(e._value)) {}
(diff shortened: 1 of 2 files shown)


libraries/libfc/src/variant_object.cpp
```diff
@@ -20,8 +20,7 @@ namespace fc
       }
       return *this;
    }
-   variant_object::entry& variant_object::entry::operator=( variant_object::entry&& e )
-   {
+   variant_object::entry& variant_object::entry::operator=( variant_object::entry&& e ) noexcept    {
```
heifner, line 23: spacing

heifner (approved):

spoonincode (approved):

libraries/libfc/src/variant_object.cpp
greg7mdp: fixed

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: In some cases use assert instead of FC_ASSERT.
Note:end

heifner (approved):

PR #2303, opened 2024-03-12
IF: Include `block_state::valid` in the snapshot.

greg7mdp:


greg7mdp:
@linh2931 this is a PR adding the snapshot fix into your branch.

heifner (approved):

linh2931 (approved):
Thanks. No changes to the tests?

greg7mdp:
> Thanks. No changes to the tests?

No change. I guess we don't have a test that uses the new `valid` member yet.

heifner:
> > Thanks. No changes to the tests?
> 
> No change. I guess we don't have a test that uses the new `valid` member yet.

Don't you need to udpate:  `unittests/snapshots/snap_v7.bin.*`

heifner:
> > Don't you need to udpate: `unittests/snapshots/snap_v7.bin.*`
> 
> I didn't have to. The test that uses those files (`test_compatible_versions` in `unittests`) doesn't run in IF mode I guess, so it must save the `block_state_legacy` in the variant.

That is a test we certainly want to run in both modes.

greg7mdp:
> That is a test we certainly want to run in both modes.

Hum, not sure about this particular one, it is meant to verify that we can load old snapshot versions.

PR #2304, opened 2024-03-12
add CI "platform" running UBSAN

spoonincode:
While there are still a few remaining violations (see comments below), I think we're in a good spot to enable UBSAN in CI builds. UBSAN is expected to generate no false positives so requiring it to pass is reasonable.

An example run with failures is here,
https://github.com/AntelopeIO/leap/actions/runs/8252261490

### DIFF
--- a/.cicd/platforms/ubsan.Dockerfile
+++ b/.cicd/platforms/ubsan.Dockerfile
@@ -1,0 +1,42 @@
+# syntax=docker/dockerfile:1
+FROM ubuntu:jammy
+ENV TZ="America/New_York"
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get upgrade -y && \
+    apt-get install -y build-essential      \
+                       cmake                \
+                       git                  \
+                       jq                   \
+                       libcurl4-openssl-dev \
+                       libgmp-dev           \
+                       llvm-11-dev          \
+                       lsb-release          \
+                       ninja-build          \
+                       python3-numpy        \
+                       software-properties-common \
+                       file                 \
+                       wget                 \
+                       zlib1g-dev           \
+                       zstd
+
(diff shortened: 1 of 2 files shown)


.cicd/platforms/ubsan.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+                       lsb-release          \
+                       ninja-build          \
+                       python3-numpy        \
+                       software-properties-common \
+                       file                 \
+                       wget                 \
+                       zlib1g-dev           \
+                       zstd
+
+RUN yes | bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)" llvm.sh 18
```
spoonincode, line 22: Use the latest clang for these builds; presumably it has the best & most reliable UBSAN yet.

.cicd/platforms/ubsan.Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+
+#make sure no confusion on what llvm library leap's cmake should pick up on
+RUN rm -rf /usr/lib/llvm-18/lib/cmake
+
+COPY <<-EOF /ubsan.supp
+  vptr:wasm_eosio_validation.hpp
+  vptr:wasm_eosio_injection.hpp
+EOF
```
spoonincode, line 30: EOS VM's WASM parser seems to do a number of invalid casts, I believe casting between `something<Foo>` and `something<Bar>` when `something<T>` is not derived from any sort of base class. Resolving this might not be straightforward; and it's not clear if we may replace EOS VM in the mid/long term; so suppress these errors for now.

greg7mdp (approved):

heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: CHORE
summary: Enable CI tests to check for undefined behavior.
Note:end

PR #2306, opened 2024-03-12
IF: Support Leap 5 Fork Database disk format and new Savanna disk format.

### LINKED ISSUE #2285: IF: Support Leap 5 Fork Database disk format and new Savanna disk format.
The fork database changed for Savanna.

It can now be either a `block_state_legacy` and/or a `block_state` container, and the `root` stores a `block_state` instead of a `block_header_state`.

We need to support a new file format which will store either one fork_db (either legacy or not) or two fork_ds (as will exist during the transition).

Leap 5 file format stored only the header part of the root `block_state_legacy`. This was changed in the `hotstuff_integration` branch (see https://github.com/AntelopeIO/leap/pull/2284).

Test `tests/ship_streamer_if_test` should also be re-enabled (uncomment in CMakeLists.txt).

greg7mdp:
Resolves #2285.

This adds support for a new `fork_database` persistent file version (`v2`), which can store two versions of the `fork_database` (the `legacy` version storing  `block_state_legacy` pointers, and the `Savanna` version storing `block_state` pointers.

In addition, the `in_use` variable indicating which version is currently in use is stored as well.

Old version (`v1`) which stores only the `legacy` version of  `fork_database`, can still be read.

Re-enabled  `ship_streamer_if_test` which now passes.


libraries/chain/fork_database.cpp
```diff
@@ -712,7 +670,33 @@ namespace eosio::chain {
...
 
    void fork_database::close() {
-      apply<void>([&](auto& forkdb) { forkdb.close(data_dir / config::forkdb_filename); });
+      auto fork_db_file {data_dir / config::forkdb_filename};
+      bool legacy_valid  = fork_db_l && fork_db_l->in_valid_state();
+      bool savanna_valid = fork_db_s && fork_db_s->in_valid_state();
+
+      // check that fork_dbs are in a consistent state
+      if (!legacy_valid && !savanna_valid) {
+         elog( "fork_database is in a bad state when closing; not writing out '${filename}'", ("filename", fork_db_file) );
```
linh2931, line 679: You can log values of `legacy_valid` and `savanna_valid` so it is easier to tell which one causing the problem when trouble shooting.

tests/CMakeLists.txt
```diff
@@ -151,8 +151,8 @@ set_property(TEST ship_if_test PROPERTY LABELS nonparallelizable_tests)
 add_test(NAME ship_streamer_test COMMAND tests/ship_streamer_test.py -v --num-clients 10 ${UNSHARE} WORKING_DIRECTORY ${CMAKE_BINARY_DIR})
 set_property(TEST ship_streamer_test PROPERTY LABELS long_running_tests)
 # uncomment after https://github.com/AntelopeIO/leap/issues/2285 implemented
```
linh2931, line 153: Remove this comment.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -136,8 +136,12 @@ namespace eosio::chain {
     * All methods assert until open() is closed.
     */
    class fork_database {
+      static constexpr uint32_t magic_number = 0x30510FDB;
```
linh2931, line 139: Only one magic number for both?

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -136,8 +136,12 @@ namespace eosio::chain {
...
+
+      enum class in_use_t : uint32_t { legacy, savanna, both };
```
linh2931, line 141: maybe `legacy_only`, `savanna_only`, `both`?

libraries/chain/fork_database.cpp
greg7mdp: done

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: yes, different versions. There is a single file.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: I don't think it is any clearer, just more verbose.

libraries/chain/controller.cpp
```diff
@@ -139,6 +139,19 @@ R apply(const block_handle& bh, F&& f) {
...
+   if constexpr (std::is_same_v<void, R>)
+      std::visit<void>(overloaded{[&](const block_state_legacy_ptr& head) { std::forward<F_L>(f_l)(head); },
+                                  [&](const block_state_ptr& head)        { std::forward<F_S>(f_s)(head); }
+                       }, bh.internal());
+   else
+      return std::visit<R>(overloaded{[&](const block_state_legacy_ptr& head) -> R { return std::forward<F_L>(f_l)(head); },
+                                      [&](const block_state_ptr& head) -> R        { return std::forward<F_S>(f_s)(head); }
+                           }, bh.internal());
+}
+
```
heifner, line 154: This does not appear to be used.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include <fc/io/cfile.hpp>
```
heifner, line 2: Move below other includes. Consider forward declare so including the header can be avoided.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -1,4 +1,5 @@
...
 #include <eosio/chain/block_state_legacy.hpp>
 #include <eosio/chain/block_state.hpp>
 #include <eosio/chain/block_handle.hpp>
```
heifner, line 5: I thought I had removed this include and the `log_fork_comparison`. I think it would be cleaner for `fork_database` to not know about the internals of `block_handle`. Seems we should only work with `block_state` and `block_state_legacy` in fork database.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -80,6 +78,8 @@ namespace eosio::chain {
 
       void remove( const block_id_type& id );
 
+      bool in_valid_state() const; // not thread safe
```
heifner, line 81: Seems safest to remove this method and just use `root()`.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -46,11 +44,11 @@ namespace eosio::chain {
...
       using branch_pair_t    = pair<branch_t, branch_t>;
 
-      explicit fork_database_t(uint32_t magic_number = legacy_magic_number);
+      explicit fork_database_t();
       ~fork_database_t();
 
-      void open( const std::filesystem::path& fork_db_file, validator_t& validator );
-      void close( const std::filesystem::path& fork_db_file );
+      void open( const std::filesystem::path& fork_db_file, fc::cfile_datastream& ds, validator_t& validator );
+      void close( std::ofstream& out );
```
heifner, line 51: I'm curious the pros/cons of putting both legacy and new into the same file. I was expecting this to use two files. This means the new format will always support reading legacy info. Is there an advantage of storing in one file instead of storing two diff fork database files?

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -157,16 +161,18 @@ namespace eosio::chain {
       // see fork_database_t::fetch_branch(forkdb->head()->id())
       block_branch_t fetch_branch_from_head() const;
 
+      void reset_root(const block_handle::block_handle_variant_t& v);
```
heifner, line 164: I really dislike that this exposes `block_handle::block_handle_variant_t`. I would prefer that type to not escape controller.cpp

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -178,11 +184,11 @@ namespace eosio::chain {
       template <class R, class F>
       R apply_s(const F& f) {
          if constexpr (std::is_same_v<void, R>) {
-            if (!legacy) {
+            if (in_use == in_use_t::savanna || in_use == in_use_t::both) {
```
heifner, line 187: This need to be something like `auto in_use_value = in_use; if (in_use_value == in_use_t::savanna || in_use_value == in_use_t::both) {` as the value can change between the `||`. Same for the rest of the checks below.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -225,6 +231,6 @@ namespace eosio::chain {
 
       // if we ever support more than one version then need to save min/max in fork_database_t
```
heifner, line 232: Update comment

libraries/chain/fork_database.cpp
```diff
@@ -712,7 +670,34 @@ namespace eosio::chain {
...
+               ("filename", fork_db_file)("l", legacy_valid)("s", savanna_valid) );
+         return;
+      }
+
+      std::ofstream out( fork_db_file.generic_string().c_str(), std::ios::out | std::ios::binary | std::ofstream::trunc );
+
+      fc::raw::pack( out, magic_number );
+      fc::raw::pack( out, max_supported_version ); // write out current version which is always max_supported_version
+                                                   // version == 1 -> legacy
+                                                   // version == 1 -> savanna (two possible fork_db, one containing `block_state_legacy`,
```
heifner, line 689: Fix comment

libraries/chain/include/eosio/chain/block_handle.hpp
```diff
@@ -8,8 +8,10 @@ namespace eosio::chain {
 // Created via controller::create_block_handle(const block_id_type& id, const signed_block_ptr& b)
 // Valid to request id and signed_block_ptr it was created from.
 struct block_handle {
+   using block_handle_variant_t = std::variant<block_state_legacy_ptr, block_state_ptr>;
```
heifner, line 11: Do not want this info leaking outside of `block_handle`.

heifner (changes requested):

libraries/chain/controller.cpp
greg7mdp: Right, thanks, removed.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: Thanks, switched to forward declaration.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: Thanks, removed that member.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: I removed the `block_handle.hpp` include.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: Updated

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: done.

libraries/chain/fork_database.cpp
greg7mdp: done.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: The new code still needs to be able to read the old format regardless. 
Since we need to save additional information, and can't re-use the old code anyways, it feels more natural to increment the version and update the file format.
Probably more user friendly as well for users who might be puzzled to see two files if the stop nodeos during the transition.
Also, because we need to store which db is in use, two files using the old read/write code would not have sufficed.

libraries/chain/include/eosio/chain/fork_database.hpp
heifner: In my mind this is even worse. I would say we either go all the way and provide every method that takes a variant or we continue to use apply/etc in controller. I say we keep with using apply in controller.

greg7mdp: Hum, I don't think it is worse.

I get that `block_handle` is its own class providing a unified  view of a `block_state` and that we don't want other things to depend on its internals.

However, it seems that there are some cases where we need the new `block_state_variant_t` type, like when we want to reset the root of our `external` fork_database which contains two fork_db pointers. I don't see why it is bad to define this type?

My intent in doing this was to try to help with the transition case where we may have two simultaneous fork_dbs. It may not be the best solution for your actual needs. If you know at this point what the interface should be, I'm happy to update this PR. 

Did you mean in your comment that I should replace all instances of `block_state_variant_t` with `std::variant<block_state_legacy_ptr, block_state_ptr>`?

libraries/chain/fork_database.cpp
```diff
@@ -712,7 +665,34 @@ namespace eosio::chain {
...
+      bool savanna_valid = fork_db_s && fork_db_s->has_root();
+
+      // check that fork_dbs are in a consistent state
+      if (!legacy_valid && !savanna_valid) {
+         elog( "fork_database is in a bad state when closing; not writing out '${filename}', legacy_valid=${l}, savanna_valid=${s}",
+               ("filename", fork_db_file)("l", legacy_valid)("s", savanna_valid) );
+         return;
+      }
+
+      std::ofstream out( fork_db_file.generic_string().c_str(), std::ios::out | std::ios::binary | std::ofstream::trunc );
```
heifner, line 679: Seems a bit odd to use `cfile` to read and `std::ofstream` to write. Might be better to use `cfile`.

libraries/chain/fork_database.cpp
```diff
@@ -703,7 +655,8 @@ namespace eosio::chain {
    fork_database::fork_database(const std::filesystem::path& data_dir)
       : data_dir(data_dir)
         // genesis starts with legacy
-      , fork_db_l{std::make_unique<fork_database_legacy_t>(fork_database_legacy_t::legacy_magic_number)}
+      , fork_db_l{std::make_unique<fork_database_legacy_t>()}
+      , fork_db_s{std::make_unique<fork_database_if_t>()}
```
heifner, line 659: I'm surprised to see we always create both. After savanna is activated we will not need legacy. Would be nice to not create either unless needed. Looks like you are creating them on open already.

libraries/chain/fork_database.cpp
```diff
@@ -288,12 +245,7 @@ namespace eosio::chain {
          fc::raw::pack( out, *(*itr) );
       }
 
-      if( head ) {
-         fc::raw::pack( out, head->id() );
-      } else {
-         elog( "head not set in fork database; '${filename}' will be corrupted",
-               ("filename", fork_db_file) );
-      }
+      fc::raw::pack( out, head ? head->id() : digest_type());
```
heifner, line 248: Seems we should fail and remove the file if no `head`. Writing a corrupt file is only setting up a time bomb.

libraries/chain/fork_database.cpp
```diff
@@ -147,109 +146,67 @@ namespace eosio::chain {
...
+   void fork_database_t<BSP>::close(std::ofstream& out) {
       std::lock_guard g( my->mtx );
-      my->close_impl(fork_db_file);
+      my->close_impl(out);
    }
 
    template<class BSP>
-   void fork_database_impl<BSP>::close_impl(const std::filesystem::path& fork_db_file) {
-      if( !root ) {
-         if( index.size() > 0 ) {
```
heifner, line 242: At one point in Savanna a node could be in a state where it had no reversible blocks. Now that we move LIB via block QCs we should always have a few reversible blocks. Likely good to allow an empty index. We will still want to always write out `head` and `root` anyway.

libraries/chain/include/eosio/chain/fork_database.hpp
```diff
@@ -149,24 +151,26 @@ namespace eosio::chain {
...
       // see fork_database_t::fetch_branch(forkdb->head()->id())
       block_branch_t fetch_branch_from_head() const;
 
+      void reset_root(const block_state_variant_t& v);
+
       template <class R, class F>
       R apply(const F& f) const {
          if constexpr (std::is_same_v<void, R>) {
-            if (legacy) {
+            if (in_use.load(std::memory_order_relaxed) == in_use_t::legacy) {
```
heifner, line 167: I don't think we can't use `memory_order_relaxed` here and below. We want to make sure access to `in_use` is ordered with access to `fork_db_l` and `fork_db_s`.

heifner (changes requested):

libraries/chain/include/eosio/chain/fork_database.hpp
heifner: > Did you mean in your comment that I should replace all instances of block_state_variant_t with std::variant<block_state_legacy_ptr, block_state_ptr>?

No. We use apply for this type of thing elsewhere, just seem to me to be consisted to keep using apply. In my PR the creation of the `new_head` is in controller.

heifner: You can leave as is, it will go away with my PR anyway.

greg7mdp: Sounds good, I'll leave as-is, feel free to update as you see fit. 
I don't think we can use controller's `apply` always because what we want to do may depend of `in_use` inside fork_db.

libraries/chain/fork_database.cpp
greg7mdp: It was that way before my change. But yes I agree it is a little bit inconsistent. I can change it if you think it is worthwhile.

libraries/chain/fork_database.cpp
greg7mdp: I don't think it matters if we create them empty, they are very small. Actually probably there is no point for them to be unique_ptr, we could store the two values directly.

libraries/chain/fork_database.cpp
greg7mdp: The reason I had a member `in_valid_state` originally is that I tried checking not only the `root` but also the `head` initially before saving a fork_db. This caused an issue, making some tests not pass, so I matched what we did before i.e. saving the fok_db even if the head is undefined.

What we did before was maybe even worse because we didn't write anything ito the file if the head was not defined, but on open we still tried to read it anyways.

libraries/chain/include/eosio/chain/fork_database.hpp
greg7mdp: OK, that doesn't buy us  much because, even with `std::memory_order_seq_cst`, another thread could interrupt us and modify `in_use` after we load the value.
If no other thread is modifying in-use, then the memory_order doesn't matter.

libraries/chain/fork_database.cpp
greg7mdp: Actually I just tried again and skipping writing the file if no `head`  seems to work fine. Pushing this to the branch now.
I also added removing the file if we try to open it and can't find the head. 
Please let me know if this is OK or I should remove it.

heifner (approved):

libraries/chain/controller.cpp
```diff
@@ -1694,12 +1689,12 @@ struct controller_impl {
       // If we start at a block during or after the IF transition, we need to provide this information
       // at startup.
       // ---------------------------------------------------------------------------------------------
-      if (fork_db.fork_db_if_present()) {
+      if (auto in_use = fork_db.version_in_use(); in_use  == fork_database::in_use_t::both || in_use  == fork_database::in_use_t::savanna) {
```
linh2931, line 1692: Maybe you can have a method `version_is_savanna()` and `version_is_legacy()` to avoid this long expression. But no need to change now.

libraries/chain/controller.cpp
```diff
@@ -3025,7 +3020,7 @@ struct controller_impl {
 
             chain_head = block_handle{std::move(new_head)};
             if (s != controller::block_status::irreversible) {
-               fork_db.switch_from_legacy(chain_head);
+               fork_db.switch_from_legacy(chain_head.internal());
```
linh2931, line 3023: Not your PR. This `internal()` implies details exposed.

linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Adds support for a new fork_database persistent file version (v2), which can store two versions of the fork_database. Needed for consensus transition. 
Note:end

PR #2309, opened 2024-03-13
[5.0] avoid using a stack variable after return

spoonincode:
`http_max_response_time` is a variable declared on that stack as part of `plugin_startup()`
https://github.com/AntelopeIO/leap/blob/c301152420c381a08aadf1f4d1f2b59414110418/plugins/producer_api_plugin/producer_api_plugin.cpp#L94
`CALL_WITH_400` captures all by reference
https://github.com/AntelopeIO/leap/blob/c301152420c381a08aadf1f4d1f2b59414110418/plugins/producer_api_plugin/producer_api_plugin.cpp#L23-L26
`INVOKE_R_R_D` then goes on to use the reference to `http_max_response_time` which is stale as `plugin_startup()` has long since returned once HTTP requests are made
https://github.com/AntelopeIO/leap/blob/c301152420c381a08aadf1f4d1f2b59414110418/plugins/producer_api_plugin/producer_api_plugin.cpp#L65-L66

This problem seemed to have been introduced in 5.0. Since this is in producer_api, which isn't expected to ever be exposed publicly, not considering this a security defect.

### DIFF
--- a/plugins/producer_api_plugin/producer_api_plugin.cpp
+++ b/plugins/producer_api_plugin/producer_api_plugin.cpp
@@ -63,6 +63,7 @@ using namespace eosio;
      auto result = api_handle.call_name(std::move(params));
 
 #define INVOKE_R_R_D(api_handle, call_name, in_param) \
+     const fc::microseconds http_max_response_time = http.get_max_response_time(); \
      auto deadline = http_max_response_time == fc::microseconds::maximum() ? fc::time_point::maximum() \
                                                                            : fc::time_point::now() + http_max_response_time; \
      auto params = parse_params<in_param, http_params_types::possible_no_params>(body);\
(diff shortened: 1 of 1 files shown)


heifner (commented):
Seems like it would be safer to capture all by value `=` in `CALL_WITH_400` instead of all by reference `&`.

https://github.com/AntelopeIO/leap/blob/c301152420c381a08aadf1f4d1f2b59414110418/plugins/producer_api_plugin/producer_api_plugin.cpp#L23-L26

greg7mdp (approved):

spoonincode:
> safer

Well, personally for non-trivial usages of lambdas -- such as this case -- I like to see explicit capture lists. So `[&producer, &http]` in this case. Does that work?

heifner:
> > safer
> 
> Well, personally for non-trivial usages of lambdas -- such as this case -- I like to see explicit capture lists. So `[&producer, &http]` in this case. Does that work?

Sounds good.

heifner (approved):

spoonincode:
that was probably a bad idea since now clang warns of unused lambda captures

PR #2314, opened 2024-03-14
Added IF ibc contract + test

systemzax:


unittests/test-contracts/if_ibc/if_ibc.cpp
```diff
@@ -0,0 +1,252 @@
...
+    //check if the target proof of inclusion correctly resolves to the root of the finality proof
+    _check_target_block_proof_of_inclusion(target_block_proof_of_inclusion, finality_proof.qc_block.finality_mroot);
+    
+    //if proof of inclusion was successful, the target block and its dynamic data have been validated as final and correct
+    block_data target_block = std::get<if_ibc::block_data>(target_block_proof_of_inclusion.target);
+
+    //if the finality_mroot we just proven is more recent than the last root we have stored, store it
+    _maybe_add_proven_root(target_block.dynamic_data.block_num, finality_proof.qc_block.finality_mroot);
+
+}
```
linh2931, line 202: I don't see base_digest being used.

unittests/test-contracts/if_ibc/if_ibc.hpp
```diff
@@ -0,0 +1,489 @@
...
+      static uint64_t clz_power_of_2(uint64_t v) {
+          return 64 - (__builtin_clzll(v)+1); 
+      }
+
+      //Compute the maximum number of layers of a merkle tree for a given number of leaves
+      static uint64_t calculate_max_depth(uint64_t node_count) {
+          if (node_count == 0) return 0;
+          auto implied_count = next_power_of_2(node_count);
+          return clz_power_of_2(implied_count) + 1;
+      }
```
greg7mdp, line 45: I was intrigued by the weird function name `clz_power_of_2`.
What that function does is more like  `log_base_2`

I think this could be replaced by:
```suggestion
      //Compute the maximum number of layers of a merkle tree for a given number of leaves
      static uint64_t calculate_max_depth(uint64_t node_count) {
          if (node_count == 0) return 0;
          return std::llround(std::log2(node_count)) + 2;
      }
```

greg7mdp:
Can you remove all the empty lines especially at the end of functions? Thanks!

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Interblock BlockChain Communication (IBC) contracts and test with Savanna Consensus.
Note:end

unittests/if_ibc_tests.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+#include <contracts.hpp>
+#include <test_contracts.hpp>
+
+#include "fork_test_utilities.hpp"
+
+using namespace eosio::chain;
+using namespace eosio::testing;
+
+// From params_test
```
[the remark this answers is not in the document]
systemzax: Removed

unittests/if_ibc_tests.cpp
```diff
@@ -0,0 +1,145 @@
...
+#include <eosio/chain/exceptions.hpp>
+using mvo = mutable_variant_object;
+
+
+BOOST_AUTO_TEST_SUITE(if_ibc)
```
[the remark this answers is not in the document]
systemzax: Done

unittests/if_ibc_tests.cpp
```diff
@@ -0,0 +1,145 @@
...
+                     ("direction", 1)
+                     ("hash", "401526ba03ec4a955c83cda131dacd3e89becaad2cf04107170e436dd90a553f")
+               }))
+
+            )
+         )
+
+      );
+
+   }
```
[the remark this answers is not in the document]
systemzax: I don't know if it is needed, since exceptions thrown by the contract do fail the test... Is there a better way to do this?

unittests/test-contracts/if_ibc/if_ibc.hpp
```diff
@@ -0,0 +1,489 @@
...
+              }
+
+          }
+
+          return hash;
+
+      }
+
+      struct quorum_certificate {
+          std::vector<uint64_t> finalizers;
```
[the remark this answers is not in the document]
systemzax: Yes, this is the underlying data type of the bitset. Since the quorum_certificate is part of the data structure that must be converted to and from binary via the ABI, I don't think we can pass the bitset type directly, unless we make it an ABI type. I'm not sure if this is desirable. But I can try if this is preferred.

unittests/test-contracts/if_ibc/if_ibc.cpp
systemzax: The base_digest is actually called witness_hash here, and its content will depend on whether or not was have an new finality policy. You can see here :

https://github.com/AntelopeIO/leap/pull/2314/files#diff-dbff99e45bcfb34c437a7a084bd325c478280c758e3282913923e42ce7afba6aR264

If you think this is preferrable, I can rename it for clarity.

unittests/test-contracts/if_ibc/bitset.hpp
```diff
@@ -0,0 +1,46 @@
+using namespace eosio;
+
+class bitset {
+public:
+    bitset(size_t size) 
+        : num_bits(size), data(new uint64_t[(size + 63) / 64]()) {}
+    
+    bitset(size_t size, uint64_t* _data) 
+        : num_bits(size), data(_data) {}
```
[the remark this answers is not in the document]
systemzax: Fixed

systemzax:
> Can you remove all the empty lines especially at the end of functions? Thanks!

Fixed

unittests/test-contracts/if_ibc/if_ibc.hpp
systemzax: That's what is, yes.

Your suggestion seems to work fine, so I adjusted that.

Originally, I ported the code that was here : 

https://github.com/AntelopeIO/leap/blob/hotstuff_integration/libraries/chain/include/eosio/chain/incremental_merkle.hpp#L59

When Bucky reviewed the contract, he told me about this __builtin_clzll function.

Maybe we want to fix that code as well?

greg7mdp: > Maybe we want to fix that code as well?

Thanks, sounds good, let me do some verification first and I'll create a PR for it.

unittests/test-contracts/if_ibc/if_ibc.cpp
linh2931: No need to rename. Probably add a comment.

unittests/if_ibc_tests.cpp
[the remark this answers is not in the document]
linh2931: Oh, you have `try ... FC_LOG_AND_RETHROW` in your test case. That would be sufficient.

unittests/test-contracts/if_ibc/if_ibc.hpp
greg7mdp: @systemzax, if you need a function which provides exactly the same number as before, please use:

```
static uint64_t calculate_max_depth(uint64_t node_count) {
   if (node_count == 0)
      return 0;
   return std::llround(std::ceil(std::log2(node_count)))+1ul;
}
```

greg7mdp: @systemzax Matt came up with a faster version, but it requires C++20 which I don't believe is supported in cdt
(See PR #2320)

spoonincode: the non c++20 way would be something like
```c++
uint64_t calculate_max_depth(uint64_t node_count) {
   if(node_count <= 1)
      return node_count;
   return 64 - __builtin_clzll(2 << (64 - 1 - __builtin_clzll ((node_count - 1))));
}
```
but this doesn't optimize quite as nicely as the c++20 variant for some reason (c++20 variant only uses a single `lzcnt` instruction instead of two as the above does). I'm not sure off hand how to improve it. Still, webassembly for the above compiles down to
```wat
calculate_max_depthr(unsigned long long):              # @calculate_max_depthr(unsigned long long)
        block           
        local.get       0
        i64.const       2
        i64.lt_u
        br_if           0                               # 0: down to label0
        i64.const       64
        i32.const       2
        i32.const       63
        local.get       0
        i64.const       -1
        i64.add 
        i64.clz 
        i32.wrap_i64
        i32.sub 
        i32.shl 
        i64.extend_i32_s
        i64.clz 
        i64.sub 
        local.set       0
        end_block                               # label0:
        local.get       0
        end_function
```
which is pretty good

greg7mdp: This version looks good, same result as prior version for all values except for `1`, where it returns `2` instead of `1`. I think it is probably fine to use.

spoonincode: > except for `1`, where it returns `2` instead of `1`

Maybe you didn't copy the entirety of the function, it starts with
```cpp
   if(node_count <= 1)
      return node_count;
```
so idk how an input of `1` could return `2`

greg7mdp: > Maybe you didn't copy the entirety of the function,

Right, my bad, thanks for the correction Matt.

@systemzax Matt's version is exactly equivalent (and simpler) than the original one, so I think it would be best to adopt it.

greg7mdp (approved):

unittests/test-contracts/svnn_ibc/bitset.hpp
```diff
@@ -0,0 +1,42 @@
+using namespace eosio;
+
+class bitset {
+public:
+    bitset(size_t size) 
+        : num_bits(size), data((size + 63) / 64) {}
+    
+   bitset(size_t size, const std::vector<uint64_t>& raw_bitset)
+       : num_bits(size), data(raw_bitset) {}
```
spoonincode, line 9: might want to consider validating that `raw_bitset` is properly sized. otherwise it seems like it's possible to construct a `bitset` where `num_bits` is greater than actual `data`.

systemzax: Done

greg7mdp (approved):

PR #2318, opened 2024-03-16
IF: Add needed dependencies to the `INSTANT_FINALITY` protocol feature

### LINKED ISSUE #2162: IF: Add the activation of WTMSIG_BLOCK_SIGNATURES  and ACTION_RETURN_VALUE features as prerequisite for the activation of the instant_finality feature
`INSTANT_FINALITY` should have prerequisite:
- `ACTION_RETURN_VALUE` is desired so that the IBC contract can use it. Also it changes the algorithm for digest computation and don't want IBC to have to handle both ways.
- `WTMSIG_BLOCK_SIGNATURES` Changes the algorithm for digest computation and don't want the IBC to handle both ways. Also we do not need to check if `WTMSIG_BLOCK_SIGNATURES` is activated in `block_header_state::next`.
- `DISALLOW_EMPTY_PRODUCER_SCHEDULE` Using the same `set_proposed_producers` which checks this protocol feature. We do not want empty proposer schedules.

greg7mdp:
Resolves #2162 

Also changed a `elog` to `wlog` in fork_database, as the check is stricter than it used to be, and the warning occurs often in normal testing.


ericpassmore:
Note:start
group: IF
category: TEST
summary: Tune logging inside fork_database to be less verbose now that stricter checks are in place. 
Note:end

unittests/protocol_feature_tests.cpp
```diff
@@ -1074,13 +1074,9 @@ BOOST_AUTO_TEST_CASE( get_sender_test ) { try {
...
 
-   const auto& pfm = c.control->get_protocol_feature_manager();
-   const auto& d = pfm.get_builtin_digest(builtin_protocol_feature_t::instant_finality);
-   // needed for bios contract
-   const auto& dp = pfm.get_builtin_digest(builtin_protocol_feature_t::bls_primitives);
-   const auto& dw = pfm.get_builtin_digest(builtin_protocol_feature_t::wtmsig_block_signatures);
-   const auto& dwk = pfm.get_builtin_digest(builtin_protocol_feature_t::webauthn_key);
-   c.preactivate_protocol_features( {*d, *dp, *dw, *dwk} );
+      // needed for bios contract
+      // const auto& dwk = pfm.get_builtin_digest(builtin_protocol_feature_t::webauthn_key);
```
heifner, line 1078: Remove commented out code and fix indent.

heifner (approved):

unittests/protocol_feature_tests.cpp
greg7mdp: Thanks, done.

PR #2320, opened 2024-03-18
Replace hacky version of `calculate_max_depth` with simpler one.

greg7mdp:
In PR #2314, I asked why we were using overly complex code to do what was basically a log base 2.
Gnome mentioned that the original code was in `incremental_merkle_tree.hpp`.

So I propose this simplification. The replacement provides exactly the same result as the previous implementation, as the following test program demonstrates:

```c++
#include <cmath>
#include <cstdlib>

static uint64_t next_power_of_2(uint64_t value) {
   value -= 1;
   value |= value >> 1;
   value |= value >> 2;
   value |= value >> 4;
   value |= value >> 8;
   value |= value >> 16;
   value |= value >> 32;
   value += 1;   return value;
}

constexpr uint64_t clz_power_2(uint64_t value) {
   int lz = 64;

   if (value) lz--;
   if (value & 0x00000000FFFFFFFFULL) lz -= 32;
   if (value & 0x0000FFFF0000FFFFULL) lz -= 16;
   if (value & 0x00FF00FF00FF00FFULL) lz -= 8;
   if (value & 0x0F0F0F0F0F0F0F0FULL) lz -= 4;
   if (value & 0x3333333333333333ULL) lz -= 2;
   if (value & 0x5555555555555555ULL) lz -= 1;

   return lz;
}

constexpr uint64_t calculate_max_depth(uint64_t node_count) {
   if (node_count == 0) {
      return 0;
   }
   auto implied_count = next_power_of_2(node_count);
   return clz_power_2(implied_count) + 1;
}

static uint64_t calculate_max_depth2(uint64_t node_count) {
   if (node_count == 0)
      return 0;
   return std::llround(std::ceil(std::log2(node_count))) + 1;
}

int main(int argc, char **argv)
{
   for (size_t i=0; i<100'000'000; ++i)
      if (auto a = calculate_max_depth(i), b = calculate_max_depth2(i); a != b) {
         printf("fails for i=%ld (%ld != %ld)\n", i, a, b);
         exit(0);
      }
   printf("success\n");
   return 0;
}
```

### DIFF
--- a/libraries/chain/include/eosio/chain/incremental_merkle.hpp
+++ b/libraries/chain/include/eosio/chain/incremental_merkle.hpp
@@ -18,7 +18,7 @@ namespace detail {
 constexpr uint64_t calculate_max_depth(uint64_t node_count) {
    if (node_count == 0)
       return 0;
-   // following is non-floating point equivalent to `std::ceil(std::log2(node_count)) + 1)
+   // following is non-floating point equivalent to `std::ceil(std::log2(node_count)) + 1)` (and about 9x faster)
    return 8*sizeof(node_count) - std::countl_zero(std::bit_ceil(node_count));
 }
 
--- a/libraries/chain/include/eosio/chain/incremental_merkle.hpp
+++ b/libraries/chain/include/eosio/chain/incremental_merkle.hpp
@@ -19,7 +19,7 @@ constexpr uint64_t calculate_max_depth(uint64_t node_count) {
    if (node_count == 0)
       return 0;
    // following is non-floating point equivalent to `std::ceil(std::log2(node_count)) + 1)` (and about 9x faster)
-   return 8*sizeof(node_count) - std::countl_zero(std::bit_ceil(node_count));
+   return std::bit_width(std::bit_ceil(node_count));
 }
 
 template<typename ContainerA, typename ContainerB>
--- a/libraries/chain/include/eosio/chain/incremental_merkle.hpp
+++ b/libraries/chain/include/eosio/chain/incremental_merkle.hpp
@@ -2,6 +2,7 @@
 #include <eosio/chain/types.hpp>
 #include <eosio/chain/merkle.hpp>
 #include <fc/io/raw.hpp>
+#include <bit>
 
 namespace eosio::chain {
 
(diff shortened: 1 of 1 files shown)


libraries/chain/include/eosio/chain/incremental_merkle.hpp
```diff
@@ -3,65 +3,21 @@
...
  */
-constexpr int calcluate_max_depth(uint64_t node_count) {
-   if (node_count == 0) {
+constexpr uint64_t calculate_max_depth(uint64_t node_count) {
+   if (node_count == 0)
       return 0;
-   }
-   auto implied_count = next_power_of_2(node_count);
-   return clz_power_2(implied_count) + 1;
+   return std::llround(std::ceil(std::log2(node_count))) + 1;
```
spoonincode, line 20: I am not so sure about using floating point math for this. How about,

```suggestion
   return 8*sizeof(node_count) - std::countl_zero(std::bit_ceil(node_count));
```

greg7mdp: Great! Even better!

spoonincode: actually even better is `return std::bit_width(std::bit_ceil(node_count))`

spoonincode (approved):

libraries/chain/include/eosio/chain/incremental_merkle.hpp
greg7mdp: Interestingly this is slower: `return std::countr_zero(std::bit_ceil(node_count)) + 1;` thank the `countl` version.

spoonincode: I'm surprised there'd be much of a difference if you've got access to `lzcnt` (either through `-mlzcnt` or something like `-march=skylake`) since they pretty much compile down to the same handful of instructions.

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Simplification of merkle tree max depth calculation.
Note:end

libraries/chain/include/eosio/chain/incremental_merkle.hpp
greg7mdp: You're right, no difference.

greg7mdp: > actually even better is `return std::bit_width(std::bit_ceil(node_count))`

using this!

Issue #2235, opened 2024-02-12
IF: SHiP changes to support IBC services


arhag:
Depends on #2080.

The newly calculation action Merkle root for each block should be provided. This can either be part of the existing stream of blocks or can be put in some new stream perhaps alongside additional data that is useful to IBC (although technically everything else needed for IBC can come from the existing block header with its extensions). Additionally, whatever structure contains the new action Merkle root will likely be expanded in the future to contain other data necessary for control Finality Tree Leaf Nodes, such as the state root.

Note that the `action_mroot` in the block header would instead contain the finality Merkle root after the IF transition, which is why the new action Merkle root must be provided separately.

Additional data to be included (to support IBC clients):
1. Protocol version
2. Active finalizer policy generation
3. Base digest

Note that the actual active finalizer policy is not required since it would have been provided earlier in a block header extension. Also the finality Merkle root does not need to be provided separately since it can already be found in the block header.

The history of new action Merkle roots will need to be stored on disk somewhere. One option is to include them in the irreversible block log by including it in a new block extension. This means when a block is validated, it would have to be mutated to add the block extension. Also, a node may receive a block with or without that block extension. If the block extension is present, it should ensure it is well-formed but it does not need to validate the data present within at the time it receives it. However, when it validates that block, if a block extension is already present, it should ensure the data within it is correct. If it is not, then it should correct the data at that time. Part of determining if this new block extension is well-formed includes determining whether it is allowed to be present in the first place; this new block extension should not be present on a block that does not have the finality header extension.

PR #2321, opened 2024-03-18
IF:  SHiP changes to support IBC services (adding finality_data log)

linh2931:
A new log `finality_data` is introduced to `SHiP` to stream finality related data. It is enabled by option `--finality-data-history` when starting a `SHiP`.

To fetch `finality_data`, use the new version of get_blocks_request:
```
struct get_blocks_request_v1 {
   uint32_t                    start_block_num        = 0;
   uint32_t                    end_block_num          = 0;
   uint32_t                    max_messages_in_flight = 0;
   std::vector<block_position> have_positions         = {};
   bool                        irreversible_only      = false;
   bool                        fetch_block            = false;
   bool                        fetch_traces           = false;
   bool                        fetch_deltas           = false;
   bool                        fetch_finality_data    = false;;
};
```

`finality_data` contains the following data:
```
struct finality_data_t {
   uint32_t     major_version{light_header_protocol_version_major};
   uint32_t     minor_version{light_header_protocol_version_minor};
   uint32_t     active_finalizer_policy_generation{0};
   digest_type  action_mroot{};
   digest_type  base_digest{};
};

```
A sample test run result looks like

```
{ "get_blocks_result_v0":
{
    "head": {
        "block_num": 143,
        "block_id": "0000008F3FF69AE70A77046027927082E922F2E42B0DAABE3A5053CB22179840"
    },
    "last_irreversible": {
        "block_num": 139,
        "block_id": "0000008B8289314ABA3DEE3B53245DA6FC74E5F9C49F6E80F5C265D50AD81805"
    },
    "this_block": {
        "block_num": 134,
        "block_id": "00000086F65F560E02C07B15B7D860BBE7D1282B4997A7538A1952B23739D535"
    },
    "prev_block": {
        "block_num": 133,
        "block_id": "000000852EE672A6249D9DE12B403B36B79449DBB1CD829890D2ABF74DA6B997"
    },
    "block": "0868165B80AE423AD15B974A0000000000852EE672A6249D9DE12B403B36B79449DBB1CD829890D2ABF74DA6B99700000000
000000000000000000000000000000000000000000000000000000002FC5444BCB1DFEBF0E4EB31ABCEC349EC55D5F70B678D796B257F3C6BC
903FD000000080000102000785000000010000002064365970301F6B7EE19CBC088869168E7BFF7CB6A826403ED3D620326119F30321923023
56C6087384941A3DEDA8ADAFBDFC57ADAE53C748E840C0E9B1A11DE700010300AB028500000001010F000000009DB726DF78ABF2BDEF6E8CEA
3B30761B2345CA2CEE61052BA62AF43FB4D68E501A7F85AD81D7E50F029CD130BD9FDE115BFF3072CD606649FECA6B8597804B79A1B356E03E
15804DA9AFD2F88B3B3D0AEF1FFAC61FAE5AB3C545AF29C13A4B068588353300903C41E9F8E2FFBA4331F879BA12CFDC6DD64FE672DF18E837
8AC6A9EAA2A72B7B45DF514794856CEB160916AA7BD9B327A11D62E09C31361AD892815F7F4AAE03B2E5F24C80CA30B9EBC6C5F64A68A96E8E
38E89D2162B94DA1180F3A3EE11AE6DB7C2BFD8C56B5D8CFE60EBB80BD159580B31CB005FC75D46D4D41CCC34674DDACFB340F29B6CB38D70E
1CF1D4D88B067533A29983B8AF350A3C1469D3A5FD0B84B41EB7F28A781C9555F443D7A64BE42E767079279479A8810E",
    "traces": "01008F9869B957E33758F276EA11400DB8E470B6B8571D81402F6E10F56EA635DE560064000000000000000000000000000
0000000000000000101010001000000000000EA3055F481D9FDB0F0AB441396E51B62DF6006B13CAB7AA3848F42AF437520F706A6E9C200000
000000000B500000000000000010000000000EA3055BC0000000000000002020000000000EA30550000000000EA305500000000221ACFA4010
000000000EA305500000000A8ED32327E0768165B70AE423AD15B974A0000000000848B75D25E89E1C801D2838C75D4BB56F9827D22563E794
769EAD53CC90000000000000000000000000000000000000000000000000000000000000000B4AA479BB182B047BE9A3F94C5CD8C4C21478CB
570F316D250DB5177238E834100000080000102000784000000010000000000000000000000000000000000000000010000000000000000000
000000000000000",
    "deltas": "02000E7265736F757263655F757361676501013B000000000000EA3055000868165B5C8A020000000000010000000000000
0000868165BA77A05000000000065000000000000003FEF06000000000000157265736F757263655F6C696D6974735F7374617465010153000
086000000D2D4D80A00000000B6000000000000000086000000DC19E51300000000B1010000000000000000000000000000000000000000000
00000000000000000B0461200000000009FB4080000000000",
    "finality_data": "010000000000000000000000EF01DE2A75105982A0D6A88FD934E3BA7C88FFC339DC1E3F33130C36717D7E245A85
1A54FA6F4E1174D92D2480D21CE66026B296803DFA12B1E3847E699EF4A0"
}
}

```
Resolved https://github.com/AntelopeIO/leap/issues/2235.


ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Add finality_data log updating SHiP to support IBC services.
Note:end

tests/ship_streamer_test.py
```diff
@@ -31,8 +31,13 @@
 
 appArgs = AppArgs()
 extraArgs = appArgs.add(flag="--num-clients", type=int, help="How many ship_streamers should be started", default=1)
+extraArgs = appArgs.add_bool(flag="--finality-data-history", help="Enable finality data history", action='store_true')
+extraArgs = appArgs.add_bool(flag="--fetch-finality-data", help="Fetch finality data", action='store_true')
```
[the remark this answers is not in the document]
linh2931: Was thinking about testing them independently. It is not necessary.

libraries/chain/block_state.cpp
```diff
@@ -297,6 +299,16 @@ digest_type block_state::get_finality_mroot_claim(const qc_claim_t& qc_claim) co
    return get_validation_mroot(next_core_metadata.final_on_strong_qc_block_num);
 }
 
+finality_data_t block_state::get_finality_data() const {
+   finality_data_t finality_data {
+      // other fields take the default values set by finality_data_t definition
+      .action_mroot = action_mroot,
+      .base_digest  = compute_base_digest()  // from block_header_state
```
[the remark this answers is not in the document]
linh2931: Thanks. This can be called multiple times. Cache it now.

heifner (approved):

libraries/chain/block_state.cpp
```diff
@@ -297,6 +303,16 @@ digest_type block_state::get_finality_mroot_claim(const qc_claim_t& qc_claim) co
...
```
greg7mdp, line 310: since `base_digest` is cached, we should use the cached value instead of recomputing it here.

libraries/chain/block_state.cpp
```diff
@@ -297,6 +303,16 @@ digest_type block_state::get_finality_mroot_claim(const qc_claim_t& qc_claim) co
...
+   };
+
+   return finality_data;
```
greg7mdp, line 313: why not: 

```
finality_data_t block_state::get_finality_data() const {
   return {
      // other fields take the default values set by finality_data_t definition
      .action_mroot = action_mroot,
      .base_digest  = base_digest
   };
}
```

libraries/chain/controller.cpp
```diff
@@ -370,7 +371,7 @@ struct assembled_block {
                                    [&](assembled_block_if& ab) {
                                       auto bsp = std::make_shared<block_state>(ab.bhs, std::move(ab.trx_metas),
                                                                                std::move(ab.trx_receipts), ab.valid, ab.qc, signer,
-                                                                               valid_block_signing_authority);
+                                                                               valid_block_signing_authority, ab.action_mroot);
```
greg7mdp, line 374: This PR needs to be caught up with the `hotstuff_integration` branch.

libraries/chain/controller.cpp
```diff
@@ -4107,6 +4110,12 @@ struct controller_impl {
       }
    }
 
+   std::optional<finality_data_t> head_finality_data() const {
+      return apply<std::optional<finality_data_t>>(chain_head,
+         overloaded{ [](const block_state_legacy_ptr& head) { return std::nullopt; },
+                     [](const block_state_ptr&        head) { return head->get_finality_data(); }});
+   }
```
greg7mdp, line 4117:
```suggestion
      return apply_s<std::optional<finality_data_t>>(chain_head, [](const block_state_ptr& head) { return head->get_finality_data(); });
   }
```

libraries/chain/block_state.cpp
linh2931: Thanks for this great catch!

libraries/chain/controller.cpp
linh2931: `apply_s` will return an empty `finality_data_t` in Legacy. `SHiP` cannot tell if it has a value or not and will proceed to  `"finality_data": null`. In Legacy, we don't want `SHiP` to generate the `finality_data` log at all so it won't break existing clients.

greg7mdp: > `apply_s` will return an empty `finality_data_t` in Legacy.

I'm pretty sure it it will return an empty optional.

linh2931: I initially used `apply_s` and tried again, the check https://github.com/AntelopeIO/leap/blob/417615de4cd035e962d8cbdf5719d517bbab8a6c/plugins/state_history_plugin/state_history_plugin.cpp#L283 always succeeded in Legacy tests.

libraries/state_history/abi.cpp
```diff
@@ -33,6 +33,19 @@ extern const char* const state_history_plugin_abi = R"({
                 { "name": "fetch_deltas", "type": "bool" }
             ]
         },
+        {
+            "name": "get_blocks_request_v1", "fields": [
```
greg7mdp, line 37: just curious, why did we need a `_v1` for `get_blocks_request_v1`, but not for `get_blocks_result_v0`?

libraries/state_history/include/eosio/state_history/types.hpp
```diff
@@ -104,6 +104,10 @@ struct get_blocks_request_v0 {
    bool                        fetch_deltas           = false;
 };
 
+struct get_blocks_request_v1 : get_blocks_request_v0 {
+   bool                        fetch_finality_data    = false;;
```
greg7mdp, line 108:
```suggestion
   bool                        fetch_finality_data    = false;
```

plugins/state_history_plugin/state_history_plugin.cpp
```diff
@@ -115,15 +117,16 @@ struct state_history_plugin_impl : std::enable_shared_from_this<state_history_pl
 
    // thread-safe
    std::optional<chain::block_id_type> get_block_id(uint32_t block_num) {
-      std::optional<chain::block_id_type> id;
       if( trace_log ) {
-         id = trace_log->get_block_id( block_num );
-         if( id )
+         if ( auto id = trace_log->get_block_id( block_num ); id )
```
greg7mdp, line 121: Small detail, you don't need the `id` here and following lines, so it could be:

```suggestion
         if ( auto id = trace_log->get_block_id( block_num ) )
```

libraries/chain/controller.cpp
greg7mdp: Which test would fail with this change?

greg7mdp:
Currently `test_chain_plugin` fails in this branch (probably an integration issue).

libraries/state_history/abi.cpp
linh2931: We need a `_v1` for request as an existing client has to add `fetch_finality_data` in a request if we put `fetch_finality_data` in the ABI definition. But we don't need a new version for `get_blocks_result` since we only add the finality_data log in new Savanna.

libraries/chain/block_state.cpp
```diff
@@ -40,6 +42,8 @@ block_state::block_state(const block_header_state&                bhs,
    , valid(valid)
    , pub_keys_recovered(true) // called by produce_block so signature recovery of trxs must have been done
    , cached_trxs(std::move(trx_metas))
+   , action_mroot(action_mroot)
+   , base_digest(compute_base_digest())
```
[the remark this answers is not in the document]
linh2931: `header_state` would be constructed first. Or move `base_digest` construction to the constructor body to save an extra `if` check in `get_base_digest`?

linh2931:
> Currently `test_chain_plugin` fails in this branch (probably an integration issue).

Hm, the PR branch passed all the tests after updating with the hotstuff_integration branch.

libraries/chain/block_state.cpp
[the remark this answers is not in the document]
linh2931: I have thought about it more. Not all `nodeos`es are `SHiP` node; `base_digest` is not always needed. I will make it as an optional.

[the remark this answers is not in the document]
heifner: That is not thread safe, make sure only ever called from the main thread.

libraries/chain/block_state.cpp
```diff
@@ -90,6 +92,8 @@ block_state::block_state(const block_state_legacy& bsp, const digest_type& actio
    validated = bsp.is_valid();
    pub_keys_recovered = bsp._pub_keys_recovered;
    cached_trxs = bsp._cached_trxs;
+   action_mroot = action_mroot_svnn;
+   base_digest = compute_base_digest();
```
greg7mdp, line 96: We don't need this line, right?

greg7mdp:
> Hm, the PR branch passed all the tests after updating with the hotstuff_integration branch.

Weird, running `./plugins/chain_plugin/test/test_chain_plugin` in my build of your branch fails every time.

libraries/chain/block_state.cpp
linh2931: Thanks for careful check.

libraries/chain/controller.cpp
linh2931: Like the legacy `ship_streamer_test` by `ctest -R ship_streamer_test`.

linh2931: The test itself does not fail, but an extra field `"finality_data": null` is added to SHiP log for Legacy, which is incorrect.

I will add a test to make sure no extra fields added for Legacy later.

greg7mdp: Looks like the extra field `"finality_data": null ` is added even with the current code in the branch.

linh2931: The extra field was caused by ABI binary to JSON converter, even though the SHiP binary log does not include `finality_data`. https://github.com/AntelopeIO/leap/issues/2340 is created to track the issue.
Thanks.

libraries/state_history/abi.cpp
spoonincode: You might be able to make `fetch_finality_data` a binary extension on `get_blocks_request_v0` to avoid needing a v1. Whether or not that's a good idea I'm not sure

heifner: Seems like we should just keep it simple and straightforward. Binary extension seems overly complex.

libraries/chain/controller.cpp
greg7mdp: Thanks for creating the new issue, Lin!

greg7mdp (approved):

heifner (approved):

PR #2326, opened 2024-03-19
IF: Create a new action receipt digest for Savanna blocks and move the digest computation to `action_trace`

### LINKED ISSUE #2161: IF: Modify the action_receipt digest to be computed by concatenating and hashing a pair constructed from act_digest and the rest of the values of the action_receipt
Depends on https://github.com/AntelopeIO/leap/issues/2162.

This allows for a smaller proof footprint, and faster verification by smart contracts and light clients.

https://github.com/AntelopeIO/leap/blob/hotstuff_integration/libraries/chain/include/eosio/chain/action_receipt.hpp#L19

greg7mdp:
Resolves #2161.

This PR:
- renames `action_receipt::digest()` to `action_trace::digest_legacy()`
- adds a new `action_trace::digest_savanna()`
- updates the existing calling places to use `action_trace::digest_legacy()`

When PR #2316 is merged, we will need to:
1. use the correct digest computation depending on Savanna/Legacy blocks.
2. store in Legacy blocks validated during the transtion pre-computed Savanna digests
3. use these Savanna digests for the conversion `block_state_legacy` -> `block_state` during the transition, so that the savanna merkle trees are computed correctly.

### DIFF
--- a/libraries/chain/apply_context.cpp
+++ b/libraries/chain/apply_context.cpp
@@ -184,9 +184,9 @@ void apply_context::exec_one()
       r.auth_sequence[auth.actor] = next_auth_sequence( auth.actor );
    }
 
-   trx_context.executed_action_receipt_digests.emplace_back( r.digest() );
-
    finalize_trace( trace, start );
+
+   trx_context.executed_action_receipt_digests.emplace_back( trace.digest_legacy() );
 
    if ( control.contracts_console() ) {
       print_debug(receiver, trace);
--- a/libraries/chain/include/eosio/chain/trace.hpp
+++ b/libraries/chain/include/eosio/chain/trace.hpp
@@ -4,7 +4,7 @@
 #include <eosio/chain/action_receipt.hpp>
 #include <eosio/chain/block.hpp>
 
-namespace eosio { namespace chain {
+namespace eosio::chain {
 
    struct account_delta {
       account_delta( const account_name& n, int64_t d):account(n),delta(d){}
@@ -45,6 +45,43 @@ namespace eosio { namespace chain {
       std::optional<fc::exception>    except;
       std::optional<uint64_t>         error_code;
       std::vector<char>               return_value;
+
+      digest_type digest_savanna() const {
+         assert(!!receipt);
+         const action_receipt& r = *receipt;
+
+         digest_type::encoder e;
+         fc::raw::pack(e, r.receiver);
+         fc::raw::pack(e, r.recv_sequence);
+         fc::raw::pack(e, act.account);
+         fc::raw::pack(e, act.name);
+         fc::raw::pack(e, r.act_digest);
+
+         {
+            digest_type::encoder e2;
+            fc::raw::pack(e2, r.global_sequence);
+            fc::raw::pack(e2, r.auth_sequence);
+            fc::raw::pack(e2, r.code_sequence);
+            fc::raw::pack(e2, r.abi_sequence);
+            fc::raw::pack(e, e2.result());
+         }
+         return e.result();
+      }
+
+      digest_type digest_legacy()const {
+         assert(!!receipt);
+         const action_receipt& r = *receipt;
+
+         digest_type::encoder e;
+         fc::raw::pack(e, r.receiver);
+         fc::raw::pack(e, r.act_digest);
+         fc::raw::pack(e, r.global_sequence);
+         fc::raw::pack(e, r.recv_sequence);
+         fc::raw::pack(e, r.auth_sequence);
+         fc::raw::pack(e, r.code_sequence);
+         fc::raw::pack(e, r.abi_sequence);
+         return e.result();
+      }
    };
 
    struct transaction_trace {
@@ -80,7 +117,7 @@ namespace eosio { namespace chain {
              auth.permission == eosio::chain::config::active_name;
    }
 
-} }  /// namespace eosio::chain
+}  /// namespace eosio::chain
 
 FC_REFLECT( eosio::chain::account_delta,
             (account)(delta) )
--- a/libraries/chain/include/eosio/chain/action_receipt.hpp
+++ b/libraries/chain/include/eosio/chain/action_receipt.hpp
@@ -1,8 +1,9 @@
 #pragma once
 
 #include <eosio/chain/types.hpp>
+#include <eosio/chain/action.hpp>
 
-namespace eosio { namespace chain {
+namespace eosio::chain {
 
    /**
     *  For each action dispatched this receipt is generated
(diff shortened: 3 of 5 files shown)


libraries/chain/apply_context.cpp
```diff
@@ -184,10 +184,10 @@ void apply_context::exec_one()
       r.auth_sequence[auth.actor] = next_auth_sequence( auth.actor );
    }
 
-   trx_context.executed_action_receipt_digests.emplace_back( r.digest() );
-
    finalize_trace( trace, start );
 
+   trx_context.executed_action_receipt_digests.emplace_back( trace.digest_legacy() );
```
heifner, line 189: This needs to be `digest_savanna()` for a proper IF block I believe. Please add a test that verifies this change.

greg7mdp: Do we not validate blocks during the transition that are not yet `proper IF`, but still use `block_state` and the new merkle tree?

heifner: The validation is going to use this same code path which is using the legacy always.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Create a new action receipt digest for Savanna blocks.
Note:end

libraries/chain/apply_context.cpp
```diff
@@ -218,17 +221,17 @@ void apply_context::exec()
...
-   }
+      for( uint32_t ordinal : _cfa_inline_actions ) {
+         trx_context.execute_action( ordinal, recurse_depth + 1 );
+      }
 
-   for( uint32_t ordinal : _inline_actions ) {
-      trx_context.execute_action( ordinal, recurse_depth + 1 );
+      for( uint32_t ordinal : _inline_actions ) {
+         trx_context.execute_action( ordinal, recurse_depth + 1 );
+      }
```
heifner, line 234: What is going on here? Did the `for` loops get moved into the `if`? How is this not tripping tests.

libraries/chain/apply_context.cpp
```diff
@@ -184,10 +184,13 @@ void apply_context::exec_one()
...
    }
 
-   trx_context.executed_action_receipt_digests.emplace_back( r.digest() );
-
    finalize_trace( trace, start );
 
+   if (trx_context.executed_action_receipts.digests_l)
+      trx_context.executed_action_receipts.digests_l->emplace_back( trace.digest_legacy() );
+   if (trx_context.executed_action_receipts.digests_s)
+      trx_context.executed_action_receipts.digests_s->emplace_back( trace.digest_savanna() );
```
heifner, line 192: What about having `trx_context.executed_action_receipts.append(trace)` and have `action_digests_t ` call the appropriate `digest_*` method and assign to appropriate digests.

libraries/chain/apply_context.cpp
greg7mdp: Yes, I moved the loops into the if. If both are empty there is no point to executing the loops.

libraries/chain/apply_context.cpp
greg7mdp: trace is a `transaction_trace`, and we compute digests for `action_trace` receipts.

libraries/chain/apply_context.cpp
heifner: Ah! That makes sense. Thanks.

libraries/chain/apply_context.cpp
greg7mdp: Beautiful! I named the function `trx_context.executed_action_receipts.compute_and_append_digests_from(trace);`

libraries/chain/apply_context.cpp
```diff
@@ -184,10 +184,10 @@ void apply_context::exec_one()
       r.auth_sequence[auth.actor] = next_auth_sequence( auth.actor );
    }
 
-   trx_context.executed_action_receipt_digests.emplace_back( r.digest() );
-
    finalize_trace( trace, start );
 
+   trx_context.executed_action_receipts.compute_and_append_digests_from(trace);
```
heifner, line 189: This should be moved back above `finalize_trace` otherwise the `action.elapsed` does not contain the time for creating the digest(s).

libraries/chain/controller.cpp
```diff
@@ -670,11 +676,11 @@ struct building_block {
                                 auto trx_merkle_fut =
                                    post_async_task(ioc, [&]() { return legacy_merkle(std::move(trx_receipts)); });
                                 auto action_merkle_fut =
-                                   post_async_task(ioc, [&]() { return legacy_merkle(std::move(action_receipts)); });
+                                   post_async_task(ioc, [&]() { return legacy_merkle(std::move(*action_receipts.digests_l)); });
```
heifner, line 679: Maybe `assert()` or even add a method to `action_receipts` that asserts.

libraries/chain/controller.cpp
```diff
@@ -670,11 +676,11 @@ struct building_block {
...
                                 return std::make_pair(trx_merkle_fut.get(), action_merkle_fut.get());
                              },
                              [&](const checksum256_type& trx_checksum) {
-                                return std::make_pair(trx_checksum, legacy_merkle(std::move(action_receipts)));
+                                return std::make_pair(trx_checksum, legacy_merkle(std::move(*action_receipts.digests_l)));
```
heifner, line 683: Maybe `assert()` or even add a method to `action_receipts` that asserts.

libraries/chain/apply_context.cpp
greg7mdp: Weird, I can't remember why I moved it! Moving it back.

libraries/chain/controller.cpp
greg7mdp: The `std::optional` already asserts that it is not empty when dereferenced, but I can add our own too.

heifner: Good point. Not needed then.

greg7mdp: OK, thanks, then I'll leave as-is.

heifner (approved):

libraries/chain/include/eosio/chain/trace.hpp
```diff
@@ -45,6 +45,43 @@ namespace eosio { namespace chain {
...
+         const action_receipt& r = *receipt;
+
+         digest_type::encoder e;
+         fc::raw::pack(e, r.receiver);
+         fc::raw::pack(e, r.recv_sequence);
+         fc::raw::pack(e, act.account);
+         fc::raw::pack(e, act.name);
+         fc::raw::pack(e, r.act_digest);
+
+         {
```
linh2931, line 60: Why is this nested encoder needed?

linh2931 (approved):

libraries/chain/include/eosio/chain/trace.hpp
greg7mdp: Because of this in the issue comment:
image

PR #2327, opened 2024-03-20
Perf issue: avoid vector copy at each insertion.

greg7mdp:
The `vector::reserve(vector::size() + 1)` call which is updated in this PR, with typical `std::vector` implementations, causes the vector to grow by a single element on every reverve, and therefore a memory allocation and full vector copy on every `reserve` call.

This updates the `reserve()` call so that the vector capacity grows quadrically.


heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: PERFORMACE
summary: Updates the reserve() call so that the vector capacity grows slower at a quadrically rate.
Note:end

spoonincode (approved):

PR #2329, opened 2024-03-21
IF: Use enum to track which action receipt digest to calculate

heifner:
Suggested alternative for tacking which digest to caculate.
Even better I think would be to create:

```
struct action_receipt_digests {
   enum class store_action_digests { legacy, savanna, both };
   // add() etc
   store_action_digests          action_digests_to_store;
   std::optional<digests_t>      executed_action_receipt_digests_l;
   std::optional<digests_t>      executed_action_receipt_digests_s;
};
```


heifner:
struct was added.

PR #2338, opened 2024-03-26
IF: Change bls_public_key & bls_signature serialization

### LINKED ISSUE #2334: IF: Ensure all serialization of g1/g2 uses serialization form
Serialization form is defined to be:
1. Non-Montgomery form and little-endian encoding for the field elements.
2. Affine form for the group element (the z component is 1 and not included in the serialization).
3. Binary serialization encodes x component first followed by y component.

All network messages that use the serialized form of g1/g2.
All files written to disk should used the serialized form of g1/g2.
Logs will need to use the base64url encoding of the serialized form of the g1/g2 logged.

The `bls_public_key` and `bls_signature` should be updated to store the binary data for the serialized form. That way `pack` and `unpack` can automatically serialize the data in the correct form. When constructing a `bls_public_key` (or `bls_signature`) either using a constructor that takes `g1` (or `g2`) or when `unpack`ing (in which case this code should be in `reflector_init`), the class must validate the serialized data and construct the `g1` (or `g2`) as a cached private field in the class. That `g1` (or `g2`) should be in Montgomery Jacobian little-endian form so that it can be used efficiently in BLS math.

Keeping the serialized data in `bls_public_key` and `bls_signature` allows for quickly writing out the serialized form when needed, e.g. in a vote message or in a QC attached to the block, and it also allows for relatively fast conversion to the base64url encoding for logging purposes. 

There are some places in ephemeral memory where it makes sense to store the `g1` (or `g2`) directly rather than store `bls_public_key` (or `bls_signature`). 

One case is when a temporary value is being mutated in a function for some use in the function to then be thrown away. For example, computing the `g1` aggregation of the public keys of the finalizers who are indicated as participating in a QC which is used to validate the aggregated `g2` signature. In that case, the `g2` can be pulled from the cache of the `bls_signature` recovered from the valid QC attached to the block. But there is no need to build up a `bls_public_key` as the accumulator when aggregating over the participating finalizers since that would require converting to serialized form for each addition, which is expensive.

Another case is when storing the aggregated signature `g2` in the pending QC within a `block_state`. For similar reasons as above, it is more efficient to store this as a raw `g2` rather than as a `bls_signature`. However, when choosing the best QC to include in a new block that is pulled from the pending QC, that accumulating raw `g2` must first be used to construct a `bls_signature` and store that as the valid QC in the `block_state`. Then that constructed `bls_signature`, which already has the serialized form within it, can be used to quickly create the QC that is attached to the block in a block extension.

heifner:
Change `bls_public_key` and `bls_signature` serialization to be:
* Non-Montgomery form and little-endian encoding for the field elements.
* Affine form for the group element (the z component is 1 and not included in the serialization).
* Binary serialization encodes x component first followed by y component.

Add `bls_aggregate_signature` which is mutable version of `bls_signature`. `bls_signature` is now immutable.

Serialize as variable sized array and instead of a fixed sized array to make it easier to use in external tools that use ABI since our ABI does not handle fixed sized arrays.

Resolves #2334


libraries/libfc/src/crypto/bls_public_key.cpp
```diff
@@ -19,40 +27,37 @@ namespace fc::crypto::blslib {
...
       return config::bls_public_key_prefix + data_str;
-
    }
 
-   bool bls_public_key::equal( const bls_public_key& pkey) const {
-      return _pkey.equal(pkey._pkey);
+   void bls_public_key::reflector_init() {
+      // called after construction, but always on the same thread and before bls_public_key passed to any other threads
+      static_assert(fc::raw::has_feature_reflector_init_on_unpacked_reflected_types,
+                    "FC unpack needs to call reflector_init otherwise unpacked_trx will not be initialized");
```
greg7mdp, line 45: update message

libraries/libfc/include/fc/crypto/bls_public_key.hpp
```diff
@@ -3,43 +3,67 @@
...
-         bls_public_key( const bls_public_key& ) = default;
-         explicit bls_public_key( const bls12_381::g1& pkey ) {_pkey = pkey;}
-         // affine non-montgomery base64url with bls_public_key_prefix
-         explicit bls_public_key(const std::string& base64urlstr);
+   // Immutable after construction.
+   // Provides an efficient wrapper around bls12_381::g1.
+   // Serialization form:
+   //   Non-Montgomery form and little-endian encoding for the field elements.
+   //   Affine form for the group element (the z component is 1 and not included in the serialization).
+   //   Binary serialization encodes x component first followed by y component.
```
greg7mdp, line 19: Please update the comment with the explanation from issue #2334 describing why we use this internal representation and this form for serialization.

libraries/libfc/include/fc/crypto/bls_signature.hpp
```diff
@@ -1,51 +1,136 @@
...
-         bls_signature() = default;
-         bls_signature( bls_signature&& ) = default;
-         bls_signature( const bls_signature& ) = default;
-         explicit bls_signature( const bls12_381::g2& sig ){_sig = sig;}
+   // Immutable after construction.
+   // Provides an efficient wrapper around bls12_381::g2.
+   // Serialization form:
+   //   Non-Montgomery form and little-endian encoding for the field elements.
+   //   Affine form for the group element (the z component is 1 and not included in the serialization).
+   //   Binary serialization encodes x component first followed by y component.
```
greg7mdp, line 18: same here, add explanation as to why. It is not obvious :-).

libraries/libfc/include/fc/crypto/bls_signature.hpp
```diff
@@ -1,51 +1,136 @@
...
+         return ds;
+      }
+
+      // Could use FC_REFLECT, but to make it obvious serialization matches bls_aggregate_signature implement via operator
+      template<typename T>
+      friend T& operator>>(T& ds, bls_signature& sig) {
+         ds.read(reinterpret_cast<char*>(sig._affine_non_montgomery_le.data()), sig._affine_non_montgomery_le.size()*sizeof(uint8_t));
+         constexpr bool check = true;  // check if invalid
+         constexpr bool raw   = false; // non-montgomery
+         std::optional<bls12_381::g2> g2 = bls12_381::g2::fromAffineBytesLE(sig._affine_non_montgomery_le, check, raw);
```
greg7mdp, line 55: I wish `fromAffineBytesLE` took `struct { bool check; bool raw; };` as parameter so we could call with:

`bls12_381::g2::fromAffineBytesLE(sig._affine_non_montgomery_le, { .check = true, .raw = false });

libraries/libfc/include/fc/crypto/bls_signature.hpp
```diff
@@ -1,51 +1,136 @@
...
+      bool equal( const bls_aggregate_signature& sig) const {
+         return _jacobian_montgomery_le.equal(sig._jacobian_montgomery_le);
+      }
+
+      template<typename T>
+      friend T& operator<<(T& ds, const bls_aggregate_signature& sig) {
+         constexpr bool raw = false;
+         std::array<uint8_t, 192> affine_non_montgomery_le = sig._jacobian_montgomery_le.toAffineBytesLE(raw);
+         ds.write(reinterpret_cast<const char*>(affine_non_montgomery_le.data()), affine_non_montgomery_le.size()*sizeof(uint8_t));
+         return ds;
```
greg7mdp, line 107: Is this OK? I though there was an issue with the `jacobian_montgomery` form being non-unique? Would a serialization + deserialization get us back to the same value?

heifner: @arhag ?

libraries/libfc/src/crypto/bls_public_key.cpp
```diff
@@ -19,40 +27,37 @@ namespace fc::crypto::blslib {
       constexpr bool raw = false;  // non-montgomery
       std::optional<bls12_381::g1> g1 = bls12_381::g1::fromAffineBytesLE(bytes, check, raw);
       FC_ASSERT(g1);
```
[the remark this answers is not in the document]
heifner: Updated `bls_public_key`. Unfortunately we need `bls_public_key::operator=` because things like `finalizer_policy` require it. Let me know if you think the `const` is worth it.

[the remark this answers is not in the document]
heifner: Actually, I forgot that fc reflection does not work with `const`, so will remove the `const`.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Update internal storage format of BLS keys to optimize usage. No change to external representation of keys.
Note:end

libraries/libfc/include/fc/crypto/bls_signature.hpp
arhag: Serialization then deserialization would not necessarily give back the original value. But that is not a problem

greg7mdp (approved):

linh2931 (commented):
Don't forget to change fixed array to vector for clients.

libraries/libfc/include/fc/crypto/bls_public_key.hpp
```diff
@@ -1,45 +1,86 @@
...
+      auto operator==(const bls_public_key& rhs) const {
+         return _affine_non_montgomery_le == rhs._affine_non_montgomery_le;
+      }
 
-         bls12_381::g1 _pkey;
+      template<typename T>
+      friend T& operator<<(T& ds, const bls_public_key& sig) {
+         // Serialization as variable length array when it is stored as a fixed length array. This makes for easier deserialization by external tools
+         fc::raw::pack(ds, fc::unsigned_int(static_cast<uint32_t>(sig._affine_non_montgomery_le.size()*sizeof(uint8_t))));
+         ds.write(reinterpret_cast<const char*>(sig._affine_non_montgomery_le.data()), sig._affine_non_montgomery_le.size()*sizeof(uint8_t));
```
greg7mdp, line 60: why not?

```suggestion
         ds.write(reinterpret_cast<const char*>(sig._affine_non_montgomery_le.data()), sizeof(sig._affine_non_montgomery_le));

```

libraries/libfc/include/fc/crypto/bls_private_key.hpp
```diff
@@ -24,7 +24,7 @@ namespace fc::crypto::blslib {
 
          bls_private_key& operator=( const bls_private_key& ) = default;
 
-         std::string to_string(const yield_function_t& yield = yield_function_t()) const;
+         std::string to_string() const;
```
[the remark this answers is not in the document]
heifner: It was not used and we have no current support for it.

libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -149,7 +149,8 @@ valid_quorum_certificate pending_quorum_certificate::to_valid_quorum_certificate
    } else if (is_quorum_met_no_lock()) {
       valid_qc._strong_votes = _strong_votes._bitset;
       valid_qc._weak_votes   = _weak_votes._bitset;
-      valid_qc._sig          = fc::crypto::blslib::aggregate(std::array{_strong_votes._sig, _weak_votes._sig});
+      valid_qc._sig          = _strong_votes._sig;
+      valid_qc._sig.aggregate(_weak_votes._sig);
```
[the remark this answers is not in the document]
heifner: Not sure I understand. This block is `is_quorum_met()` which means `s == state_t::weak_achieved || s == state_t::weak_final` because `strong` is the `if` of this `else if`. So there has to be week signature, right?

libraries/chain/block_state.cpp
```diff
@@ -184,21 +184,24 @@ void block_state::verify_qc(const valid_quorum_certificate& qc) const {
                   ("s", strong_weights)("w", weak_weights)("t", active_finalizer_policy->threshold) );
    }
 
-   std::vector<bls_public_key> pubkeys;
+   // no reason to use bls_public_key wrapper
```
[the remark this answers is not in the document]
heifner: Could create a `bls_aggregate_public_key` similar to `bls_aggregate_signature` for this use case. Without that you would pay for expensive creation of the `bls_public_key` affine data. If we want to hide the `bls12_381` use in our code I could certainly add the additional wrappers needed.

libraries/chain/hotstuff/hotstuff.cpp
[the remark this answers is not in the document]
linh2931: Oh I missed the `else if`.

libraries/chain/block_state.cpp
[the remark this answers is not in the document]
linh2931: It might be consistent to hide `bls12_381` library, but no need to do it now.

linh2931 (approved):

greg7mdp (approved):

PR #2344, opened 2024-03-27
IF: Set correct proposed producer schedule version

### LINKED ISSUE #2313: IF: Modify set_proposed_producers() to make sure propser_schedule.version numbers not being skipped
In reviewing https://github.com/AntelopeIO/leap/pull/2282, @heifner commented:

> I believe this change means it is now possible to skip propser_schedule.version numbers. We should modify the set_proposed_producers logic so we don't skip version numbers. Need a test to cover that case so we don't break it again.
> 
> We need to make sure that if a proposer policy is scheduled for the same proposer_policy.active_time that it replaces the current one there but doesn't increment the schedule version. I think we will need to change controller::set_proposed_producers to check if there is an existing one that would have the same active_time and replace it and use its version. We should return the correct version to the contract.

Please see https://github.com/AntelopeIO/leap/pull/2282/#discussion_r1525082170

heifner:
When `set_proposed_producers` replaces an existing in-flight producer schedule, use the existing in-flight producer schedule version so that versions are not skipped.

Resolves #2313

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -479,6 +479,10 @@ struct building_block {
 
       uint32_t get_next_proposer_schedule_version() const {
          if (!parent.proposer_policies.empty()) {
+            block_timestamp_type active_time = detail::get_next_next_round_block_time(timestamp);
+            if (auto itr = parent.proposer_policies.find(active_time); itr != parent.proposer_policies.cend()) {
+               return itr->second->proposer_schedule.version; // will replace so retrun same version
+            }
             return (--parent.proposer_policies.end())->second->proposer_schedule.version + 1;
          }
          assert(active_proposer_policy);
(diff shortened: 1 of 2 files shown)


linh2931 (approved):

libraries/chain/controller.cpp
```diff
@@ -479,6 +479,10 @@ struct building_block {
 
       uint32_t get_next_proposer_schedule_version() const {
          if (!parent.proposer_policies.empty()) {
+            block_timestamp_type active_time = detail::get_next_next_round_block_time(timestamp);
+            if (auto itr = parent.proposer_policies.find(active_time); itr != parent.proposer_policies.cend()) {
+               return itr->second->proposer_schedule.version; // will replace so retrun same version
```
greg7mdp, line 484:
```suggestion
               return itr->second->proposer_schedule.version; // will replace so return same version
```

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Improvement to prevent proposer schedule version numbers from being skipping when an in-flight schedule is replaced.  
Note:end

PR #2347, opened 2024-03-27
add ASAN CI "platform"

spoonincode:
Adds a new "platform" that builds & tests the software with ASAN.

It's possible to enable a single build with UBSAN+ASAN simultaneously, but this further increases both the compile time and run time overhead. So do UBSAN & ASAN separately for now.

And speaking of ASAN build overhead, I was unable to get NP or LR tests to complete in CI without either running out of memory or having some sort of timing problem. I even tried increasing the resources for the run but that still wasn't enough. I think it's likely worthwhile to go ahead with what we've got though as at least we'll still get unit test coverage, and reevaluate whatever the performance problem is later (I was able to complete them successfully locally).

Example run of ASAN failures:
https://github.com/AntelopeIO/leap/actions/runs/8458172328/job/23172568986

### DIFF
--- a/.cicd/platforms/asan.Dockerfile
+++ b/.cicd/platforms/asan.Dockerfile
@@ -1,0 +1,37 @@
+# syntax=docker/dockerfile:1
+FROM ubuntu:jammy
+ENV TZ="America/New_York"
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get upgrade -y && \
+    apt-get install -y build-essential      \
+                       cmake                \
+                       git                  \
+                       jq                   \
+                       libcurl4-openssl-dev \
+                       libgmp-dev           \
+                       llvm-11-dev          \
+                       lsb-release          \
+                       ninja-build          \
+                       python3-numpy        \
+                       software-properties-common \
+                       file                 \
+                       wget                 \
+                       zlib1g-dev           \
+                       zstd
+
+RUN yes | bash -c "$(wget -O - https://apt.llvm.org/llvm.sh)" llvm.sh 18
+
+#make sure no confusion on what llvm library leap's cmake should pick up on
+RUN rm -rf /usr/lib/llvm-18/lib/cmake
+
+ENV LEAP_PLATFORM_HAS_EXTRAS_CMAKE=1
+COPY <<-EOF /extras.cmake
+  set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "" FORCE)
+
+  set(CMAKE_C_COMPILER "clang-18" CACHE STRING "")
+  set(CMAKE_CXX_COMPILER "clang++-18" CACHE STRING "")
+  set(CMAKE_C_FLAGS "-fsanitize=address -fno-omit-frame-pointer" CACHE STRING "")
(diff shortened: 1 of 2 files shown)


.cicd/platforms/asan.Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+COPY <<-EOF /extras.cmake
+  set(CMAKE_BUILD_TYPE "RelWithDebInfo" CACHE STRING "" FORCE)
+
+  set(CMAKE_C_COMPILER "clang-18" CACHE STRING "")
+  set(CMAKE_CXX_COMPILER "clang++-18" CACHE STRING "")
+  set(CMAKE_C_FLAGS "-fsanitize=address -fno-omit-frame-pointer" CACHE STRING "")
+  set(CMAKE_CXX_FLAGS "-fsanitize=address -fno-omit-frame-pointer" CACHE STRING "")
+EOF
+
+ENV ASAN_OPTIONS=detect_leaks=0
```
spoonincode, line 37: There are some WAVM memory "leaks" that have been known about for 5+ years. It'd be nice to resolve them one day to reenable the leak test.

greg7mdp (approved):

heifner (commented):
> I was unable to get NP or LR tests to complete in CI without either running out of memory or having some sort of timing problem. I even tried increasing the resources for the run but that still wasn't enough. I think it's likely worthwhile to go ahead with what we've got though as at least we'll still get unit test coverage, and reevaluate whatever the performance problem is later (I was able to complete them successfully locally).

Is this still true, looks like NP and LR tests are running.

linh2931:
> > I was unable to get NP or LR tests to complete in CI without either running out of memory or having some sort of timing problem. I even tried increasing the resources for the run but that still wasn't enough. I think it's likely worthwhile to go ahead with what we've got though as at least we'll still get unit test coverage, and reevaluate whatever the performance problem is later (I was able to complete them successfully locally).
> 
> Is this still true, looks like NP and LR tests are running.

I got a failure in Ubsan CI https://github.com/AntelopeIO/leap/issues/2350

spoonincode:
> Is this still true, looks like NP and LR tests are running.

Not for `asan` platform.

I had tried in a different branch to give NP/LR asan more resources, for example by doing for NP,
https://github.com/AntelopeIO/leap/blob/f1ded48eaf73e1eab81cda0777ac84a4d615b90e/.github/workflows/build.yaml#L178-L184
but `plugin_http_api_test` failed (doesn't locally) and it looks like a timing problem,
https://github.com/AntelopeIO/leap/actions/runs/8430283230/job/23086076054

and for LR doing something similar such as
https://github.com/AntelopeIO/leap/blob/f1ded48eaf73e1eab81cda0777ac84a4d615b90e/.github/workflows/build.yaml#L220-L226
but the box fell over running out of memory
https://github.com/AntelopeIO/leap/actions/runs/8430283230/job/23086075824

maybe giving them _even more_ resources would work.

heifner (approved):

ericpassmore:
Note:start
group: STABILITY
category: CHORE
summary: Adds test pass with AddressSanitizer.
Note:end

Issue #2340, opened 2024-03-27
IF: add get_blocks_result_v1 to SHiP


linh2931:
This is a placeholder for a possible issue.

https://github.com/AntelopeIO/leap/pull/2321 adds `SHiP` support for IBC. In the ABI definition and type, we introduce a new get_blocks_request_v1 which includes a field fetch_finality_data, but we reuse get_blocks_result_v0 by adding an { "name": "finality_data", "type": "bytes?" } to the bottom. Our code makes sure no finality_data be written to binary SHiP log in Legacy mode. But if a client uses ABI bin_to_json() to convert the binary to JSON based on get_blocks_result_v0, the ABI converter will add an extra "finality_data": null to the JSON output. 

A solution would be to add `get_blocks_result_v1` to include finality_data and leave get_blocks_result_v0 alone.

PR #2349, opened 2024-03-27
IF: add get_blocks_result_v1 support to SHiP

linh2931:
https://github.com/AntelopeIO/leap/pull/2321 added SHiP support for IBC. A new `get_blocks_request_v1` was introduced to include a field `fetch_finality_data`, but existing `get_blocks_result_v0` was reused to return the result of `get_blocks_request_v1`. That can cause confusions to the users, and a client's ABI `bin_to_json()` would add an extra `"finality_data": null` to the JSON output in Legacy (as `finality_data` does not exist).

This PR adds `get_blocks_result_v1` which stores the result of `get_blocks_request_v1`; `get_blocks_result_v0` is restored to its original use: only storing the result of `get_blocks_result_v0`.

`get_blocks_request_v1` adds extra `get_blocks_request_v1` to `get_blocks_request_v0`; `get_blocks_result_v1` adds extra `finality_data` to `get_blocks_result_v0`.

```
struct get_blocks_request_v1 : get_blocks_request_v0 {
   bool                        fetch_finality_data    = false;
};

struct get_blocks_result_v1 : get_blocks_result_v0 {
   std::optional<bytes>        finality_data;
};
```

Tests are updated to check both `get_blocks_result_v0` and `get_blocks_result_v1` work.

Resolved https://github.com/AntelopeIO/leap/issues/2340

### DIFF
--- a/libraries/state_history/include/eosio/state_history/serialization.hpp
+++ b/libraries/state_history/include/eosio/state_history/serialization.hpp
@@ -709,11 +709,7 @@ datastream<ST>& operator<<(datastream<ST>& ds, const history_context_wrapper_sta
 
 template <typename ST>
 datastream<ST>& operator<<(datastream<ST>& ds, const eosio::state_history::get_blocks_result_v0& obj) {
-   fc::raw::pack(ds, obj.head);
-   fc::raw::pack(ds, obj.last_irreversible);
-   fc::raw::pack(ds, obj.this_block);
-   fc::raw::pack(ds, obj.prev_block);
-   history_pack_big_bytes(ds, obj.block);
+   ds << static_cast<const eosio::state_history::get_blocks_result_base&>(obj);
    history_pack_big_bytes(ds, obj.traces);
    history_pack_big_bytes(ds, obj.deltas);
    return ds;
@@ -721,13 +717,7 @@ datastream<ST>& operator<<(datastream<ST>& ds, const eosio::state_history::get_b
 
 template <typename ST>
 datastream<ST>& operator<<(datastream<ST>& ds, const eosio::state_history::get_blocks_result_v1& obj) {
-   fc::raw::pack(ds, obj.head);
-   fc::raw::pack(ds, obj.last_irreversible);
-   fc::raw::pack(ds, obj.this_block);
-   fc::raw::pack(ds, obj.prev_block);
-   history_pack_big_bytes(ds, obj.block);
-   history_pack_big_bytes(ds, obj.traces);
-   history_pack_big_bytes(ds, obj.deltas);
+   ds << static_cast<const eosio::state_history::get_blocks_result_v0&>(obj);
    history_pack_big_bytes(ds, obj.finality_data);
    return ds;
 }
--- a/libraries/state_history/include/eosio/state_history/serialization.hpp
+++ b/libraries/state_history/include/eosio/state_history/serialization.hpp
@@ -716,6 +716,18 @@ datastream<ST>& operator<<(datastream<ST>& ds, const eosio::state_history::get_b
    history_pack_big_bytes(ds, obj.block);
    history_pack_big_bytes(ds, obj.traces);
    history_pack_big_bytes(ds, obj.deltas);
+   return ds;
+}
+
+template <typename ST>
+datastream<ST>& operator<<(datastream<ST>& ds, const eosio::state_history::get_blocks_result_v1& obj) {
+   fc::raw::pack(ds, obj.head);
+   fc::raw::pack(ds, obj.last_irreversible);
+   fc::raw::pack(ds, obj.this_block);
+   fc::raw::pack(ds, obj.prev_block);
+   history_pack_big_bytes(ds, obj.block);
+   history_pack_big_bytes(ds, obj.traces);
+   history_pack_big_bytes(ds, obj.deltas);
    history_pack_big_bytes(ds, obj.finality_data);
    return ds;
 }
--- a/plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
+++ b/plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
@@ -320,9 +320,11 @@ public:
       std::visit(
          chain::overloaded{
             [&](state_history::get_blocks_result_v0& r) {
+               static_assert(std::is_same_v<state_history::get_blocks_result_v0, std::variant_alternative_t<1, state_history::state_result>>);
                pack_result_base(r, 1); // 1 for variant index of get_blocks_result_v0 in state_result
             },
             [&](state_history::get_blocks_result_v1& r) {
+               static_assert(std::is_same_v<state_history::get_blocks_result_v1, std::variant_alternative_t<2, state_history::state_result>>);
                pack_result_base(r, 2); // 2 for variant index of get_blocks_result_v1 in state_result
             }
          },
--- a/plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
+++ b/plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
@@ -194,7 +194,7 @@ public:
 template <typename Session>
 class blocks_result_send_queue_entry : public send_queue_entry_base, public std::enable_shared_from_this<blocks_result_send_queue_entry<Session>> {
    std::shared_ptr<Session>                                        session;
-   state_history::get_blocks_result_v0                             r;
+   state_history::get_blocks_result                                result;
    std::vector<char>                                               data;
    std::optional<locked_decompress_stream>                         stream;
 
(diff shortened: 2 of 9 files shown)


plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -264,46 +264,68 @@ class blocks_result_send_queue_entry : public send_queue_entry_base, public std:
...
-         me->send_finality_data();
-      });
+      std::visit(chain::overloaded{
+                    [&](state_history::get_blocks_result_v0& r) {
+                        send_log(session->get_delta_log_entry(r, stream), true, [me=this->shared_from_this()]() {
+                           me->stream.reset();
+                           me->session->session_mgr.pop_entry();}); },
+                    [&](state_history::get_blocks_result_v1& r) {
+                        send_log(session->get_delta_log_entry(r, stream), false, [me=this->shared_from_this()]() {
+                           me->send_finality_data(); }); }},
```
greg7mdp, line 288: Do we still have to do `me->send_finality_data()` even though it is included in `state_history::get_blocks_result_v1`?

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -264,46 +264,68 @@ class blocks_result_send_queue_entry : public send_queue_entry_base, public std:
...
+
       // pack the state_result{get_blocks_result} excluding the fields `traces` and `deltas`
       fc::datastream<size_t> ss;
-      fc::raw::pack(ss, fc::unsigned_int(1)); // pack the variant index of state_result{r}
-      fc::raw::pack(ss, static_cast<const state_history::get_blocks_result_base&>(r));
+      if(std::holds_alternative<state_history::get_blocks_result_v0>(result)) {
+         fc::raw::pack(ss, fc::unsigned_int(1)); // pack the variant index of state_result{r}, 1 for get_blocks_result_v0
+      } else {
+         fc::raw::pack(ss, fc::unsigned_int(2)); // pack the variant index of state_result{r}, 2 for get_blocks_result_v1
+      }
```
greg7mdp, line 315: why not one `std::visit` for lines 311 to 328?

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
linh2931: Yes we still need to call `me->send_finality_data()`. It populates and sends finality data log.

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
linh2931: I was trying to reduce duplicated code but that was even messier. Will change.

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -264,46 +264,68 @@ class blocks_result_send_queue_entry : public send_queue_entry_base, public std:
...
 public:
-   blocks_result_send_queue_entry(std::shared_ptr<Session> s, state_history::get_blocks_result_v0&& r)
+   blocks_result_send_queue_entry(std::shared_ptr<Session> s, state_history::get_blocks_result&& r)
        : session(std::move(s)),
-         r(std::move(r)) {}
+         result(std::move(r)) {}
 
    void send_entry() override {
+      assert(std::holds_alternative<state_history::get_blocks_result_v0>(result) ||
+             std::holds_alternative<state_history::get_blocks_result_v1>(result));
```
heifner: Seems it would be better to `static_assert` [variant_alternative](https://en.cppreference.com/w/cpp/utility/variant/variant_alternative) instead.

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -264,46 +264,68 @@ class blocks_result_send_queue_entry : public send_queue_entry_base, public std:
...
+
       // pack the state_result{get_blocks_result} excluding the fields `traces` and `deltas`
       fc::datastream<size_t> ss;
-      fc::raw::pack(ss, fc::unsigned_int(1)); // pack the variant index of state_result{r}
-      fc::raw::pack(ss, static_cast<const state_history::get_blocks_result_base&>(r));
+      if(std::holds_alternative<state_history::get_blocks_result_v0>(result)) {
+         fc::raw::pack(ss, fc::unsigned_int(1)); // pack the variant index of state_result{r}, 1 for get_blocks_result_v0
+      } else {
+         fc::raw::pack(ss, fc::unsigned_int(2)); // pack the variant index of state_result{r}, 2 for get_blocks_result_v1
```
heifner: I think with the `static_assert` we could do `fc::raw::pack(ss, fc::unsigned_int(result.index() + 1);`

Same for below.

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -515,7 +537,8 @@ struct session : session_base, std::enable_shared_from_this<session<Plugin, Sock
       current_request = std::move(req);
    }
 
-   void send_update(state_history::get_blocks_request_v0& request, bool fetch_finality_data, state_history::get_blocks_result_v0 result, const chain::signed_block_ptr& block, const chain::block_id_type& id) {
+   template<typename T> // get_blocks_result_v0 or get_blocks_result_v1
+   void send_update(state_history::get_blocks_request_v0& request, bool fetch_finality_data, T result, const chain::signed_block_ptr& block, const chain::block_id_type& id) {
```
heifner, line 544: Should be `T&& result` and caller should `std::move` it.

heifner: Probably doesn't matter much, but we are using `variant` elsewhere, why switch to `template` and `constexpr` here?

tests/ship_streamer.cpp
```diff
@@ -142,7 +142,11 @@ int main(int argc, char* argv[]) {
          eosio::check(!result_document.HasParseError(),                                      "Failed to parse result JSON from abieos");
          eosio::check(result_document.IsArray(),                                             "result should have been an array (variant) but it's not");
          eosio::check(result_document.Size() == 2,                                           "result was an array but did not contain 2 items like a variant should");
-         eosio::check(std::string(result_document[0].GetString()) == "get_blocks_result_v0", "result type doesn't look like get_blocks_result_v0");
+         if( fetch_finality_data ) {
+            eosio::check(std::string(result_document[0].GetString()) == "get_blocks_result_v1", "result type doesn't look like get_blocks_result_v1");
+         } else {
+            eosio::check(std::string(result_document[0].GetString()) == "get_blocks_result_v0", "result type doesn't look like get_blocks_result_v0");
+         }
```
heifner, line 149: Not sure we need this handling of `get_blocks_result_v0`. I think we should always send a `get_blocks_request_v1` and always expect a `get_blocks_result_v1`.

tests/ship_streamer_test.py
```diff
@@ -268,11 +271,14 @@ def getLatestSnapshot(nodeId):
...
-            this_block_num = i['get_blocks_result_v0']['this_block']['block_num']
+            this_block_num = i['get_blocks_result_v1']['this_block']['block_num'] if args.finality_data_history else i['get_blocks_result_v0']['this_block']['block_num']
             if this_block_num < block_num:
                 block_num = this_block_num
             assert block_num == this_block_num, f"{block_num} != {this_block_num}"
-            assert isinstance(i['get_blocks_result_v0']['deltas'], str) # verify deltas in result
+            if args.finality_data_history:
+                assert isinstance(i['get_blocks_result_v1']['deltas'], str) # verify deltas in result
+            else:
+                assert isinstance(i['get_blocks_result_v0']['deltas'], str) # verify deltas in result
```
heifner, line 281: Same comment for these as above. Just support `v1` version of request and result.

heifner (changes requested):

libraries/state_history/include/eosio/state_history/serialization.hpp
```diff
@@ -709,6 +709,18 @@ datastream<ST>& operator<<(datastream<ST>& ds, const history_context_wrapper_sta
...
+   fc::raw::pack(ds, obj.this_block);
+   fc::raw::pack(ds, obj.prev_block);
+   history_pack_big_bytes(ds, obj.block);
+   history_pack_big_bytes(ds, obj.traces);
+   history_pack_big_bytes(ds, obj.deltas);
+   return ds;
+}
+
+template <typename ST>
+datastream<ST>& operator<<(datastream<ST>& ds, const eosio::state_history::get_blocks_result_v1& obj) {
```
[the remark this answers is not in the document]
linh2931: Will also change `get_blocks_result_v0` serilaization to use `get_blocks_result_base`.

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
linh2931: This code was re-factored while you were reviewing it. It is now simpler.

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
linh2931: The difference in `send_update` for handling `get_blocks_request_v0` and `get_blocks_request_1` is deep inside the middle of code (adding finality_data_log). Using template here seems less messy to try to reuse the same code for both `get_blocks_request_v0` and `get_blocks_request_1`

tests/ship_streamer.cpp
linh2931: To support Legacy test `ship_streamer_test` and newer Savanna tests, we need to handle both versions.

tests/ship_streamer_test.py
linh2931: Same comment above.

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
```diff
@@ -264,45 +264,70 @@ class blocks_result_send_queue_entry : public send_queue_entry_base, public std:
...
-      fc::raw::pack(ds, fc::unsigned_int(1)); // pack the variant index of state_result{r}
-      fc::raw::pack(ds, static_cast<const state_history::get_blocks_result_base&>(r));
+      std::visit(
+         chain::overloaded{
+            [&](state_history::get_blocks_result_v0& r) {
+               pack_result_base(r, 1); // 1 for variant index of get_blocks_result_v0 in state_result
+            },
+            [&](state_history::get_blocks_result_v1& r) {
+               pack_result_base(r, 2); // 2 for variant index of get_blocks_result_v1 in state_result
+            }
```
heifner, line 327: Would like to see a `static_assert` [variant_alternative](https://en.cppreference.com/w/cpp/utility/variant/variant_alternative) that `1` and `2` are the correct values.

tests/ship_streamer.cpp
heifner: No, you can just use v1 always everywhere. No need to test v0.

plugins/state_history_plugin/include/eosio/state_history_plugin/session.hpp
linh2931: Thanks! This makes it safer.

tests/ship_streamer.cpp
```diff
@@ -142,7 +138,9 @@ int main(int argc, char* argv[]) {
          eosio::check(!result_document.HasParseError(),                                      "Failed to parse result JSON from abieos");
          eosio::check(result_document.IsArray(),                                             "result should have been an array (variant) but it's not");
          eosio::check(result_document.Size() == 2,                                           "result was an array but did not contain 2 items like a variant should");
-         eosio::check(std::string(result_document[0].GetString()) == "get_blocks_result_v0", "result type doesn't look like get_blocks_result_v0");
+         if( fetch_finality_data ) {
+            eosio::check(std::string(result_document[0].GetString()) == "get_blocks_result_v1", "result type doesn't look like get_blocks_result_v1");
+         }
```
heifner, line 143: don't need the `if`

linh2931: My bad. Missed that one. Thanks.

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Adds get_blocks_result_v1 to SHiP as part of IBC support; get_blocks_result_v0 is restored to its original use.
Note:end

Issue #2354, opened 2024-03-28
IF: Test Failure: auto_bp_peering_if_test 


heifner:
https://github.com/AntelopeIO/leap/actions/runs/8470775216/job/23212479735  `ubsan`

Appears to have failed because the producer was so busy processing vote messages it was unable to send out block.

```
debug 2024-03-28T18:15:06.584 net-1     net_plugin.cpp:2659           operator()           ] ["localhost:9876 - da03a39" - 4 127.0.0.1:33032] bcast block 172
```
Received on client:
```
debug 2024-03-28T18:15:13.643 net-2     net_plugin.cpp:3115           process_next_block_m ] ["localhost:9877 - 777d221" - 2 127.0.0.1:55490] received block 172, id 7c9d32fc29ab578c..., latency: 7143ms, head 171
```

Potentially related to: #2351

heifner:
https://github.com/AntelopeIO/leap/actions/runs/8481930721/job/23240489887
This one failed because it forked between `get_info` and `get_block`. The `head` block reported by `get_info` was not available when `get_block` tried to retrieve it. The fork was caused by extreme slowness it appears. So not complete like the other failure, but seems close enough to just report under this issue.

heifner:
https://github.com/AntelopeIO/leap/actions/runs/8485923201/job/23251802866

heifner:
ci/cd perf increase seems to have fixed the issue.

Issue #2351, opened 2024-03-28
IF: Test Failure: nodeos_startup_catchup_if_lr_test


heifner:
`AssertionError: Expected to at least receive 9.0 transactions per block, but only getting 7.6`

It is not clear what is going on here. Could just mean our ci/cd could not keep up. Or maybe all the voting is slowing it down enough it can't process trxs quick enough. Would be worth looking closer at the logs. Might need more net threads to allow trx processing along with vote processing. We should consider processing votes in its own thread pool so they can't clog the net thread pool. 

Happened at the same time: 
https://github.com/AntelopeIO/leap/actions/runs/8467659168/job/23199344820 ubsan
https://github.com/AntelopeIO/leap/actions/runs/8467659168/job/23199345763 ubnuntu22repo

heifner:
Happened at the same time: 
https://github.com/AntelopeIO/leap/actions/runs/8467659168/job/23199344820
https://github.com/AntelopeIO/leap/actions/runs/8467659168/job/23199345763

heifner:
https://github.com/AntelopeIO/leap/actions/runs/8467659168/job/23202047049

greg7mdp:
Issue was caused by the addition of two new tests in https://github.com/AntelopeIO/leap/pull/2335.

Two workflow runs with these tests removed passed without problems. https://github.com/AntelopeIO/leap/actions/runs/8523025337

#### Resolution:

@spoonincode will beef up the instances  the workflows are run on for: `NP`, `LR`, `ubsan`, `asan`, `asserton`

PR #2355, opened 2024-03-29
IF: net: Add extra logging around async write

heifner:
Add additional debug logging around async write.
Make all logs that contain a connection id use a prefix of `- ` so they are easy to grep for in the log.

```
debug 2024-03-28T20:58:09.489 net-2     net_plugin.cpp:2654           operator()           ] socket_is_open true, state connected, syncing false, connection - 4
```
The `- ` matches default output of `peer_*` logging:
```
debug 2024-03-28T20:58:09.489 net-0     net_plugin.cpp:2668           operator()           ] ["localhost:9876 - b648e56" - 4 127.0.0.1:9876] bcast block 7
```

Should help verify issue of #2354 and #2351.

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -723,10 +723,17 @@ namespace eosio {
          return _out_queue.empty();
       }
 
-      bool ready_to_send() const {
-         fc::lock_guard g( _mtx );
+      // called from connection strand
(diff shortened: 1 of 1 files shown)


linh2931 (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Add additional debug logging around async write. To make log grep easier, all logs with connection id now contain prefix of - .
Note:end

PR #2357, opened 2024-03-29
IF: Ignore proposed producers during transition

### LINKED ISSUE #2057: IF: Implement proper IF transition

heifner:
Adds ignoring proposed producers during transition.

Resolves #2057

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -512,10 +512,10 @@ struct building_block {
    R apply_l(F&& f) {
       if constexpr (std::is_same_v<void, R>)
          std::visit(overloaded{[&](building_block_legacy& bb) { std::forward<F>(f)(bb); },
-                               [&](building_block_if& bb)   {}}, v);
+                               [&](building_block_if&)        {}}, v);
       else
          return std::visit(overloaded{[&](building_block_legacy& bb) -> R { return std::forward<F>(f)(bb); },
-                                      [&](building_block_if& bb)   -> R { return {}; }}, v);
+                                      [&](building_block_if&)        -> R { return {}; }}, v);
    }
 
    void set_proposed_finalizer_policy(const finalizer_policy& fin_pol) {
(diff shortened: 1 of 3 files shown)


linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Ignores proposed producers during transition.
Note:end

PR #2359, opened 2024-03-29
IF: Verify bls signature of vote while not holding a mutex

### LINKED ISSUE #2289: IF: Update libtester framework to guarantee deterministic vote processing
Options:
1. Provide a way to wait on votes (which happen on another thread normally).
2. Or, provide a flag for the voting to occur on the main thread for tests.

This should resolve test instability in our Boost IF tests such as https://github.com/AntelopeIO/leap/issues/2267.

heifner:
Since bls signature verification can take 6-8ms on ci/cd machines (~2ms on i7 or i9 12th gen), perform the verify while not holding a mutex. This reduces the contention on the `pending_quorum_certificate` mutex.

I believe we still need to do the verify off the `net` thread, but this is a simple improvement we can get in now.

This PR severely aggravated #2289, so added a fix for that issue.

Resolves #2289

### DIFF
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3491,6 +3491,24 @@ struct controller_impl {
       return fork_db.apply<vote_status>(aggregate_vote_legacy, aggregate_vote);
    }
 
+   bool node_has_voted_if_finalizer(const block_id_type& id) const {
+      if (my_finalizers.finalizers.empty())
+         return true;
+
+      std::optional<bool> voted = fork_db.apply_s<std::optional<bool>>([&](auto& forkdb) -> std::optional<bool> {
+         auto bsp = forkdb.get_block(id);
+         if (bsp) {
+            for (auto& f : my_finalizers.finalizers) {
+               if (bsp->has_voted(f.first))
+                  return {true};
+            }
+         }
+         return {false};
+      });
+      // empty optional means legacy forkdb
+      return !voted || *voted;
+   }
+
    void create_and_send_vote_msg(const block_state_ptr& bsp) {
       if (!bsp->block->is_proper_svnn_block())
          return;
@@ -5067,6 +5085,10 @@ void controller::set_proposed_finalizers( const finalizer_policy& fin_pol ) {
 vote_status controller::process_vote_message( const vote_message& vote ) {
    return my->process_vote_message( vote );
 };
+
+bool controller::node_has_voted_if_finalizer(const block_id_type& id) const {
+   return my->node_has_voted_if_finalizer(id);
+}
 
 const producer_authority_schedule& controller::active_producers()const {
    return my->active_producers();
--- a/libraries/chain/controller.cpp
+++ b/libraries/chain/controller.cpp
@@ -3498,12 +3498,11 @@ struct controller_impl {
       std::optional<bool> voted = fork_db.apply_s<std::optional<bool>>([&](auto& forkdb) -> std::optional<bool> {
          auto bsp = forkdb.get_block(id);
          if (bsp) {
-            for (auto& f : my_finalizers.finalizers) {
-               if (bsp->has_voted(f.first))
-                  return {true};
-            }
+            return std::ranges::all_of(my_finalizers.finalizers, [&bsp](auto& f) {
+               return bsp->has_voted(f.first);
+            });
          }
-         return {false};
+         return false;
       });
       // empty optional means legacy forkdb
       return !voted || *voted;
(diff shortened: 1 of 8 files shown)


libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -125,15 +123,32 @@ vote_status pending_quorum_certificate::add_weak_vote(std::span<const uint8_t> p
...
+         s = strong ? add_strong_vote(index, sig, weight)
+                    : add_weak_vote(index, sig, weight);
+         post_state = _state;
+         g.unlock();
+      }
+   }
+
    dlog("block_num: ${bn}, vote strong: ${sv}, status: ${s}, pre-state: ${pre}, post-state: ${state}, quorum_met: ${q}",
-        ("bn", block_num)("sv", strong)("s", s)("pre", pre_state)("state", _state)("q", is_quorum_met_no_lock()));
+        ("bn", block_num)("sv", strong)("s", s)("pre", pre_state)("state", post_state)("q", is_quorum_met(post_state)));
```
greg7mdp, line 151:
```suggestion
        ("bn", block_num)("sv", strong)("s", s)("pre", pre_state)("state", post_state)("q", is_quorum_met_no_lock(post_state)));
```

heifner: There is no `is_quorum_met_no_lock(state_t)`. `is_quorum_met(post_state)` is the static one that does not lock.

libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -20,17 +20,19 @@ inline std::vector<uint32_t> bitset_to_vector(const hs_bitset& bs) {
...
+      return _strong_votes._bitset[index];
    }
-   if (!fc::crypto::blslib::verify(pubkey, proposal_digest, new_sig)) {
-      wlog( "signature from finalizer ${i} cannot be verified", ("i", index) );
-      return vote_status::invalid_signature;
+   return _weak_votes._bitset[index];
+}
+
+vote_status pending_quorum_certificate::votes_t::add_vote(size_t index, const bls_signature& sig) {
+   if (_bitset[index]) { // check here as could have come in while unlocked
```
linh2931, line 31: This is not needed as you have already checked duplication.

heifner: The comment explains why it is needed.

linh2931: In which case `add_vote` come in while unlocked? I did not find it.

heifner:
```
Thread 1:  pending_quorum_certificate::add_vote(key1)
Thread 1:     checks has_voted_no_lock() return false
Thread 1:     unlocks mutex
Thread 2:  pending_quorum_certificate::add_vote(key1)
Thread 2:     checks has_voted_no_lock() return false
Thread 2:     unlocks mutex
Thread 1:  locks mutex, calls add_strong_vote(key1), unlocks mutex
Thread 2:  locks mutex, calls add_strong_vote(key1), needs to double check that it has not already voted.
```

heifner: This is an example of https://en.wikipedia.org/wiki/Double-checked_locking . Incidentally, I learned the pattern from Doug Schmidt which is my only claim to fame, lol. (see first reference in the wikipedia article.

libraries/chain/controller.cpp
```diff
@@ -3491,6 +3491,24 @@ struct controller_impl {
...
+   bool node_has_voted_if_finalizer(const block_id_type& id) const {
+      if (my_finalizers.finalizers.empty())
+         return true;
+
+      std::optional<bool> voted = fork_db.apply_s<std::optional<bool>>([&](auto& forkdb) -> std::optional<bool> {
+         auto bsp = forkdb.get_block(id);
+         if (bsp) {
+            for (auto& f : my_finalizers.finalizers) {
+               if (bsp->has_voted(f.first))
+                  return {true};
```
greg7mdp, line 3503: No need for curly braces, but ok I guess.

libraries/chain/controller.cpp
```diff
@@ -3491,6 +3491,24 @@ struct controller_impl {
...
 
+   bool node_has_voted_if_finalizer(const block_id_type& id) const {
+      if (my_finalizers.finalizers.empty())
+         return true;
+
+      std::optional<bool> voted = fork_db.apply_s<std::optional<bool>>([&](auto& forkdb) -> std::optional<bool> {
+         auto bsp = forkdb.get_block(id);
+         if (bsp) {
+            for (auto& f : my_finalizers.finalizers) {
+               if (bsp->has_voted(f.first))
```
greg7mdp, line 3502: Really should return true only if the last finalizer has voted, not the first one.

linh2931 (approved):

greg7mdp (approved):

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Reduce lock contention during BLS signature verification.
Note:end

PR #2361, opened 2024-03-29
IF: Create new efficient incremental Merkle tree

### LINKED ISSUE #2333: IF: Create new efficient incremental Merkle tree
The existing incremental Merkle tree, even for the non-legacy version, is designed in a way that unnecessarily stores extra digests and does extra hashes. This version is consistent with `calculate_merkle` (which is the non-legacy version).

This behavior is also how the legacy version works except it also adds the complexity of modifying the MSB of the digests to indicate left and right pairs.

We do not want to touch the legacy versions at all for backwards compatibility. It would be good to rename the types to include legacy in them and also rename the file names to indicate legacy as well.

But the non-legacy versions need to be replaced to use a more efficient incremental Merkle tree structure that does not have the additional digests/hashes. This should be stored in a separate file to keep it clearly separated from the legacy versions.

Information on how the new efficient incremental Merkle tree should work:

The incremental Merkle tree stores at a minimum `(std::popcount(digests.size()) + (std::has_single_bit(digests.size()) ? 0 : 1))` number of hashes.

And when you append a new digest to the incremental Merkle tree, that increments `digests.size()` which indicates through the the change in 1s and 0s of `digests.size()` how to aggregate the existing hashes together (and remove them from the structure) to form the new hash you need to store in the structure.

So when `digests.size() == 3`, you have `0b0011` which means you have three hashes:
1. The root (included since `std::has_single_bit(digests.size()) == false`) is `hash(hash(digests[0], digests[1]), digests[2])`.
2. The one represented by `0b0010` is `hash(digests[0], digests[1])`.
3. The one represented by `0b0001` is `digests[2]`.

When you add a new digest and get `digests.size() == 4`, you have `0b0100`, which means the two digests represented by `0b0010` and `0b0001` are replaced by a single digest:
1. The one represented by `0b0100` is `hash(hash(digests[0],digests[1]), hash(digests[2],digests[3]))`.

Note that you can also store the root explicitly, but in this case, since `std::has_single_bit(digests.size()) == true`, it would be redundant.

The implementation is free to always explicitly store a root (even if redundant) or only store the minimally needed number of digests, whichever is easier to implement.

The implementation should try to use C++17 since we will need a version of this implementation in CDT as well.

This new incremental Merkle tree should be used for the Validation Trees.

There should also be a non-incremental version of calculating the Merkle tree as well that is consistent with this new more efficient way of computing the root. There should be tests to ensure they are consistent with each other.

This non-incremental version of calculating the Merkle root should be used to calculate the the roots used for `action_mroot` and `transaction_mroot` of Proper IF Blocks only. The incremental Merkle tree could be used as well, but since we already have the list of digests used to calculate the root, it is faster to use the non-incremental version on a separate thread (and potentially in parallel).

greg7mdp:
Resolves #2333 

This PR provides a new Merkle tree implementation which does not unnecessarily store extra digests and does not do extra hashes.

Previous versions are renamed: `incremental_merkle_tree_legacy`, `calculate_merkle_legacy`.
New versions are named: `incremental_merkle_tree`, `calculate_merkle`.

Note: Even though the files have been renamed `incremental_merkle_legacy.hpp` and `merkle_legacy.hpp`, the implementation has not been modified (besides remoning the `canonical` template parameter).

#### New `incremental_merkle_tree`
- does not store extra hashes (root matches `calculate_merkle`)
- provides the root on demand, does not compute it on `append()`.
- is about 20x faster than `incremental_merkle_tree_legacy`.
- storage is minimal. Just one `uint64_t` and a `vector<digest_type>` whose size() is `std::popcount(num_digest_appended)`.

#### New `calculate_merkle`
- does not store extra hashes. (root matches `incremental_merkle_tree`).
- is about 25% faster than `calculate_merkle_legacy` without multithreading
- supports internal multithreading for computing roots of collections whose `size() > 4096`, in which case it is 2x to almost 4x faster (2x for sequences of at least 10k digests, 4x for very large sequences).
- doesn't mutate the passed `deque<digest_type>`.

#### New performance tests
This PR adds two performance tests comparing the `legacy` and `new` merkle tree implementations
- `./unittests/unit_test "--run_test=merkle_tree_tests/perf_test_one_large"`
- `./unittests/unit_test "--run_test=merkle_tree_tests/perf_test_many_small"`

#### New consistency test
This PR adds a test checking consistency between `incremental_merkle_tree` and `calculate_merkle` when adding from 1 to 1000 digests:
- `./unittests/unit_test "--run_test=merkle_tree_tests/consistency_over_large_range"`


libraries/chain/controller.cpp
```diff
@@ -4068,7 +4068,7 @@ struct controller_impl {
       if (if_active) {
          return calculate_merkle( std::move(digests) );
```
heifner, line 4069: Search for `calculate_merkle` and you will find a few places where copies are made so they can be moved into `calculate_merkle`. Those copies are no longer needed.

libraries/chain/include/eosio/chain/merkle.hpp
```diff
@@ -1,29 +1,82 @@
...
-    *  Calculates the merkle root of a set of digests, if ids is odd it will duplicate the last id.
-    *  Uses make_canonical_pair which before hashing sets the first bit of the previous hashes
-    *  to 0 or 1 to indicate the side it is on.
-    */
-   digest_type legacy_merkle( deque<digest_type>&& ids );
+   if (size == 2)
+      return hash_combine(start[0], start[1]);
+   else {
+      if (async && size >= 4096) {                    // below 4096, starting async threads is overkill
+         std::array<std::future<digest_type>, 4> fut; // size dictates the number of threads (must be power of two)
```
heifner, line 40: This is likely fine as it is only 4 threads. We do need to be mindful of thread usage as we don't want to overwhelm a system so that it slows down the main thread. Other thread usage is setup at startup. Would be nice to be able to provide an `io_context` for these tasks so they could run on the chain thread pool; however likely not worth the effort.

libraries/chain/include/eosio/chain/merkle_legacy.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+   inline auto make_legacy_digest_pair(const digest_type& l, const digest_type& r) {
+      return make_pair(make_legacy_left_digest(l), make_legacy_right_digest(r));
+   };
+
+}
+
+/**
+ *  Calculates the merkle root of a set of digests, if ids is odd it will duplicate the last id.
+ *  Uses make_canonical_pair which before hashing sets the first bit of the previous hashes
```
heifner, line 37: Fix comment

libraries/chain/include/eosio/chain/merkle_legacy.hpp
```diff
@@ -0,0 +1,57 @@
+#pragma once
+#include <eosio/chain/types.hpp>
+#include <fc/io/raw.hpp>
+
+namespace eosio::chain {
+
+namespace detail {
+
+   inline digest_type make_legacy_left_digest(const digest_type& val) {
+      digest_type canonical_l = val;
```
heifner, line 10: Rename variable to `left`.

libraries/chain/include/eosio/chain/merkle_legacy.hpp
```diff
@@ -0,0 +1,57 @@
...
+      canonical_l._hash[0] &= 0xFFFFFFFFFFFFFF7FULL;
+      return canonical_l;
+   }
+
+   inline digest_type make_legacy_right_digest(const digest_type& val) {
+      digest_type canonical_r = val;
```
heifner, line 16: Rename variable to `right`.

libraries/chain/include/eosio/chain/incremental_merkle_legacy.hpp
```diff
@@ -0,0 +1,205 @@
...
+ *
+ * to achieve this new nodes can either imply an set of future nodes
+ * that achieve a balanced tree OR realize one of these future nodes.
+ *
+ * Once a sub-tree contains only realized nodes its sub-root will never
+ * change.  This allows proofs based on this merkle to be very stable
+ * after some time has past only needing to update or add a single
+ * value to maintain validity.
+ *
+ * @param canonical if true use the merkle make_canonical_pair which sets the left/right bits of the hash
```
heifner, line 51: Remove

libraries/chain/include/eosio/chain/incremental_merkle.hpp
```diff
@@ -6,212 +6,70 @@
...
-
-      }
-
-      /**
-       * return the current root of the incremental merkle
-       */
-      DigestType get_root() const {
-         if (_node_count > 0) {
-            return _active_nodes.back();
+class incremental_merkle_tree {
```
heifner, line 9: A class comment would be nice.

libraries/chain/include/eosio/chain/merkle.hpp
```diff
@@ -1,29 +1,82 @@
...
+         for (size_t i=0; i<fut.size(); ++i)
+            res[i] = fut[i].get();
+
+         return calculate_merkle_pow2(res.begin(), res.end());
+      } else {
+         auto mid = start + size / 2;
+         return hash_combine(calculate_merkle_pow2(start, mid), calculate_merkle_pow2(mid, end));
+      }
+   }
+}
```
heifner, line 57: Add comment that `detail` ends and this is the public interface. I think `calculate_merkle(start, end)` should be part of the public interface along with `calculate_merkle(deque)`

libraries/chain/include/eosio/chain/merkle.hpp
```diff
@@ -1,29 +1,82 @@
...
+   if (size == midpoint)
+      return calculate_merkle_pow2<It, true>(start, end);
+
+   auto mid = start + midpoint;
+   return hash_combine(calculate_merkle_pow2<It, true>(start, mid), calculate_merkle(mid, end));
+}
+
+}
+
+inline digest_type calculate_merkle(const deque<digest_type>& ids) {
```
heifner, line 77: Seems this should be a template so it could take any range that provides `cbegin(), cend()`.

heifner (changes requested):

libraries/chain/include/eosio/chain/merkle_legacy.hpp
greg7mdp: done

libraries/chain/include/eosio/chain/merkle_legacy.hpp
greg7mdp: done

libraries/chain/include/eosio/chain/merkle_legacy.hpp
greg7mdp: done.

libraries/chain/controller.cpp
greg7mdp: done.

libraries/chain/include/eosio/chain/incremental_merkle_legacy.hpp
greg7mdp: done.

libraries/chain/controller.cpp
greg7mdp: done

libraries/chain/include/eosio/chain/merkle.hpp
greg7mdp: done

libraries/chain/include/eosio/chain/merkle.hpp
greg7mdp: done.

libraries/chain/include/eosio/chain/merkle.hpp
greg7mdp: I updated it so that it uses only 2 threads up to 4096 digests, and 4 for 4096 and up.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Improved Merkle Tree, no longer storing unneeded digests, switched to incremental calculations.
Note:end

libraries/chain/include/eosio/chain/incremental_merkle.hpp
greg7mdp: done, please let me know if you would add/change something.

heifner (approved):

linh2931 (approved):

Issue #2360, opened 2024-03-29
IF: Provide finality_data in SHiP for Transition Savanna Blocks


linh2931:
SHiP support of IBC was implemented before proper Savanna transition was implemented. It was based on instant transition where Transition Savanna Blocks were not applicable. Now that proper Savanna transition has been implemented, @systemzax reports the `finality_data` for the second block after Savanna Genesis Block is not provided by SHiP. 

Need to update code to provide `finality_data` for Transition blocks.

PR #2362, opened 2024-03-29
IF: SHiP -- Provide finality_data for Transition blocks

linh2931:
Current implementation of SHiP support of IBC was based on instant transition where Transition Savanna Blocks were not applicable. Now that proper Savanna transition has been implemented, @systemzax reports the `finality_data` for the second block after Savanna Genesis Block is not provided by SHiP. This PR adds `finality_data` for Transition blocks.

https://github.com/AntelopeIO/leap/issues/2360


libraries/chain/controller.cpp
```diff
@@ -4301,7 +4301,15 @@ struct controller_impl {
    }
 
    std::optional<finality_data_t> head_finality_data() const {
-      return apply_s<std::optional<finality_data_t>>(chain_head, [](const block_state_ptr& head) { return head->get_finality_data(); });
+      if (fork_db.version_in_use() == fork_database::in_use_t::both) {
+         // During transition to Savanna, need to fetch the head from fork_db which
+         // contains Savanna information for finality_data
+         return fork_db.apply_s<std::optional<finality_data_t>>([&](const auto& forkdb) {
+            return forkdb.head()->get_finality_data(); });
```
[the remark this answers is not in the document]
heifner: We don't populate forkdb_s as we go because we only wanted the branch with IF Critical Block in it.

heifner (changes requested):

libraries/chain/controller.cpp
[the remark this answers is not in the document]
linh2931: Oh I thought we populate both DBs while in transition. So we have to store all information required by `base_digest` (including `header`, `core`, `finalizer_policies`, ...) in `block_header_state_legacy`?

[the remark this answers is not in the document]
heifner: That or build it up right here from IF Genesis Block for each transition block.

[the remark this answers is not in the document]
linh2931: Probably it is cleaner to just build here, to avoid clutter `block_header_state_legacy` with too much less used data. The downside is to have to rebuild all previous Transition blocks for a particular block. But we should not have too many Transitions blocks. I can traverse the branch from IF Genesis block to current block in `forkdb_l`?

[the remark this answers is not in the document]
heifner: Yes, similar to how it is done in replay.

ericpassmore:
Note:start
group: IF
category: INTERNALS
summary: Update SHiP, restore missing transitions blocks.
Note:end

heifner (approved):

libraries/chain/controller.cpp
```diff
@@ -4316,8 +4316,72 @@ struct controller_impl {
...
+      });
+
+      block_state_ptr prev;
+      auto bitr = legacy_branch.rbegin();
+
+      // search for the first block having instant_finality_extension
+      // and create the Savanna Genesis Block
+      if (legacy_root->header.contains_header_extension(instant_finality_extension::extension_id())) {
+         prev = block_state::create_if_genesis_block(*legacy_root);
+      } else {
```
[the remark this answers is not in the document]
linh2931: Thanks. I have added a comment about the reason searching both places.

libraries/chain/controller.cpp
```diff
@@ -4316,8 +4316,77 @@ struct controller_impl {
       }
    }
 
+   // This is only used during Savanna transition, which is a one-time occurrance
+   // and the number of Transition block is small.
```
greg7mdp, line 4320:
```suggestion
   // This is only used during Savanna transition, which is a one-time occurrence
   // and the number of Transition blocks is small.
```

libraries/chain/controller.cpp
```diff
@@ -4316,8 +4316,77 @@ struct controller_impl {
...
+      });
+
+      block_state_ptr prev;
+      auto bitr = legacy_branch.rbegin();
+
+      // get_transition_block_finality_data is called by SHiP as a result
+      // of receiving accepted_block signal. That is before
+      // the call to log_irreversible where root() is updated.
+      // Search both root and legacy_branch for the first block having
+      // instant_finality_extension -- the the Savanna Genesis Block.
```
greg7mdp, line 4337:
```suggestion
      // instant_finality_extension -- the Savanna Genesis Block.
```

libraries/chain/controller.cpp
```diff
@@ -4316,8 +4316,77 @@ struct controller_impl {
...
+      for (; bitr != legacy_branch.rend(); ++bitr) {
+         auto new_bsp = std::make_shared<block_state>(
+               *prev,
+               (*bitr)->block,
+               protocol_features.get_protocol_feature_set(),
+               validator_t{}, skip_validate_signee);
+
+         assert((*bitr)->action_receipt_digests_savanna);
+         auto digests = *((*bitr)->action_receipt_digests_savanna);
+         new_bsp->action_mroot = calculate_merkle(std::move(digests)); // required by finality_data
```
greg7mdp, line 4364:
```suggestion
         const auto& digests = *((*bitr)->action_receipt_digests_savanna);
         new_bsp->action_mroot = calculate_merkle(digest)); // required by finality_data
```

greg7mdp (changes requested):

libraries/chain/controller.cpp
heifner: yep, depends on your definition of "small"

greg7mdp: I'm a little worried about doing 200+ `calculate_merkle` in in a row... couldn't that take more than the block time?

linh2931: Good point. We could cache `action_mroot_savanna` instead of `action_receipt_digests_savanna` in legacy block state. That would add some cost when building legacy block state, but the cost is amortized.

greg7mdp: Yes, I think that's clearly what we should do.

linh2931: Will do it shortly.

libraries/chain/block_state_legacy.cpp
```diff
@@ -75,7 +75,7 @@ namespace eosio::chain {
    ,block( std::move(b) )
    ,_pub_keys_recovered( true ) // called by produce_block so signature recovery of trxs must have been done
    ,_cached_trxs( std::move(trx_metas) )
-   ,action_receipt_digests_savanna( std::move(action_receipt_digests_savanna) )
+   ,action_mroot_savanna( action_receipt_digests_savanna ? std::optional<digest_type>(calculate_merkle(*action_receipt_digests_savanna)) : std::nullopt )
```
greg7mdp, line 78: let's make the `action_receipt_digests_savanna` parameter to the constructor be a `const&`, we don't need it ro be a rvalue anymore.

libraries/chain/controller.cpp
```diff
@@ -4316,8 +4313,76 @@ struct controller_impl {
...
+
+      for (; bitr != legacy_branch.rend(); ++bitr) {
+         auto new_bsp = std::make_shared<block_state>(
+               *prev,
+               (*bitr)->block,
+               protocol_features.get_protocol_feature_set(),
+               validator_t{}, skip_validate_signee);
+
+         assert((*bitr)->action_mroot_savanna);
+         new_bsp->action_mroot = *((*bitr)->action_mroot_savanna); // required by finality_data
```
greg7mdp, line 4360: I see we set `new_bsp->action_mroot` here, but I don't see where it is used when we create the new `block_state` at the next loop iteration?

linh2931: `action_mroot` is not used during this process of transitions; it is used when we do `prev->get_finality_data();` at line 4366.

greg7mdp: So if we use only the one from the last block_state, why do we populate (and did `calculate_merkle`) for all the preceding blocks?

linh2931: Oha, we could have checked if it is the last block in legacy_branch, save it, and do the copy after the for loop. I think we should still cache action_mroot in block_state_legacy because it is used several places.

greg7mdp: Wait, I'm totally confused. Should,'t the `action_mroot` be used in `block_state::new_valid` to compute the new `valid_t` stored in the block state, which is then used for the next `valid` computation.
I thought we had to go through the bsp from the genesis block to recreate this. If we don't why do we need the loop at line 4352?

linh2931: For finality_data, we don't need `valid`. We only need information from header state for `base_digest` plus new action_mroot. That's why `valid` is not re-created.

greg7mdp: At line 4360, we set `new_bsp->action_mroot = *((*bitr)->action_mroot_savanna);`.
``new_bsp->action_mroot` is not the result of `calculate_merkle(action_receipt_digests)`, it should be `parent.get_finality_mroot_claim(qc_data.qc_claim);` I think.

linh2931: No, for finality_data sent to SHiP, the action_mroot is the mroot of the actual actions, not the action_mroot in header state changed to represent qc_claim. The actual mroot is defined in line 81 of block_state.hpp

greg7mdp: OK, thanks!

PR #2381, opened 2024-04-04
IF: switch variant index of new `get_blocks_request_v1`

spoonincode:
The order of message types in the `state_request` variant is part of the binary protocol. I suggest we add the new `get_blocks_request_v1` to the end. Otherwise, unmodified ship clients that both use `get_blocks_ack_request_v0` and do not consume the ABI from the websocket endpoint (I believe abieos or eos-go based readers) will misbehave.


heifner (approved):

linh2931 (approved):
Thanks for catching this.

PR #2385, opened 2024-04-08
IF: Populated fork db ASAP; Vote processing off the main thread

### LINKED ISSUE #2102: IF: Guarantee net_plugin processes votes after it has has processed the block the vote was for
The net_plugin will be responsible for respecting this order per connection. It can leverage the existing cache of block IDs received and sent per connection to know when it is the appropriate time to process the votes per each connection.

We also need to make sure we process the block (add to fork database) before we process the votes for that block, otherwise they will be dropped.

heifner:
- Add blocks to the fork db as soon as header validation is complete. They are added off the main thread (net threads).
  - Move logging of received blocks to the controller since many are processed in `maybe_switch_forks` and would not be logged by the `producer_plugin` after a call to `push_block` otherwise.
- Call `log_irreversible` even when head is not updated as processed qcs can move irreversible at any time.
- Integrate QC and consider voting as soon as header validation is complete off the main thread (net threads).
- Add logging to `app().quit()` to be able to determine if nodeos is stopped via a SIGTERM.

Does not fully resolve #2102. Still need to move the processing into a voting thread pool. That will come as a separate PR as the changes here already are rather extensive.


libraries/chain/controller.cpp
```diff
@@ -3676,23 +3720,29 @@ struct controller_impl {
...
       EOS_ASSERT( id == bsp->id(), block_validate_exception,
                   "provided id ${id} does not match block id ${bid}", ("id", id)("bid", bsp->id()) );
+
+      if constexpr (savanna_mode) {
+         integrate_received_qc_to_block(bsp); // Save the received QC as soon as possible, no matter whether the block itself is valid or not
+         consider_voting(bsp);
+      }
+
+      if (conf.terminate_at_block == 0 || bsp->block_num() <= conf.terminate_at_block) {
+         forkdb.add(bsp, mark_valid_t::no, ignore_duplicate_t::yes);
```
linh2931, line 3730: So even for Legacy, blocks are added to forkdb earlier?

libraries/chain/controller.cpp
```diff
@@ -4775,9 +4833,21 @@ void controller::assemble_and_complete_block( block_report& br, const signer_cal
    br = my->pending->_block_report;
 }
 
-void controller::commit_block() {
+void controller::commit_block(block_report& br) {
+   fc::time_point start = fc::time_point::now();
+
    validate_db_available_size();
    my->commit_block(block_status::incomplete);
+
```
linh2931, line 4841: It would be nice those logging are done inside `my-commit_block`.

libraries/chain/controller.cpp
```diff
@@ -3753,17 +3803,17 @@ struct controller_impl {
          return;
       }
 
-      // Don't save the QC from block extension if the claimed block has a better valid_qc.
-      if (claimed->valid_qc && (claimed->valid_qc->is_strong() || received_qc.is_weak())) {
+      // Don't save the QC from block extension if the claimed block has a better or same valid_qc.
+      if (received_qc.is_weak() || claimed->valid_qc_is_strong()) {
```
linh2931, line 3807: Maybe `claimed->valid_qc_is_strong()` should be placed the first, as strong is common.

libraries/chain/include/eosio/chain/thread_utils.hpp
```diff
@@ -12,6 +12,74 @@
...
+      }
+
+      class accessor {
+         std::lock_guard<std::mutex> g;
+         T& v;
+      public:
+         accessor(std::mutex& m, T& v)
+            : g(m), v(v) {}
+         T& value() { return v; }
+      };
```
greg7mdp, line 48: This seems like a potentially dangerous design. I could do:

`auto& id = if_irreversible_block_id.make_accessor().value();`

and then access `id` outside of the lock.

I think if we use it only for `digest_type`, it would be preferable to call `load` and copy the 32 bytes, it is no big deal.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -68,22 +70,38 @@ namespace eosio::chain {
...
       fsi_map                           inactive_safety_info;  // loaded at startup, not mutated afterwards
       fsi_t                             default_fsi = fsi_t::unset_fsi(); // default provided at leap startup
       mutable bool                      inactive_safety_info_written{false};
 
-      template<class F>
+   public:
+      my_finalizers_t(block_timestamp_type startup_time, const std::filesystem::path& persist_file_path)
+         : t_startup(startup_time)
+         , persist_file_path(persist_file_path)
+      {}
```
greg7mdp, line 87: Why add this constructor which prevents aggregate initioalization?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -3847,7 +3841,7 @@ namespace eosio {
 
       uint32_t lib = cc.last_irreversible_block_num();
       try {
-         if( blk_num <= lib || cc.block_exists(blk_id) ) {
+         if( blk_num <= lib || cc.validated_block_exists(blk_id) ) {
```
greg7mdp, line 3844: why `validated`?

libraries/chain/block_state.cpp
```diff
@@ -85,7 +85,6 @@ block_state_ptr block_state::create_if_genesis_block(const block_state_legacy& b
    // TODO: https://github.com/AntelopeIO/leap/issues/2057
    // TODO: Do not aggregate votes on blocks created from block_state_legacy. This can be removed when #2057 complete.
    result.pending_qc = pending_quorum_certificate{result.active_finalizer_policy->finalizers.size(), result.active_finalizer_policy->threshold, result.active_finalizer_policy->max_weak_sum_before_weak_final()};
-   result.valid_qc = {}; // best qc received from the network inside block extension, empty until first savanna proper IF block
```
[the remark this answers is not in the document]
heifner: To me at least, `result` seems an easier to understand name than `genesis_bsp` as it is clearer it is what is being returned.

libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -21,43 +21,51 @@ inline std::vector<uint32_t> bitset_to_vector(const hs_bitset& bs) {
 }
 
 bool pending_quorum_certificate::has_voted(size_t index) const {
-   std::lock_guard g(*_mtx);
-   return _strong_votes._bitset.at(index) || _weak_votes._bitset.at(index);
+   return _strong_votes.has_voted(index) || _weak_votes.has_voted(index);
```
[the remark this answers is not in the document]
heifner: No, each is an atomic bool.

libraries/chain/controller.cpp
heifner: yes

libraries/chain/controller.cpp
heifner: It should only be done for `commit` called by the producer plugin. We could tell that in `my->commit_block()` via `block_status::incomplete`, but I don't see how that is better than logging it here.

libraries/chain/controller.cpp
heifner: I placed `received_qc.is_weak()` because first because it can be checked without acquiring a mutex.

libraries/chain/controller.cpp
```diff
@@ -3318,16 +3335,47 @@ struct controller_impl {
...
+
    template<class BSP>
    void apply_block( controller::block_report& br, const BSP& bsp, controller::block_status s,
                      const trx_meta_cache_lookup& trx_lookup ) {
       try {
          try {
             auto start = fc::time_point::now();
+
+            const bool already_valid = bsp->is_valid();
+            if (!already_valid) // has not been validated (applied) before, only in forkdb, see if we can vote now
```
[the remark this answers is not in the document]
heifner: I updated the comment:
```
            // not have been validated and we could not vote. At this point bsp->final_on_strong_qc_block_ref has been validated and we can vote.
            // Only need to consider voting if not already validated, if already validated then we have already voted.
```

libraries/chain/controller.cpp
```diff
@@ -3716,25 +3766,25 @@ struct controller_impl {
...
       }
-      return fork_db.apply_s<std::optional<block_handle>>(f);
+      return fork_db.apply<std::optional<block_handle>>(unlinkable, f);
    }
 
-   // expected to be called from application thread as it modifies bsp->valid_qc and if_irreversible_block_id
+   // thread safe
    void integrate_received_qc_to_block(const block_state_ptr& bsp_in) {
       // extract QC from block extension
       const auto& block_exts = bsp_in->block->validate_and_extract_extensions();
```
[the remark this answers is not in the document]
heifner: Doesn't matter. The `const&` keeps the temp object around until end of scope.

libraries/chain/controller.cpp
```diff
@@ -3954,19 +3996,20 @@ struct controller_impl {
...
-            }
-
             auto branches = forkdb.fetch_branch_from( new_head->id(), chain_head.id() );
 
-            if( branches.second.size() > 0 ) {
+            bool switch_fork = !branches.second.empty();
+            if( switch_fork ) {
+               auto head_fork_comp_str = apply<std::string>(chain_head, [](auto& head) -> std::string { return log_fork_comparison(*head); });
+               ilog("switching forks from ${current_head_id} (block number ${current_head_num}) ${c} to ${new_head_id} (block number ${new_head_num}) ${n}",
+                    ("current_head_id", chain_head.id())("current_head_num", chain_head.block_num())("new_head_id", new_head->id())("new_head_num", new_head->block_num())
```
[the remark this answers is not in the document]
heifner: Your clang format file has `ColumnLimit: 148` which this is longer than. I'll shorten it.

libraries/chain/hotstuff/finalizer.cpp
```diff
@@ -99,7 +99,7 @@ std::optional<vote_message> finalizer::maybe_vote(const bls_public_key& pub_key,
       } else {
          sig =  priv_key.sign({(uint8_t*)digest.data(), (uint8_t*)digest.data() + digest.data_size()});
       }
-      return vote_message{ bsp->id(), decision == vote_decision::strong_vote, pub_key, sig };
+      return std::optional{vote_message{ bsp->id(), decision == vote_decision::strong_vote, pub_key, sig }};
```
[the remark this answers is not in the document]
heifner: https://www.youtube.com/watch?v=0yJk5yfdih0

libraries/chain/hotstuff/hotstuff.cpp
```diff
@@ -133,34 +141,29 @@ vote_status pending_quorum_certificate::add_vote(block_num_type block_num, bool
...
+      wlog( "signature from finalizer ${i} cannot be verified", ("i", index) );
+      return vote_status::invalid_signature;
+   }
+
+   std::unique_lock g(*_mtx);
+   state_t pre_state = _state;
+   s = strong ? add_strong_vote(index, sig, weight)
+              : add_weak_vote(index, sig, weight);
+   state_t post_state = _state;
+   g.unlock();
```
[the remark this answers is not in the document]
heifner: `pre_state` and `post_state` used in the `dlog`.

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -78,16 +78,27 @@ namespace eosio::chain {
...
          bls_aggregate_signature _sig;
+         std::vector<std::atomic<bool>> _processed; // avoid locking mutex for _bitset duplicate check
 
-         void resize(size_t num_finalizers) { _bitset.resize(num_finalizers); }
-         size_t count() const { return _bitset.count(); }
+         void reflector_init();
+      public:
+         explicit votes_t(size_t num_finalizers)
+            : _bitset(num_finalizers)
+            , _processed(num_finalizers) {}
```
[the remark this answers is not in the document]
heifner: It makes sure that both `_bitset` and `_processed` are initialized to the same size.

libraries/chain/include/eosio/chain/hotstuff/hotstuff.hpp
```diff
@@ -113,19 +124,22 @@ namespace eosio::chain {
...
+      std::optional<valid_quorum_certificate> _valid_qc; // best qc received from the network inside block extension
       uint64_t             _quorum {0};
       uint64_t             _max_weak_sum_before_weak_final {0}; // max weak sum before becoming weak_final
       state_t              _state { state_t::unrestricted };
       uint64_t             _strong_sum {0}; // accumulated sum of strong votes so far
       uint64_t             _weak_sum {0}; // accumulated sum of weak votes so far
-      votes_t              _weak_votes;
-      votes_t              _strong_votes;
+      votes_t              _weak_votes {0};
+      votes_t              _strong_votes {0};
```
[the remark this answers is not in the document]
heifner: Because of the added constructor.

libraries/chain/controller.cpp
linh2931: Maybe put a comment there.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: Oh, please change to 120 in the `.clang_format`. I had changed it in mine.

libraries/chain/include/eosio/chain/thread_utils.hpp
heifner: It is not about avoiding the copy. It is to avoid `if_irreversible_block_id` being set to a lower block number.

```c++
   void set_if_irreversible_block_id(const block_id_type& id) {
      const block_num_type id_num = block_header::num_from_id(id);
      const bock_id_type current_lib_id = if_irreversible_block_id.load();
      const block_num_type current_num = block_header::num_from_id(current_lib_id);
      // thread 1 is setting LIB from 3 to 7
      // thread 2 is setting LIB from 3 to 5
      if( id_num > current_num ) {
         // thread 2: which is setting LIB to 5 pauses here
         // thread 1: which is setting LIB to 7 comes in and sets to 7
         // thread 1: unpauses and sets LIB back to 5
         dlog("set irreversible block ${bn}: ${id}, old ${obn}: ${oid}", ("bn", id_num)("id", id)("obn", current_num)("oid", accessor.value()));
         if_irreversible_block_id.store(id);
      }
   }



```

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: Yes, but this is a very weird behavior IMO which is not without problems.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
heifner: Because the attributes are now private. They are prviate so their access can be protected by a mutex.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: Wait, here we check `bsp->is_valid()`, not the `bsp->final_on_strong_qc_block_ref`?

libraries/chain/controller.cpp
linh2931: All other public controller APIs are uniformly one-line code calling the corresponding function in `my`, hiding the implementation. But you don't have to change.

plugins/net_plugin/net_plugin.cpp
heifner: Because `cc.block_exists(blk_id)` is always true here. We added to the fork_db when the block_state was created. However, if validated then that means its state has been applied we don't need this block.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
heifner: Ok, changed. Now I have many lines past that.

libraries/chain/hotstuff/finalizer.cpp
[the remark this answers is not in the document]
greg7mdp: I watched this video a couple weeks ago when you sent it and I had already forgotten about it! Thanks.

libraries/chain/controller.cpp
heifner: Makes sense, I'll move it.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
heifner: It is according to the spec. We actually rely on that behavior in other locations. This doesn't need that but it has the nice property that if `validate_and_extract_extensions()` is changed to return a `const&` then this code is still optimal.

libraries/chain/include/eosio/chain/thread_utils.hpp
greg7mdp: Got it, makes sense. I think I would have preferred a custom struct, with a mutex, and a member function `advance_lib(block_num_type block_num);` which would do it safely.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
greg7mdp: Do they need to be protected by a mutex?  I believe they are set at startup time and never change.

heifner: Those two do not, but the other attributes do need the protection.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: Thanks.

Not only it is nicer for ne, but also lines less than 120 characters are also nicer to view in github's split view as they fit on one line.

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: Still I'm not getting the check at line 3372 (`if (!already_valid)`).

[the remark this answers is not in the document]
heifner: If we have already called `apply_block` for this block then we have already consider voting for this block. At the point of `apply_block` the `bsp->final_on_strong_qc_block_ref` has to have been applied so we have already voted if we could. There is no reason to call `consider_voting` and look up the `final_on_strong_qc_block_ref` as we know that we have already voted if we could have.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
greg7mdp: got it, thanks.

libraries/chain/include/eosio/chain/hotstuff/finalizer.hpp
```diff
@@ -95,20 +113,28 @@ namespace eosio::chain {
...
          }
       }
 
-      size_t  size() const { return finalizers.size(); }
-      void    set_keys(const std::map<std::string, std::string>& finalizer_keys);
+      size_t  size() const { return finalizers.size(); }   // doesn't change, thread safe
+      bool    empty() const { return finalizers.empty(); } // doesn't change, thread safe
+
+      template<typename F>
+      bool all_of_public_keys(F&& f) const { // only access keys which do not change, thread safe
```
greg7mdp, line 126: This should probably lock the mutex.

linh2931 (approved):

libraries/chain/controller.cpp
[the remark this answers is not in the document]
greg7mdp: Thanks, got it!

greg7mdp (approved):

PR #2400, opened 2024-08-05
[5.0] compute_transaction should return failure trace

### LINKED ISSUE #2281: The `compute_transaction` API endpoint response changed
Somewhere between 4.x and 5.x the way the `/v1/chain/compute_transaction` API endpoint changed how it responds when errors occur, which is causing it to be more difficult to use the endpoint to calculate resource requirements. 

Previously it would allow the transaction to "process" and return any errors that occurred in the `processed.except` field with a nice error stack. If there was a `tx_cpu_usage_exceeded` exception, it would return the data alongside afield to easily access the `stack[0].data.billed` required for the transaction.

Now in post 5.x, it's throwing an actual 500 error as the response with more limited data as an error string, and no error stack.

What are the chances we can revert and go back to the previous behavior?

heifner:
`/v1/chain/compute_transaction` API endpoint should return a failure trace. 
This was accidently changed to false when refactoring send_transaction in 5.0.

See 4.0 impl: https://github.com/AntelopeIO/leap/blob/01dab51ac00aed6730700b82943c30ccfd854fce/plugins/chain_plugin/chain_plugin.cpp#L2562-L2562

Resolves #2281

### DIFF
--- a/plugins/chain_plugin/chain_plugin.cpp
+++ b/plugins/chain_plugin/chain_plugin.cpp
@@ -2552,7 +2552,7 @@ read_only::get_required_keys_result read_only::get_required_keys( const get_requ
 }
 
 void read_only::compute_transaction(compute_transaction_params params, next_function<compute_transaction_results> next) {
-   send_transaction_params_t gen_params { .return_failure_trace = false,
+   send_transaction_params_t gen_params { .return_failure_trace = true,
                                           .retry_trx            = false,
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):

spoonincode (approved):

ericpassmore:
Note:start
category: Other
component: Plugin-Chain
summary: Restore failure trace to `compute_transaction`.
Note:end

PR #2403, opened 2024-08-20
[5.0] fix `true_lowest()` for `uint128[2]` index; possibly resolving some secondary table row RPC queries

spoonincode:
Default parameters for the lower bound of a `uint128[2]` ("`checksum256`") index used for the `get_table` RPC call were uninitialized and might cause unexpected results to be returned. Initialize this to zero.


heifner (approved):

ericpassmore:
Note:start
category: Other
component: Internal
summary: Fix `true_lowest()` index to possibly resolve some secondary table row RPC queries.
Note:end

Issue #2330, opened 2024-03-22
abi_tests/abi_large_array test fails with swap space less than 2GB


Duncan-Ultra:
OS: Ubuntu-22.04
Leap: v5.0.2

Where this happens: https://github.com/AntelopeIO/leap/blob/13b1c03c29801d13bd87d2d46cb778b05fbb74e4/unittests/abi_tests.cpp#L2254

Expected exception when I run with 8GB swap space so it will pass as `fc::exception`
```
3015013 unpack_exception: Unpack data exception
Unable to unpack 'ARRAY[0]' from stream
    {"p":"ARRAY[0]"}
    unit_test  abi_serializer.cpp:457 _binary_to_variant
rethrow
    {}
    unit_test  abi_tests.cpp:2258 test_method
``` 
Actual exception:
```
13 St9bad_alloc: std::bad_alloc
rethrow std::bad_alloc:
    {"what":"std::bad_alloc"}
    unit_test  abi_tests.cpp:2257 test_method
```

My swap space for actual exception:
```
➜ free -m
               total        used        free      shared  buff/cache   available
Mem:           32059         608       30373           3        1077       31016
Swap:           2048           0        2048
```

I cannot find any information about the minimum requirement for running leap or these tests. So, should you guy add exception for this case, or I'm expected to increase my swap space instead?

arhag:
Thanks for reporting. This issue will be fixed in Leap 5.0.3 (see https://github.com/AntelopeIO/leap/pull/2405) and Spring 1.0.0 (see https://github.com/AntelopeIO/spring/pull/603).

ericpassmore:
Note:start
group: STABILITY
category: INTERNALS
summary: Limit vector size that can be reserved.
Note:end

PR #2404, opened 2024-08-20
Fix issue #2330 by limiting the vector size that can be reserve'd

greg7mdp:
Resolves [#2330](https://github.com/AntelopeIO/leap/issues/2330).

The deserialization of arrays involves reading a size, and then the array elements which are `emplace`'d back into a `std::vector`. For normal vectors, since we know the size in advance, it is beneficial to call `std::vector::reserve()` to prevent quadratic reallocations of the vector's buffer as the values are read. Still, limit the size that can be reserved to `1024` `fc::variant` values, or 16KB since `sizeof(fc::variant) == 16`).


heifner:
Should target `release/5.0`

PR #2408, opened 2024-09-25
[5.0.3] P2P: Resolve on reconnect

heifner:
Resolve address on reconnect.

See https://github.com/AntelopeIO/spring/issues/525

Instead of:
````
info  2024-09-25T21:50:50.686 net-1     net_plugin.cpp:2797           operator()           ] connection failed to p2p.telos-bad.zenblocks.io:9876, Element not found
info  2024-09-25T21:50:50.686 net-1     net_plugin.cpp:1461           _close               ] ["p2p.telos-bad.zenblocks.io:9876" - 1 <unknown>:<unknown>] closing
````
Now you get:
```
warn  2024-09-25T21:42:48.642 net-0     net_plugin.cpp:4524           operator()           ] Unable to resolve p2p.telos-bad.zenblocks.io:9876 Host not found (authoritative)
```

### DIFF
--- a/plugins/net_plugin/net_plugin.cpp
+++ b/plugins/net_plugin/net_plugin.cpp
@@ -348,7 +348,6 @@ namespace eosio {
          std::string host;
          connection_ptr c;
          tcp::endpoint active_ip;
-         tcp::resolver::results_type ips;
       };
 
       using connection_details_index = multi_index_container<
@@ -418,7 +417,7 @@ namespace eosio {
       string connect(const string& host, const string& p2p_address);
       string resolve_and_connect(const string& host, const string& p2p_address);
       void update_connection_endpoint(connection_ptr c, const tcp::endpoint& endpoint);
-      void connect(const connection_ptr& c);
+      void reconnect(const connection_ptr& c);
       string disconnect(const string& host);
       void close_all();
 
@@ -912,7 +911,7 @@ namespace eosio {
 
       fc::sha256              conn_node_id;
       string                  short_conn_node_id;
-      string                  listen_address; // address sent to peer in handshake
+      const string            listen_address; // address sent to peer in handshake
       string                  log_p2p_address;
       string                  log_remote_endpoint_ip;
       string                  log_remote_endpoint_port;
@@ -2764,6 +2763,10 @@ namespace eosio {
             fc_dlog( logger, "Skipping connect due to go_away reason ${r}",("r", reason_str( no_retry )));
             return false;
       }
+
+      if (incoming())
+         return false;
+
       if( consecutive_immediate_connection_close > def_max_consecutive_immediate_connection_close || no_retry == benign_other ) {
          fc::microseconds connector_period = my_impl->connections.get_connector_period();
          fc::lock_guard g( conn_mtx );
@@ -2771,9 +2774,10 @@ namespace eosio {
             return true; // true so doesn't remove from valid connections
          }
       }
+
       connection_ptr c = shared_from_this();
       strand.post([c]() {
-         my_impl->connections.connect(c);
+         my_impl->connections.reconnect(c);
       });
       return true;
    }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: plugins/net_plugin/net_plugin.cpp [hunk 1]
```diff
@@ -4493,32 +4497,35 @@ namespace eosio {
...
             connection_ptr c = std::make_shared<connection>( peer_address, listen_address );
-            c->set_heartbeat_timeout( heartbeat_timeout );
-            std::lock_guard g( connections_mtx );
-            auto [it, inserted] = connections.emplace( connection_detail{
-               .host = peer_address,
-               .c = std::move(c),
-               .ips = results
+            c->strand.post([this, resolver, c, err, results, host, port, peer_address]() {
+               c->set_heartbeat_timeout( heartbeat_timeout );
+               std::unique_lock g( connections_mtx );
```
greg7mdp, line 4514: same here with using `lock_guard` within a `{}` block.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4536,13 +4543,12 @@ namespace eosio {
...
 
-   void connections_manager::connect(const connection_ptr& c) {
-      std::lock_guard g( connections_mtx );
-      const auto& index = connections.get<by_connection>();
-      const auto& it = index.find(c);
-      if( it != index.end() ) {
-         it->c->connect( it->ips );
-      }
+   void connections_manager::reconnect(const connection_ptr& c) {
+      std::unique_lock g( connections_mtx );
```
greg7mdp, line 4547: same here

greg7mdp (approved):

linh2931 (approved):
Can a test be added?

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4493,32 +4497,37 @@ namespace eosio {
...
+                     .c = c,
+                  });
+               }
+               if( !err ) {
+                  c->connect( results );
+               } else {
+                  fc_wlog( logger, "Unable to resolve ${host}:${port} ${error}",
+                           ("host", host)("port", port)( "error", err.message() ) );
+                  c->set_state(connection::connection_state::closed);
+                  ++(c->consecutive_immediate_connection_close);
```
spoonincode, line 4528: What does increasing this do? I thought it was going to cause some sort of cool down when reaching `def_max_consecutive_immediate_connection_close` but doesn't seem to matter

heifner: Need to look into it, but I think the change that broke this re-connect also broke this back off of `consecutive_immediate_connection_close`

heifner: Hmm, actually this a reason to reuse the connection object. I guess I should maintain the reuse of the connection object for this.

spoonincode: It seems like now (on 51fd8fa) after failing to resolve a host `def_max_consecutive_immediate_connection_close` times it just never tries it again. But if the resolve completes and simply unable to connect to the remote host, it will try retry indefinitely. I can't tell if this discrepancy is intentional or not. It makes me wonder if we should _not_ be increasing this counter here for consistency between the two, but realistically if someone has a completely bad hostname it's probably not fixing itself.

plugins/net_plugin/net_plugin.cpp
```diff
@@ -4486,39 +4490,48 @@ namespace eosio {
...
+                  std::lock_guard g( connections_mtx );
+                  connections.emplace( connection_detail{
+                     .host = peer_address,
+                     .c = c,
+                  });
+               }
+               if( !err ) {
+                  c->connect( results );
+               } else {
+                  fc_wlog( logger, "Unable to resolve ${host}:${port} ${error}",
```
spoonincode, line 4529: fwiw this is a warning log, where failing to connect is an info log. Not sure if you want to make consistent or not

spoonincode (approved):

plugins/net_plugin/net_plugin.cpp
heifner: If the `async_connect` fails it calls `c->close(false)` which calls `_close()` which increments `consecutive_immediate_connection_close`.

heifner: I think a `warn` is appropriate if unable to resolve.

spoonincode: > If the `async_connect` fails it calls `c->close(false)` which calls `_close()` which increments `consecutive_immediate_connection_close`.

There is something different between the two though. For example when I run nodeos with ` --p2p-peer-address fdsfdsfds.invalid:4444 --p2p-peer-address fddsfdsfdsfds.invalid:4444 --p2p-peer-address 127.0.0.1:2121 --p2p-peer-address www.google.com:4444` you'll see how the latter 2 will try to reconnect forever where the first 2 won't.

There is also some log oddness  at shutdown,
```
info  2024-09-26T02:45:43.993 nodeos    net_plugin.cpp:4578           close_all            ] close all 4 connections
info  2024-09-26T02:45:43.993 net-1     net_plugin.cpp:1460           _close               ] ["127.0.0.1:2121" - 1 <unknown>:<unknown>] closing
info  2024-09-26T02:45:43.993 net-2     net_plugin.cpp:1460           _close               ] ["fddsfdsfdsfds.invalid:4444" - 2 <unknown>:<unknown>] closing
info  2024-09-26T02:45:43.993 nodeos    net_plugin.cpp:4385           plugin_shutdown      ] exit shutdown
```

heifner: This different in backoff behavior seems to match Leap 4.0 behavior.

ericpassmore:
Note:start
category: Other
component: P2P
summary: When reconnecting using the net plugin resolve endpoint.
Note:end

Issue #2417, opened 2026-07-21
Add result limits and pagination to get_accounts_by_authorizers


pur3miish:
## Problem

`/v1/chain/get_accounts_by_authorizers` currently returns every matching account permission authority for the supplied accounts or public keys. Unlike other collection endpoints, the request does not accept a result limit or pagination cursor.

A public key with an unusually large authority set can therefore cause the endpoint to perform substantial work, allocate and serialize a very large response, and keep the request pending for a long time. Clients cannot bound the result set at the nodeos API boundary.

## WAX example

The following public key is associated with a very large number of accounts on the WAX blockchain:

```
PUB_K1_5E84rb8xKvRks4iUpQvChnG7F35CJvF3SaGXpEHiEkHTeWkGg9
```

Example request:

```bash
curl -X POST https://<wax-node>/v1/chain/get_accounts_by_authorizers \
  -H 'content-type: application/json' \
  -d '{
    "accounts": [],
    "keys": [
      "PUB_K1_5E84rb8xKvRks4iUpQvChnG7F35CJvF3SaGXpEHiEkHTeWkGg9"
    ]
  }'
```

Independent WAX indexers currently associate this key with tens of thousands of account names. Many names appear to be small variations or transpositions of other account names.

One possible explanation is typo-squatting: creating look-alike account names in the hope that a sender mistypes a recipient and sends funds to the wrong account. That motivation has not been proven and is not necessary to reproduce the endpoint problem. Regardless of why these accounts exist, one shared key creates a pathological, attacker-influenceable result set.

In practice, a lookup for this key can remain pending for an excessive period or fail to complete within a normal client timeout. A downstream client-side limit does not solve the problem because nodeos has already traversed, allocated, and serialized the complete result before the client can discard excess rows.

## Impact

- A single public key can trigger an unbounded response.
- Public API consumers cannot specify how much work or data they are willing to accept.
- Large authority sets can consume HTTP worker time, CPU, memory, and bandwidth.
- Client disconnection or proxy timeouts may not immediately stop server-side work.
- This creates an availability and resource-amplification concern for public API nodes.
- Applications that only need the first few matching accounts cannot safely use the endpoint with adversarial or unexpectedly reused keys.

## Proposed behavior

Add bounded, server-enforced result handling to `get_accounts_by_authorizers`. For example:

1. Accept an optional `limit` parameter.
2. Enforce a node-configured maximum even when the caller requests a larger value or omits `limit`.
[...]

PR #2418, opened 2026-07-21
Limit get_accounts_by_authorizers results

pur3miish:
## Summary

- add a caller-supplied `limit` to `get_accounts_by_authorizers`
- enforce a node-side maximum of 1,000 results regardless of the requested limit
- stop traversing the account-query indices as soon as the effective limit is reached
- return `more: true` when matching authorities were truncated
- document the request and response fields in the Chain API schema
- add focused coverage for limited and complete result sets

## Why

A public key can be assigned to an unusually large number of account permissions. The endpoint currently traverses and serializes every match, so a pathological or attacker-influenced authority set can consume significant CPU, memory, HTTP worker time, and bandwidth.

Applying a limit downstream does not protect nodeos because the complete response has already been generated. This change enforces the bound while iterating the node's authority index.

This addresses the node-side result-limit portion of #2417. Cursor pagination and cancellation-aware deadlines can be considered separately.

## Validation

- `git diff --check`
- Chain API Swagger YAML parsed successfully with Ruby/Psych
- added `get_accounts_by_authorizers_limit_test` for truncated and complete results

The focused C++ test was not executed locally because this clean checkout does not include Leap's initialized submodules and pinned build toolchain; CI should compile and run it.


pur3miish:
Superseded by the maintained Spring implementation: AntelopeIO/spring#1894. The corresponding Spring issue is AntelopeIO/spring#1893. Closing this draft because Leap is no longer supported.

